PR #166, opened 2016-01-22
Fixed SNAP-471

nthanvi:
the storeCallbackImpl is not shared on executors in nonembedded mode
and hence reading RDD ID from there is not possible directly.
Currently setting RDDId to -1 in  non embedded mode but later we need to pass the
RDDId from driver to executors so that UI display could be better.


sumwale:
@nthanvi have created https://snappydata.atlassian.net/browse/SNAP-502 to track the comments mentioned above. Would appreciate if you can respond to comments in future in terms of: a) either comments are not valid and reasons for the same, b) orl tackle in followup checkins and mention the same, c) or file an issue to track. After providing comments, it is not expected of the reviewer/contributor to keep track of followup so please do not ignore such.

nthanvi:
@sumwale  thanks for opening jira to track the issues. missed  on few review comments  due to some loose  gmail filters. corrected them now. 
ExternalShellDUnitTest.VerifyEmbeddedTablesAndCreateNewInShell   does create tables with shell SparkContext.

PR #185, opened 2016-02-22
Handle all the store boot properties

nthanvi:
Handle all the gemfirexd store boot properties
minor changes for scala code styling

### DIFF
--- a/snappy-tools/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
+++ b/snappy-tools/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
@@ -19,7 +19,7 @@ package io.snappydata
 
 import java.util.Properties
 
-import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
 
 import com.gemstone.gemfire.distributed.DistributedMember
 import com.pivotal.gemfirexd.internal.engine.distributed.utils.GemFireXDUtils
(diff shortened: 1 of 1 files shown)


snappy-tools/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
```diff
@@ -37,16 +37,21 @@ object ToolsCallbackImpl extends ToolsCallback {
...
-            "be provided in the format host[port] or host:port", null)
-      }
-      properties.setProperty("locators", locator)
-      sc.getConf.getOption(Property.mcastPort).map(
-        properties.setProperty("mcast-port", _))
-    }.getOrElse(properties.setProperty("mcast-port",
-      sc.getConf.get(Property.mcastPort)))
+    sc.getConf.getAll.filter(
+      p => p._1.startsWith(Constant.STORE_PROPERTY_PREFIX)).
+        foreach(p => p._1 match {
```
hbhanawat: Use different names of the variables for the inner and outer blocks.

hbhanawat: okay they are different blocks. But changing name would still be good from reading perspective.

hbhanawat:
Looks good. But we wanted an overall strategy for handling properties for job-server, spark and gemxd.

nthanvi:
@hbhanawat  - SNAP-322 - enables user to pass all the gemxd store related properties from the spark conf file to gemxd. if you think we need strategy/handling for job server and spark properties as well then please open respective jira to track those gaps.

PR #186, opened 2016-02-26
Snap 565

rishitesh:
Streaming API changes.
1) SnappyStreamingContext is to be more alike StreamingContext. The access pattern used by users should be easy to understand and be alike with StreamingContext.
   Most of the constructor are now similar to StreamingContext. One can also give check point directory to initialize SnappyStreamingContext.

2) We have made SQL commands around StreamingContext lifecycle. This mandates us to extend some of the StreamingContext access mechanism. For example I have added SnappyStreamingContext.getInstance() to get an initialized or started instance.

3) SnappyStreamingContext.getActive() will behave same as StreamingContext.getActive().

4) Package of SnappyStreamingContext is changed to org.apache.spark.streaming rather than org.apache.spark.sql.streaming. This context interacts with SQL no doubt, but it can also be used as a normal StreamingContext.

5) Added an adapter StreamSqlAdapter to overcome package restrictions for streaming & sql


ymahajan:
I think we should keep the packages as it is. sql package is more appropriate as we will be enhancing sql and cq and will be dependent on internal catalyst classes. We are not going to enhance spark streaming as such. This is also inline with streaming Dataframes which is coming in Spark.

SnappyStreamingContext can be more like StreamingContext even it is in sql package. We can change access modifiers in the StreamingContext if required. 

Also we are adding new StreamSqlAdapter just to get around of package restructuring. 

Is this required for 0.5, as these APIs are going to change anyways once we get the querying support on streams in spark itself.

rishitesh:
Precisely my point. If Spark adds sql support for ( which we can see in StreamFrame PRs) , we will be enhancing mostly SQL related classes. And it might happen that we may not need to change StreamingContext at all.. 


Changing access modifiers is easy, but we are not keeping the sanctity of the package restrictions , in a case where it can be easily done. Once we are compatible with Spark 2.0 we can see if we need SnappyStreamingContext. If not we can phase it out. But till then users should be able use SnappyStreamingContext exactly like StreamingContext with some additional benefits. 

But IF AT ALL we need to extend StreamingContext we need to put that in streaming package, as we can't give half cooked streaming functionality to users. Needs to be a full fledged StreamingContext  with future maintainability. How many classes we keep on making public ?


That's my humble opinion. Others opinion is awaited.

ymahajan:
I am also questioning the need of this considering the APIs will be modified in future. I thought Jags also clarified that we should not spend time on streaming in this release considering the amount of other high priority items for the 15th Apr deadline. The new StreamSqlAdapter is something misleading, in my opinion. Is this a default implementation of any trait?  In fact i am asking to reduce public classes here.

snappy-core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
```diff
@@ -0,0 +1,290 @@
...
+
+  def getOrCreate(
+      checkpointPath: String,
+      creatingFunc: () => SnappyStreamingContext,
+      hadoopConf: Configuration = SparkHadoopUtil.get.conf,
+      createOnError: Boolean = false
+      ): SnappyStreamingContext = {
+    val checkpointOption = CheckpointReader.read(
+      checkpointPath, new SparkConf(), hadoopConf, createOnError)
+    checkpointOption.map(new SnappyStreamingContext(null, _, null)).getOrElse(creatingFunc())
```
hbhanawat: 1. Throw an exception if a streaming context is already created. 
2. set this newly created streaming context as the current instance

hbhanawat:
Looks good to me.

PR #187, opened 2016-03-01
Python API support for Snappy

nthanvi:


nthanvi:
@rishitesh  - please review

snappy-core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -322,20 +328,21 @@ class SnappyContext protected[spark](@transient override val sparkContext: Spark
...
+    * val DF = snappyContext.createTable(stagingAirline, "parquet", Map("path" -> airlinefilePath))
+    *
+    * }}}
+    *
+    * @param tableName Name of the table
+    * @param provider  Provider name such as 'COLUMN', 'ROW', 'JDBC', 'PARQUET' etc.
+    * @param options   Properties for table creation
+    * @return DataFrame for the table
+    */
   def createTable(
```
sumwale: Please revert all these formatting changes. I believe current formatting of these is proper as per Spark guidelines that we have been following. Please ensure the application of codeStyleSettings.xml (as in docs/build-instructions.md).

snappy-core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -347,29 +354,84 @@ class SnappyContext protected[spark](@transient override val sparkContext: Spark
...
+    * @return DataFrame for the table
+    */
+  def createTable(
+      tableName: String,
+      provider: String,
+      schema: StructType,
+      options: java.util.Map[String, String]): DataFrame = {
+    createTable(tableName, provider, schema, options.asScala.toMap);
+  }
+
```
sumwale: Don't we need similar overloads for createSampleTable, createApproxTSTable?

sumwale:
Looks good. See couple of comments above regarding indentation changes and other overloads.

snappy-core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
nthanvi: Yes. Those changes will be needed there as well. Rishi also mentioned some Optional class instead of Option,  that probably fit both for java and scala (need to check for python) . if that works out then we will not need to overload these methods.

snappy-core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
nthanvi: ok

PR #188, opened 2016-03-09
Merge1.6

rishitesh:


snappy-core/src/main/scala/org/apache/spark/sql/execution/datasources/StoreDataSourceStrategy.scala
```diff
@@ -77,9 +77,11 @@ private[sql] object StoreDataSourceStrategy extends Strategy with Logging {
       filterPredicates,
       numPartition,
       partitionColumns,
-      (requestedColumns, pushedFilters) => {
-        scanBuilder(requestedColumns, selectFilters(pushedFilters).toArray)
+      (requestedColumns, _, pushedFilters) => {
+        scanBuilder(requestedColumns, pushedFilters.toArray)
       })
+
```
sumwale: @rishitesh Don't we need selectFilters() to restrict Expressions to Filters? Or is it now being done at some higher level?

snappy-tools/build.gradle
```diff
@@ -103,7 +103,9 @@ task docs(type: ScalaDoc) {
         !it.getPath().matches('.*/internal/.*') && !it.getPath().contains('com/gemstone/gemfire/cache/operations/')
       })
       if (project.plugins.hasPlugin('scala')) {
-        allSource.addAll(project.sourceSets.main.allScala)
+        allSource.addAll(project.sourceSets.main.allScala.findAll {
+          !it.getPath().matches('.*org/apache/spark/sql/execution/joins/HashedRelation.*') 
+        })
       }
```
sumwale: What's the problem with scaladoc generation for HashedRelation and shouldn't that be fixed instead of excluding it? Spark should already be creating scaladoc for that in their maven build, so are their our changes causing an issue?

snappy-core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -249,7 +249,7 @@ object ExternalStoreUtils extends Logging {
         case _ =>
       }
     }
-    JdbcUtils.createConnection(url, connProperties)
+    JdbcUtils.createConnectionFactory(url, connProperties).apply()
```
sumwale: Better to create the factory only once from calling classes.

snappy-tools/src/main/scala/org/apache/spark/sql/columntable/ColumnFormatRelation.scala
```diff
@@ -264,8 +267,8 @@ class ColumnFormatRelation(
    */
   override def destroy(ifExists: Boolean): Unit = {
     // use a non-pool connection for operations
-    val conn = JdbcUtils.createConnection(externalStore.connProperties.url,
-      externalStore.connProperties.connProps)
+    val conn = JdbcUtils.createConnectionFactory(externalStore.connProperties.url,
```
sumwale: @rishitesh See the comment before. Better to create the connection factory once and keep as a private member in the class.

snappy-tools/src/main/scala/org/apache/spark/sql/columntable/ColumnFormatRelation.scala
```diff
@@ -350,7 +353,7 @@ class ColumnFormatRelation(
     StoreCallbacksImpl.stores.get(table) match {
       case Some((_, _, rddId)) =>
         val rddInfo = new RDDInfo(rddId, table, numPartitions,
-          StorageLevel.OFF_HEAP, Seq(), None)
+          StorageLevel.OFF_HEAP, Seq())
```
sumwale: Can keep the empty Seq() as a static constant.

snappy-tools/build.gradle
rishitesh: Spark ignores internal packages while generating scaldoc . See Unidoc.ignoreUndocumentedPackages. I dont know whey we decided to document all. Ideally we should also ignore internal classes. I will file a bug and track it in a JIRA

snappy-core/src/main/scala/org/apache/spark/sql/execution/datasources/StoreDataSourceStrategy.scala
rishitesh: This the area of code , which signals which predicates have been pushed down , so that filter does not apply on them again. I have that in mind and going forward will resolve this.

snappy-core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
rishitesh: Will change accordingly.

snappy-tools/src/main/scala/org/apache/spark/sql/columntable/ColumnFormatRelation.scala
rishitesh: Will change accordingly.

snappy-tools/src/main/scala/org/apache/spark/sql/columntable/ColumnFormatRelation.scala
rishitesh: Will change accordingly.

snappy-core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
sumwale: I have done that as part of the merge for #207 so ignore

snappy-core/src/main/scala/org/apache/spark/sql/execution/datasources/StoreDataSourceStrategy.scala
sumwale: @rishitesh Ok. Btw, I also noticed this in ColumnFormatRelation:

```
  override def partitionColumns: Seq[String] = {
    connectionType match {
      case ConnectionType.Embedded => partitioningColumns
      case _ =>   Seq.empty[String] // Temporary fix till we fix Non-EMbededed join
    }
  }
```

Is this issue being tracked somewhere in JIRA? I believe the external shell mode will become important going forward (already we have multiple prospectives who want to run with stock spark in shell mode).

hbhanawat: Opened https://snappydata.atlassian.net/browse/SNAP-657.

PR #190, opened 2016-03-11
We can avoid array object creation while scanning column table in an …

rishitesh:
We can avoid array object creation while scanning column table in an iterator.
Also changed the name of snappyApi to snappyJavaUtil , which I missed in earlier merge.


snappy-core/src/main/scala/org/apache/spark/sql/store/JDBCSourceAsStore.scala
```diff
@@ -151,8 +151,11 @@ final class CachedBatchIteratorOnRS(conn: Connection,
 
   override def hasNext: Boolean = _hasNext
 
+  val numCols = requiredColumns.length
+  val colBuffers = new Array[Array[Byte]](numCols)
+
```
sumwale: Better to make these private.

When branch-0.1-preview having SNAP-608 is merged to master then some conflicts will have be to resolved (like the next() method below is gone to base class which need not be altered).

PR #191, opened 2016-03-12
Fixes and refactoring for SNAP-608

sumwale:
* moved all ResultSet iteration code into a ResultSetIterator class taken mostly from InternalRowIteratorOnRS
   for cleanly closing it even if not iterated fully (like for LIMIT); also simplified handling a bit
   and now both InternalRowIteratorOnRS and CachedBatchIteratorOnRS extend it. This fixes SNAP-608
 * added a test section in existing QueryRoutingDUnitTest.testSNAP193 to check the fix
 * moved all disparate pool property handling (max/idle sizes etc) into ExternalStoreUtils.getAllPoolProperties
   where the respective properties will not be overwritten if explicitly provided
 * removed DriverRegistry.register call in every ConnectionPool get and instead have it only in few places
   like creation of Relation objects on driver, and ExternalStoreUtils.getConnector call on executors
 * updated tomcat pool and hikariCP versions; added the choice of default Tomcat vs Hikari in Constant

### DIFF
--- a/snappy-core/src/main/scala/org/apache/spark/sql/store/JDBCSourceAsStore.scala
+++ b/snappy-core/src/main/scala/org/apache/spark/sql/store/JDBCSourceAsStore.scala
@@ -25,6 +25,7 @@ import scala.collection.mutable.ArrayBuffer
 import scala.language.implicitConversions
 import scala.reflect.ClassTag
 import scala.util.Random
+import scala.util.control.NonFatal
 
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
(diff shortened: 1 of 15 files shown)


snappy-core/src/main/scala/org/apache/spark/sql/store/JDBCSourceAsStore.scala
```diff
@@ -143,36 +144,70 @@ class JDBCSourceAsStore(override val connProperties: ConnectionProperties,
...
+}
+
+final class CachedBatchIteratorOnRS(conn: Connection,
+    requiredColumns: Array[String],
+    stmt: Statement, rs: ResultSet, context: TaskContext)
+    extends ResultSetIterator[CachedBatch](conn, stmt, rs, context) {
+
+  protected override def getNextValue(rs: ResultSet): CachedBatch = {
     // it will be having the information of the columns to fetch
     val numCols = requiredColumns.length
```
rishitesh: Please see https://github.com/SnappyDataInc/snappydata/pull/190. We can avoid array creation for each row.
You can take that code from the PR

sumwale: Thanks @rishitesh. I will go ahead with merge of this because zoomdata is waiting for fix for this and I have subsequent fixes lined up for other reported issues. Can you update your PR after this merge?

PR #192, opened 2016-03-12
Fix the datatype mismatch issue when number of cols > 8 (SNAP-607)

sumwale:
* use the correct index into the datatype array and not the current group index
 * simplified the logic of splitting into groups of 8 to be just /8 and %8 to go with corresponding changes in snappy-store; updated snappy-store link for (https://github.com/SnappyDataInc/snappy-store/pull/33)
 * added test for SNAP-607 in QueryRoutingDUnitTest.testSNAP193_608 and also enabled the test for SNAP-609

Also see https://github.com/SnappyDataInc/snappy-store/pull/33 for corresponding PR on snappy-store.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -246,7 +246,6 @@ subprojects {
       }
     }
   })
-  check.dependsOn test, scalaTest
 
   // apply default manifest
   if (rootProject.hasProperty('enablePublish')) {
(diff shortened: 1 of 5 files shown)


sumwale:
Discussed with Soubhik. He will be providing detailed review comments later.

PR #195, opened 2016-03-15
Merge 0.2 preview branch into master

ashetkar:
Merge 0.2 preview branch into master.
It is showing old changes in master, which is a bit confusing to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ SnappyData is a **distributed in-memory data store for real-time operational ana
 ## Download binary distribution
 You can download the latest version of SnappyData here:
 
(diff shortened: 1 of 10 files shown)


sumwale:
Amogh, I had deliberately asked you to do it so that there is some additional know-how of that apart from myself. All you should need to do is an "git merge origin/branch-0.x-preview" on latest master. The only conflict you should need to resolve is in build.gradle where version will be at 0.2-SNAPSHOT rather than PREVIEW.

Let me know if you have any other trouble and we can sit and sort it. Thanks.

ashetkar:
This has been merge manually, see f0212fb750b9f6c6ec59dd95a763d751b4eb5351.
Closing the PR.

Issue #183, opened 2016-02-18
Running local on MacOsX El Capitain


nxbdi:
snappy> connect client 'localhost:1527';
ERROR 08006: (SQLState=08006 Severity=40000) A communications error has been detected: Failed after trying all available servers: [], for control URL: localhost[1527].
Caused by: ConnectException: Connection refused
	at java.net.PlainSocketImpl.socketConnect(Native Method)
ERROR 08006: (SQLState=08006 Severity=40000) A communications error has been detected: Failed after trying all available servers: [localhost[1527]], for control host[port]: localhost[1527].
Caused by: ConnectException: Connection refused
	at java.net.PlainSocketImpl.socketConnect(Native Method)
ERROR 08001: (SQLState=08001 Severity=40000) java.net.ConnectException : Error connecting to server localhost on port 1527 with message Connection refused.
Caused by: ConnectException: Connection refused
	at java.net.PlainSocketImpl.socketConnect(Native Method)

16/02/18 14:17:39.384 EST serverConnector<tid=0xc> INFO snappystore: Initializing region _DDL_STMTS_META_REGION
16/02/18 14:17:39.385 EST serverConnector<tid=0xc> INFO snappystore: Region /_DDL_STMTS_META_REGION recovered from the local disk. Old persistent ID: /127.0.0.1:/Users/timothyspann/Downloads/snappydata-0.1.0-PREVIEW/work/localhost-locator-1/./datadictionary created at timestamp 1455822888412 version 0 diskStoreId bd4a1e44-f164-46d2-8855-281e40271e87 name , new persistent ID /127.0.0.1:/Users/timothyspann/Downloads/snappydata-0.1.0-PREVIEW/work/localhost-locator-1/./datadictionary created at timestamp 1455823059383 version 0 diskStoreId bd4a1e44-f164-46d2-8855-281e40271e87 name
16/02/18 14:17:40.018 EST serverConnector<tid=0xc> INFO snappystore: FabricDatabase: waiting for index loading from GFXD-DEFAULT-DISKSTORE
16/02/18 14:17:40.019 EST serverConnector<tid=0xc> INFO snappystore: FabricDatabase: Index loading completed for GFXD-DEFAULT-DISKSTORE in 0 ms
16/02/18 14:17:40.021 EST Idle OplogCompactor<tid=0x37> INFO snappystore: Oplog::recoverValuesIfNeeded: recovering values from oplog#1
[...]

PR #196, opened 2016-03-15
Put the hostname in quotes to prevent zsh issues.

radutanasa:
See issue #183


hbhanawat:
Tested this on bash and zsh. Works good. Merged.

PR #197, opened 2016-03-15
Typos, spacings & splitting words

radutanasa:
I suggest we do not split words at the end of lines because it makes the docs harder to maintain. Each time we add / remove something we need to go through the entire document and make sure that the splitted words are not sitting in the middle of a line.


hbhanawat:
Looks good to me. Merged.

PR #200, opened 2016-03-17
1) Added support for preserve partitions in Snappy column tables

rishitesh:
1) Added support for preserve partitions in Snappy column tables
2) Removed the code which used to store data in row buffer in case of a dataframe insert and partition column empty.
3) We can define an option for table as PRESERVE_PARTITION. It will always try to do a local insert. In rare cases it will do a remote put( bucket moved cases)

API for the change is more or less same as before. One thing we need to check is how to provide PRESERVE_PARTITION option. Right now I have added it in table level option.

### DIFF
--- a/snappy-dunits/src/test/scala/io/snappydata/dunit/externalstore/ColumnTableDUnitTest.scala
+++ b/snappy-dunits/src/test/scala/io/snappydata/dunit/externalstore/ColumnTableDUnitTest.scala
@@ -92,7 +92,7 @@ class ColumnTableDUnitTest(s: String) extends ClusterManagerTestBase(s) {
 
   // changing the test to such that batches are created
   // and looking for column table stats
-  def testSNAP205_InsertLocalBucketsNonPartitioning(): Unit = {
+  def testSNAP205_InsertLocalBucketsNonPartitioning(): Unit =   {
     val snc = SnappyContext(sc)
 
     var data = Seq(Seq(1, 2, 3), Seq(7, 8, 9), Seq(9, 2, 3),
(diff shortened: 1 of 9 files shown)


rishitesh:
@jramnara @suranjan @ymahajan @hbhanawat Please review.

snappy-dunits/src/test/scala/io/snappydata/dunit/externalstore/ColumnTableDUnitTest.scala
```diff
@@ -92,7 +92,7 @@ class ColumnTableDUnitTest(s: String) extends ClusterManagerTestBase(s) {
 
   // changing the test to such that batches are created
   // and looking for column table stats
-  def testSNAP205_InsertLocalBucketsNonPartitioning(): Unit = {
+  def testSNAP205_InsertLocalBucketsNonPartitioning(): Unit =   {
```
sumwale: unnecessary space

snappy-tools/src/main/scala/org/apache/spark/sql/columntable/ColumnFormatRelation.scala
```diff
@@ -158,21 +159,38 @@ class ColumnFormatRelation(
...
-      //TODO - rddID should be passed to the executors so that cachedBatch size can be shown be correctly even for non embedded mode
+    //TODO - rddID should be passed to the executors so that cachedBatch size can be shown be correctly even for non embedded mode
+
+    //Change for preserve partition : {
+    // We are inserting a cached batch to column table directly
+    // in case of preserve partition. We are not considering small batches.
+    // We will be simply putting the data as CachedBatches. This will
+    // ensure the query results will be ok even though the storage is suboptimal.
+    // Once we come up with a cleaner solution we can start inserting
+    // residual rows to row buffer
```
sumwale: Why does it matter whether row goes to row buffer or directly? All we need to ensure is that the bucketId column (which is used for partitioning in row buffer) is the same as partition ID. I think it should be quite simple to add row buffer support which will be important especially for streaming cases (the most important use-case for preservePartitioning) where lots of small batches will be the norm rather than exception.

sumwale:
@rishitesh See some comments -- haven't reviewed all. However, I think we should discuss whether we need this at all or not with unified partitioner in master.

In my opinion we can forgo this (too many options already for user) and just have the common way of specifying same partitioning columns as in the source RDD partitioning. This is required in any case for proper behaviour if user were to insert in some other way like JDBC or SnappyContext.insert.

sumwale:
@rishitesh what we do need instead is some way of having source RDD route partitions as per buckets. In other words somehow passing the preferredLocations of buckets above to parent (if its preferredLocations is Nil) from a DataFrameWriter.

sumwale:
@rishitesh Last comment about routing source RDD partitions as per buckets has already been taken care of in the ColumnInsertExec plan (and an EXCHANGE if it does not match). I don't think we need the preserve-partitioning anymore so this should be closed.

What we do need is to somehow allow matching the Kafka partitioning with table partitioning (and still be consistent with Catalyst partitioner) but that is a different discussion.

rishitesh:
Closing this PR as it is not needed.

PR #201, opened 2016-03-18
Linking snappy-store with latest SNAP-517

rk-geode:


sumwale:
@rk-geode I would suggest that if a change is just for linking then there is no use of doing that here because after merge to master the merge revision will be a different on master in any case. If need be, then just commit the new link version directly on master for such cases (i.e. when there is no code change otherwise in snappydata).

PR #203, opened 2016-03-21
script to collect stack dumps and debug information of all the nodes

kneeraj:
* added a script to collect logs and stats of all the nodes of the system.

* modified the start script to add the hostname and working dir folder for
each of the components.

@hbhanawat you also please look into this.


snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+  srv_num=$3
+
+  if [ "${VERBOSE}" = "1" ]; then
+    echo "Collecting data for process running on ${host} with working_dir ${wd}"
+  fi
+
+  # make a sub directory for this host and pid
+  out_dir="${PWD}/debug_data/${host}-${srv_num}"
+
+  collector_host="pnq-kneeraj3"
```
soubhik-c: this should be $host. right ?

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+
+  # get the stack dumps first
+  dump_num=1
+  for i in {1..${num_stack_dumps}}
+  do
+    dump_num=`expr ${dump_num} + 1`
+    kill -URG $proc_id
+    sleep $int_stack_dumps
+  done
+
```
soubhik-c: Stack dump should be optional because of DLock dump that freezes the system during SIGURG, customers don't want stack dump to happen in production.

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+for p in "${all_pids[@]}"
+do
+  if [ "${VERBOSE}" = "1" ]; then
+    echo "Waiting for pid ${p}"
+  fi
+  wait $p 2> /dev/null
+done
+
+# make zipped tar ball
+tar -zcf debug_data.tar.gz debug_data
```
soubhik-c: can we have an option to create individual zip per server because logs can be quite big at times.

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+    copyright_headers=`grep 'Copyright (C)' ${l}`
+    if [ ! -z "$copyright_headers" ]; then
+      # also check for the pid line and get the pid
+      proc_id=`sed -n 's/.*Process ID: \([0-9]\+\)$/\1/p' ${l}`
+      if [ "${verbose}" = "1" ]; then
+        echo "Addiing file ${l} to the array"
+      fi
+      files+=($l)
+      last_restart_log=$l
+      break
```
soubhik-c: can we have an option to collect just an additional file just before the file containing Copyright information. During crash analysis log file just before relaunch of server is most interesting.

soubhik-c:
Overall the script is quite clean and does exactly as per the defined scope. Using typeset is nifty and avoids the need for the script file accessible on remote host.

I would like to have an option to sort by name instead of time for log files because at times ls -t may not be reliable in the sense user touched that log file or copied that file without preserve changing the timestamp. In such situations, logs collected will be misleading and hard to detect this human error. 
gfs file time based is fine.

This can be merged and enhancements can be taken up next although my personal preference will be doing the enhancement here (needed for smart situation) and merge it in a single shot. especially now we have squash merge :).

snappy-tools/sbin/collect-debug-artifacts.sh
kneeraj: No. This should be the collector host, on which the script is invoked. For testing I changed it to pnq-kneeraj. Will rectify it.

snappy-tools/sbin/collect-debug-artifacts.sh
kneeraj: ok.

snappy-tools/sbin/collect-debug-artifacts.sh
kneeraj: This I have left it. If user wants he can do it.

snappy-tools/conf/debug.conf.template
```diff
@@ -0,0 +1,3 @@
+MEMBERS_INFO_FILE=members_info
```
sumwale: Can we use something other than *info :-)

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+usage="Usage: collect-debug-artifacts [ -c|--conf|--config conffile ] [ -h|--help ] \
+       [ -a|--all ] [ -v|--verbose ]"
+
+PWD=`pwd`
+while [ "$1" != "" ]; do
+key="$1"
+
+case $key in
+  -c|--conf|--config)
+  CONF_FILE="$2"
```
sumwale: usual convention with double hyphen "--" is to have the value with an '=' rather than in next argument. So I will suggest like:

```sh
case $key in
  -c)
    CONF_FILE="$2"
    shift ;;
  --conf=*|--config=*)
    CONF_FILE="`echo "$2" | sed 's/^[^=]*=//'`" ;;
```

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+#!/usr/bin/env bash
+
+usage="Usage: collect-debug-artifacts [ -c|--conf|--config conffile ] [ -h|--help ] \
+       [ -a|--all ] [ -v|--verbose ]"
+
+PWD=`pwd`
+while [ "$1" != "" ]; do
+key="$1"
+
+case $key in
```
sumwale: Put in double quotes (since with long opts it can also have value separated by =): `case "$key" in`

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+  exit 0
+  ;;
+  -a|--all)
+  GET_EVERYTHING=1
+  ;;
+  -v|--verbose)
+  VERBOSE=1
+  ;;
+esac
+shift # past argument or value
```
sumwale: why not indent one level more inside while loop?

snappy-tools/sbin/collect-debug-artifacts.sh
sumwale: If you do that then update usage accordingly.

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+# permissions and limitations under the License. See accompanying
+# LICENSE file.
+#
+
+#!/usr/bin/env bash
+
+usage="Usage: collect-debug-artifacts [ -c|--conf|--config conffile ] [ -h|--help ] \
+       [ -a|--all ] [ -v|--verbose ]"
+
+PWD=`pwd`
```
sumwale: Use a pattern like this for the directory of this script:
```sh
SCRIPT_DIR="`dirname "$0"`"
SCRIPT_DIR="`cd "$SCRIPT_DIR" && pwd`"
```

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+  VERBOSE=1
+  ;;
+esac
+shift # past argument or value
+done
+
+# Check configurations and assign defaults
+function check_configs {
+  if [ -z "${CONF_FILE}" ]; then
+    CONF_FILE="../conf/debug.conf.template"
```
sumwale: prefix for SCRIPT_DIR mentioned at the start `"${SCRIPT_DIR}/../conf/debug.conf.template"`

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+  fi
+
+  if [ ! -f "${CONF_FILE}" ]; then
+    echo "Config file ${CONF_FILE} does not exists"
```
sumwale: "does not exist"

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+    exit 1
+  fi
+
+  source $CONF_FILE
+
+  if [ ! -f "${MEMBERS_INFO_FILE}" ]; then
+    echo "MEMBERS_INFO_FILE file ${MEMBERS_INFO_FILE} does not exists"
```
sumwale: "does not exist"

sumwale: can we have a default template MEMBERS_FILE which matches our defaults in snappy-start-all.sh? i.e. one lead, locator and server

snappy-tools/sbin/collect-debug-artifacts.sh
sumwale: Also the name "key" can be something more generic like "option" since it can be both key or key=value

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+    echo NUM_STACK_DUMPS=$NO_OF_STACK_DUMPS
+    echo INTERVAL_BETWEEN_DUMPS=$INTERVAL_BETWEEN_DUMPS
+    echo GET_EVERYTHING=${GET_EVERYTHING}
+  fi
+}
+
+function collect_data {
+  host=$1
+  wd=$2
+  srv_num=$3
```
sumwale: lets use double quotes for uniformity: `host="$1"` etc

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+
+  if [ "${VERBOSE}" = "1" ]; then
+    echo "Collecting data for process running on ${host} with working_dir ${wd}"
+  fi
+
+  # make a sub directory for this host and pid
+  out_dir="${PWD}/debug_data/${host}-${srv_num}"
+
+  collector_host="pnq-kneeraj3"
+  mkdir -p $out_dir
```
sumwale: double quotes

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+    echo "arg1 wd = ${wd}" 
+    echo "arg2 num_stack_dumps = ${NO_OF_STACK_DUMPS}" 
+    echo "arg3 interval_dumps = ${INTERVAL_BETWEEN_DUMPS}" 
+    echo "arg4 GET_EVERYTHING = ${GET_EVERYTHING}" 
+    echo "arg5 collector_host = ${collector_host}" 
+    echo "arg6 out dir = ${out_dir}"
+  fi
+
+  typeset -f | ssh $host "$(cat);collect_on_remote ${wd} ${NO_OF_STACK_DUMPS} \\
+      ${INTERVAL_BETWEEN_DUMPS} ${GET_EVERYTHING} ${collector_host} ${out_dir} ${VERBOSE}"
```
sumwale: Great use of typset

sumwale: Will be preferable to double quote args `wd`, `collector_host` and `out_dir`. Can be tricky inside ssh. Check if backslash quoted works: `ssh $host "$(cat);collect_on_remote \"${wd}\" ..."

If it doesn't then use: `ssh $host "$(cat);eval collect_on_remote \"${wd}\" ...`

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+}
+
+function collect_on_remote {
+  data_dir=$1
+  num_stack_dumps=$2
+  int_stack_dumps=$3
+  get_all=$4
+  collector_host=$5
+  collector_dir=$6
+  verbose=$7
```
sumwale: double quotes

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+
+  # first get the pid. The latest log file with the header will have the pid
+  host=`hostname`
+  if [ ! -d $data_dir ]; then
```
sumwale: double quotes (not necessary for host but will be good to have for $data_dir)

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+    echo "${data_dir} not found on host: ${host}"
+  fi
+
+  cd $data_dir
```
sumwale: double quotes

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+  files=()
+  last_restart_log=""
+  for l in $( ls -t *.log )
+  do
+    copyright_headers=`grep 'Copyright (C)' ${l}`
```
sumwale: I think there are some variants here in our previous releases. In some the unicode copyright symbol was used instead of `(C)`. Maybe a little more generic pattern: `grep 'Copyright [ ]*([ ]*.[ ]*)'`

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+    if [ ! -z "$copyright_headers" ]; then
+      # also check for the pid line and get the pid
+      proc_id=`sed -n 's/.*Process ID: \([0-9]\+\)$/\1/p' ${l}`
+      if [ "${verbose}" = "1" ]; then
+        echo "Addiing file ${l} to the array"
```
sumwale: typo

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+      fi
+      files+=($l)
+      last_restart_log=$l
```
sumwale: double quote

snappy-tools/sbin/collect-debug-artifacts.sh
sumwale: Probably Soubhik means the latest file in addition to the one containing copyright+settings information. However, that will be better served by allowing for a time interval to grab the files (which Soubhik has mentioned later I think).

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+    kill -URG $proc_id
+    sleep $int_stack_dumps
+  done
+
+  if [ "${get_all}" = "1" ]; then
+    if [ "${verbose}" = "1" ]; then
+      rsync -avrz $data_dir "${collector_host}:${collector_dir}/"
+    else
+      rsync -avrzq $data_dir "${collector_host}:${collector_dir}/"
+    fi
```
sumwale: Won't it be better to tar gzip it and then ship? That can be provided to us directly instead of user having to tar it:

```sh
tar cvzf "${data_dir}.tar.gz" "$data_dir" && \
rsync -av "${data_dir}.tar.gz" "${collector_host}:${collector_dir}/" && \
rm -f "${data_dir}.tar.gz"
```

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+      files+=($l)
+      if [ "${l}" = "${last_restart_log}" ]; then
+        break
+      fi
+    done
+    # get the latest stats file
+    for l in $( ls -t *.gfs 2> /dev/null )
+    do
+      if [ "${verbose}" = "1" ]; then
+        echo "Addiing file ${l} to the array"
```
sumwale: typo

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+      fi
+      files+=($l)
```
sumwale: Is the idea to get only the latest gfs file? In that case this should break out immediately after this. Time range based will be best though.

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+  fi
+  if [ "${verbose}" = "1" ]; then
+    echo "FILES=${files[@]} will be rsynced to ${collector_host}:$collector_dir}"
+  fi
+  if [ "${verbose}" = "1" ]; then
+    rsync -avz ${files[@]} "${collector_host}:${collector_dir}"
+  else
+    rsync -avzq ${files[@]} "${collector_host}:${collector_dir}"
+  fi
+}
```
sumwale: Again tarball here upfront will be better:

```sh
tar cvzf "${data_dir}_files.tar.gz" "${files[@]}" && \
rsync -av "${data_dir}_files.tar.gz" "${collector_host}:${collector_dir}/" && \
rm -f "${data_dir}_files.tar.gz"
```

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+      files+=($l)
+    done
+  fi
+  if [ "${verbose}" = "1" ]; then
+    echo "FILES=${files[@]} will be rsynced to ${collector_host}:$collector_dir}"
+  fi
+  if [ "${verbose}" = "1" ]; then
+    rsync -avz ${files[@]} "${collector_host}:${collector_dir}"
+  else
+    rsync -avzq ${files[@]} "${collector_host}:${collector_dir}"
```
sumwale: Both "v" and "q"? Both can be removed.

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,230 @@
...
+}
+
+check_configs
+# Assuming each line in the members info file has the following format
+# host pid cwd
+
+all_pids=()
+# Make output directory
+rm -rf debug_data 2> /dev/null
+mkdir debug_data
```
sumwale: Suffix with some uid:
```sh
collector_dir="$(mktemp --tmpdir="`pwd`" debug_data.XXXXXX)"
```

cluster/sbin/snappy-nodes.sh
```diff
@@ -141,6 +143,8 @@ function startComponent() {
     ssh $SPARK_SSH_OPTS "$host" $"${@// /\\ } ${args}" < /dev/null \
       2>&1 | sed "s/^/$host: /"
   fi
+
+  echo "${host} ${dirfolder}" >> $MEMBERS_FILE
```
sumwale: I think it will be better for this to into work dir instead of current dir (i.e. "$SPARK_HOME/work")

cluster/sbin/snappy-start-all.sh
```diff
@@ -23,6 +23,11 @@
 sbin="`dirname "$0"`"
 sbin="`cd "$sbin"; pwd`"
 
+MEMBERS_FILE=members.txt
+
+rm -rf $MEMBERS_FILE 2> /dev/null
+touch $MEMBERS_FILE
```
sumwale: see comment about having this in work dir

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,273 @@
...
+    echo "Config file ${CONF_FILE} does not exist"
+    exit 1
+  fi
+
+  source $CONF_FILE
+
+  if [ ! -f "${MEMBERS_FILE}" ]; then
+    echo "members file ${MEMBERS_FILE} does not exist"
+    exit 1
+  fi
```
sumwale: see comment about having this in work dir

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,273 @@
...
+function collect_on_remote {
+  data_dir="$1"
+  num_stack_dumps="$2"
+  int_stack_dumps="$3"
+  get_all="$4"
+  collector_host="$5"
+  collector_dir="$6"
+  verbose="$7"
+
+  tmp_dir="$(mktemp -d --tmpdir="`pwd`" data.XXXX)"
```
sumwale: perhaps you want this relative to data_dir?

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,273 @@
...
+    echo "FAILED TO CREATE tmp dir on ${host} at ${data_dir} with errno ${retval}"
+  fi
+
+  # first get the pid. The latest log file with the header will have the pid
+  host=`hostname`
+  if [ ! -d "$data_dir" ]; then
+    echo "${data_dir} not found on host: ${host}"
+  fi
+
+  cd "$data_dir"
```
sumwale: since this has cded into data_dir won't usages of tmp_dir will cause trouble below (unless tmp_dir itself is created inside data_dir as mentioned in the comment above (i.e. --tmpdir="$data_dir")

snappy-tools/sbin/collect-debug-artifacts.sh
```diff
@@ -0,0 +1,273 @@
...
+    for l in $( ls *.log 2> /dev/null )
+    do
+      files+=($l)
+    done
+    for l in $( ls *.gfs 2> /dev/null )
+    do
+      files+=($l)
+    done
+  else
+    logs_sorted_reverse=`ls *.log | sed 's/\([0-9]\)/;\1/' | sort -r -n -t\; -k2,2 | tr -d ';'`
```
sumwale: This will replace digits if they happen to be part of filename. I think a better pattern will be: `sed 's/[0-9]\+\(-[0-9]\+\)\?\.log$/;\0/'`

sumwale:
Looks good. See few comments (hopefully last ones :-)

PR #204, opened 2016-03-21
SNAP-584

dshirish:
Modified snappy-tools/sbin/snappy-start-all.sh not to start lead node when rowstore arg is passed as discussed earlier.

Added a test in QueryRoutingDUnitTest.scala to confirm that GemXD URL works.

[Update]: As discussed updated the scripts to use GfxdLauncher when started in rowstore mode.

@kneeraj @sumwale please take a look.


snappy-tools/sbin/snappy-start-all.sh
```diff
@@ -34,4 +34,6 @@ sbin="`cd "$sbin"; pwd`"
 "$sbin"/snappy-servers.sh start
 
 # Start Leads
-"$sbin"/snappy-leads.sh start
+if [ "$1" != "rowstore" ]; then
+  "$sbin"/snappy-leads.sh start
```
kneeraj: For stopping if we maintain that the cluster was started with rowstore option then skip stopping the lead node.

kneeraj:
Looks good to me.

dshirish:
@sumwale 
As discussed updated the scripts to use GfxdLauncher when started in rowstore mode.
Please review.

snappy-tools/bin/snappy-shell
```diff
@@ -27,4 +27,19 @@ then
...
+
+function setnewargs() { newargs="$@"; }
+
+if echo $@ | grep -qw "rowstore"; then
+  #using rowstore launcher
+  newargs=
+  for arg in "$@"; do
+    if [[ "$arg" != "rowstore" ]] ; then
+      setnewargs "$newargs" "$arg"
+    fi
```
sumwale: Won't this be better as "-rowstore"? Latter is our normal convention for other options while "start/stop/status" denote actions to be done.

snappy-tools/sbin/snappy-start-all.sh
sumwale: See above comment about "-rowstore"

sumwale:
Looks good. I had added comment about "-rowstore" but removed it since "snappy-shell rowstore start" also seems fine.

PR #206, opened 2016-03-28
Snap 634

rishitesh:
Modified code of LocalJoin for better performance. In Local join each partition of Streamed RDD is joined against a HashRelation computed from replicated table.  For each partition the replicated table was scanned & a HashedRelation was made out of it.  It's a costly operation.

Changes : Cached the computed HashedRelation with Spark BlockManager. This will ensure faster execution of replicated table joins. As the cached data is with Spark BlockManager it will be evicted in LRU fashion.


rishitesh:
@sumwale Please review.

snappy-core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -66,48 +73,108 @@ case class LocalJoin(leftKeys: Seq[Expression],
     }
     val numOutputRows = longMetric("numOutputRows")
 
-    narrowPartitions(buildPlan.execute(), streamedPlan.execute(), true) {
-      (buildIter, streamIter) => {
-        val hashed = HashedRelation(buildIter, numBuildRows, buildSideKeyGenerator)
+    val h1 = (buildIter: Iterator[InternalRow]) => Iterator[HashedRelation] {
```
hbhanawat: can we rename h1 to something else?

snappy-core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -66,48 +73,108 @@ case class LocalJoin(leftKeys: Seq[Expression],
...
+    ) extends RDD[HashedRelation](sc, Seq(new OneToOneDependency(buildRDD))) {
+
+  override def compute(s: Partition, context: TaskContext): Iterator[HashedRelation] = {
+
+    val blockId = RDDBlockId(this.id, s.index)
+
+    val rel1 = SparkEnv.get.blockManager.getSingle(blockId) match {
+      case Some(x) => x.asInstanceOf[HashedRelation]
+      case None => {
+        HashRelationRDD.synchronized {
```
hbhanawat: Since all threads wait until the first task has created a HashedRelation, I doubt that this will result in any performance gain on clusters where enough cores are available to process all tasks simultaneously. What do you think? We should have a strategy to create hashedRelation beforehand on replicated tables. Something like index relation that we are working on for column tables.

snappy-core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -66,48 +73,108 @@ case class LocalJoin(leftKeys: Seq[Expression],
...
+          SparkEnv.get.blockManager.getSingle(blockId) match {
```
hbhanawat: Will this block be visible in UI in storage tab?

snappy-core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -66,48 +73,108 @@ case class LocalJoin(leftKeys: Seq[Expression],
...
+            case Some(x) => x.asInstanceOf[HashedRelation]
+            case None => {
+              val hashedRelation = f(buildRDD.iterator(s, context))
+              val rel = hashedRelation.next()
+
+              SparkEnv.get.blockManager.putSingle(
+                blockId, rel, StorageLevel.MEMORY_AND_DISK_SER, tellMaster = false)
```
hbhanawat: Given that block ids are based on partition id, can they clash when more than one replicated tables are being cached?

snappy-core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -66,48 +73,108 @@ case class LocalJoin(leftKeys: Seq[Expression],
...
+      case Some(x) => x.asInstanceOf[HashedRelation]
+      case None => {
+        HashRelationRDD.synchronized {
+          SparkEnv.get.blockManager.getSingle(blockId) match {
+            case Some(x) => x.asInstanceOf[HashedRelation]
+            case None => {
+              val hashedRelation = f(buildRDD.iterator(s, context))
+              val rel = hashedRelation.next()
+
+              SparkEnv.get.blockManager.putSingle(
```
hbhanawat: who is removing this block from memory after the query is done?

rishitesh: BlockManager removes based on LRU and time to live. A background cleaner thread does it.

snappy-core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
rishitesh: No, As each query creates a new RDD, it will be a new RDD id + partition ID.

snappy-core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
rishitesh: Have not checked it.. Will check and update.

snappy-core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
rishitesh: Will do it.

snappy-core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
rishitesh: Yes pre-creating will help. But we cant do that as replicated tables are mutable and subject to change. Even though all threads wait for HashedRelation to be created on each node , the join operation becomes significantly faster. And we have to keep in mind that we will be supporting concurrent queries in future , which itself will leave our query with fewer cores.

sumwale: Also I think other cores should be busy in processing other parts of the query (or not)? Ideally we could parallelize this on the node by having multiple threads scan the hashtable of the region with each thread scanning one (or more) segment of the CHM instead of waiting on the lock. This will require a bit of internal region APIs (or extension to the "SELECT" with a hint to scan only some segments). Can be noted for future extension.

snappy-core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -66,48 +73,108 @@ case class LocalJoin(leftKeys: Seq[Expression],
...
+    val rel1 = SparkEnv.get.blockManager.getSingle(blockId) match {
+      case Some(x) => x.asInstanceOf[HashedRelation]
+      case None => {
+        HashRelationRDD.synchronized {
+          SparkEnv.get.blockManager.getSingle(blockId) match {
+            case Some(x) => x.asInstanceOf[HashedRelation]
+            case None => {
+              val hashedRelation = f(buildRDD.iterator(s, context))
+              val rel = hashedRelation.next()
+
```
sumwale: Why returning an iterator here when it is always going to be a single object?

snappy-core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
rishitesh: Also I think other cores should be busy in processing other parts of the query (or not)? [Rishi] I need to change the order of iterator , so that scanning of partition table will be first. Then the hashed relation. This will allow threads to go parallel without waiting. First thread finishing partitioned scan will also scan the replicated table and cache it. Others when finished with their scanning need to wait till HashedRelation is available.

sumwale: We need concurrent scanning as a general perf improvement for replicate table queries in any case. Currently any query on those will be single threaded.

About other parts of query, I meant whether in a multi-way join like TPCH ones can other threads work on other parts of the query (joins, filters etc) or you think all will have to be doing LocalJoin only?

PR #207, opened 2016-03-29
SNAP-643: support complex types (except UDT) for column tables

sumwale:
* support for complex types when creating column tables with mapping of these to BLOB
    types in the row buffer table (in GemFireXDDialect.scala)
  * use the existing code generation of Spark for UnsafeArrayData, UnsafeMapData, UnsafeRow
    to convert complex types into byte[] efficiently; for this purpose all the calls
    that set columns in PreparedStatement are now done via code generation to make it both
    efficient and work easily with the mentioned code generation piece of Spark;
    the code now is in sql/store/CodeGeneration.scala to set the values in PreparedStatement
    reading from InternalRows; for public Row objects, the WrappedRow mentioned below is used
  * consolidate saveTable/savePartition calls inside jdbcExtensions.scala with upsert support and
    use those consistently everywhere (which has support for complex types via code generation mentioned above)
  * avoid double Row->InternalRow conversions and back with CatalystTypeConverters in multiple places
  * cleanup executor code generation cache on SnappyContext.stop and individual cleanup when table is dropped
  * added a WrappedRow class that wraps a Row exposing an InternalRow; this is used
    to pass along Row objects for the code generation above without having to copy
    columns or write similar code generation for Rows
  * added support for BinaryType in query routing of SparkSQLExecuteImpl to expose it as a BLOB

@hbhanawat This has the code generation changes for all other types (required it for Array/Struct/Map types to reuse Spark serialization code for those).

@suranjan @soubhik-c @rishitesh please take a look too


snappy-core/src/main/scala/org/apache/spark/sql/store/CodeGeneration.scala
```diff
@@ -0,0 +1,261 @@
...
+        s"""
+          if (!row.isNullAt($col)) {
+            $nonNullCode
+          } else {
+            $nullCode
+          }
+        """)
+    }
+    val evaluator = new CompilerFactory().newScriptEvaluator()
+    evaluator.setClassName("io.snappydata.execute.GeneratedEvaluation")
```
hbhanawat: Same class name for all the statements? Classloader playing some role here?

hbhanawat:
Looks good.

snappy-core/src/main/scala/org/apache/spark/sql/store/CodeGeneration.scala
sumwale: Yes, janino uses different classloader for each generated class. I guess this also helps with their GC with <=JDK7 that used PermGen (which used to be a problem with Derby generated classes).

snappy-core/src/main/scala/org/apache/spark/sql/store/CodeGeneration.scala
```diff
@@ -0,0 +1,261 @@
...
+  override def hashCode(): Int = name.hashCode
+
+  override def equals(other: Any): Boolean = other match {
+    case o: ExecuteKey => name == o.name
+    case s: String => name == s
+    case _ => false
+  }
+}
+
+object CodeGeneration extends Logging {
```
rishitesh: Sumedh , why we could not follow the same pattern as that of Spark's CodeGenerator. Any custom generator needs to extend CodeGenerator.  Was there any specific reason to bypass that ?

rishitesh: Is it because we wanted to avoid code string creation in case of a cache hit ?

sumwale: Yes, Spark's current way will generate code separately for every plan creation. This takes quite a bit of time so didn't want to do that and instead let it be per table based. Can change to be plan based for SnappyContext.update if we are not worried about the performance of that.

Ideally we should be going the whole way and caching the whole plan (with code generation) per PreparedStatement (or statement with tokenization) like done in snappy-store. Then it also requires a mechanism to invalidate plans (we added some start of that already in DependencyCatalog). It will be a good idea to contribute back to Spark but will require more work for proper plan caching which is what we should be doing ultimately.

PR #210, opened 2016-04-05
Snap 646: Unified Partitioner For Spark & Snappy Store

rishitesh:


rishitesh:
@kneeraj @suranjan @hbhanawat @soubhik-c Please review.

snappy-tools/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
```diff
@@ -0,0 +1,80 @@
...
+          case sb: SQLBoolean => if (sb.getBoolean) 0 else 1
+          case sd: SQLDate => sd.getDate(Calendar.getInstance()).hashCode()
+          case sd: SQLBit => java.util.Arrays.hashCode(sd.getBytes())
+          case sf: SQLReal => java.lang.Float.floatToIntBits(sf.getFloat)
+          case clob: SQLClob => utfStringHashCode(clob.getString.getBytes("utf-8")) //@TODO
+          // Inefficient, we should use character array
+          case varchar: SQLVarchar => utfStringHashCode(varchar.getString.getBytes("utf-8"))
+          case time: SQLTimestamp => time.getTimestamp(Calendar.getInstance()).hashCode()
+          case decimal :SQLDecimal => decimal.getObject().hashCode()
+          case other => other.hashCode()
```
sumwale: As we discussed, all these methods should go into respective SQL* implementations (with base in DataType that can do a getObject().hashCode()) that can efficiently implement for each type without deserialization where possible (starting with implementation SQLChar). This can then become the default instead of existing computeHashCode() for isSnappyStore so it will retain the current behaviour for existing customers that boot it as rowstore.

snappy-tools/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
```diff
@@ -0,0 +1,80 @@
...
+        result = 31 * result + a(i)
+      }
+      ({
+        i += 1;
+        i - 1
+      })
+    }
+
+    return result
+  }
```
sumwale: We can do this in one pass from raw data in SQLChar without first deserializing into String, then serializing back to UTF-8.

snappy-tools/src/main/scala/org/apache/spark/sql/store/StorePartitioner.scala
```diff
@@ -39,3 +40,7 @@ class ColumnPartitioner (partitions: Int) extends Partitioner {
   override def hashCode: Int = numPartitions
 
 }
+
+
+
+
```
sumwale: unneeded newlines

snappy-tools/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -211,7 +211,7 @@ object StoreUtils extends Logging {
                 throw new DDLException("Column table cannot be partitioned on" +
                     " PRIMARY KEY as no primary key")
               }
-            case _ => s"COLUMN ($v) "
+            case _ =>  s"(${v.replace(",", "+")})"
```
sumwale: why this replace by "+"?

sumwale:
Will review tomorrow. Please note that couple of tests are failing in SnappyJoinSuite after my changes (to use upstream Spark) due to this. Two row table joins with different number of partitions is doing exchange for one but not for other (which is proper) but the join gives wrong results because of different hashing. I will mark the two tests "More than two table joins", "Row PR table join with PR Table with unequal partitions", as ignored. Please uncomment those two after this merge.

rishitesh:
yes. I am aware of that issue. As Spark has added a feature to avoid
unnecessary shuffle in Exchange.scala, we are getting this problem, With
this PR this problem is gone.

Regards,
Rishitesh Mishra,
SnappyData . (http://www.snappydata.io/)

https://in.linkedin.com/in/rishiteshmishra

core/src/main/scala/org/apache/spark/sql/CatalystHashFunction.scala
```diff
@@ -0,0 +1,96 @@
...
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+
+package org.apache.spark.sql
+
+/**
+ * This class acts as a public interface to the hashcode logic implemented at catalyst layer.
+ * It ensures Spark's partitioning and store's partitioning follows the same logic.
+ * This helps in reducing shuffle operations when Spark's datafarme is joined with store data.
```
sumwale: Typo

core/src/main/scala/org/apache/spark/sql/CatalystHashFunction.scala
```diff
@@ -0,0 +1,96 @@
...
+    val numBytes = a.length
+
+    var i: Int = 0
+    while (i < numBytes) {
+      {
+        result = 31 * result + a(i)
+      }
+
+      i += 1;
+      i - 1
```
sumwale: Useless i - 1

core/src/main/scala/org/apache/spark/sql/CatalystHashFunction.scala
```diff
@@ -0,0 +1,96 @@
...
+          case b: Byte => b.toInt
+          case s: Short => s.toInt
+          case i: Int => i
+          case l: Long => (l ^ (l >>> 32)).toInt
+          case f: Float => java.lang.Float.floatToIntBits(f)
+          case d: Double =>
+            val b = java.lang.Double.doubleToLongBits(d)
+            (b ^ (b >>> 32)).toInt
+          case a: Array[Byte] => java.util.Arrays.hashCode(a)
+          case str: java.lang.String => utfStringHashCode(str.getBytes("utf-8"))
```
sumwale: As I commented earlier we need to add a method that can calculate string hash without serialisation (or deserialisation). This will likely have big impact on performance.

sumwale: Can be tracked in separate issue.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
```diff
@@ -125,6 +128,14 @@ object StoreCallbacksImpl extends StoreCallbacks with Logging with Serializable
     schemas.add(ColumnFormatRelation.INTERNAL_SCHEMA_NAME)
     schemas
   }
+
+  override def getHashCodeSnappy(dvd: scala.Any): Int = {
+    partioner.hashValue(dvd)
```
sumwale: See comment above. Definitely need method at least for string to calculate hash without deserialisation. Right now it is deserializing to create string then serializing again in the method body.

core/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
```diff
@@ -0,0 +1,91 @@
...
+          result = 31 * result + b
+        }
+        else {
+          b = (0xC0 | ((c >> 6) & 0x1F)).toByte
+          result = 31 * result + b
+          b = (0x80 | ((c >> 0) & 0x3F)).toByte
+          result = 31 * result + b
+        }
+      }
+    }
```
sumwale: This is modified UTF8 used by DataOutput and GemXD. Incompatible with full UTF8 for 3-4byte encodings. You should be able to find code for full UTF8 on net.

core/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
```diff
@@ -0,0 +1,91 @@
...
+          case l: Long => (l ^ (l >>> 32)).toInt
+          case f: Float => java.lang.Float.floatToIntBits(f)
+          case d: Double =>
+            val b = java.lang.Double.doubleToLongBits(d)
+            (b ^ (b >>> 32)).toInt
+          case a: Array[Byte] => java.util.Arrays.hashCode(a)
+          case str: java.lang.String => utfStringHashCode(str.getBytes("utf-8"))
+          //Custom type checks for Store
+          case sb: SQLBoolean => if (sb.getBoolean) 0 else 1
+          case sd: SQLDate => sd.getDate(Calendar.getInstance()).hashCode()
```
sumwale: I think this is wrong. InternalRow stores date as int.

core/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
```diff
@@ -0,0 +1,91 @@
...
+          case i: Int => i
+          case l: Long => (l ^ (l >>> 32)).toInt
+          case f: Float => java.lang.Float.floatToIntBits(f)
+          case d: Double =>
+            val b = java.lang.Double.doubleToLongBits(d)
+            (b ^ (b >>> 32)).toInt
+          case a: Array[Byte] => java.util.Arrays.hashCode(a)
+          case str: java.lang.String => utfStringHashCode(str.getBytes("utf-8"))
+          //Custom type checks for Store
+          case sb: SQLBoolean => if (sb.getBoolean) 0 else 1
```
sumwale: Reverse?

core/src/main/scala/org/apache/spark/sql/CatalystHashFunction.scala
```diff
@@ -0,0 +1,96 @@
...
+ */
+class CatalystHashFunction {
+
+  def computeHash(key: Any): Int = {
+    val update: Int =
+      if (key == null) {
+        0
+      } else {
+        key match {
+          case b: Boolean => if (b) 0 else 1
```
sumwale: Reverse?

core/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
```diff
@@ -0,0 +1,91 @@
...
+          case sd: SQLDate => sd.getDate(Calendar.getInstance()).hashCode()
+          case sd: SQLBit => java.util.Arrays.hashCode(sd.getBytes())
+          case sf: SQLReal => java.lang.Float.floatToIntBits(sf.getFloat)
+          case clob: SQLClob => hashClob(clob.getCharArray(true))
+          case varchar: SQLVarchar => hashClob(varchar.getCharArray(true))
+          case time: SQLTimestamp => time.getTimestamp(Calendar.getInstance()).hashCode()
```
sumwale: Timestamp is long in InternalRow

core/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
```diff
@@ -0,0 +1,91 @@
...
+          case decimal: SQLDecimal => decimal.getObject().hashCode()
```
sumwale: We do want optimised non deserializing version for decimals too. Can be tracked separately.

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -190,7 +190,7 @@ object StoreUtils extends Logging {
       val primaryKey = {
         v match {
           case PRIMARY_KEY => ""
-          case _ => s"$PRIMARY_KEY ($v, $SHADOW_COLUMN_NAME)"
+          case _ => s"$PRIMARY_KEY ($SHADOW_COLUMN_NAME)"
```
sumwale: Why this change? Don't understand the PK setting here.

core/src/test/scala/org/apache/spark/sql/store/UnifiedPartitionerTest.scala
```diff
@@ -0,0 +1,350 @@
...
+        "PERSISTENT 'ASYNCHRONOUS')")
+
+    val regtwo: Region[_, _] = Misc.getRegion("/APP/COLUMNTABLE1", true, true)
+    val rattr: RegionAttributes[_, _] = regtwo.getAttributes
+    val pr: PartitionResolver[_, _] = rattr.getPartitionAttributes.getPartitionResolver
+    val rpr: GfxdPartitionByExpressionResolver = pr.asInstanceOf[GfxdPartitionByExpressionResolver]
+    assert(rpr != null)
+
+    // Check All Datatypes
+    val row = new GenericMutableRow(1)
```
sumwale: Is it row we need to compare against or InternalRow? Exchange partitioning is on InternalRow. See comments about date, timestamp above.

core/src/test/scala/org/apache/spark/sql/store/UnifiedPartitionerTest.scala
```diff
@@ -0,0 +1,350 @@
...
+    dvd = new SQLReal(10.5F)
+    row.update(0, 10.5F)
+    assert(rpr.getRoutingKeyForColumn(dvd) == row.hashCode)
+
+    dvd = new SQLLongint(479L)
+    row.update(0, 479L)
+    assert(rpr.getRoutingKeyForColumn(dvd) == row.hashCode)
+
+    dvd = new SQLVarchar("xxxx");
+    row.update(0, UTF8String.fromString("xxxx")) // As catalyst converts String to UtfString
```
sumwale: Let's just use mutable internalrow throughout this test. Then you would have discovered issues with date, timestamp and possibly others I may have missed.

core/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
rishitesh: Yes Spark does it like this. Ours is reverse.

core/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
rishitesh: InternalRow takes epoch time . SQLTimestamp hashCode() also uses the same to compute hashCode.

core/src/test/scala/org/apache/spark/sql/store/UnifiedPartitionerTest.scala
rishitesh: Will change.

core/src/test/scala/org/apache/spark/sql/store/UnifiedPartitionerTest.scala
rishitesh: Will change it to InternalRow.

core/src/main/scala/org/apache/spark/sql/CatalystHashFunction.scala
rishitesh: Added SNAP-710

core/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
rishitesh: Added SNAP-711

core/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
sumwale: Ok, but avoid creating Timestamp object here. What you need is this fragement from getTimestamp:
```
    public long getEpochTime() {
      if (cal == null) {
          cal = ClientSharedData.getDefaultCalendar();
      }
      setCalendar(cal);
      return cal.getTimeInMillis();
    }
```

Also this and similar methods can return a null. Is null handling being done elsewhere?

core/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
sumwale: For example see Charsets_toUtf8Bytes here: https://android.googlesource.com/platform/libcore/+/jb-mr2-release/luni/src/main/native/java_nio_charset_Charsets.cpp

(note: we can't use OpenJDK code due to licensing issues; android or jsr166 usually provide good replacements)

sumwale: Similarly see the UTF8 encoder around line 250 here: https://android.googlesource.com/platform/libcore/+/jb-mr2-release/luni/src/main/java/java/lang/String.java

core/src/main/scala/org/apache/spark/sql/store/StoreHashFunction.scala
rishitesh: I missed the null handling . Will add.

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
rishitesh: Will revert it back . Not necessary

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -162,7 +162,11 @@ object StoreUtils extends Logging {
       val primaryKey = {
         v match {
           case PRIMARY_KEY => ""
-          case _ => s"$PRIMARY_KEY ($v, $SHADOW_COLUMN_NAME)"
+          //case _ => s"$PRIMARY_KEY ($v, $SHADOW_COLUMN_NAME)"
+          case _ => s"$PRIMARY_KEY ($SHADOW_COLUMN_NAME)" // Removing partition column for
+          // primary key. This has been done as we dont have support for CLOB in primary key in
+          // desrailized form. Once that's fixed we should uncomment the above line and comment
+          // this line
```
sumwale: We should keep it for cases other than LOB and complex types (CLOB, BLOB, ArrayType, MapType, StructType, UDT).

Also we need to find a way to carry through VARCHAR at Spark layer to here. Right now VARCHAR just gets through as a StringType which gets mapped to CLOB in GemXD. Clean way would have been to add a new DataType at our end, but then it could be used anywhere in Spark code and we would be in trouble. So we will need to have it as a private DataType which somehow can be used only in createTable, and that gets changed to a StringType once we have done the mapping.

rishitesh: First of all we probably don't support partition by complex type( BLOB, ArrayType, MapType, StructType, UDT)  at Gem layer, or do we ? Only remaining thing is CLOB, which we map if  StringType is mentioned in column table. There is no equivalent of CLOB at Spark side . 

What about user partitions based on  (StringType, IntegerType) columns ? We would still like to skip adding them to PK.

Agree to 2nd point for VARCHAR having a first class support at Spark layer or atleast a mechanism to put some metadata to indicate it as a VARCHAR.

sumwale: Yes, hash partitioning does support LOBs/UDTs just fine because it uses a hashCode which is implemented for all types.

> There is no equivalent of CLOB at Spark side.

StringType is the equivalent because it provides unlimited size strings.

> What about user partitions based on (StringType, IntegerType) columns?

As discussed in email chain, we want to keep these for now otherwise it will end up creating a global index that can have a major insert performance hit.

sumwale: @rishitesh Can we track the VARCHAR issue in JIRA (or is it already being tracked)?

rishitesh: Created one issue SNAP-735

rishitesh:
I have incorporated all review comments .. Would you like to check it again ?

PR #212, opened 2016-04-07
Snap 626

nthanvi:

1. Added a new service that read row and column table size from store at regular interval.

2. the column table and row table value count is provided to sizeInBytes field so that query optimization can calculate the query cost correctly

3. Added a new tab "Snappy" in spark UI tab and it show rows and column tables sizes  from snappy store.
rows table shows only the table name and size
column table shows table name , row  buffer size, column table size and total size

## Patch testing
added a unit test

## Other PRs 

No


nthanvi:
@rishitesh  - please review

snappy-core/src/main/scala/org/apache/spark/sql/SnappyAnalyticsService.scala
```diff
@@ -0,0 +1,154 @@
...
+				val entrySize = convertToBytes(rs.getString(2))
+				val keySize = convertToBytes(rs.getString(3))
+				val valueSize = convertToBytes(rs.getString(4))
+				val valueSizeOffHeap = convertToBytes(rs.getString(5))
+				val totalSize = convertToBytes(rs.getString(6))
+				currentTableStats.put(name,
+					new MemoryAnalytics(entrySize, keySize, valueSize, valueSizeOffHeap, totalSize))
+			}
+
+			readWriteLock.executeInWriteLock(tableStats = currentTableStats)
```
rishitesh: Why we need this lock?

nthanvi: while this service is updating tableStats structure other threads like from UI or  query cost calculator might be reading these stats.

rishitesh: Cant it be a concurrent map? I think UI and cost calculator can manage with a little old data. If we can avoid the lock overhead it will be good.

snappy-core/src/main/scala/org/apache/spark/sql/SnappyAnalyticsService.scala
```diff
@@ -0,0 +1,154 @@
...
+			rs.close()
+			connection.close()
```
rishitesh: We should move it to some finally block. rs.close() and conn.close()

snappy-core/src/main/scala/org/apache/spark/sql/SnappyAnalyticsService.scala
```diff
@@ -0,0 +1,154 @@
...
+
+	private def convertToBytes(value: String): Long = (value.toDouble * 1024).toLong
+
+	private def getRowBufferName(columnStoreName: String): String = {
+		DEFAULT_SCHEMA +
+				(columnStoreName.replace(SHADOW_TABLE_SUFFIX, "").
+						replace(INTERNAL_SCHEMA_NAME, "")).trim
+	}
+
+	private def isColumnTable(tablename: String): Boolean =
```
rishitesh: This class needs information about row table & column table. Can't we move it to snappy-tool or once Sumedh's checkin comes in we can decide.

snappy-core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -47,6 +47,15 @@ object Constant {
...
   val DEFAULT_USE_HIKARICP = false
+
+  // Interval in ms  to run the SnappyAnalyticsService
+  val DEFAULT_ANALYTICS_SERVICE_INTERVAL: Long = 5000
+
+  // Internal Column table store schema
+  final val INTERNAL_SCHEMA_NAME = "SNAPPYSYS_INTERNAL"
+
+  // Internal Column table store suffix
+  final val SHADOW_TABLE_SUFFIX = "_COLUMN_STORE_"
```
rishitesh: We need to discuss these constants , once Sumedh's refactoring is in. Right now snappy-core does not look like a place to have these constants or any reference to Column & Row table.

snappy-core/src/main/scala/org/apache/spark/sql/SnappyAnalyticsService.scala
```diff
@@ -0,0 +1,154 @@
...
+					new MemoryAnalytics(entrySize, keySize, valueSize, valueSizeOffHeap, totalSize))
+			}
+
+			readWriteLock.executeInWriteLock(tableStats = currentTableStats)
+			rs.close()
+			connection.close()
+		}
+	}
+
+	private def convertToBytes(value: String): Long = (value.toDouble * 1024).toLong
```
rishitesh: put a null check for value.

snappy-core/src/main/scala/org/apache/spark/sql/SnappyAnalyticsService.scala
```diff
@@ -0,0 +1,154 @@
...
+				(columnStoreName.replace(SHADOW_TABLE_SUFFIX, "").
+						replace(INTERNAL_SCHEMA_NAME, "")).trim
+	}
+
+	private def isColumnTable(tablename: String): Boolean =
+		tablename.startsWith(INTERNAL_SCHEMA_NAME) && tablename.endsWith(SHADOW_TABLE_SUFFIX)
+
+
+	private def getValue(tableName: String, isColumnTable: Boolean = false): Long = {
+		val defaultStats = new MemoryAnalytics(0, 0, 0, 0, 0)
```
rishitesh: val defaultStats = new MemoryAnalytics(0, 0, 0, 0, 0)
Make it a constant. This will  avoid creating an object everytime getValue() is called.

snappy-core/src/main/scala/org/apache/spark/sql/SnappyAnalyticsService.scala
```diff
@@ -0,0 +1,154 @@
...
+		DEFAULT_SCHEMA +
+				(columnStoreName.replace(SHADOW_TABLE_SUFFIX, "").
+						replace(INTERNAL_SCHEMA_NAME, "")).trim
+	}
+
+	private def isColumnTable(tablename: String): Boolean =
+		tablename.startsWith(INTERNAL_SCHEMA_NAME) && tablename.endsWith(SHADOW_TABLE_SUFFIX)
+
+
+	private def getValue(tableName: String, isColumnTable: Boolean = false): Long = {
```
rishitesh: getValue() sounds a generic name and suggests all kind of metrics can be obtained, where as it only returns size. If possible declare constants for metrics and pass the required metrics to this method. 

private def getValue(tableName: String, isColumnTable: Boolean = false, metricsName : String):

snappy-core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -796,17 +797,21 @@ object GlobalSnappyInit {
         // prior to `new SnappyContext(sc)` after this
         // method ends.
         ToolsCallbackInit.toolsCallback.invokeLeadStartAddonService(sc)
+        SnappyAnalyticsService.start(sc)
```
rishitesh: We can put another class like SnappyDaemons.start() . Inside it we can start SnappyAnalyticService. As tomorrow this list might grow :)

nthanvi:
@rishitesh  - handle all the review comments.

core/src/main/scala/io/snappydata/SnappyAnalyticsService.scala
```diff
@@ -0,0 +1,196 @@
+/*
```
rishitesh: Please change the file name to SnappyDaemons,scala.

PR #214, opened 2016-04-12
Snap 659

hbhanawat:

1. Adding API for creating indexes of Column tables. This required changes in the row tables index API to be consistent. However there has been no change made to the row table index functionality. The API is based on the discussion that we are having in the document https://docs.google.com/document/d/1hwR_JKzvkRMvesr2mlTqIlUahxfSiKjHbuKhM8S-08k/edit#.
2. added missing support for creation of global hash and unique indexes from spark for row tables.
3. added missing support of 'if exists' clause for drop index ddl.
4. added support for column table DDL and APIs.
5. added the functionality of adding index information to the the parent column table. When an index is created on a table, the index table is created which is a column table and then the properties of the main table is updated to store the index information in it.
6. A new IndexColumnFormatRelation is added that derives from ColumnFormatRelation. IndexColumnFormatRelation implements a DependentRelation while ColumnFormatRelation implements ParentRelation. This is done to introduce parent-dependent relationship between the two so that it can be taken care during drop commands. The index name stored in the table properties is used to create parent-child relationship between the two when the table relation is created.
7. Functionality of the drop index for the column table and row table is added.

Next tasks:
Inserts in column tables.
Adding a plan to use index instead of tables for a join.

## Patch testing

Have updated CreateIndexTest to test all the scenarios. 

## Other PRs 
No other associated PRs.


hbhanawat:
@sumwale  @rishitesh  please have a look.

hbhanawat:
@sumwale @rishitesh I fixed the conflicts after code reorganization. Please review now.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -544,41 +543,108 @@ class SnappyContext protected[spark](
...
+        // Main table is updated to store the index information in it. We could have instead used
+        // createIndex method of HiveClient to do this. But, there were two main issues:
+        // a. Schema needs to be added to the HiveTable to supply indexedCols while creating
+        // the index. But, when schema is added to the HiveTable object, it throws a foreign
+        // key constraint failure on the serdes table. Need to look into this.
+        // b. The bigger issue is that the createIndex needs an indexTable for creating this
+        // index. Also, there are multiple things (like implementing HiveIndexHandler)
+        // that are hive specific and can create issues for us from maintenance perspective
+        catalog.alterTableToAddIndexProp(tableIdent, getIndexTable(indexIdent).toString)
+
```
sumwale: Why isn't this encapsulated in JDBCAppendableRelation.createIndex. Then the code here would be:
```
case LogicalRelation(i: IndexableRelation, _) =>
        i.createIndex(indexIdent, tableIdent, indexColumns, options)
```

like before

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -49,6 +49,10 @@ object ExternalStoreUtils {
   final val DEFAULT_SAMPLE_TABLE_BUCKETS = "53"
   final val DEFAULT_TABLE_BUCKETS_LOCAL_MODE = "11"
   final val DEFAULT_SAMPLE_TABLE_BUCKETS_LOCAL_MODE = "7"
+  final val INDEX_TYPE = "INDEX_TYPE"
+  final val PREFIX_INDEX_TABLE = "SNCIDXTBL_"
+  final val INDEX_NAME = "INDEX_NAME"
+
```
sumwale: Lets use a common convention for prefixing so that it can be documented as out of bounds for users. Column tables uses ColumnFormatRelation.INTERNAL_SCHEMA_NAME. So index can use similar internal schema with some suffix to this (e.g. SNAPPYSYS_INTERNAL_INDEX). Then we can document that everything in SNAPPYSYS_INTERNAL* schema is out of bounds.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -544,41 +543,108 @@ class SnappyContext protected[spark](
...
+    * @param indexName Index name which goes in the catalog
+    * @param baseTable Fully qualified name of table on which the index is created.
+    * @param indexColumns Columns on which the index has to be created
+    * @param options Options for indexes. For e.g.
+    *                column table index - ("COLOCATE_WITH"->"CUSTOMER").
+    *                row table index - ("INDEX_TYPE"->"GLOBAL HASH") or ("INDEX_TYPE"->"UNIQUE")
+    */
+  def createIndexOnTable(indexName: String,
+                         baseTable: String,
+                         indexColumns: Seq[String],
```
sumwale: I think OnTable is redundant. Lets lose it. Particularly the "dropIndexOnTable" looks odd.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -544,41 +543,108 @@ class SnappyContext protected[spark](
...
   /**
-   * Drop Index of a SnappyData table (created by a call to createIndexOnTable).
-   */
-  private[sql] def dropIndexOfTable(sql: String): Unit = {
+    * Drops an index on a table
+    * @param indexName Index name which goes in catalog
+    * @param ifExists Drop if exists, else exit gracefully
+    */
+  def dropIndexOnTable(indexName: String, ifExists: Boolean): Unit = {
+
```
sumwale: Just dropIndex should be fine as mentioned before.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -544,41 +543,108 @@ class SnappyContext protected[spark](
...
+    val indexIdent = getIndexTable(catalog.newQualifiedTableName(indexName))
+
+    // Since the index does not exist in catalog, it may be a row table index.
+    if (!catalog.tableExists(indexIdent)) {
+      dropRowStoreIndex(indexName, ifExists)
+    } else {
+      catalog.lookupRelation(indexIdent) match {
```
sumwale: What are the implications? Is it possible to have same index as both row store and column store. I guess with a unique schema prefix in storage, it should not cause trouble in operation. Still it seems a bit of a usability problem that same index name can be for both row and column indexes.

sumwale: We need to track display of these column indexes in "show indexes", "select * from sys.indexes" in snappy-shell.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -544,41 +543,108 @@ class SnappyContext protected[spark](
...
+        case LogicalRelation(dr: DependentRelation, _) =>
+          // Remove the index from the bse table props
+          val baseTableIdent = catalog.newQualifiedTableName(dr.baseTable.get)
+          catalog.alterTableToRemoveIndexProp(baseTableIdent, indexIdent.toString)
+          // Remove the actual index
+          dropTable(indexIdent, ifExists)
```
sumwale: Why not encapsulate this in ParentRelation.removeDependent? Lets keep the code common for dependency tracking for dropTable/dropIndex into a separate method (e.g. in DependencyCatalog.dropDependent or something).

core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCAppendableRelation.scala
```diff
@@ -40,19 +40,20 @@ import org.apache.spark.sql.types.StructType
...
+ origOptions: Map[String, String],
+ externalStore: ExternalStore,
+ @transient override val sqlContext: SQLContext)
+  extends BaseRelation
+  with PrunedFilteredScan
+  with InsertableRelation
+  with DestroyRelation
+  with IndexableRelation
+  with Logging
+  with Serializable {
```
sumwale: indentation change? Original was more in line with Spark's convention.

core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCAppendableRelation.scala
```diff
@@ -276,6 +277,13 @@ case class JDBCAppendableRelation(
   def flushRowBuffer(): Unit = {
     // nothing by default
   }
+
+  override def createIndex(indexIdent: QualifiedTableName,
+                           tableIdent: QualifiedTableName,
+                           indexColumns: Seq[String],
+                           options: Map[String, String]): Unit = {
+    throw new UnsupportedOperationException("Indexes are not supported");
+  }
```
sumwale: Inconsistent indentation of functions? Use the codeStyleSettings we have been using which is as per Spark's conventions.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -400,7 +405,7 @@ class ColumnFormatRelation(
       connProperties.hikariCP)
     try {
       val stmt = connection.prepareStatement(sql)
-      //stmt.setSt
+      // stmt.setSt
       val result = stmt.executeUpdate()
```
sumwale: looks useless commented out code -- remove

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -417,8 +422,90 @@ class ColumnFormatRelation(
...
+    * clause in its options. Also add GEM_INDEXED_TABLE parameter to
+    * indicate that this is an index table.
+    */
+  private def createIndexTable(indexIdent: QualifiedTableName,
+                               tableIdent: QualifiedTableName,
+                               tableRelation: JDBCAppendableRelation,
+                               indexColumns: Seq[String],
+                               options: Map[String, String]): Unit = {
+
+
```
sumwale: indentation?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -417,8 +422,90 @@ class ColumnFormatRelation(
...
+    val caseInsensitiveMap = new CaseInsensitiveMutableHashMap(tableRelation.origOptions)
+    val tempOptions = caseInsensitiveMap.filterNot(pair => {
+      pair._1.equals(Utils.toLowerCase(StoreUtils.PARTITION_BY)) ||
+        pair._1.equals(Utils.toLowerCase(StoreUtils.COLOCATE_WITH)) ||
+        pair._1.equals(Utils.toLowerCase(JdbcExtendedUtils.DBTABLE_PROPERTY)) ||
+        pair._1.equals(Utils.toLowerCase(ExternalStoreUtils.INDEX_NAME))
+    }).toMap + (StoreUtils.PARTITION_BY -> indexColumns.mkString(",")) +
+      (StoreUtils.GEM_INDEXED_TABLE -> tableIdent.toString) +
+      (JdbcExtendedUtils.DBTABLE_PROPERTY -> indexTblName)
+
```
sumwale: For row tables, see the full CREATE INDEX syntax: https://db.apache.org/derby/docs/10.4/ref/rrefsqlj20937.html
Notice the ASC/DESC specifications for individual columns. We need that for column indexes too (at least going forward). The proper way in API to accept that will be to use the "Column" class (optionally) like in DataFrame. See the DataFrame.sort() overloads and lets match that.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -417,8 +422,90 @@ class ColumnFormatRelation(
...
+      tableRelation.schema,
+      indexOptions)
+  }
+
+  override def createIndex(indexIdent: QualifiedTableName,
+                           tableIdent: QualifiedTableName,
+                           indexColumns: Seq[String],
+                           options: Map[String, String]): Unit = {
+
+    createIndexTable(indexIdent, tableIdent, this, indexColumns, options)
```
sumwale: indentation

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -417,8 +422,90 @@ class ColumnFormatRelation(
...
+                            catalog: SnappyStoreHiveCatalog): Boolean =
+    DependencyCatalog.addDependent(table, dependent.name)
+
+  override def removeDependent(dependent: DependentRelation,
+                               catalog: SnappyStoreHiveCatalog): Boolean =
+    DependencyCatalog.removeDependent(table, dependent.name)
+
+  override def getDependents(catalog: SnappyStoreHiveCatalog): Seq[String] =
+    DependencyCatalog.getDependents(table)
 }
```
sumwale: indentation

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -417,8 +422,90 @@ class ColumnFormatRelation(
...
+                            _mode: SaveMode,
+                            _userSchema: StructType,
+                            _schemaExtensions: String,
+                            _ddlExtensionForShadowTable: String,
+                            _origOptions: Map[String, String],
+                            _externalStore: ExternalStore,
+                            _blockMap: Map[InternalDistributedMember, BlockManagerId],
+                            _partitioningColumns: Seq[String],
+                            _context: SQLContext,
+                            baseTableName: String)
```
sumwale: indentation

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -417,8 +422,90 @@ class ColumnFormatRelation(
...
+  extends ColumnFormatRelation(_table, _provider, _mode, _userSchema,
+    _schemaExtensions, _ddlExtensionForShadowTable, _origOptions,
+    _externalStore, _blockMap, _partitioningColumns, _context)
+ with DependentRelation {
+
```
sumwale: This is a bit of a problem since ColumnFormatRelation is a ParentRelation. So Index becomes both parent and dependent? Perhaps we should stick trait in ColumnFormatRelation elsewhere when creating, or split into base class.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -447,22 +534,22 @@ object ColumnFormatRelation extends Logging with StoreCallback {
...
 
     val tableName = if (table.indexOf('.') > 0) {
       table.replace(".", "__")
     } else {
       table
     }
     INTERNAL_SCHEMA_NAME + "." + tableName +
-        SHADOW_TABLE_SUFFIX
+      SHADOW_TABLE_SUFFIX
   }
```
sumwale: check indentation

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -506,20 +593,53 @@ final class DefaultSource extends ColumnarRelationProvider {
...
+        sqlContext)
+    }
+
     try {
       relation.createTable(mode)
+      var indexes: Array[String] = Array()
+      try {
+        val options = new CaseInsensitiveMutableHashMap(relation.origOptions)
+        indexes = options(ExternalStoreUtils.INDEX_NAME).split(",")
+      } catch {
```
sumwale: How multiple indexes are possible here?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -506,20 +593,53 @@ final class DefaultSource extends ColumnarRelationProvider {
...
+        case e: NoSuchElementException =>
+        }
+
+      indexes.foreach(index => {
+        val sncCatalog = sqlContext.asInstanceOf[SnappyContext].catalog
+        val dr = sncCatalog.lookupRelation(sncCatalog.newQualifiedTableName(index)) match {
+          case LogicalRelation(r: DependentRelation, _ ) => r
+        }
+        relation.addDependent(dr, sncCatalog)
+      })
```
sumwale: Lets make this a generic thing. Whenever a DependentRelation is created, then it needs to be registered against its Parent, so this can be handled by a generic method somewhere. We could even add a separate class like "DependencyManager" or else let DependencyCatalog take care of such.

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatRelation.scala
```diff
@@ -160,6 +160,19 @@ class RowFormatRelation(
       connection.close()
     }
   }
+  override protected def constructSQL(indexName: String,
+                           baseTable: String,
+                           indexColumns: Seq[String],
+                           options: Map[String, String]): String = {
+
```
sumwale: indentation

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatRelation.scala
```diff
@@ -160,6 +160,19 @@ class RowFormatRelation(
...
+    val parameters = new CaseInsensitiveMutableHashMap(options)
+    val columns = indexColumns.reduceLeft[String](_ + "," + _)
+    val indexType = parameters.get(ExternalStoreUtils.INDEX_TYPE) match {
```
sumwale: Hm, reduceLeft vs mkString. I think latter will be more efficient in general because of the use of a single StringBuilder so I would always prefer mkString for string concat with separator (though efficiency is not important here).

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -516,6 +522,7 @@ class SnappyStoreHiveCatalog(context: SnappyContext)
         client = newClient().asInstanceOf[ClientWrapper]
         clientDropTable(dbName, tableIdent.table)
     }
+    cachedDataSourceTables.invalidate(tableIdent)
```
sumwale: already done before in the method. why again?

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -615,6 +622,44 @@ class SnappyStoreHiveCatalog(context: SnappyContext)
...
+      try {
+        indexes = hiveTable.serdeProperties(ExternalStoreUtils.INDEX_NAME) + ","
+      } catch {
+        case e: NoSuchElementException =>
+      }
+      client.alterTable(
+        hiveTable.copy(serdeProperties = hiveTable.serdeProperties +
+          (ExternalStoreUtils.INDEX_NAME -> (indexes + indexName)))
+      )
+    }
```
sumwale: note the code in registerDataSourceTable/unregisterDataSourceTable. We catch HiveException and then retry with a new client. This is to handle "stale" connections which are particularly the case in our unit tests with repeatedly start/shutdown the database (there are other cases too). Add similar handling here.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -615,6 +622,44 @@ class SnappyStoreHiveCatalog(context: SnappyContext)
     }
   }
 
+  def alterTableToAddIndexProp(inTable: QualifiedTableName, indexName: String): Unit = {
+    alterTableLock.synchronized {
```
sumwale: Don't use indexName as Strings at this level. We should "normalize" the names to create QualifiedTableName's (can rename the class to just QualifiedName) at the very start. That takes care of case-sensitivity of names.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -615,6 +622,44 @@ class SnappyStoreHiveCatalog(context: SnappyContext)
...
+        hiveTable.copy(serdeProperties = hiveTable.serdeProperties +
+          (ExternalStoreUtils.INDEX_NAME -> (indexes + indexName)))
+      )
+    }
+    cachedDataSourceTables.invalidate(inTable)
+  }
+
+  def alterTableToRemoveIndexProp(inTable: QualifiedTableName, indexName: String): Unit = {
+    alterTableLock.synchronized {
+      val hiveTable = inTable.getTable(client)
```
sumwale: see the comment before about using QualifiedTableName for indexName rather than String.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -615,6 +622,44 @@ class SnappyStoreHiveCatalog(context: SnappyContext)
...
+        client.alterTable(
+          hiveTable.copy(
+            serdeProperties = hiveTable.serdeProperties - ExternalStoreUtils.INDEX_NAME)
+        )
+      } else {
+        client.alterTable(
+          hiveTable.copy(serdeProperties = hiveTable.serdeProperties +
+            (ExternalStoreUtils.INDEX_NAME -> (newindexes)))
+        )
+      }
```
sumwale: see the comment before about handling HiveException for stale connections

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -626,6 +671,32 @@ class SnappyStoreHiveCatalog(context: SnappyContext)
...
 
+  class SnappyIndexHandler extends TableBasedIndexHandler {
+    override def usesIndexTable() : Boolean = false
+
+    @throws(classOf[HiveException])
+    override def analyzeIndexDefinition(table: Table, index: Index, table1: Table) : Unit = { }
+
+    def generateIndexQuery(var1: List[Index], var2: ExprNodeDesc,
+                           var3: ParseContext, var4: HiveIndexQueryContext): Unit = {}
+
```
sumwale: Is this class required? I thought you mentioned in the PR description that it is not.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -626,6 +671,32 @@ class SnappyStoreHiveCatalog(context: SnappyContext)
...
+  }
+
+  def createIndex(tableIdent: QualifiedTableName, indexName: String,
+                  indexedCols: List[String],
+                  idxProps: java.util.Map[String, String]): Unit = {
```
sumwale: Both indexName and indexedCols should be normalized at this level.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -626,6 +671,32 @@ class SnappyStoreHiveCatalog(context: SnappyContext)
...
+    val dbName = tableIdent.getDatabase(client)
+
+    hiveClient.createIndex(dbName + "." + tableIdent.table,
+      dbName + "." + indexName,
+      "org.apache.spark.sql.hive.SnappyIndexHandler", indexedCols,
+      dbName + "." + indexName, false, null, null, null, null, null, idxProps,
+      null, null, null, null, null, null, null, null)
+
+  }
+
```
sumwale: I thought the idea was to not use this?

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -727,6 +798,7 @@ object SnappyStoreHiveCatalog {
 
   private[this] var relationDestroyVersion = 0
   private val relationDestroyLock = new ReentrantReadWriteLock()
+  private val alterTableLock = new Object
```
sumwale: Purpose of this lock? The relationDestroyLock was for a different reason, namely to protect the relationDestroyVersion so that any drops from other catalog instances cause all other catalog instances to also discard their cached value in cachedDataSourceTables.

core/src/main/scala/org/apache/spark/sql/row/JDBCMutableRelation.scala
```diff
@@ -274,12 +276,25 @@ class JDBCMutableRelation(
...
   }
 
-  override def createIndex(tableName: String, sql: String): Unit = {
+  protected def constructSQL(indexName: String,
+                           baseTable: String,
+                           indexColumns: Seq[String],
+                           options: Map[String, String]): String = {
+
+    ""
+  }
```
sumwale: indentation

core/src/main/scala/org/apache/spark/sql/row/JDBCMutableRelation.scala
```diff
@@ -274,12 +276,25 @@ class JDBCMutableRelation(
...
+
+  override def createIndex(indexIdent: QualifiedTableName,
+                           tableIdent: QualifiedTableName,
+                           indexColumns: Seq[String],
+                           options: Map[String, String]): Unit = {
     val conn = connFactory()
```
sumwale: indentation

core/src/main/scala/org/apache/spark/sql/snappyParsers.scala
```diff
@@ -931,23 +955,26 @@ private[sql] case class TruncateTable(
   }
 }
 
-private[sql] case class CreateIndex(
-    tableIdent: QualifiedTableName,
-    sql: String) extends RunnableCommand {
+private[sql] case class CreateIndex(indexName: String,
+                                    baseTable: QualifiedTableName,
+                                    indexColumns: Seq[String],
+                                    options: Map[String, String]) extends RunnableCommand {
```
sumwale: indentation

core/src/main/scala/org/apache/spark/sql/snappyParsers.scala
```diff
@@ -796,17 +802,35 @@ private[sql] class SnappyDDLParser(caseSensitive: Boolean,
     }
 
   protected lazy val createIndex: Parser[LogicalPlan] =
-    (CREATE ~> INDEX ~> tableIdentifier) ~ (ON ~> tableIdentifier) ~ wholeInput ^^ {
-      case indexName ~ tableName ~ sql =>
-        CreateIndex(tableName, sql)
+    (CREATE ~> (GLOBAL ~ HASH | UNIQUE).? <~ INDEX) ~
+      (tableIdentifier) ~ (ON ~> tableIdentifier) ~
+      commaSepIdents ~ (OPTIONS ~> options).? ^^ {
+      case indexType ~ indexName ~ tableName ~ cols ~ opts =>
```
sumwale: Need optional ASC/DESC specification per column too.

core/src/main/scala/org/apache/spark/sql/snappyParsers.scala
```diff
@@ -796,17 +802,35 @@ private[sql] class SnappyDDLParser(caseSensitive: Boolean,
...
+        val parameters = opts.getOrElse(Map.empty[String, String])
+        if (indexType.isDefined) {
+          val typeString = indexType match {
+            case Some("unique") =>
+              "unique"
+            case Some(x) if x.toString.equals("(global~hash)") =>
+              "global hash"
```
sumwale: "global~hash"??

core/src/main/scala/org/apache/spark/sql/snappyParsers.scala
```diff
@@ -931,23 +955,26 @@ private[sql] case class TruncateTable(
...
+    snc.createIndexOnTable(indexName, baseTable.toString,
+      indexColumns, options)
     Seq.empty
   }
 }
 
 private[sql] case class DropIndex(
-    sql: String) extends RunnableCommand {
+                                   indexName: String,
+                                   ifExists : Boolean) extends RunnableCommand {
```
sumwale: indentation

core/src/main/scala/org/apache/spark/sql/sources/interfaces.scala
```diff
@@ -182,9 +182,19 @@ trait DestroyRelation {
...
+    * @param indexColumns Columns on which the index has to be created
+    * @param options Options for indexes. For e.g.
+    *                column table index - ("COLOCATE_WITH"->"CUSTOMER").
+    *                row table index - ("INDEX_TYPE"->"GLOBAL HASH") or ("INDEX_TYPE"->"UNIQUE")
+    */
+  def createIndex(indexIdent: QualifiedTableName,
+                  tableIdent: QualifiedTableName,
+                  indexColumns: Seq[String],
+                  options: Map[String, String]): Unit
+
```
sumwale: indentation

core/src/main/scala/org/apache/spark/sql/sources/jdbcExtensions.scala
```diff
@@ -137,7 +137,7 @@ object JdbcExtendedUtils extends Logging {
     }
     val tableName = if (dotIndex > 0) table.substring(dotIndex + 1) else table
     try {
-      val rs = conn.getMetaData.getTables(null, schemaName, tableName, null)
+      val rs = conn.getMetaData.getTables(null, schemaName.toUpperCase, tableName.toUpperCase, null)
       rs.next()
```
sumwale: This is a side-affect of skipping normalization of names. We should never need to do such code which is also incorrect when "spark.sql.caseSensitive=true" in SQLConf -- please remove.

sumwale:
Overall idea looks fine. See comments.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
rishitesh: Please add a java compatible version to API as well. Each time one adds an API , he/she should also think about how it can be accessed by Java code. Seq[String] can be List<String> and Map can be java.util.Map<String.String>

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -544,41 +543,108 @@ class SnappyContext protected[spark](
...
+    }
     if (!catalog.tableExists(tableIdent)) {
       throw new AnalysisException(
-        s"$tableIdent is not an indexable table")
+        s"Could not find $tableIdent in catalog")
     }
-
-    //println("qualifiedTable=" + qualifiedTable)
+    import scala.collection.JavaConverters._
     catalog.lookupRelation(tableIdent) match {
```
rishitesh: Action after both the relation seems similar . Can't we club them to a single  case LogicalRelation(ir: JDBCMutableRelation, _) | LogicalRelation(ir: JDBCAppendableRelation, _) =>

rishitesh: Ok . Missed Sumedh's comment here.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -417,8 +422,90 @@ class ColumnFormatRelation(
...
+      case Some(value) => tempOptions + (StoreUtils.COLOCATE_WITH -> value)
+      case _ => tempOptions
+    }
+    snc.createTable(
+      indexTblName,
+      "column",
+      tableRelation.schema,
+      indexOptions)
+  }
+
```
sumwale: I am assuming you plan on adding local CachedBatch sorting and merge iterator later. In fact it can be a general table level thing (so table can also be sorted per CachedBatch on the partitioning column) using a common property. For the merge iterator we may need some hint from the higher level that this is from sorting or not else it will be unnecessary merge sorting.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
sumwale: Btw, I have fixed this in the new parser itself where the result of "identifier" non-terminal is always properly cased (as per conf setting, and whether identifier was quoted or unquoted because for former it should always be taken as is). However, still need to get the new parser done for the DDLParser. I am waiting for AQP to finish merge first else my changes will again increase their work and merge will again get stalled. In the meantime, we have to take care of that explicitly using catalog.processIdentifier even though it is not proper to do that for quoted identifiers as I mentioned.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
hbhanawat: Is there a way to directly update sys.indexes without adding an index actually? Creating a dummy index on the cached batch table may not be a great idea.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
hbhanawat: I am not sure if I get this. here when a ParentRelation is getting created, it is creating DependentRelations and adding them. Finding out DependentRelation of ParentRelation would differ for each ParentRelation. Can you clarify?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
hbhanawat: Yes. I will take it when I work on it.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
hbhanawat: Since while are altering the tables we are maintaining a comma separted list.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
hbhanawat: Well, may be there are two thread trying to create index at the same time and one overwrites the index of another.

hbhanawat:
@sumwale  @rishitesh I have fixed issues other than the ones on which I have commented. Please review.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
sumwale: We will have to put a hook in the sys.indexes VTI (see IndexInfo class). @kneeraj can you comment on this?

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
sumwale: Since they will create a column table, the GemXD layer will ultimately fail. Note that this can happen across clients too e.g. via embedded mode vs in split cluster mode which this lock cannot prevent. So what we do need to ensure is that if the index table creation in GemXD does fail with any exception, then the catalog artifacts should be cleaned properly.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -546,41 +548,114 @@ class SnappyContext protected[spark](
...
+    * @param options Options for indexes. For e.g.
+    *                column table index - ("COLOCATE_WITH"->"CUSTOMER").
+    *                row table index - ("INDEX_TYPE"->"GLOBAL HASH") or ("INDEX_TYPE"->"UNIQUE")
+    */
+  def createIndex(indexName: String,
+                  baseTable: String,
+                  indexColumns: java.util.Map[String, Option[SortDirection]],
+                  options: java.util.Map[String, String]): Unit = {
+    createIndex(indexName, baseTable, indexColumns.asScala.toMap, options.asScala.toMap)
+  }
```
sumwale: indentation

sumwale: I think its time to enable scalaStyle on core and cluster too so that such issues are caught beforehand.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -546,41 +548,114 @@ class SnappyContext protected[spark](
...
     }
   }
 
+  private[sql] def getIndexTable(indexIdent: QualifiedTableName): QualifiedTableName = {
+    new QualifiedTableName(Some(ColumnFormatRelation.INTERNAL_SCHEMA_NAME),
+      indexIdent.table)
+  }
+
+  private def constructDropSQL(indexName: String,
+              ifExists : Boolean): String = {
```
sumwale: indentation

hbhanawat: Spark is following this standard. What are we following?

core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCAppendableRelation.scala
```diff
@@ -280,6 +283,19 @@ case class JDBCAppendableRelation(
...
+               indexColumns: Map[String, Option[SortDirection]],
+               options: Map[String, String]): Unit = {
+    throw new UnsupportedOperationException("Indexes are not supported")
+  }
+
+  override def dropIndex(indexIdent: QualifiedTableName,
+                tableIdent: QualifiedTableName,
+                ifExists: Boolean): Unit = {
+    throw new UnsupportedOperationException("Indexes are not supported")
+  }
```
sumwale: indentation

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
hbhanawat: Agreed. I will fix this.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
sumwale: After our conversation on phone, I checked both the Spark coding standards as well as existing Spark code. See: https://cwiki.apache.org/confluence/display/SPARK/Spark+Code+Style+Guide -- clearly says 4 indent. Also looked at some methods like StaticMemoryManager.acquireStorageMemory which are using 4 indent only. I had created the original codeStyleSettings.xml looking at the guide so looks like that is the proper standard. Not clear what indentation Reynold's talked about.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
sumwale: Generic code can be about getting the parent from DependentRelation and then adding to parent's addDependent.

But I don't quite get what this is doing here. Isn't this code fragment for index table creation? How can indexes be non-existent when base table is getting created?

hbhanawat: This code fragment is executed when column table Relation is being created (both indexes and column tables are already created). This code finds out all the indexes of a column table and create relation for them if not already created and then add dependents.

hbhanawat:
@sumwale Please review

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
sumwale: Ok, so in that case the generic code can look like:

```
// relation recovered that implements ParentRelation (in cachedDataSourceTables CacheLoader.load)
// when ParentRelation is stored in hive catalog then it should have a generic property like "DEPENDENTS" that contains all the dependent relation names with type names (e.g. INDEX can be a type)
// the dependent types will be an enum in Scala code but in hive catalog its name will be stored like done for ExternalTableType)
// for each of the retrieved dependents, add to DependencyCatalog (the dependency type can also be added to DependencyCatalog)
```

core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCAppendableRelation.scala
```diff
@@ -143,8 +146,8 @@ case class JDBCAppendableRelation(
 
   def uuidBatchAggregate(accumulated: ArrayBuffer[UUIDRegionKey],
       batch: CachedBatch): ArrayBuffer[UUIDRegionKey] = {
-    //TODO - currently using the length from the part Object but it needs to be handled more generically
-    //in order to replace UUID
+    // TODO - currently using the length from the part Object but
+    // it needs to be handled more generically    // in order to replace UUID
```
sumwale: two comments in one line not merge properly

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -478,20 +633,55 @@ final class DefaultSource extends ColumnarRelationProvider {
...
+        }
+
+        indexes.foreach(index => {
+          val sncCatalog = sqlContext.asInstanceOf[SnappyContext].catalog
+          val dr = sncCatalog.lookupRelation(sncCatalog.newQualifiedTableName(index)) match {
+            case LogicalRelation(r: DependentRelation, _) => r
+          }
+          relation.asInstanceOf[ColumnFormatRelation].addDependent(dr, sncCatalog)
+        })
+      }
```
sumwale: Previous comment that got folded as outdated.

We should look at adding generic dependency management rather than specific for index (and previous stream tables). For this case it can look like:

```
// relation recovered that implements ParentRelation (in cachedDataSourceTables CacheLoader.load)
// when ParentRelation is stored in hive catalog then it should have a generic property like "DEPENDENTS" that contains all the dependent relation names with type names (e.g. INDEX can be a type)
// the dependent types will be an enum in Scala code but in hive catalog its name will be stored like done for ExternalTableType)
// for each of the retrieved dependents, add to DependencyCatalog (the dependency type can also be added to DependencyCatalog)
```

If you don't want to burden this PR with above, then please track as a JIRA.

hbhanawat: Ok. I created a new function recoverDependentsRelation that recreates any DependentRelation and add them as dependent. So I have folded the two step strategy you have suggested in single step. However, I have added a new ExternalTableType for indexes which is not getting used now but can be used in future.

PR #215, opened 2016-04-12
Snap 677

rishitesh:

While saving a dataframe to row table if a varchar field contains null it gives an exception saying wrong access to CLOB type. We use code generation to set PrepearedStatement values. 
This code ignores the fact that row table might contain an actual VARCHAR and not a CLOB type as is done in column tables.
Changed the code to set java.sql.Types.NULL if a row value is null. 

## Patch testing
Added a test in RowTableTest

## Other PRs 

NA


sumwale:
looks good for merge

PR #216, opened 2016-04-13
[SNAP-606] fixes to allow working core, store against stock Spark

sumwale:

- Rewritten SQL parser based using parboiled2.org which ls also combinatorial PEG grammar like current Spark SQL one. Optimized version which is more than two orders of magnitude faster than Spark SQL combinatorial parser.
- Refactored code to remove dependencies on snappy-spark extensions everywhere in core. Moved store dependencies to be inside core and everything else apart from cluster and memory related additions.
- Introduced reflection in CodeGeneration to access some private methods since there does not seem to be a cleaner way to handle that.
- Moved around code from ToolsCallback to ServiceUtils. Moved code from Services/ServiceManager into core to allow starting server. Leads/locators will continue to be available only in cluster.
- Moved the WHERE subquery, EXISTS/IN support into SnappyExtendedParser only available in cluster mode
- Build changes to allow core compilation with stock Spark 1.6.1. Fixes to DataWriter usage, appendable temp tables and others to allow doing this. (see more details in https://github.com/SnappyDataInc/snappydata/commit/20dc6c27509013c4a3966f03efa1695dc55ab493)

Includes pending code reorganization:

```
snappy-aqp               => aqp (currently defunct)
snappy-core              => core
snappy-tools             => cluster, datasource implementations moved into core
snappy-examples          => examples
snappy-spark             => spark
snappy-store             => store
snappy-tools/benchmark   => tests/benchmark
tests-common             => tests/common
snappy-dunits            => tests/dunit
```

Moved around tests that are supposed to work only in embedded mode into cluster project, vs others that should work against stock Spark into core project. Dunits are still all running in embedded mode though some could be further split which can be done later.

## Patch testing

All existing tests in core as well as most from store now run against stock Spark. Added a test "compare parser performance" in ColumnTableTest which checks for a decently complex TPCH q7 correctness and performance of Spark SQL vs Snappy vs GemXD vs Hive.

## Other PRs 

Minor changes to store in https://github.com/SnappyDataInc/snappy-store/pull/40


sumwale:
@rishitesh @soubhik-c please take a look too

core/src/main/scala/org/apache/spark/streaming/StreamUtils.scala
```diff
@@ -0,0 +1,51 @@
...
+  /** Invoke <code>DStream.getOrCompute</code> */
+  def getOrCompute[T: ClassTag](dStream: DStream[T],
+      time: Time): Option[RDD[T]] = dStream.getOrCompute(time)
+
+  def getGeneratedRDDs[T: ClassTag](dStream: DStream[T]): mutable.Map[Time,
+      RDD[T]] = {
+    // using reflection here since Spark's object is a HashMap while it is
+    // a ConcurrentHashMap in snappydata's version of Spark
+    // [TODO PR] it should be a concurrent map in Apache Spark too
+    dStream.getClass.getMethod("generatedRDDs").invoke(dStream)
```
rishitesh: Why it should be CHM ? Rather should it not be guarded by a lock , if multiple threads are invoking action on the DStream at the same time ?
Can't we cache the method instance  here and only invoke method.invoke(dstream)  ?

core/src/main/scala/org/apache/spark/sql/sources/StoreStrategy.scala
```diff
@@ -52,9 +52,9 @@ object StoreStrategy extends Strategy {
     case DMLExternalTable(name, storeRelation: LogicalRelation, insertCommand) =>
       ExecutedCommand(ExternalTableDMLCmd(storeRelation, insertCommand)) :: Nil
 
-    case InsertIntoTable(l @ LogicalRelation(t: RowPutRelation, _),
-    part, query, overwrite, false) if part.isEmpty =>
-      ExecutedCommand(PutIntoDataSource(l, query, overwrite)) :: Nil
+    case InsertIntoTable(l@LogicalRelation(t: RowPutRelation, _),
```
rishitesh: I guess this still does not fix the PUT conversion of INSERT statements , right ?

rishitesh:
Rest all looks OK to me. New parser concept yet to understand. But we can go for the merge.

core/src/main/scala/org/apache/spark/streaming/StreamUtils.scala
sumwale: I don't think there can be a case where multiple threads are writing to the same map -- only one driver ingester for a single DStream. This CHM was added primarily to safeguard concurrent readers as the writer is going on which can always happen.

>  Can't we cache the method instance here and only invoke method.invoke(dstream) ?

will do

core/src/main/scala/org/apache/spark/sql/sources/StoreStrategy.scala
sumwale: This kind of fixes the issue seen in SNAP-677 because now PutIntoDataSource will only be invoked if it is a insertInto with SaveMode==Overwrite which will not happen. However, as we discussed we don't want to change the behaviour of overwrite to be like put (rather than a truncate which is what happens for other data sources). This is handled by the other PR I have created.

PR #218, opened 2016-04-14
[SNAP-677] Create a separate PutIntoTable plan for PUT operations

sumwale:
## Changes proposed in this pull request

Avoid overloading the InsertIntoTable for puts into row tables else primary key violations are never detected for bulk inserts.

Uses reflection for DataWriter implicit putInto method since DataFrame is private (and not exposed at all).

## Patch testing

Existing tests like RowTableTest that use putInto.

## Other PRs 

Based off the PR for SNAP-606 #216

PR #219, opened 2016-04-14
[SNAP-677] Create a separate PutIntoTable plan for PUT operations

sumwale:

Avoid overloading the InsertIntoTable for puts into row tables else primary key violations are never detected for bulk inserts.

Uses reflection for DataWriter implicit putInto method since DataFrame is private (and not exposed at all).

## Patch testing

Existing tests like RowTableTest that use putInto.

## Other PRs

Fix to other part of SNAP-677 in #215


sumwale:
Merging as per @rishitesh's comment in previous PR: https://github.com/SnappyDataInc/snappydata/pull/218#issuecomment-209967752

PR #222, opened 2016-04-18
a) added overriden method in SnappyContext for createTable, createSam…

rishitesh:

a) added overriden method in SnappyContext for createTable, createSampleTable and createApproxTSTopK to take java.util.Map as options parameter.
b) Removed ifExist parameter from createSampleTable & createApproxTSTopK as its not being used.
c) removed Option(schema) createSampleTable & createApproxTSTopK and added two methods one taking schema and others without schema. This is much cleaner pattern followed by Spark as well.
d) added a quickstart Java class examples/src/main/java/io/snappydata/examples/JavaAirlineDataJob.java. 

## Patch testing
Ran precheckin

## Other PRs 
NA


rishitesh:
@sumwale @hbhanawat @nthanvi also please have a look.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -218,41 +219,125 @@ class SnappyContext protected[spark](
...
    * @param schema
    * @param samplingOptions
-   * @param ifExists
    */
-  def createSampleTable(tableName: String, schema: Option[StructType],
-      samplingOptions: Map[String, String],
-      ifExists: Boolean = false): DataFrame = {
+  def createSampleTable(tableName: String, schema: StructType,
+      samplingOptions: Map[String, String]): DataFrame = {
     val plan = createTable(catalog.newQualifiedTableName(tableName),
```
sumwale: Don't remove ifExists parameter. A bug that it was being ignored. Set the SaveMode appropriately (to Ignore or ErrorIfExists). Add it to the new APIs as well.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -218,41 +219,125 @@ class SnappyContext protected[spark](
...
-      SnappyContext.SAMPLE_SOURCE, schema, schemaDDL = None,
+      SnappyContext.SAMPLE_SOURCE, Some(schema), schemaDDL = None,
       SaveMode.ErrorIfExists, samplingOptions,
```
sumwale: Why is schema not optional now? I think the idea was to inherit from "baseTable" if no explicit schema has been provided.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -218,41 +219,125 @@ class SnappyContext protected[spark](
...
+   * @param topKName the qualified name of the top-K structure
+   * @param keyColumnName
+   * @param topkOptions
+   */
+  def createApproxTSTopK(topKName: String, keyColumnName: String,
+      topkOptions: Map[String, String]): DataFrame = {
     val plan = createTable(catalog.newQualifiedTableName(topKName),
-      SnappyContext.TOPK_SOURCE, inputDataSchema, schemaDDL = None,
+      SnappyContext.TOPK_SOURCE, None, schemaDDL = None,
       SaveMode.ErrorIfExists, topkOptions + ("key" -> keyColumnName),
```
sumwale: See the comment before about adding back ifExists.

Why inputDataSchema has been removed? I think it was being used for the case when there is no "baseTable" (and user may want to insert DFs explicitly)

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
sumwale: Ok, I see you have moved it into a separate overload. Any specific reason?

rishitesh: Yes, First of all it is helpful for Java APIs.( Handling Option in Java is messy) .  Second I observed Spark's createExternalTable() also follows the same pattern. Perhaps they also faced the same issue with Java API. More importantly I feel its good to have two separate API , rather than forcing the user to pass None in case he does not want to pass a schema.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
rishitesh: We don't have same parameter in createTable() as well. Also Spark createExternalTable() also sets allowExisting=false always. Do you see any other usage of this ? One can always use dropTable() and createTable() to recreate the table.

sumwale: @rishitesh

That's an oversight because DDL does allow for it (in Spark SQL too), so we should also add to createTable.
Dropping the table and recreating is no solution since the data will be lost. One can instead catch the exception and ignore in code but that is ugly. As I said DDL does allow for it so the API should also allow.

rishitesh: Rather than ifExist, how do you think allowExisting would work ? 

  @param allowExisting When set to true it will ignore if a table with the same name is
                      present , else it will throw table exist exception

PR #225, opened 2016-04-21
Fixed SnappyJoinSuite to not allow broadcasthash join .

rishitesh:
Changed SnappyJOinSuite to disable broadcast hash join. After fixing sizes for snappy relations the small sizes of tests will always cause a broadcast hash join

## Patch testing
ran precheckin

## Other PRs 
NA

### DIFF
--- a/core/src/test/scala/org/apache/spark/sql/store/SnappyJoinSuite.scala
+++ b/core/src/test/scala/org/apache/spark/sql/store/SnappyJoinSuite.scala
@@ -17,16 +17,32 @@
 package org.apache.spark.sql.store
 
 import io.snappydata.SnappyFunSuite
-import io.snappydata.core.{FileCleaner, RefData, TestData2}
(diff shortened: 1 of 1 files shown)


core/src/test/scala/org/apache/spark/sql/store/SnappyJoinSuite.scala
```diff
@@ -17,17 +17,33 @@
...
+  }
+
+  override def afterAll(): Unit = {
+    super.afterAll()
+    this.stopAll()
+  }
+
+  protected override def newSparkConf(addOn: (SparkConf) => SparkConf): SparkConf = {
+    val conf = super.newSparkConf(addOn)
+    conf.set("spark.sql.autoBroadcastJoinThreshold", "-1")
```
sumwale: @rishitesh Isn't this too small? Properly speaking to validate Namrata's PR it should be some value > 0 but less than the data size in table. I guess can be done after Namrata's PR is merged.

PR #226, opened 2016-04-25
 [SNAP-606] Support for "spark.snappydata" props & compound type client serialization

sumwale:

This changeset adds support for "spark.snappydata.*" property names (in addition to current "snappydata.*") to work with upstream Apache Spark distribution (SparkConf system props and SparkSubmit arguments are filtered as per "spark." prefix).

- Changed io.snappydata.Property into an enum with a Value class that allows an alternative name (prefixed with "spark.") for each of the properties
- Updated LeadImpl, ExternalStoreUtils etc to make use of above and allow for "spark." prefix when filtering properties; changed filter+map to collect calls to simplify and refactored common code a bit into ServiceUtils
- Added new callback class implementation ComplexTypeSerializerImpl to allow serializing ARRAY, MAP, STRUCT types over a JDBC client connection that is compatible with Spark serialized forms as stored in BLOBs in GemXD; the main code is in the new methods added in CodeGeneration that generate code for only the serialization portion
- Removed separate tests/dunit project and instead moved into core and cluster
- Added "dunitTest" tasks for above which run the *DUnitTest test classes with appropriate parameters
- Split the ExternalShellDUnitTest into SplitClusterDUnitTest and SnappyResourceEventsDUnitTest
  where former is in core and runs the split cluster against stock Spark while latter against snappy-spark; the test in core starts a separate snappy embedded cluster and uses JDBC connection into it to do the operations (and the new ComplexTypeSerializer added for serialization)
- Added target to download upstream spark distribution and unpack in dist/ (only once) for above test in core
- Renamed some classes from '*Shell*' to '*Split*' as per agreed convention
- Increasing the default DECIMAL/NUMERIC (precision,scale) to (38,18) as per usual convention in most databases

## Patch testing

New test SplitClusterDUnitTest uses JDBC connection into a separate SnappyData cluster for embedded ops while using normal Spark API for the Spark cluster.
System property for locators is set with "spark.snappydata.store.locators" for latter to let SparkConf pick it up after filter.

## Other PRs 

The base class for ComplexTypeSerializer is in the store PR: https://github.com/SnappyDataInc/snappy-store/pull/48


sumwale:
@kneeraj please take a look too (with the other store PR)

kneeraj:
I think the changes make for at least three different pull request.
1. for serialization changes 2. Test org changes and 3. Property related changes

Too much in one pull request.

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -63,19 +71,58 @@ object Constant {
...
-object Property {
+object Property extends Enumeration {
+
+  final class ValueAlt(name: String, altName: String)
+      extends Property.Val(name) {
+
+    def getOption(conf: SparkConf): Option[String] = if (altName == null) {
+      conf.getOption(name)
+    } else {
+      conf.getOption(name).orElse(conf.getOption(altName))
```
kneeraj: Should we throw error here in case a property is present with both name and altName? If name and altName are both given by the user then altName will never be picked up. Might lead to some unintentional configuration?

core/src/main/scala/io/snappydata/impl/ComplexTypeSerializerImpl.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  private[this] val (serializer, converter) = CodeGeneration
+      .getComplexTypeSerializer(dataType)
+
+  private[this] val struct = dataType match {
+    case s: StructType => Some(s)
+    case _ => None
+  }
+
+  private[this] val bufferHolder = new BufferHolder()
```
kneeraj: Just to understand. It looks like there should be one such serializer per thread and we can reuse it specially since we are using this bufferHolder. Still yo go through the rest. But how and when this serializer object be created. Everytime or one per type or one per type per thread?

sumwale:
@kneeraj 

> I think the changes make for at least three different pull request. 1. for serialization changes 2. Test org changes and 3. Property related changes

I agree that its too much in one pull request. However, working on original issue 3), the problem is that need to add test for 3) in the PR. So how do we write a dunit test that runs against stock Spark? That requires 2).

After having done reorg, if one has to write a similar test to existing one, then need to perform operations in embedded mode, then verify from split mode and vice-versa. To do the former there are two options: a) use a job, b) use JDBC. Now to do a), will have to first add test infrastructure to submit jobs, check success/failure status etc. For b) needed to add 1) which is what I proceeded with.

So the choice is: 2 + 3 + (test infrastructure for running jobs)
or 1 + 2 + 3

core/src/main/scala/io/snappydata/Literals.scala
sumwale: Ok

core/src/main/scala/io/snappydata/impl/ComplexTypeSerializerImpl.scala
sumwale: The CodeGeneration.getComplexTypeSerializer is one per JVM. The ComplexTypeSerializer can be created as user sees fit for each type being used. Normal way would be to create one per type in code-fragment (e.g. a method, "session" etc) of user code.

tests/dunit/src/test/scala/io/snappydata/dunit/streaming/StreamingDUnitTest.scala
```diff
@@ -1,45 +0,0 @@
-package io.snappydata.dunit.streaming
```
kneeraj: I think for this only you said this file is not showing as rename.

sumwale: After looking at docs, git actually does not have any rename operation. It just detects rename automatically based on content where it was failing for this case. Vivek has moved this to AQP so will delete it in any case.

sumwale:
Closing this PR and splitting into five separate PRs (after even more changes) as suggested by @kneeraj

PR #228, opened 2016-04-27
copy the scripts to the correct location

nthanvi:

should read from  debug.conf by default
should not process the files  if specified directory is not available on remote server or
unable to create temp directory on remote

## Patch testing

manual testing

## Other PRs 
No

should read from  debug.conf by default
should not process the files  if specified directory is not available on remote server or
unable to create temp directory on remote


cluster/sbin/collect-debug-artifacts.sh
```diff
@@ -88,6 +101,28 @@ function check_configs {
...
+    echo START TIME = "${START_TIME}"
+    echo END   TIME = "${END_TIME}"
+  fi
+
+  if [ -z "${START_TIME}" ]; then
+    START_EPOCH=0
+  else
+    START_EPOCH=$(date +%s --date "${START_TIME}" 2>/dev/null)
+    if ! [[ "$START_EPOCH" =~ $num_regex ]] ; then
+      echo "${START_TIME} error: Not expected date format"
```
sumwale: Error message will seem a bit odd with time at the start like: `Thu 5 05:36:04 error: Not expected date format`
Instead something like `echo "Error: not expected date format '${START_TIME}'"

cluster/sbin/collect-debug-artifacts.sh
```diff
@@ -88,6 +101,28 @@ function check_configs {
...
+      exit 1
+    fi
+  fi
+
+  if [ -z "${END_TIME}" ]; then
+    END_EPOCH=0
+  else
+    END_EPOCH=`date +%s --date "${END_TIME}" 2>/dev/null`
+    if ! [[ $END_EPOCH =~ $num_regex ]] ; then
+      echo "${END_TIME} error: Not expected date format"
```
sumwale: see comment above

cluster/sbin/collect-debug-artifacts.sh
```diff
@@ -134,41 +170,44 @@ function collect_on_remote {
...
   verbose="$7"
+  start_epoch="$8"
+  end_epoch="$9"
 
-  tmp_dir="$(mktemp -d --tmpdir="`pwd`" data.XXXX)"
+  tmp_dir="$(mktemp -d --tmpdir="$data_dir" data.XXXX)"
   retval=$?
   if [ ! -d ${tmp_dir} ]; then
     echo "FAILED TO CREATE tmp dir on ${host} at ${data_dir} with errno ${retval}"
+    return
```
sumwale: The better way to break out would be "exit 1" (or some other non-zero code)
Otherwise function will simply return and caller will continue execution thereon.

cluster/sbin/collect-debug-artifacts.sh
```diff
@@ -134,41 +170,44 @@ function collect_on_remote {
...
   fi
 
   # first get the pid. The latest log file with the header will have the pid
   host=`hostname`
   if [ ! -d "$data_dir" ]; then
     echo "${data_dir} not found on host: ${host}"
+    return
```
sumwale: see comment above

cluster/sbin/collect-debug-artifacts.sh
```diff
@@ -186,18 +225,51 @@ function collect_on_remote {
...
+          echo "${l} MOD TIME = ${file_mod_epoch}"
+          echo "Adding file ${l} to the array"
+        fi
+        files+=($l)
+      fi
+    done
+
+    for l in $( ls -t *.gfs* 2>/dev/null )
+    do
+      file_mod_epoch=`date +%s --date "$(ls  -l --time-style=long-iso $l | cut -d' ' -f6-7)"`
```
sumwale: Using stat will be much simpler: stat -c %Y $l

Also the time-style=long-iso is a GNU extension which may not be available on other Unixes like Mac (when we support other platforms) but above should work on all Unixes.

cluster/sbin/collect-debug-artifacts.sh
```diff
@@ -186,18 +225,51 @@ function collect_on_remote {
...
+      if [ "${file_mod_epoch}" -ge "${start_epoch}" -a "${file_mod_epoch}" -le "${end_epoch}" ]; then
```
sumwale: This condition may be bit trickier. What we really want is that given a time-range `start-end`, the `start` should also consider the log which contains that `start` (and likewise for `end`). This means that if a log was created on `start-400` and continued till `start+400` then it should be included and similar for `end`.

There is a creation timestamp available in stat (%W) which could have been used but unfortunately most filesystems do not record this and it will give output as zero. Hence to accomplish this need to consider the modification timestamp of the previous file in the same directory and check if `start` lies between previous file's mod-time and this file's, and include this file if that is the case (likewise for `end`).

sumwale: In other words if we keep `prev_file_mod_epoch`, then the condition will be:
```sh
if [ "${file_mod_epoch}" -ge "${start_epoch}" -a "${prev_file_mod_epoch}" -le "${end_epoch}" ]; then
```

Or in other words if there is any overlap of `[prev_file_mod_epoch - file_mod_epoch]` and `[start - end]` intervals. Don't need the next file time since this file's time is its last modification time.

cluster/conf/debug.conf.template
```diff
@@ -1,3 +1,3 @@
-MEMBERS_FILE=members.txt
+MEMBERS_FILE=../work/members.txt
```
sumwale: Shouldn't the default be "$SPARK_HOME/work/members.txt"? Since the script just sources the conf file, the variable should expand just fine. Otherwise "../work" is dependent on the current directory of the user which can be somewhere else.

cluster/sbin/collect-debug-artifacts.sh
sumwale: In your code, the listing will have to be first reversed in time: `ls -tr ...`

Then keep the previous mod-time in a variable (with some initial value like 0, so will need to check for that to set it to `file_mod_epoch` if it is 0).

sumwale:
@nthanvi The move of scripts to correct location was already part of my PR. But other changes in the script are required so you may want to reword the PR and description.

PR #230, opened 2016-04-28
SNAP-652 - Buckets with empty host-list are provided with other hosts.

ashetkar:
- Partial fix for SNAP-652. Buckets with empty host-list (network servers) are given other buckets host-list for executors to process them on. See Other PRs section for the other part of the fix.

- Added a unit test for remote iteration in both the - unified and split cluster - modes.

## Patch testing
- precheckin
- New dunit test: ExternalShellDUnitTest.testRemoteIteratorSNAP652()

## Other PRs 
- https://github.com/SnappyDataInc/snappy-store/pull/52

### DIFF
--- a/core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
+++ b/core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
@@ -56,11 +56,14 @@ final class SparkShellRDDHelper {
     DriverRegistry.register(Constant.JDBC_CLIENT_DRIVER)
     val resolvedName = StoreUtils.lookupName(tableName, conn.getSchema)
     val par = split.index
+
+    if (!useLocatorURL) {
+      val ps = conn.createStatement()
+      ps.execute(s"call sys.SET_BUCKETS_FOR_LOCAL_EXECUTION('$resolvedName', $par)")
+      ps.close()
+    }
+
     val statement = conn.createStatement()
-
-    if (!useLocatorURL)
-      statement.execute(s"call sys.SET_BUCKETS_FOR_LOCAL_EXECUTION('$resolvedName', $par)")
-
     val rs = statement.executeQuery(query)
     (statement, rs)
   }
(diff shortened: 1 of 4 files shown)


ashetkar:
@hbhanawat you too may want to take a look

ashetkar:
@sumwale  you too may want to take a look at this.

core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
```diff
@@ -56,11 +56,14 @@ final class SparkShellRDDHelper {
     DriverRegistry.register(Constant.JDBC_CLIENT_DRIVER)
     val resolvedName = StoreUtils.lookupName(tableName, conn.getSchema)
     val par = split.index
-    val statement = conn.createStatement()
 
-    if (!useLocatorURL)
-      statement.execute(s"call sys.SET_BUCKETS_FOR_LOCAL_EXECUTION('$resolvedName', $par)")
+    if (!useLocatorURL) {
```
rishitesh: Any specific reason why this change ?

core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
```diff
@@ -149,6 +152,8 @@ object SparkShellRDDHelper {
         ClientAttribute.LOAD_BALANCE + "=false"
     val membersToNetServers = GemFireXDUtils.getGfxdAdvisor.
         getAllDRDAServersAndCorrespondingMemberMapping
+    var availableNetUrls = ArrayBuffer.empty[(String, String)]
```
rishitesh: it can be a val

core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
ashetkar: Intention was to have separate statement for each query to make it look neat. But I now realize that it is not supposed to be efficient.
I'll revert this specific change.

core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
ashetkar: Will change it.

PR #231, opened 2016-04-28
[SNAP-606] Support for "spark.snappydata" properties

sumwale:

This changeset adds support for "spark.snappydata.\*" property names (in addition to current "snappydata.*") to work with upstream Apache Spark distribution (SparkConf system props and SparkSubmit arguments are filtered as per "spark." prefix).

- Changed io.snappydata.Property into an enum with a Value class that allows an alternative name (prefixed with "spark.") for each of the properties
- Updated LeadImpl, ExternalStoreUtils etc to make use of above and allow for "spark." prefix when filtering properties; changed filter+map to collect calls to simplify and refactored common code a bit into ServiceUtils
- Renamed some classes from '*Shell*' to '*Split*' as per agreed convention
- Increasing the default DECIMAL/NUMERIC (precision,scale) to (38,18) as per usual convention in most databases

## Patch testing

Proper testing of this is blocked by other PRs SNAP-753, SNAP-749, SNAP-750 that require reorganization of dunits (so that core module can run dunit test against upstream Spark), ability to download and run dunits against upstream Spark, ability to fetch complex types from JDBC. See https://github.com/SnappyDataInc/snappydata/pull/226#issuecomment-214635203 for more details.

The full test will be added as part of SNAP-753.

## Other PRs

(will link others as they are created)


sumwale:
@kneeraj Split original PR into multiple. Please start with this one.

sumwale:
@kneeraj: Now in 5 PRs + 1 store PR. Please see in order: #231, #232, #233, #234, #235
Store PR: https://github.com/SnappyDataInc/snappy-store/pull/53

nthanvi:
looks good

PR #232, opened 2016-04-28
[SNAP-753] Reorganize tests-dunits to core+cluster

sumwale:

This changeset reorganizes the dunits from the separate tests/dunit project into the respective core and cluster projects. This allows adding core and cluster specific dunit (where former can run against upstream Spark while latter against snappy-spark as in other tests).

- Removed separate tests/dunit project and instead moved into core and cluster
- Added "dunitTest" tasks for above which run the *DUnitTest test classes with appropriate parameters
- Split the ExternalShellDUnitTest into SplitClusterDUnitTest and SnappyResourceEventsDUnitTest
  where former is in core and runs the split cluster against stock Spark while latter against snappy-spark; the test in core starts a separate snappy embedded cluster and uses JDBC connection into it to do the operations (and the new ComplexTypeSerializer added for serialization)
- Added target to download upstream spark distribution and unpack in dist/ (only once) for above test in core

## Patch testing

As of now this patch will only allow running existing tests. Support for new split cluster mode tests will make it in with SNAP-751.

## Other PRs 

N/A


sumwale:
@kneeraj split out from original #226 as suggested

PR #233, opened 2016-04-29
[SNAP-750] Serialization of complex types from JDBC

sumwale:

This adds support for serialization and deserialization of complex types (ARRAY, MAP, STRUCT) from JDBC clients for column tables. The storage in tables is as BLOB in the table and as returned in query results (or for inserts), so this is needed to allow both inserts from JDBC or get results properly instead of as a serialized blob.

- Added new callback class implementation ComplexTypeSerializerImpl to allow serializing ARRAY, MAP, STRUCT types over a JDBC client connection that is compatible with Spark serialized forms as stored in BLOBs in GemXD
- the main code is in the new methods added in CodeGeneration that generate code for the serialization portion while deserialization is handled
- use CodeGeneration of serialization code for complex types with handling in SparkSQLExecuteImpl
- also add checks to validate meta-data of the complex type only in first call or in all calls of
  ComplexTypeSerializerImpl.deserialize
- implementation of the new methods of ExternalCatalog in SnappyHiveCatalog to fetch the table schema
  as JSON string; refactored schema loading code in SnappyStoreHiveCatalog for the same
- use the new snappy SQL query hint "complexTypeAsClob" to use string representation of ARRAY, MAP, STRUCT
  instead of fetching as blobs
- multiple fixes to conversion of complex types to string representation
- new tests in SplitClusterDUnit for the above: a) complex type serialization/deserialization,
  b) testing new query hint to fetch values as strings instead, c) schema validation and expected failures
  in mismatch
- also removed the workaround of the system property "scalaTest" that skipped persistence of Hive metadata

## Patch testing

The new split mode test in core, SplitClusterDUnitTest handles various scenarios of serialization, meta-data compatibility checks.

## Other PRs 

I have branched this one from #232 so as to use the new test organization. Also depends on store https://github.com/SnappyDataInc/snappy-store/pull/53

### DIFF
--- a/core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
+++ b/core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
@@ -32,13 +32,14 @@ import com.pivotal.gemfirexd.jdbc.ClientAttribute
 import io.snappydata.Constant
 
 import org.apache.spark.Partition
-import org.apache.spark.sql.collection.ExecutorLocalShellPartition
+import org.apache.spark.sql.collection.{Utils, ExecutorLocalShellPartition}
 import org.apache.spark.sql.execution.ConnectionPool
 import org.apache.spark.sql.execution.columnar.{ExternalStore, ExternalStoreUtils}
 import org.apache.spark.sql.execution.datasources.jdbc.DriverRegistry
 import org.apache.spark.sql.row.GemFireXDClientDialect
 import org.apache.spark.sql.sources.ConnectionProperties
 import org.apache.spark.sql.store.StoreUtils
+import org.apache.spark.sql.types.{StructField, DataType}
 
 final class SparkShellRDDHelper {
 
@@ -177,5 +178,12 @@ object SparkShellRDDHelper {
       case r => sys.error("unexpected region with dataPolicy=" +
           s"${r.getAttributes.getDataPolicy} attributes: ${r.getAttributes}")
     }
+  }
+
+  def getDataType(tp: java.lang.Object): DataType = tp match {
+    case d: DataType => d
+    case s: StructField => s.dataType
+    case _ => throw Utils.analysisException(
+      s"Expected DataType object for the compound type but got $tp")
   }
 }
--- a/core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
+++ b/core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
@@ -32,21 +32,20 @@ import com.pivotal.gemfirexd.jdbc.ClientAttribute
 import io.snappydata.Constant
 
 import org.apache.spark.Partition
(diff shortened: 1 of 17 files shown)


core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
```diff
@@ -32,13 +32,14 @@ import com.pivotal.gemfirexd.jdbc.ClientAttribute
 import io.snappydata.Constant
 
 import org.apache.spark.Partition
-import org.apache.spark.sql.collection.ExecutorLocalShellPartition
+import org.apache.spark.sql.collection.{Utils, ExecutorLocalShellPartition}
```
kneeraj: Looks like unused import.

core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
```diff
@@ -32,13 +32,14 @@ import com.pivotal.gemfirexd.jdbc.ClientAttribute
...
 import org.apache.spark.sql.execution.ConnectionPool
 import org.apache.spark.sql.execution.columnar.{ExternalStore, ExternalStoreUtils}
 import org.apache.spark.sql.execution.datasources.jdbc.DriverRegistry
 import org.apache.spark.sql.row.GemFireXDClientDialect
 import org.apache.spark.sql.sources.ConnectionProperties
 import org.apache.spark.sql.store.StoreUtils
+import org.apache.spark.sql.types.{StructField, DataType}
```
kneeraj: Looks like unused import.

May be we can just revert the file.

core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
```diff
@@ -462,78 +462,84 @@ object Utils {
...
+      }
+      case ArrayType(elementType, _) => value match {
         case data: ArrayData =>
           result.append('[')
-          val etype = atype.elementType
           val len = data.numElements()
           if (len > 0) {
-            val ebuilder = dataTypeStringBuilder(etype, result)
-            ebuilder(data.get(0, etype))
+            val elementBuilder = dataTypeStringBuilder(elementType, result)
```
kneeraj: If the Array size is very big then can this recursive way result in stack overflow situation?

kneeraj:
I looked at the change and it looks good as far as I have understood the code. To do a more thorough review I need more time.

core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
sumwale: It is not recursive in the size of the Array but the type of Array. If we have:

- Array[Long] -- level 1
- Array[Array[Long] -- level 2
- Array[Struct[Long, Int, Array[Int]]] -- level 3

So there is practically no chance of stack overflow.

PR #234, opened 2016-04-29
[SNAP-749] Support for query hints in snappy SQL parser

sumwale:

The query hint in snappydata parser follows the convention used by Oracle, Hive (and possibly others). It is expressed in SQL comments in the form of "key(value)" entries where the comment has to be followed by a "+". For example:

```
SELECT * /*+ hint1(value1) hint2(value2) */ FROM t1

SELECT * FROM t1 --+ hint(value)

```

## Patch testing

Test in SplitClusterDUnitTest will be enabled with fix for SNAP-751 that actually makes use of query hint to fetch complex types as strings.

## Other PRs 

#233 #232


core/src/main/scala/org/apache/spark/sql/snappyParsers.scala
```diff
@@ -630,14 +633,21 @@ class SnappyParser(caseSensitive: Boolean)
  * Snappy dialect uses a much more optimized parser and adds SnappyParser
```
rishitesh: Would a thread local not work here ? Sometimes deep down inside rules it becomes cumbersome to access a SnappyContext.

sumwale: You mean for SnappyContext? Thread local is nothing but heaps of trouble in such cases -- we have seen it over the years with Gem TXState and similar. This should be a general rule for us with very careful consideration if an exception has to be made. I can see a use in SnappyHiveCatalog though not sure if we really need that -- @kneeraj  can you tell the reason?

There is no restriction that user cannot have two SnappyContext's in the same flow of code, and then with ThreadLocal's we have to engage in a never-ending game of fixing the threadlocal from context and vice-versa (e.g. see GemFireTransaction.reattachTransaction, its call hierarchy from TransactionResourceImpl.setupContextStackAndReattach in snappy-store). Contexts in general are there to avoid this ugliness and we should stick by them. If need be, lets have explicit refs of context in analyser and other parts where required.

sumwale: However, I do agree that SnappyContext is not the proper Spark's way of doing this. The SnappyContext is more like a context at execution time whereas query hint is a compilation time context. In other words it should be possible to pick a plan created by some other SnappyContext and run in some other (derby has separate execution and compilation contexts for that reason). Ideally this should be encapsulated in a wrapper LogicalPlan which is then captured in execution plan in our analyser and then picked up where required. This was quite some bit of work so I skipped (the original PR already lead to 7 PRs which are still big).

PR #235, opened 2016-04-29
[SNAP-751] Allow fetching complex types as strings from JDBC clients

sumwale:

Use the new snappy SQL query hint "complexTypeAsClob" to use string representation of ARRAY, MAP, STRUCT instead of blobs.

Enabled test fragment in SplitClusterDUnitTest for the above query hint to fetch values as strings instead.

## Patch testing

Test in SplitClusterDUnitTest uncommented out that fetches and verifies results for complex types as strings.

## Other PRs 

#234 that adds support for query hints in the snappydata's SQL parser


kneeraj:
ok for merge.

PR #237, opened 2016-04-29
Update SYS.GET_COLUMN_TABLE_SCHEMA to use CLOB as registered out type

sumwale:

As part of SNAP-757, the snappy-store procedure implementation has been updated to use CLOB type return value for the schema JSON string. This change is corresponding update to the calling point to use Types.CLOB as the registered output parameter.

## Patch testing

Existing test SplitClusterDUnitTest that uses the above procedure.

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/54


core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
```diff
@@ -575,6 +575,39 @@ object Utils {
     SQLExecution.withNewExecutionId(ctx, queryExecution)(body)
   }
 
+  def immutableMap[A, B](m: mutable.Map[A, B]): Map[A, B] = new Map[A, B] {
```
kneeraj: Just to understand. Why these '+' and '-' needed to be overridden and defaults were not ok for immutable map.

kneeraj:
Looks ok.

core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
sumwale: Not clear on question. The '+' and '-' are abstract methods in immutable.Map that need to be implemented. In mutable map these will change the original map itself but want to avoid for this, hence this wrapper (like Java's Collections.unmodifiable...)

The conversion of mutable to immutable using toMap creates a copy which I wanted to avoid since it can be invoked by many modules in future.

PR #238, opened 2016-05-02
Added interface JavaSnappySQLJob to work easily with SnappyContext fr…

rishitesh:
Added interface JavaSnappySQLJob to work easily with SnappyContext from Java and job server. 

We needed to add a new interface in Java as its not possible to define an concrete class , which implements an interface in scala with abstract type members in this case SparkJobBase.

Added a test examples/src/main/java/io/snappydata/examples/JavaCreateAndLoadAirlineDataJob.java to test the new interface.

## Patch testing

Tested with the new Java based job from job server

## Other PRs 

NA

### DIFF
--- a/cluster/src/main/scala/org/apache/spark/sql/SnappyContextFactory.scala
+++ b/cluster/src/main/scala/org/apache/spark/sql/SnappyContextFactory.scala
@@ -19,13 +19,29 @@ package org.apache.spark.sql
 import com.typesafe.config.Config
 import io.snappydata.impl.LeadImpl
 import spark.jobserver.context.SparkContextFactory
-import spark.jobserver.{ContextLike, SparkJobBase}
+import spark.jobserver.{SparkJobValid, SparkJobInvalid, SparkJobValidation, SparkJob, ContextLike, SparkJobBase}
 
-import org.apache.spark.SparkConf
+import org.apache.spark.api.java.JavaSparkContext
+import org.apache.spark.streaming.api.java.JavaDStream
+import org.apache.spark.{SparkContext, SparkConf}
 
 trait SnappySQLJob extends SparkJobBase {
   type C = SnappyContext
 }
+
+object JavaJobValidate{
+  def validate(status :JSparkJobValidation) : SparkJobValidation ={
+    status match {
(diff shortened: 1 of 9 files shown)


rishitesh:
@soubhik-c @hbhanawat @nthanvi also have a look..

ymahajan:
Looks good Rishi.

cluster/src/main/scala/org/apache/spark/sql/SnappyContextFactory.scala
```diff
@@ -19,14 +19,30 @@ package org.apache.spark.sql
...
+    status match {
+      case j : JSparkJobValid => SparkJobValid
+      case j : JSparkJobInvalid => SparkJobInvalid(j.reason)
+      case _ => SparkJobInvalid("isValid method is not correct")
+    }
+  }
+}
+trait JSparkJobValidation
+case class JSparkJobValid extends JSparkJobValidation
+case class JSparkJobInvalid(reason : String) extends JSparkJobValidation
```
sumwale: This is also using traits and case class? Is the reason for adding this the "case object" for SparkJobValid?

sumwale: Why not make JSparkJobValid a java singleton (or explicit singleton using scala to keep java compatibility)? This will make it consistent with SparkJobValid.

sumwale:
@rishitesh Didn't get the description. What's the problem in extending SnappySQLJob in Java? Your implementation of JSnappySQLJob is already java code which is doing exactly that. Is the only reason the type-erasure leading to context as Object in runJob?

rishitesh:
When java implements a trait with abstract member it can't understand the type of that member. It always thinks it as an Object. It would be ugly to ask user to typecast the Object to SnappyContext. Hence the indirection.

cluster/src/main/scala/org/apache/spark/sql/SnappyContextFactory.scala
rishitesh: Its more for JSparkJobInvalid which takes a reason string as a constructor argument. Scala case class is cleaner than a java bean.

PR #239, opened 2016-05-03
Enabled code generation in column table scan.

rishitesh:

Enabled code generation for column table scan.

## Patch testing
As its only scanning running pre checkin is sufficient. 

## Other PRs 
NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
@@ -410,14 +410,13 @@ object ExternalStoreUtils {
       cachedBatches: Iterator[CachedBatch],
       requestedColumns: Array[String],
       schema: StructType): Iterator[InternalRow] = {
-    /* TODO: (native code gen does not work for some reason)
     // check and compare with InMemoryColumnarTableScan
     val numColumns = requestedColumns.length
     val columnIndices = new Array[Int](numColumns)
     val columnDataTypes = new Array[DataType](numColumns)
     for (index <- 0 until numColumns) {
       val fieldIndex = schema.fieldIndex(requestedColumns(index))
-      columnIndices(index) = fieldIndex
+      columnIndices(index) = index
       schema.fields(fieldIndex).dataType match {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -410,29 +410,23 @@ object ExternalStoreUtils {
...
       schema: StructType): Iterator[InternalRow] = {
-    /* TODO: (native code gen does not work for some reason)
     // check and compare with InMemoryColumnarTableScan
     val numColumns = requestedColumns.length
     val columnIndices = new Array[Int](numColumns)
     val columnDataTypes = new Array[DataType](numColumns)
     for (index <- 0 until numColumns) {
       val fieldIndex = schema.fieldIndex(requestedColumns(index))
-      columnIndices(index) = fieldIndex
+      columnIndices(index) = index
```
sumwale: Don't get this change. This should be the index of the projected column in CachedBatch i.e. fieldIndex. The code in GenerateColumnAccessor also assumes the same (GenerateColumnAccessor lines 173-175):

```scala
          for (int i = 0; i < columnIndexes.length; i ++) {
            buffers[i] = batch.buffers()[columnIndexes[i]];
          }
```
Can you explain this change and why this works?

rishitesh: This is because , we already have a pruned CachedBatches while selecting from gem layer.

sumwale: Ah of course. So how did previous code work which is also using schema.getFieldIndex?

sumwale:
Looks good for merge.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
rishitesh: We take care by having a "bufferIndex", instead of actual requested column indices.

PR #240, opened 2016-05-04
TestUtil is creating a new SnappyContext while dropping tables. This …

rishitesh:

TestUtil is creating a new SnappyContext while dropping tables. This is incorrect , as we need any existing SnappyContext.

## Patch testing
Ran precheckin

## Other PRs 
NA

### DIFF
--- a/core/src/test/scala/io/snappydata/util/TestUtils.scala
+++ b/core/src/test/scala/io/snappydata/util/TestUtils.scala
@@ -30,7 +30,7 @@ object TestUtils {
   def dropAllTables(snc: => SnappyContext): Unit = {
     val sc = SnappyContext.globalSparkContext
     if (sc != null && !sc.isStopped) {
-      val snc = SnappyContext(sc)
+      val snc = SnappyContext.getOrCreate(sc)
       try {
         // drop all the stream tables that can have dependents at the end
         // also drop parents in colocated chain last (assuming chain length = 1)


svbhokare:
Looks good.

sumwale:
@rishitesh Can you tell why creating a new SnappyContext is incorrect? The idea here was to drop all the non-temporary tables in teardown (the temp tables should go away automatically as instance created by test is GCed). So tests should also not be using SnappyContext.getOrCreate() so that everything works as expected. Besides there is nothing to say that test is using the SnappyContext returned by SnappyContext.getOrCreate().

I have been saying this for very long now. We need to remove SnappyContext.getOrCreate() completely because the instance it returns is not thread-safe. Unless someone has a justification for retaining it with reasoning on why lack of thread-safety is not an issue, @rishitesh please file a JIRA to track removal of that.

rishitesh:
Does that mean we are expecting user can create multiple SnappyContext and all our apis would work as expected ? I agree that it would be much cleaner to make SnappyContext completely stateless. 
But problems with such approach would be with mixing temp tables and data source tables.

e.g. 1st step : Create a new SnappyContext. 
                       create a temp table
         2nd step : Create another context
                         select from the earlier temp table.

This was precisely my problem with the issue.

sumwale:
> Does that mean we are expecting user can create multiple SnappyContext and all our apis would work as expected?

This is the **only** supported mode of operation. For each flow of code, user has to create a new SnappyContext. Sharing contexts is not thread-safe and will lead to nasty issues. We need to document this very clearly for users and take care of this correctly in our internal handling like JDBC, jobs, streaming. I believe it is proper for first two but I think the third one is still a problem.

>  But problems with such approach would be with mixing temp tables and data source tables.

This is why they are temp tables in the first place, i.e. they should be accessible only in that context (or connection for JDBC, session for REST whatever be the appropriate abstraction). If a unit test is expecting to access temp table created by another test then either:

a) wrong expectation, change test design

b) keep the tests in the same class and share the context between them by creating own class specific one

I can see that SnappyFunSuite.snc is using getOrCreate one. This is likely old code when we were just trying to get something going and needs to be fixed (perhaps a lazy val if sharing between tests of same class is desirable).

> This was precisely my problem with the issue. 

Please revert this change. We need to fix as mentioned above and also fix SnappyFunSuite.

sumwale:
@rishitesh if you are looking at the problem with sharing in AQP, then in my opinion this needs to be fixed in AQP with highest priority (higher than bootstrap or whatever else in AQP). The sampling reservoir and TopK structures need to be fixed to be in column/row store rather than as temporary structures which is a fatal shortcoming in my opinion. Unfortunately it does not look to be on the radar of AQP team for some reason. Regardless, the tests should be fixed as mentioned in my previous comment.

rishitesh:
I see what you are trying to say. The problems I see needs to be fixed before are.

1) in memory structures like reservoir , etc to use a data source based structure. 

2) There should be no assumptions in our code that temp tables will be available. e.g. sample table's base table can not be a temp table or the validation of base table schema( where we check either schema or baseTable is provided) needs to be gone/ changed.

3) We are ok with the performance hit of relation resolution as cachedDataSource will not be of much use now, or that needs to be made a singleton out of the catalog.

4) relation resolution should be idempotent. (which is required anyway)

rishitesh:
Also as SnappyContext is our main API we should remove the getOrCreate() API right away.

sumwale:
Removal is blocked by AQP issue I think. About 3, yes we can make it singleton and add cleanup in SnappyContext.stop

PR #243, opened 2016-05-06
[SNAP-750] Merge of query hint support from SNAP-749

sumwale:

The query hint in snappydata parser follows the convention used by Oracle, Hive (and possibly others). It is expressed in SQL comments in the form of "key(value)" entries where the comment has to be followed by a "+". For example:

```
SELECT * /*+ hint1(value1) hint2(value2) */ FROM t1

SELECT * FROM t1 --+ hint(value)

```

## Patch testing

Test in SplitClusterDUnitTest will be enabled with fix for SNAP-751 that actually makes use of query hint to fetch complex types as strings.

## Other PRs 

#232 #233 #234


jramnara:
yes, this syntax is better but what are the available options for hints? is this all the GemXD hints(index, etc)?

sumwale:
@jramnara No GemXD hints right now since this is for Spark execution. Right now one hint being used is "complexTypesAsClob" which tells whether complex types should be returned back as CLOBs to the driver (e.g for ODBC that doesn't have such types) or as serialized objects (we have explicit API in Java to deserialize as a list/map on client).

Going forward we can add hints related to join order, index selection etc. This will be important while we work on optimizer to implement the best plan selection especially for the new index support.

PR #244, opened 2016-05-09
If analytic thread can not get data for any table store in snappy, it…

rishitesh:

  If analytic thread can not get data for any table store in snappy, it should default the size one more than broadcast join threshold.
After SNAP-473, this might not be needed.

## Patch testing
Tested with pre checkin

## Other PRs 
NA

### DIFF
--- a/core/src/main/scala/io/snappydata/SnappyDaemons.scala
+++ b/core/src/main/scala/io/snappydata/SnappyDaemons.scala
@@ -30,6 +30,7 @@ import scala.reflect.ClassTag
 import com.google.common.util.concurrent.ThreadFactoryBuilder
 import io.snappydata.Constant._
 
+import org.apache.spark.sql.SQLContext
 import org.apache.spark.sql.execution.ConnectionPool
 import org.apache.spark.sql.execution.columnar.ExternalStoreUtils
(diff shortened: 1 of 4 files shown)


core/src/main/scala/io/snappydata/SnappyDaemons.scala
```diff
@@ -111,22 +112,27 @@ object SnappyAnalyticsService extends Logging {
...
         +currentTableStats.get(tableName).get.valueSize
       } else {
         currentTableStats.get(tableName).get.valueSize
       }
     }
   }
 
+  def defaultSizeForPlanning(sqlContext: SQLContext) : Long =
+    (sqlContext.getConf("spark.sql.autoBroadcastJoinThreshold") + 1).toLong
+
```
sumwale: Better to use: sqlContext.conf.defaultSizeInBytes

Also I think it will be cleaner to not do this here at all and instead let getTableSize return Option[Long], and in the calling points:

```scala
  override def sizeInBytes: Long = SnappyAnalyticsService.getTableSize(table,
    isColumnTable = true).getOrElse(super.sizeInBytes)
```

What do you think?

rishitesh: yes handling from calling point would be more cleaner.. Will do that.

PR #245, opened 2016-05-10
While chunking data from lead node to executor node, we send metadata…

rishitesh:
While chunking data from lead node to executor node, we send metadata once. At receiving side we check if metadata flag is on , then we try to read metadata. But if we are not sending metadata in subsequent chunks we don't set the metadata flag to false. Changed the code to set the flag to false if metadata is not being sent.

## Patch testing
pre-checkin

## Other PRs 
NA


sumwale:
Are you going to add a junit test for this as you had mentioned? Looks good but adding a junit will be better.

rishitesh:
Added the test.

sumwale:
good for merge

PR #246, opened 2016-05-10
Snap 732

SachinJanani:

Corrected API createIndex to make it easier for Java developers

## Patch testing

precheckin ran successfully

## Other PRs 


core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -796,16 +797,32 @@ class SnappyContext protected[spark](
...
+   *                     if value is true and descending when value is false.
+   *                     Direction can be specified as null
     * @param options Options for indexes. For e.g.
     *                column table index - ("COLOCATE_WITH"->"CUSTOMER").
     *                row table index - ("INDEX_TYPE"->"GLOBAL HASH") or ("INDEX_TYPE"->"UNIQUE")
     */
   def createIndex(indexName: String,
       baseTable: String,
-      indexColumns: java.util.Map[String, Option[SortDirection]],
+      indexColumns: java.util.Map[String, lang.Boolean],
```
sumwale: looks a bit odd; better to use java.lang.Boolean

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -796,16 +797,32 @@ class SnappyContext protected[spark](
...
+
+      if(kv._2 == null){
+        indexCol.put(kv._1,None)
+      } else if(kv._2){
+        indexCol.put(kv._1,Some(Ascending))
+      }else{
+        indexCol.put(kv._1,Some(Descending))
+      }
+    )
+    createIndex(indexName, baseTable, indexCol.toMap, options.asScala.toMap)
```
sumwale: A cleaner "scala-friendly" way:

```scala
    val indexCol = indexColumns.asScala.mapValues {
      case null => None
      case true => Some(Ascending)
      case false => Some(Descending)
    }.toMap

```

core/src/test/java/io/snappydata/app/JavaCreateIndexTestSuite.java
```diff
@@ -0,0 +1,138 @@
+/*
```
sumwale: Do we want to put this into "app" package? That was originally meant for stand-alone many test programs and not junit tests. Perhaps something like io.snappydata.api will be better.

sumwale: I think the overall formatting of this file is not proper. Please close Idea, copy "codeStyleSettings.xml" inside .idea directory in the checkout, then restart. Hitting a Ctrl-Shift-F will format the entire file (when using eclipse key bindings) or as per key binding you have set.

sumwale:
Looks good to me. @rishitesh please comment

rishitesh:
Looks good Sachin. You can merge it..

PR #248, opened 2016-05-11
Added installPulse target to copy pulse.war to snappy lib folder.

rishitesh:
Added installPulse target to copy pulse.war to snappy lib folder.
This target assumes the user has a checkout of Pulse repository . We can also automate this if preferable.

## Patch testing
Tested Pulse after running the target

## Other PRs 
NA

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -91,7 +91,7 @@ allprojects {
     hadoopVersion = '2.4.1'
     jettyVersion = '8.1.14.v20131031'
     gemfireXDVersion = '1.5.0-SNAPSHOT'
-    pulseVersion     = '1.4.1'
+    pulseVersion     = '1.5.0'
     buildFlags = ''
     createdBy = System.getProperty("user.name")
   }
@@ -884,21 +884,21 @@ gradle.taskGraph.whenReady({ graph ->
   tasks.getByName('allReports').reportOn rootProject.subprojects.collect{ it.tasks.withType(Test) }.flatten()
 })
 
-task installPulse {
+task packagePulse {
   def pulseWarName = "pulse-${pulseVersion}.war"
-  def pulseDir = System.env.pulsedir
+  def pulseDir = System.env.PULSEDIR
 
   if (pulseDir == null || pulseDir.length() == 0) {
     pulseDir = "${projectDir}/../pulse"
   }
 
-  if (pulseDir == null || pulseDir.length() == 0) {
+  if (!file(pulseDir).canWrite()){
     throw  new Exception(" Pulse directroy can not be empty while building product")
   }
 
   println "Using Pulse location ${pulseDir}"
 
-  println runScript("${pulseDir}/build.sh", pulseDir, ["clean", "build-all"]);
+  runScript("${pulseDir}/build.sh", pulseDir, ["clean", "build-all"])
 
   doFirst {
     delete "${snappyProductDir}/lib/pulse.war"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -91,6 +91,7 @@ allprojects {
     hadoopVersion = '2.4.1'
     jettyVersion = '8.1.14.v20131031'
     gemfireXDVersion = '1.5.0-SNAPSHOT'
+    pulseVersion     = '1.4.1'
```
sumwale: Since the release is going to be 1.5.0, we should change the pulse version to be 1.5.0 here and in pulse repo.

build.gradle
```diff
@@ -883,6 +884,36 @@ gradle.taskGraph.whenReady({ graph ->
...
+    pulseDir = "${projectDir}/../pulse"
+  }
+
+  if (pulseDir == null || pulseDir.length() == 0) {
+    throw  new Exception(" Pulse directroy can not be empty while building product")
+  }
+
+  println "Using Pulse location ${pulseDir}"
+
+  println runScript("${pulseDir}/build.sh", pulseDir, ["clean", "build-all"]);
```
sumwale: println? should be removed and also trailing semi-colon is not useful

build.gradle
```diff
@@ -883,6 +884,36 @@ gradle.taskGraph.whenReady({ graph ->
...
+task installPulse {
+  def pulseWarName = "pulse-${pulseVersion}.war"
+  def pulseDir = System.env.pulsedir
+
+  if (pulseDir == null || pulseDir.length() == 0) {
+    pulseDir = "${projectDir}/../pulse"
+  }
+
+  if (pulseDir == null || pulseDir.length() == 0) {
+    throw  new Exception(" Pulse directroy can not be empty while building product")
```
sumwale: This check will always fail. Better to check write permission to pulseDir here: `if !(file(pulseDir).canWrite()) { ... }`

build.gradle
```diff
@@ -883,6 +884,36 @@ gradle.taskGraph.whenReady({ graph ->
   tasks.getByName('allReports').reportOn rootProject.subprojects.collect{ it.tasks.withType(Test) }.flatten()
 })
 
+task installPulse {
+  def pulseWarName = "pulse-${pulseVersion}.war"
+  def pulseDir = System.env.pulsedir
```
sumwale: usually convention for environment vars is upper-case like PULSEDIR

build.gradle
```diff
@@ -883,6 +884,36 @@ gradle.taskGraph.whenReady({ graph ->
   tasks.getByName('allReports').reportOn rootProject.subprojects.collect{ it.tasks.withType(Test) }.flatten()
 })
 
+task installPulse {
```
sumwale: perhaps better name is "packagePulse" since there is no system installation here

build.gradle
```diff
@@ -883,6 +884,36 @@ gradle.taskGraph.whenReady({ graph ->
...
+  doFirst {
+    delete "${snappyProductDir}/lib/pulse.war"
+  }
+
+  doLast {
+    copy {
+      from "${pulseDir}/build-artifacts/linux/dist/${pulseWarName}"
+      into "${snappyProductDir}/lib/"
+      include "${pulseWarName}"
+      rename { filename -> "pulse.war" }
```
sumwale: Any specific reason for the rename? Does the product depend on the name being exactly pulse.war?

sumwale:
See some minor comments.

build.gradle
rishitesh: Yes, GemXD require this exact name.

PR #249, opened 2016-05-12
Fixed the build failure , which was caused by validating pulseDir. Mo…

rishitesh:

Fixed the build failure , which was caused by validating pulseDir. Moved the validation to a method so that it will be called only when packagePulse is called.

## Patch testing

Checked gradle task without pulse dir

## Other PRs 

NA

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -892,15 +892,14 @@ task packagePulse {
     pulseDir = "${projectDir}/../pulse"
   }
 
-  if (!file(pulseDir).canWrite()){
-    throw  new Exception(" Pulse directroy can not be empty while building product")
-  }
-
-  println "Using Pulse location ${pulseDir}"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -892,15 +892,14 @@ task packagePulse {
...
   }
 
-  if (!file(pulseDir).canWrite()){
-    throw  new Exception(" Pulse directroy can not be empty while building product")
-  }
-
-  println "Using Pulse location ${pulseDir}"
+  doFirst {
+    if (!file(pulseDir).canWrite()){
+      throw  new Exception(" Pulse directroy can not be empty while building product")
```
sumwale: The message should be corrected to something like: s"Pulse checkout directory $pulseDir not writable"

PR #250, opened 2016-05-12
Snap 657

rishitesh:
Allowed column table collocated join in split mode
Added test for collocated join in split mode.

## Patch testing
Added dunit tests

## Other PRs 
NA


hbhanawat:
Looks good

PR #251, opened 2016-05-16
Snap 733

nthanvi:
1. size from SnappyAnalytics is fetched on demand for UI interactions.
2. the table value size used f for query optimization is calculated on regular intervals by adding all primary bucket sizes for Partitioned regions. for replicated regions default is returned for broadcast join.
3. minor change in shadow table name to support schema.

## Patch testing
precheckin
new FunSuite to test functionality
(Fill in the details that how was this patch tested)

## Other PRs 
**Yes - https://github.com/SnappyDataInc/snappy-store/pull/63**


cluster/src/dunit/scala/io/snappydata/cluster/QueryRoutingDUnitTest.scala
```diff
@@ -337,12 +354,12 @@ class QueryRoutingDUnitTest(val s: String) extends ClusterManagerTestBase(s) {
     }
     assert(foundTable)
 
-    val rSet2 = dbmd.getTables(null, JDBCAppendableRelation.INTERNAL_SCHEMA_NAME, null,
+    val rSet2 = dbmd.getTables(null, INTERNAL_SCHEMA_NAME + "APP", null,
```
sumwale: not clear, why this change was required?

sumwale: Ok, looks like you are now adding table schema name to the internal column table schema name. Is that correct?

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
```diff
@@ -0,0 +1,194 @@
...
+  private var tableSizeInfo = Map[String, Long]()
+  private var timer: SystemTimer = null
+
+
+  def start(sc: SparkContext): Unit = {
+    val delay =
+      sc.getConf.getOption("spark.snappy.calcTableSizeInterval")
+          .getOrElse(DEFAULT_CALC_TABLE_SIZE_SERVICE_INTERVAL).toString.toLong
+    timer = new SystemTimer(Misc.getGemFireCache.getDistributedSystem,
+      true, Misc.getGemFireCache.getLoggerI18n)
```
sumwale: Why a new timer? Did GemFireCacheImpl.getCCPTimer() not work for some reason?

cluster/src/dunit/scala/io/snappydata/cluster/QueryRoutingDUnitTest.scala
nthanvi: If user  provides fully qualified tableName then there was a inconsistency in the product while creating the internal column table name and retrieving the row buffer name out of this.

This pull request also fixes that bug by using the naming  new convention  INTERNAL_SCHEMA_NAME + "fully qualifuied tableName" +  SHADOW_TABLE_SUFFIX.
This change done to support the default schema as well.

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
```diff
@@ -0,0 +1,194 @@
...
+
+        result.asInstanceOf[ArrayList[HashMap[String, Long]]]
+            .asScala.foreach(_.asScala.foreach(row => {
+          var totalSize: Long = row._2
+          if (currentTableSizeInfo.contains(row._1)) {
+            totalSize = currentTableSizeInfo.get(row._1).get + row._2
+          }
+          currentTableSizeInfo.put(row._1, totalSize)
+        }))
+
```
sumwale: a more scala-like code can be:

```scala
        .asScala.foreach(_.asScala.foreach { case (k, v) =>
      currentTableSizeInfo += (k -> result.get(k).map(_ + v).getOrElse(v))
    }
```
(also does in a single lookup instead of two, one for contains and other get)

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
```diff
@@ -0,0 +1,194 @@
...
+          tableSizeInfo = currentTableSizeInfo.toMap
```
sumwale: Had added `Utils.immutableMap(currentTableSizeInfo)` to avoid copying like toMap does for better performance (though that slight perf difference probably won't matter here).

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
```diff
@@ -0,0 +1,194 @@
...
+        return Misc.getGemFireCache.getLoggerI18n
+      }
+    }
+  }
+
+  def getTableSize(tableName: String, isColumnTable: Boolean = false):
+  Option[Long] = {
+    val currentTableSizeInfo = tableSizeInfo
+    if (currentTableSizeInfo == null || !currentTableSizeInfo.contains(tableName)) {
+      None
```
sumwale: Is the contains check useful for a column table?

sumwale: Should this not be adding up size of column table and its row buffer?

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
```diff
@@ -0,0 +1,188 @@
...
+          Some(optStat.get + currentTableSizeInfo.get(tableName).get)
+        } else {
+          None
+        }
+      } else {
+        Some(currentTableSizeInfo.get(tableName).get)
+      }
+    }
+  }
+
```
sumwale: I think if there is no column table for some reason then can return row buffer. Somewhat simpler code:
```scala
if (currentTableSizeInfo == null) {
  None
} else currentTableSizeInfo.get(tableName) match {
  case v if isColumnTable =>
    val rowBufferSize = v.getOrElse(0)
    currentTableSizeInfo.get(ColumnFormatRelation.cachedBatchTableName(tableName)).map(_ + rowBufferSize)

  case v => v
}
```

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
```diff
@@ -0,0 +1,188 @@
...
+      val rowBufferSize = currentTableStats.get(rowBuffer).get
+      columnTableNames = columnTableNames.+:(rowBuffer).+:(entry._1)
+      new UIAnalytics(rowBuffer, rowBufferSize, entry._2, true)
+    })
+        ++ currentTableStats.filter(
+      entry => {
+        !columnTableNames.contains(entry._1)
+      }).map(rowEntry =>
+      new UIAnalytics(rowEntry._1, rowEntry._2, 0, false))).toSeq
+  }
```
sumwale: Can't you do in one pass instead of double filter like:
```scala
currentTableStats.map {
  case (isColumnTable(k), v) =>
      val rowBuffer = getRowBufferName(entry._1)
      val rowBufferSize = currentTableStats.get(rowBuffer).get 
      new UIAnalytics(rowBuffer, rowBufferSize, v, true)
  case (k, v) => new UIAnalytics(k, v, 0, false)
}
```

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
```diff
@@ -0,0 +1,188 @@
...
+      }
+    }
+  }
+
+}
+
+object StoreTableSizeProvider {
+
+  def getTableSizes: Seq[UIAnalytics] = {
+    val currentTableStats = tryExecute(conn => getMemoryAnalyticsDetails(conn))
```
sumwale: Wasn't this being removed in favour of the function? Or is this used for obtaining on-the-fly with details in UI pane that other service does not provide?

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
```diff
@@ -0,0 +1,188 @@
...
+
+  private def convertToBytes(value: String): Long = {
+    if (value == null) 0 else (value.toDouble * 1024).toLong
+  }
+
+
+  private def getRowBufferName(columnStoreName: String): String = {
+    columnStoreName.replace(SHADOW_TABLE_SUFFIX, "").
+        replace(INTERNAL_SCHEMA_NAME, "").
+        replaceFirst(".", "").
```
sumwale: why this replaceFirst?

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
```diff
@@ -0,0 +1,188 @@
...
+    tablename.startsWith(INTERNAL_SCHEMA_NAME) && tablename.endsWith(SHADOW_TABLE_SUFFIX)
+
+
+  final def tryExecute[T: ClassTag](f: Connection => T,
+      closeOnSuccess: Boolean = true): T = {
+    val connProperties = ExternalStoreUtils.validateAndGetAllProps(SnappyContext.globalSparkContext
+      , mutable.Map.empty[String, String])
+    val conn = ConnectionPool.getPoolConnection("SYS.MEMORYANALYTICS",
+      connProperties.dialect, connProperties.poolProps, connProperties.connProps,
+      connProperties.hikariCP)
```
sumwale: use ExternalStoreUtils.getConnector here for consistency

sumwale:
Looks good overall. See some code simplification comments.

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
nthanvi: internal column table format is INTERNAL_SCHEMA_NAME.${TABLE_SCHESHA}__{$
TABLE_NAME}SHADOW_TABLE_SUFFIX and just converting it to ${TABLE_SCHESHA}.${TABLE_NAME}

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
nthanvi: it is used  for UI on demand as it shows the correct values and the daemon is to provide the values for the query optimization

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
sumwale: Perhaps I am missing something here. The format of column table name is as you say, then where is a "." there which this replaceFirst is trying to remove?

sumwale: Ah ok, ignore last comment.

sumwale: Ah ok. Perhaps it will be clearer to use: `replace(INTERNAL_SCHEMA_NAME + '.', "")`

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
nthanvi: I dont think it can be done in single pass as we need to delete the column table entries  from the currentTableStats  during the map operation and it should be visible immediately.

sumwale: Can you tell why deleting from currentTableStats is required? The code snippet I gave looks to be doing the same thing of returning seq of UIAnalytics one for each of column and row tables. What is the issue in the code snippet?

nthanvi: the currentTableStats will be having entry for column table and  it's row buffer table, but got the idea what you were saying and converted the code to


    currentTableStats.filter(entry => !isColumnTable(entry._1)).map(details => {
      val columnTableName = ColumnFormatRelation.cachedBatchTableName(details._1)
      val columnTableSize: Long = currentTableStats.getOrElse(columnTableName, 0)
      val isColumnTable = currentTableStats.contains(columnTableName)
      new UIAnalytics(details._1, details._2, columnTableSize, isColumnTable)
    }).toSeq

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
```diff
@@ -88,20 +84,14 @@ object StoreTableValueSizeProviderService extends Logging {
     if (currentTableSizeInfo == null || !currentTableSizeInfo.contains(tableName)) {
```
sumwale: @nthanvi the contains check is redundant here with the below pattern match for get

core/src/main/scala/io/snappydata/SnappyStoreTableSizeProviders.scala
sumwale: Ok, this looks good.

sumwale: @nthanvi Though would have preferred bit more funky code over filter+map :-)

```scala
currentTableStats.collect {
  case (name, v) if !isColumnTable(name) =>
    val columnTableName = ColumnFormatRelation.cachedBatchTableName(name)
    val (columnTableSize, isColumnTable) = currentTableState.get(columnTableName).map(size => (size, true)).getOrElse((0L, false))
    new UIAnalytics(name, v, tableSize, isColumnTable)
}
```

Something like above avoids double pass over list (for fiter and then map) and also double lookup into currentTableStats (one for columnTableSize and another for isColumnTable). Not important here since it is not perf sensitive, but in other code paths which are, the cumulative effect of such code can be substantial, so IMO its good to have a habit of always trying to write tighter code.

cluster/src/dunit/scala/io/snappydata/externalstore/ColumnTableDUnitTest.scala
```diff
@@ -160,8 +175,7 @@ class ColumnTableDUnitTest(s: String) extends ClusterManagerTestBase(s) {
     val p = Map.empty[String, String]
     snc.createTable(tableName, "column", dataDF.schema, p)
 
-    val tName = s"${Constant.INTERNAL_SCHEMA_NAME}.${
-      tableName.toUpperCase}${Constant.SHADOW_TABLE_SUFFIX}"
+    val tName = ColumnFormatRelation.cachedBatchTableName(tableName.toUpperCase())
```
sumwale: @nthanvi Missed this in initial review. Both original and updated code should never be using toUpperCase which breaks case-sensitivity SQLConf setting. Can you check precheckin removing toUpperCase? I believe all case-sensitivity issues should be now fixed in current code.

nthanvi: ok. will update this

PR #256, opened 2016-05-18
Snap 771

rishitesh:
Removed SnappyContext.getOrCreate() method. We intend to keep SnappyContext stateless , which can interact with persistent catalog. 

## Patch testing
precheckin

## Other PRs 
NA


cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
```diff
@@ -483,7 +483,7 @@ object SnappyContextPerConnection {
     val context = connectionIdMap.get(connectionID)
     if (context != null) context
     else {
-      val snc = SnappyContext.getOrCreate(null: SparkContext).newSession()
+      val snc = SnappyContext(null: SparkContext).newSession()
```
sumwale: Since this is creating a new SnappyContext anyway, any need of newSession?

examples/src/main/java/io/snappydata/examples/JavaAirlineDataJob.java
```diff
@@ -28,7 +28,7 @@
   public static void main(String[] args) throws Exception {
     SparkConf sparkConf = new SparkConf().setAppName("JavaSparkSQL").setMaster("local[2]");
     JavaSparkContext jsc = new JavaSparkContext(sparkConf);
-    SnappyContext snc = SnappyContext.getOrCreate(jsc);
+    SnappyContext snc = new SnappyContext(jsc.sc());
```
sumwale: "new" seems unnecessary. Perhaps can make it protected so everywhere apply is used in uniform way.

PR #259, opened 2016-05-19
Removed some of the thread unsafe code from SnappyStoreHiveCatalog.

rishitesh:
  -----------
  Removed some of the thread unsafe code from SnappyStoreHiveCatalog.
a) tempTable structures made concurrent hash map.
b) SessionState in hive client is a thread local variable and need to be set for every thread. HiveContext does it by HiveQL. The same need to be done for SnappyContext. As SnappyContext also has helper APIs, I have taken care of setting a SessionState inside the catalog.
c) If we do this probably we don't need to remove SnappyContext.getOrCreate();

d) Need to add withSessionState for all the public methods of SnappyStoreHiveCatalog , which access ClientWrapper directly or indirectly. 

This code might need some rework once we merge Spark2.0.

## Patch testing

I am adding a test which will test concurrency in accessing the catalog. 

## Other PRs 

We won't need PR SNAP-771 after this.


sumwale:
Some background: various thread-safety issues in HiveContext were fixed in 1.5 and 1.6 like https://issues.apache.org/jira/browse/SPARK-7792

However, there is a basic design issue here rather than an implementation one. Sharing a context among multiple threads means the following:

a) Single-threaded blocking access to the parser (see top-level synchronized in AbstractSparkSQLParser.parse used by HiveContext and likewise synchronized in our parse). This severely limits the scalability especially going forward when we want to gradually move OLTP workloads to snappy-spark too.

b) Temp tables don't have a defined meaning when contexts/catalogs that hold them are shared. Normally a temp table should mean a context specific entity visible only in that context. Especially for JDBC connections the semantics have to follow the SQL semantics, meaning the temp table should be visible only on that "connection" and should die when the connection is closed. Similar considerations hold for web sessions. Imagine one user creating a temp table as a dumping ground for some intermediate results, and then suddenly other users start seeing it, clashing with it and perhaps also modifying it (both inadvertently or maliciously which is both a usability and security problem).

c) Likewise any execution specific things that make sense in that execution context only will have trouble with multi-threaded access (e.g. SnappyContext.queryHints) unless the entire execution is also made single-threaded.

The hive client was another implementation problem which is corrected with your changes above. However, also look at ClientWrapper.conf which was changed to not be thread-local one rather taken from SessionState.getConf to fix the design problem where execution thread (e.g. from query routing) could be different from the thread that created it initially. The "withSessionState" will workaround that problem by setting thread-local temporarily where required but it is an unnecessary overhead (multiple thread-local map lookups that may not be good for perf of OLTP workloads in future). I had looked at it but saw it as a design problem with context itself. Please check how you intend to have the two changes work together properly.

Making SnappyContext thread-safe is good because users may in fact share the SnappyContext and land up into trouble, but it does not obviate the need to get rid of getOrCreate due to above.

In summary, the 2.0 approach of SparkSession is what these contexts should have been in the first place. I have not yet taken a close look at SparkSession but that is what SnappyContext should be ideally be extending from.

sumwale:
I think there is not a lot of meaning in playing around with loaders/thread-locals in withSessionState. We could instead just make it a sync block -- see the previous comment about the change to ClientWrapper.conf to fix the broken thread-local model of ClientWrapper.

rishitesh:
getOrCreate() as an API is one thing, but our design assumption is SnappyContext is singleton . Some of the clean up code like ConnectionPool.clear(), stopping fabric server etc also needed to be taken into account, We can't stop a fabric server if a SnappyContext is stopped. If that’s the case who is the owner of the Fabric server or Connection pool.

Similarly while creating SnappyContext we do a bunch of initialization like spark listener, cluster initialization etc..
I still believe SnappyContext design is mostly correct. What we are missing is a SnappySession which will be the owner of catalog, parser etc which we can't be shared due to the reasons you mentioned.. Right now we can safely say we have only a single session, and we can agree to make that thread safe. 

Then the question becomes how to create a session, which we can discuss more looking at Spark 2.0 design.

sumwale:
> but our design assumption is SnappyContext is singleton

Where so? We are using first initialization to initialize a bunch of Spark layer stuff but that's about it.

> If that’s the case who is the owner of the Fabric server or Connection pool.

The administrators that manage the cluster. This is a long running data store, so no app/context is owner (and also the lead nodes in some sense).

> Right now we can safely say we have only a single session, and we can agree to make that thread safe. 

No, no, that's incorrect both design-wise and behaviour-wise. I don't understand why we are back to square one.
Agreement to make it thread-safe is only because users may share contexts even if we have docs that recommend otherwise.

rishitesh:
Ah, Got it. Was thinking global init is from SnappyCOntext object rather than the SnappyContext class. I have raised a PR to remove SnappyContext.getOrCreate() SNAP-771_2. @sumwale please have a look.

rishitesh:
Spark master branch has this code  , which is similar to what we have done.
/** Returns the configuration for the current session. */
  def conf: HiveConf = state.getConf

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -123,7 +124,14 @@ class SnappyStoreHiveCatalog(context: SnappyContext)
   protected[sql] var client: ClientWrapper =
     newClient().asInstanceOf[ClientWrapper]
 
-  private def newClient(): ClientInterface = {
+  // As long as threads using same IsolatedClientLoader, they should
+  // use the cached Hive client and not create a new one. we will be setting the cached client as
+  // soon as the ClientWrapper is initialized, so that any thread accessing Hive client after
+  // that will get the cached client. This change will be redundant after Spark 2.0 merge. But
+  // for the time being it will avoid ThreadLocal access to set SessionState.
+  protected val internalHiveclient = this.client.client
```
sumwale: unused?

sumwale:
Looks okay. Also need to fix SnappyStoreAQPHiveCatalog.mainDFToSampleDFs

However, does not make SnappyContext fully thread-safe because there is at least SnappyContext.getPreviousQueryHints which is trouble when shared between threads. It cannot be fixed by synchronization because it is set in parsing and accessed later after execution, and is a purely contextual thing like in a session.

One solution is to wrap such entities in a wrapper LogicalPlan and insert into the tree, followed by a custom analyse pass that will convert it into wrapper execution plan that in turn will result in a wrapper RDD. This can then be accessed later by searching for either wrapper LogicalPlan, execution plan or RDD as appropriate. However, to me this looks very convoluted way of doing something which needs to be carried in an execution context (which should be SnappyContext for us or perhaps SnappySession in 2.0). In other words sharing the context has inherent problems even with this fix as I have mentioned -- hopefully will be taken care of in 2.0 if session object is something that is available to all phases of query processing.

If you do decide to merge this, please try to understand the points raised before and above, and track it in JIRA.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
rishitesh: It is just to initialize ClientWrapper.cachedClient, so that other thread do not go to recreate their own, which causes in NPE, while accessing SessionState.getConf with stock spark. In our Spark it works fine as you had already done the fix.

rishitesh:
Added SNAP-847 for SnappyContext.getPreviousQueryHints or any other thing making SnappyContext thread unsafe.

PR #260, opened 2016-05-20
SnappyUnifiedMemoryManager

hbhanawat:
Added a SnappyUnifiedMemoryManager. The idea is that when there is request for execution or storage memory, critical up and eviction up events are checked. If they are set, we try to free the memory cached by Spark rdds by calling memoryStore.evictBlocksToFreeSpace. If enough memory cannot be freed, we return the call and let Spark take a corrective action. In such cases Spark either fails the task or move the current RDDs data to disk. 

With 1.6, Spark is accounting memory for execution and storage pools. When one pool grows, it can borrow from another pool but keeps track of it. This accounting is lost with Snappy and we need to fix it with a new resource manager. 

## Patch testing

Tested it using the existing SnappyResourceEventsDUnitTest.  

## Other PRs 
NA


hbhanawat:
@sumwale  @dshirish please have a look.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -0,0 +1,127 @@
...
+  private val listener = Misc.getMemStoreBooting.thresholdListener()
+
+  private def freeMemory(blockId: Option[BlockId], numBytes: Long,
+      evictedBlocks: mutable.Buffer[(BlockId, BlockStatus)]): Boolean = {
+    val freeMemory = Runtime.getRuntime.freeMemory()
+    if (freeMemory < numBytes) {
+
+      storageMemoryPool.memoryStore.evictBlocksToFreeSpace(blockId,
+        numBytes - freeMemory,
+        evictedBlocks) || !listener.isCritical
```
dshirish: Why does freeMemory() return true when it fails to evict blocks but listener.critical is false?

In SnappyUnifiedMemoryManager. acquireStorageMemory / acquireExecutionMemory, this will mean that when eviction is up, if Spark mem allocation is below spark.memory.fraction, even when no block was evicted memory will be acquired due to calls to super.acquireStorageMemory/super.acquireExecutionMemory.   Do we want Spark to continue acquiring storage memory in such case and only GemXD keeps evicting table data?

cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
```diff
@@ -45,7 +45,7 @@ import org.apache.spark.{Logging, SparkCallbacks, SparkConf, SparkEnv}
  */
 object ExecutorInitiator extends Logging {
 
-  val SNAPPY_MEMORY_MANAGER = "org.apache.spark.memory.SnappyStaticMemoryManager"
+  val SNAPPY_MEMORY_MANAGER = "org.apache.spark.memory.SnappyUnifiedMemoryManager"
```
dshirish: ExecutorInitiator will always set memory manager to SnappyUnifiedMemoryManager? Do we want this configurable with default being SnappyUnifiedMemoryManager?

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
hbhanawat: My opinion is that Spark tasks should not fail until critical up is set. Eviction up is just seen as a warning and Snappy tries to free the blocks but doesn't fail the task. @sumwale, @jramnara thoughts?

dshirish: Ok, I thought at least for storage memory, job will not fail if enough memory is not acquired in acquireStorageMemory(). It will just move the RDD to disk (if possible) or recompute RDD on the fly when needed.

hbhanawat: Yes but that would not be desired for a running job when we have the option of evicting older Spark blocks and least recently used Gemfire xd entries.

cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
hbhanawat: Done.

jramnara:
Is this according to how we defined the semantics in the earlier spec? if not, we should ALWAYS spec and review. Don't mean to be anal but this discipline is important. 

Now, to this PR - it seems to imply when memory becomes scarce we prefer allocating memory to the store than the use of temporary memory for spark caching. Is this correct? I thought we made the decision to be fair across spark temporary caches and the store. First come first serve. 

Though, if memory is insufficient for execution (like a sort or shuffle) I suppose eviction should be attempted up to a point.

hbhanawat:
>>Now, to this PR - it seems to imply when memory becomes scarce we prefer allocating memory to the store than the use of temporary memory for spark caching. Is this correct?
When memory becomes scarce, and critical event is set, we fail the tasks that ask for execution memory and the requests for rdd caching are disallowed. 
When memory becomes scarce, and eviction event is set while critical is not, we allow task execution and allow the requests for rdd caching. This means Snappy in the backend is evicting the entries to disk.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
sumwale: > Ok, I thought at least for storage memory, job will not fail if enough memory is not acquired in acquireStorageMemory(). It will just move the RDD to disk (if possible) or recompute RDD on the fly when needed.

I think in the new model if storage pool fails to acquire the required blocks, then there are no more blocks available in Spark layer (since it can steal from execution pool).

@hbhanawat Assuming that to be the case, I think Shirish is right that it should better start dropping cached blocks in that case for storage blocks. As per the default setting of maxMemory==GemXD_eviction below, similar is the effect that beyond eviction limit start dropping storage blocks (otherwise you should also change maxMemory below to criticalLimit to keep it consistent with the reasoning here). So a flag can be passed to freeMemory for the two cases:

- if freeMemory() has been invoked from acquireStorageMemory, return false if evictBlocksToFreeSpace fails and evictionUp is still true
- if freeMemory() has been invoked from acquireExecutionMemory then only fail if criticalUp is still true after evictBlocksToFreeSpace

Secondly should we allow it to fail the Spark way or should we fail by explicitly throwing LowMemoryException. Issue is that if the failure comes from GemXD layer it will be a LowMemoryException, but from Spark layer it will be like "unable to acquire ..." error. Perhaps we should somehow have Spark layer throw LowMemoryException after it has done retries and all instead of "unable to acquire ..." for consistency (message can still be same as "unable to acquire xxx bytes")

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -0,0 +1,127 @@
...
+    }
+    val usableMemory = systemMemory - reservedMemory
+    // GemFireXD is required to be already booted before constructing
+    // snappy memory manager
+    var evictFraction: Double = GemFireCacheImpl.getExisting.getResourceManager
+        .getEvictionHeapPercentage
+    if (evictFraction <= 0.0) {
+      evictFraction = 0.75
+    }
+    val memoryFraction = conf.getDouble("spark.memory.fraction", evictFraction)
```
sumwale: see comment above that this looks inconsistent with freeMemory that only fails after criticalFraction

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
hbhanawat: Ok. I have made the changes for the first point. For second point, I will take it later because it may need changes in Spark side and is not a high priority.  Please let me know if you don't agree.

sumwale: Sounds good.

sumwale:
@jramnara 

> Now, to this PR - it seems to imply when memory becomes scarce we prefer allocating memory to the store than the use of temporary memory for spark caching. Is this correct? I thought we made the decision to be fair across spark temporary caches and the store. First come first serve. 

To add to Hemant's points: when memory becomes scarce (i.e. EVICTION_UP) we start evicting both from snappy-store as well as spark layer as above. Both happen independently of each other where store will have explicit background threads doing the eviction till EVICTION_DOWN is not hit, while in Spark layer any further requests will push out blocks to disk where possible as in this patch.

When things are really bad (CRITICAL_UP) then operations in both layers will start failing.

PR #261, opened 2016-05-20
Removed SnappyContext.getOrCreate method.

rishitesh:
Removed SnappyContext.getOrCreate method. Please see PR SNAP-771_1 for detailed explanation.

## Patch testing
precheckin

## Other PRs 

Related PRs SNAP-771_1 which deals with concurrency of catalog.  AQP SNAP-771_2

### DIFF
--- a/cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
@@ -483,7 +483,7 @@ object SnappyContextPerConnection {
     val context = connectionIdMap.get(connectionID)
     if (context != null) context
     else {
-      val snc = SnappyContext.getOrCreate(null: SparkContext).newSession()
+      val snc = SnappyContext(null: SparkContext).newSession()
       val oldContext = connectionIdMap.putIfAbsent(connectionID, snc)
       if (oldContext == null) snc else oldContext
     }
--- a/cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
@@ -483,7 +483,7 @@ object SnappyContextPerConnection {
     val context = connectionIdMap.get(connectionID)
     if (context != null) context
     else {
-      val snc = SnappyContext(null: SparkContext).newSession()
+      val snc = SnappyContext(null: SparkContext)
       val oldContext = connectionIdMap.putIfAbsent(connectionID, snc)
       if (oldContext == null) snc else oldContext
     }
--- a/core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
@@ -92,7 +92,7 @@ class SnappyContext protected[spark](
   self =>
 
   protected[spark] def this(sc: SparkContext) {
-    this(sc, SQLContext.createListenerAndUI(sc), true)
+    this(sc, SnappyContext.sqlListener(sc), true)
   }
 
   @transient private[spark] lazy val snappyContextFunctions:
@@ -1141,6 +1141,7 @@ object GlobalSnappyInit {
   private val aqpContextFunctionImplClass =
     "org.apache.spark.sql.execution.SnappyContextAQPFunctions"
 
+
   private[spark] def getSnappyContextFunctionsImpl: SnappyContextFunctions = {
     Try {
       val mirror = u.runtimeMirror(getClass.getClassLoader)
@@ -1162,6 +1163,7 @@ object SnappyContext extends Logging {
 
   @volatile private[this] var _anySNContext: SnappyContext = _
   @volatile private[this] var _clusterMode: ClusterMode = _
+
   @volatile private[this] var _globalSNContextInitialized: Boolean = false
   private[this] val contextLock = new AnyRef
 
@@ -1193,11 +1195,26 @@ object SnappyContext extends Logging {
       throw new IllegalStateException("Invalid SparkConf")
   }
 
+
   /** Returns the current SparkContext or null */
   def globalSparkContext: SparkContext = try {
     SparkContext.getOrCreate(INVALID_CONF)
   } catch {
     case _: IllegalStateException => null
+  }
+
+
+  @volatile private[this] var _sqlListener : SQLListener = _
+
+  def sqlListener(sc : SparkContext) : SQLListener = {
+    if(_sqlListener == null){
+      synchronized {
+        if(_sqlListener == null) {
+          _sqlListener = SQLContext.createListenerAndUI(sc)
+        }
+      }
+    }
+    _sqlListener
   }
 
   private def newSnappyContext(sc: SparkContext) = {
@@ -1372,6 +1389,7 @@ object SnappyContext extends Logging {
     }
     _clusterMode = null
     _anySNContext = null
+    _sqlListener  = null
     _globalSNContextInitialized = false
 
   }
--- a/core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
@@ -1237,12 +1237,12 @@ object SnappyContext extends Logging {
 
   /**
    * @todo document me
-   * @param sc
+   * @param jsc
    * @return
    */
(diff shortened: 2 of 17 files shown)


rishitesh:
Closed SNAP-771 and raised this PR , as it solves the problem with least amount of changes to tests and is much cleaner. I still need to figure out how to close Hive connections for all the threads . We will deal with this with SNAP-771_1

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
```diff
@@ -483,7 +483,7 @@ object SnappyContextPerConnection {
     val context = connectionIdMap.get(connectionID)
     if (context != null) context
     else {
-      val snc = SnappyContext.getOrCreate(null: SparkContext).newSession()
+      val snc = SnappyContext(null: SparkContext).newSession()
```
sumwale: since a new SnappyContext is being created, so what's the need for newSession?

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1240,31 +1240,15 @@ object SnappyContext extends Logging {
...
-  /**
-   * Returns an existing SnappyContext or create one if does not exists
-   * @param jsc
-   * @return SnappyContext
-   */
-  def getOrCreate(jsc: JavaSparkContext): SnappyContext = {
+  def apply(jsc: JavaSparkContext): SnappyContext = {
     if (jsc != null) {
-      getOrCreate(jsc.sc)
+      newSnappyContext(jsc.sc)
```
sumwale: better have `apply(jsc.sc)` in case someone modifies apply behaviour later (and will almost surely forget updating this java one)

core/src/test/scala/io/snappydata/SnappyFunSuite.scala
```diff
@@ -40,23 +40,48 @@ abstract class SnappyFunSuite
...
+      val gnc = cachedContext
+      if (gnc != null) gnc
+      else {
+        cachedContext = SnappyContext(sc)
+        cachedContext
+      }
+    }
+  }
+
+  protected def snc: SnappyContext = getOrCreate(sc)
```
sumwale: Does this really require cachedContext and all this, and not be handled by a simple lazy val? Since the lifetime of lazy val is the life of the class, it should be valid for only the tests of one class and then be released when class object goes out of scope (which will happen post afterAll).

core/src/test/scala/io/snappydata/SnappyFunSuite.scala
```diff
@@ -136,6 +161,7 @@ abstract class SnappyFunSuite
     // GemFireXD stop for local mode is now done by SnappyContext.stop()
     println(" Stopping spark context = " + SnappyContext.globalSparkContext)
     SnappyContext.stop()
```
sumwale: We should really get rid of this. I had suggested it to Namrata where she had changed snappy cleanup to happen in SparkContext stop via listener bus. Its unnecessary and confusing to users because it is exactly equivalent to SparkContext stop.

sumwale:
Looks good -- some comments.

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
hbhanawat: newSession is anyway creating a new Object. What purpose this explicit object creation is solving? We added newSession instead of explicit object creation to share the UI tab.  IMO, we should leave this code untouched.

sumwale: @hbhanawat I don't think separate JDBC connections should be sharing the same UI tab in any case which is currently wrong behaviour. Only SQL from the same connection should share which was one of the purpose of connectionIdMap (to have same SnappyContext for same JDBC connection). So we should just remove newSession() here in my opinion.

sumwale: We can add something to UI to group multiple connections in a single UI tab to reduce number of tabs if number of connections is large. Can be tracked separately for 0.6/1.0

hbhanawat: >>I don't think separate JDBC connections should be sharing the same UI tab 

It would mean that for every client connection we will be creating a new tab. Also, there would be a different tab for every job? Why would we want to do that when Storage and Job tabs are already shared. 
IMO, this is a hack but from usability point of view it is needed. My vote would be to keep the newSession as is for now and make the necessary changes only when we merge with 2.0. 

>>We can add something to UI to group multiple connections in a single UI tab to reduce number of tabs if number of connections is large. Can be tracked separately for 0.6/1.0 

UI Listener has been moved to SparkSession in 2.0. So 2.0, I believe, we will have a clean solution. 
I would recommend deferring this PR.

sumwale: I don't understand this reasoning. SparkSession will also be one for every connection. So its behaviour will be exactly the same i.e. one tab per connection unless some separate provision of grouping tabs is made.

> It would mean that for every client connection we will be creating a new tab.

Yes, thats how it should be. Refer to SYS.SESSIONS table in GemXD. Otherwise all queries will get mixed in a single tab making it useless. As I said we can look at clean grouping of tabs later.

hbhanawat: >>Yes, thats how it should be. Refer to SYS.SESSIONS table in GemXD. Otherwise all queries will get mixed in a single tab making it useless. 

Hmm. I agree that one cannot make out the queries fired from one connection if the tab is shared. But having multiple tabs is no solution either. See the picture on https://jira.snappydata.io/browse/SNAP-183. How can one easily figure out the SQL tab for the connection that he is looking for? I would still recommend postponing this PR and not change the UI behaviour for now. We can do it in a better way with 2.0, I suppose.

rishitesh: Spark 2.0  has very clear definition of SharedState and SessionState. SessionState is one which accepts a SQLConf (understandable) . But Spark 2.0 has one listener across all sessions. Its a bit weird, as different SQL conf might result in different query plans (e.g broadcast limit). How can a user make out , how his query got executed. May be its a Spark oversight.

sumwale: Yes, the tab should also list the username used for connection (later also the clientID esp for thrift connections which provide that info). That is a more imp issue to fix to make the UI usable.

How many concurrent connections are we expecting that route queries? We can suppress tabs that have not fired anything for a while.

sumwale: @rishitesh I think in Spark world they are not expecting many concurrent queries. We will have to work it out for our usage patterns in any case regardless of Spark2.0 changes.

rishitesh: That's correct Sumedh, as I have given an example, two different SqlContext with different conf will create confusion even now with Spark. I think their UI is not suited for multiple live sessions/connection etc.. Probably we need to work for an extension to Spark UI.

rishitesh: So , what is the final conclusion @hbhanawat @sumwale , should we go ahead and merge it or wait for Spark2.0 ?

hbhanawat: :-) As I have already said, my recommendation - wait for 2.0. Also, I don't think there is an urgency to push this in 0.5.

sumwale: And mine is the opposite :)

core/src/test/scala/io/snappydata/SnappyFunSuite.scala
sumwale: @rishitesh Did you try with lazy val like I suggested?

rishitesh: Yes Sumedh, I forgot to mention that here. That becomes a bit problematic when tests stops SparkContext on their own. In that case the lazy val still points to the stale SnappyContext.

PR #263, opened 2016-05-24
Bucket size should be calculated only for the datastores

nthanvi:

instead of calling  function on all server it should be called only on datastore 

## Patch testing

clean precheckin

## Other PRs 
No


nthanvi:
change as requested in https://github.com/SnappyDataInc/snappy-store/pull/69

PR #265, opened 2016-05-24
[SNAP-837] Add packageTests to dtests archives to enable publishing of tests jar

sumwale:

Addition of packageTests task to dtests archives enables the "-tests" jar to be published in local and remote maven.

- moved the dangling copy task in dtests/build.gradle inside testsClass.doLast
- also moved the download plugin separately into core so as to not conflict with
  store/native declaration (which runs both as sub-project or independent project)

## Patch testing

Checked publishLocal that it does place the gemfirexd-scala-tests...-tests jar in local m2 repository.

## Other PRs 

NA

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,6 @@ buildscript {
     classpath 'io.snappydata:gradle-scalatest:0.13-1'
     classpath 'org.github.ngbinh.scalastyle:gradle-scalastyle-plugin_2.10:0.8.2'
     classpath 'com.github.jengelman.gradle.plugins:shadow:1.2.3'
-    classpath 'de.undercouch:gradle-download-task:3.0.0'
   }
 }
 
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -1,6 +1,9 @@
+plugins {
+  id "de.undercouch.download" version "3.0.0"
+}
+
 apply plugin: 'scala'
 apply plugin: 'com.github.johnrengelman.shadow'
-apply plugin: 'de.undercouch.download'
 
 compileScala.options.encoding = 'UTF-8'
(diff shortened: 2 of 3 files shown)


sumwale:
@KishorBachhav @hbhanawat I moved the dangling copy task in dtests/build.gradle in the doLast section. But it still bothers me that the task:

a) Copies to store/tests/sql which is not a dependent sub-project. It should really be copying to its own area and script should pick up from there into classpath or resource search path (or create a mega-shadow jar that includes gemfirexd-tests but that will not be preferable).

b) Uses custom hackish way to construct the full build path to copy instead of using regular project(...).buildDir.

Do you guys know the reasons for the above?

sumwale:
@KishorBachhav @hbhanawat To clarify point a) further: the resources are all copied to the store tests, but the tests jar remains separate. This indicates buggy scripts/mechanism and also implies that hydra framework is really not extendible at all in that all confs/bts have to go into base framework directory. If that be the case, then hydra search for confs/bts should be fixed to use regular resources mechanism or something.

KishorBachhav:
@swatisawant can speak more on this. I think she has added this copy task.

sumwale:
I have replaced the hard-coded paths in the copy task with appropriate gradle properties.

Will go ahead with the merge if there are no comments. Will wait for comments from @swatisawant on why the config has to be copied to store/tests/sql

swatisawant:
@sumwale - Yes, I am suppose to add the support in hydra to include DTESTS path for searching required conf and inc files similar to JTESTS and EXTRA_JTESTS.
This is the temporary solution I had used to proceed further fast and will be working on this once I have free cycle for doing the same.

PR #268, opened 2016-05-25
helper method to check if relation exists in store

nthanvi:
disabling the concurrency checks for the row buffers so that tombstone creation can be avoided.

Before table creation if this method can be used to check if the table
is column table or not simply by looking the entry into the store.
currently it contains hack for default schema but in next commit will
fix  this issue by setting the default schema

## Patch testing
precheckin

## Other PRs 
Yes.
https://github.com/SnappyDataInc/snappy-store/pull/74

Before table creation if this method can be used to check if the table
is column table or not simply by looking the entry into the store.
currently it contains hack for default schema but in next commit will
fix  this issue by setting the default schema

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
@@ -16,8 +16,8 @@
  */
 package org.apache.spark.sql.execution.columnar.impl
 
-import java.lang
 import java.util.{Collections, UUID}
+
 import scala.collection.JavaConversions
 import scala.collection.concurrent.TrieMap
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
```diff
@@ -137,6 +137,16 @@ object StoreCallbacksImpl extends StoreCallbacks with Logging with Serializable
   override def getHashCodeSnappy(dvds: scala.Array[Object]): Int = {
     partioner.hashValue(dvds)
   }
+
+  override def haveRegisteredExternalStore(tableName: String): lang.Boolean = {
+    // TODO -  remove below that deals with default schema and all
```
sumwale: Why java.lang.Boolean object? This can't return null so just Boolean should be good.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
```diff
@@ -137,6 +137,16 @@ object StoreCallbacksImpl extends StoreCallbacks with Logging with Serializable
...
+    // entries in store should come with fully qualified tableName
+    val table = if (tableName.startsWith(Constant.DEFAULT_SCHEMA + ".")) {
+      tableName.split("\\.")(1)
```
sumwale: Bit more efficient will be: tableName.substring(Constant.DEFAULT_SCHEMA.length() + 1)

sumwale:
Looks good - see couple of comments

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
nthanvi: using Boolean throws compilation error

PR #269, opened 2016-05-26
Snap 852

nthanvi:

on the Snappy spark UI along with store table sizes we should also display the rowcount for each table.

## Patch testing

manual testing,
running precheckin

## Other PRs 

No.


ymahajan:
looks good. We will keep adding meaningful stats to UI

PR #272, opened 2016-05-27
Added python api documentation

nthanvi:

documentation python streaming api

## Patch testing
NA
(Fill in the details that how was this patch tested)

## Other PRs 
NA

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -33,7 +33,8 @@
     * [Top-K Elements in a Stream](#top-k-elements-in-a-stream-explanation-1)
     * [Step 5 - Create and Query Stream Table and Top-K](#step-5---create-and-query-stream-table-and-top-k)
     * [Working with Spark shell and spark-submit](#working-with-spark-shell-and-spark-submit)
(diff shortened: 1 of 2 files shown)


nthanvi:
@ymahajan  I think this document should be split in two documents. one containing the API info and another containing the job info. what you think?

ymahajan:
'schema Dstream' should be 'SchemaDStream'
'SnappyStreaming Context' should be 'SnappyStreamingContext'

ymahajan:
@nthanvi, we can move these to gettingStarted.md as you suggested

ymahajan:
Instead of 
'The Snappy Python Streaming Api have limited features in compare to it's corresponding scala API. It does not support TopK structures and registerCQ.'
We can add following 
'Currently we don't have Python APIs added for continuous queries and AQP/Sampling.'

nthanvi:
@ymahajan  - I have taken care of the review comments except that the API's are yet not added to the GettingStarted Doc As I think that before doing that it would be a better to discuss in the scrum meeting and take others opinion as well.

Merging the commit for now so that at least we  have some documentation of python API in place.

PR #273, opened 2016-05-27
* Addresses the second part of the changes required in SNAP-658.

ashetkar:
* Addresses the second part of the changes required in SNAP-658.
* A map of <InternalDistributedMember, BlockManagerId> in StoreUtils is updated when
  1. an executor is initialized and it sends this info to driver via
     SnappyExecutorMessage.
  2. the scheduler detects a departed member and removes its entry from this map.

## Patch testing
* Added a unit testcase 'ExecutorMessageDUnitTest.scala' for verifying this fix.

## Other PRs 
NA


ashetkar:
@rishitesh @hbhanawat,
FYI, I have now updated the fix to rely on the spark message and listener to update the block map. Got rid of the changes in snappy-store layer.

cluster/src/dunit/scala/io/snappydata/externalstore/ExecutorMessageDUnitTest.scala
```diff
@@ -0,0 +1,165 @@
...
+
+  def testStore01BlockMapUpdatesWithExecutorDown(): Unit = {
+    import ExecutorMessageDUnitTest._
+    executeSomething()
+    verifyMap("stopExecutor")
+    restartExecutors(locatorPort, bootProps)
+    getLogWriter.info("testStore01BlockMapUpdatesWithExecutorDown() Successful")
+  }
+
+  // TODO The next test that runs after this one hangs/fails because we stop the
```
hbhanawat: Why do you have to stop the process? Can we just stop snappy like ClusterManagerTestBase.stopAny?

cluster/src/dunit/scala/io/snappydata/externalstore/ExecutorMessageDUnitTest.scala
```diff
@@ -0,0 +1,165 @@
...
+  // server process in this test. Enable it after resolving the issue.
+  // When run individually, this test passes.
+  def _testStore02BlockMapUpdatesWithNodeDown(): Unit = {
```
hbhanawat: None of the tests executes any queries after the executor is shutdown. We need to ensure that queries run ok after an executor is shutdown.

cluster/src/dunit/scala/io/snappydata/externalstore/ExecutorMessageDUnitTest.scala
```diff
@@ -0,0 +1,165 @@
...
+          loc.stop(null)
+        }
+      }
+    })
+    // Delete all oplog files
+    deleteOplogFiles(true)
+  }
+
+  @throws(classOf[IOException])
+  def deleteOplogFiles(includeDataDictionary: Boolean): Unit = {
```
hbhanawat: If we drop all the tables after the test, why is this explicit cleaning needed? If this is a workaround for any issue, we should mention the JIRA of that issue.

cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
```diff
@@ -32,6 +29,8 @@ import com.gemstone.gemfire.distributed.internal.membership.InternalDistributedM
 import com.gemstone.gemfire.internal.cache.GemFireCacheImpl
 import com.pivotal.gemfirexd.internal.engine.Misc
 import com.pivotal.gemfirexd.internal.engine.distributed.utils.GemFireXDUtils
+import com.pivotal.gemfirexd.internal.engine.store.ServerGroupUtils
+import io.snappydata.gemxd.ClusterCallbacksImpl
```
hbhanawat: Unused imports?

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -132,12 +128,10 @@ object StoreUtils extends Logging {
...
       schema: Option[StructType],
       partitions: Int,
-      connProperties: ConnectionProperties): Map[InternalDistributedMember,
-      BlockManagerId] = {
+      connProperties: ConnectionProperties): Unit = {
     // TODO for SnappyCluster manager optimize this . Rather than calling this
-    val blockMap = new StoreInitRDD(sqlContext, table,
-      schema, partitions, connProperties).collect()
-    blockMap.toMap
+    new StoreInitRDD(sqlContext, table, schema, partitions, connProperties)
```
hbhanawat: Why do we still have it? The goal of this PR is to remove StoreInitRDD, isn't it? :-)

cluster/src/dunit/scala/io/snappydata/externalstore/ColumnTableDUnitTest.scala
```diff
@@ -550,8 +550,8 @@ class ColumnTableDUnitTest(s: String) extends ClusterManagerTestBase(s) {
 
     assert(r.length == 10012)
 
-    println("startSparkJob6 " + region.size())
-    println("startSparkJob6 " + shadowRegion.size())
+    println("startSparkJob7 " + region.size())
```
hbhanawat: Why do we need this change?

hbhanawat:
The changes look good.

cluster/src/dunit/scala/io/snappydata/externalstore/ColumnTableDUnitTest.scala
ashetkar: Not related to this PR, but this is inside def startSparkJob7 and not startSparkJob6.

cluster/src/dunit/scala/io/snappydata/externalstore/ExecutorMessageDUnitTest.scala
ashetkar: That is the second testcase which is currently commented out as it causes subsequent testcase to fail.

cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
ashetkar: Not unused, it is showing changes in this file because I had earlier made changes to it which were removed later.

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
ashetkar: This PR addresses the second part of that goal i.e. updating the block map outside StoreInitRDD. Will file another sub-task/PR for the first part i.e. getting rid of the need to boot up the store in embedded mode via StoreInitRDD. (I remember some tests failing when StoreInitRDD was removed, even with these changes).

cluster/src/dunit/scala/io/snappydata/externalstore/ExecutorMessageDUnitTest.scala
ashetkar: This was missing and hiding a small issue too. Thanks for pointing this out. Will update the code.

cluster/src/dunit/scala/io/snappydata/externalstore/ExecutorMessageDUnitTest.scala
ashetkar: That is the commented code, retained until SNAP-907 gets resolved.

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
ashetkar: Opened SNAP-906 to track and resolve the first part. I think it should not block this PR.

ashetkar:
@hbhanawat 
Incorporated your comments and clarified the concerns.

cluster/src/dunit/scala/io/snappydata/externalstore/ExecutorMessageDUnitTest.scala
hbhanawat: SNAP-907 says that "This test shuts down a vm during the testcase for some validation". I had mentioned somewhere else that vms should not be shutdown, you should rather stop the snappy component -fabricservice.stop. SNAP-907 is non issue, I believe your test should be fixed.

ashetkar: okay, the wording was not correct.
The testcase does not shuts down the vm but only stops the snappy process. I have tried the stopAny() method too with no luck before commenting this testcase. Also, during one of the stand-ups, @ymahajan mentioned seeing similar issue with other dunit testcase.

hbhanawat: Also, we should be very wary about start and stop issues in snappy as that has been the pain point for gemxd customers.

ashetkar:
The test issue has now been resolved. Merging the PR.

PR #282, opened 2016-06-20
Added support for collecting stats in the framework

swatisawant:

- Added support for collecting stats in the framework
- Modified the existing tests to use new methods for creating the configuration data for writting to the respective files under conf directory
- Added a check if in case appPropsForJobServer parameter is not provided in conf file. The method should use default APP_PROPS in this case

## Patch testing

Verified the changes by running sample.bt

## Other PRs 

No


KishorBachhav:
Merge It

PR #286, opened 2016-06-24
[SNAP-416] Fix for SNAP-416 Support for create table as Select

SachinJanani:

This PR resolves issue with Create table as Select when Using..Options is not provided

## Patch testing

Written a Unit test for this fix in RowTest.scala

## Other PRs 

N/A


ymahajan:
It looks like the target table will be created using default options. So will it create the target table as row table?  For example, in default case, we will create row table as target table even for the column table case. 
Does it make sense to re-use the base table's option here?

ymahajan:
We have discussed that for the default options, we are not changing anything there. So this PR looks good for merge.

PR #287, opened 2016-06-24
Added the conf param spark.sql.autoBroadcastJoinThreshold to -1 , so …

rishitesh:

  Added the conf param spark.sql.autoBroadcastJoinThreshold to -1 , so that broadcast join is disabled during the scope of the test. Or else the size of the table determines shuffle or broadcast. As we are using a small table size this is necessary.

## Patch testing
precheckin is clean

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


ymahajan:
Looks good Rishi. Please go ahead with merge asap, as this test is failing in precheckin.

PR #291, opened 2016-06-28
Gradle script fixes

ashetkar:
* Correcting the artifact names from snappydata-store* to
  gemfirexd* in various build.gradle files.
* In earlier commits, only jar names were renamed to
  snappydata-store-* and not artifact names.
* Fixes the issue of './gradlew precheckin -Pstore' not running
  unit tests of snappy-store.

## Patch testing
./gradlew precheckin (running)

## ReleaseNotes.txt changes
NA

## Other PRs 
NA

### DIFF
--- a/dtests/build.gradle
+++ b/dtests/build.gradle
@@ -26,8 +26,6 @@ dependencies {
     compile 'org.scala-lang:scala-library:' + scalaVersion
     compile 'org.scala-lang:scala-reflect:' + scalaVersion
     compile 'org.scala-lang:scala-compiler:' + scalaVersion
-    compile project(subprojectBase + 'gemfire-core')
-    compile project(subprojectBase + 'gemfire-tests')
 
     compile(group: 'com.databricks', name: 'spark-csv_2.10', version: '1.2.0') {
         exclude(group: 'org.scala-lang', module: 'scala-library')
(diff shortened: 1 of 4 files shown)


ymahajan:
changes look good Amogh.

dtests/build.gradle
```diff
@@ -26,8 +26,6 @@ dependencies {
     compile 'org.scala-lang:scala-library:' + scalaVersion
     compile 'org.scala-lang:scala-reflect:' + scalaVersion
     compile 'org.scala-lang:scala-compiler:' + scalaVersion
-    compile project(subprojectBase + 'gemfire-core')
-    compile project(subprojectBase + 'gemfire-tests')
```
sumwale: why this is removed?

sumwale:
@ashetkar I think we decided to also change the published artifact names hence had changed it. We should do it immediately after final release if not possible now.

ashetkar:
While trying to do the build with only the maven artifacts on master, observed that this was broken because (among other things) we had not published the artifacts with changed names until then.
So reverted these changes on master but put them on release branch. The release branch is now merged into master after publishing the artifacts.

dtests/build.gradle
ashetkar: It used to fail when there was not snappy-store checked out.

PR #293, opened 2016-06-30
[SNAP-859] Moved registration of InternalDistributedMember from Spark…

SachinJanani:

-Changes related to SNAP-859
-Moved registration of InternalDritributedManager to SnappyKryoRegistrator instead of spark code base
-Changed LeadImpl to register SnappykryoRegistrator
## Patch testing

Added DUnitTest for this change
## ReleaseNotes.txt changes

## Other PRs

Yes
 [snappy-aqp PR-70](https://github.com/SnappyDataInc/snappy-aqp/pull/70) ,[SnappySpark PR-37](https://github.com/SnappyDataInc/snappy-spark/pull/37)

### DIFF
--- a/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
@@ -357,9 +357,9 @@ class LeadImpl extends ServerImpl with Lead with Logging {
   def getSnappyKryoRegistrarClass(): String = {
 
     Try {
-      val miror = Class.forName("org.apache.spark.sql.execution.serializer.SnappyKryoAQPRegistrator")
+      val miror = Class.forName("org.apache.spark.sql.execution.serializer.SnappyAQPKryoRegistrator")
     } match {
-      case Success(v) => "org.apache.spark.sql.execution.serializer.SnappyKryoAQPRegistrator"
+      case Success(v) => "org.apache.spark.sql.execution.serializer.SnappyAQPKryoRegistrator"
       case Failure(_) => "io.snappydata.serializer.SnappyKryoRegistrator"
     }
   }
--- a/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
@@ -128,7 +128,7 @@ class LeadImpl extends ServerImpl with Lead with Logging {
       /**
        * If kryoserializer is being used then register snappyclasses for kryoserializer.
        * The snappyclasses will not get registered if user is creating its own spark context.
-       * In that case user has to specify the snappy kryo registrator registrator externally
+       * In that case user has to specify the snappy kryo registrator externally
        */
       if (conf.contains("spark.serializer") && conf.get("spark.serializer").equals("org.apache.spark.serializer.KryoSerializer")) {
         conf.set("spark.kryo.registrator", getSnappyKryoRegistrarClass())
--- a/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
@@ -21,6 +21,7 @@ import java.util.Properties
 import java.util.concurrent.atomic.AtomicReference
 
 import scala.collection.JavaConverters._
+import scala.util.{Failure, Success, Try}
 
 import akka.actor.ActorSystem
 import com.gemstone.gemfire.distributed.internal.DistributionConfig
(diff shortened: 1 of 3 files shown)


ymahajan:
@SachinJanani  We need to identify if there are more classes like InternalDistributedMember which are not KryoSerializable. Can you run precheckin with KryoSerializer enabled in SnappyContext/SparkContext ?

ymahajan:
@SachinJanani Please add a test with sample table and approx query.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -124,6 +125,15 @@ class LeadImpl extends ServerImpl with Lead with Logging {
         conf.set(key, v)
       })
 
+      /**
+       * If kryoserializer is being used then register snappyclasses for kryoserializer.
+       * The snappyclasses will not get registered if user is creating its own spark context.
+       * In that case user has to specify the snappy kryo registrator registrator externally
+       */
```
ymahajan: duplicate registrator word

hbhanawat:
Looks good.

SachinJanani:
@ymahajan As discussed ran the precheckin by enabling kryoserializer and all tests passed

ymahajan:
@SachinJanani looks good for merge.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -124,6 +125,15 @@ class LeadImpl extends ServerImpl with Lead with Logging {
         conf.set(key, v)
       })
 
+      /**
+       * If kryoserializer is being used then register snappyclasses for kryoserializer.
+       * The snappyclasses will not get registered if user is creating its own spark context.
+       * In that case user has to specify the snappy kryo registrator externally
+       */
+      if (conf.contains("spark.serializer") && conf.get("spark.serializer").equals("org.apache.spark.serializer.KryoSerializer")) {
+        conf.set("spark.kryo.registrator", getSnappyKryoRegistrarClass())
```
rishitesh: What happens when user also specifies any registrator ? Can we have multiple registrator ? I suppose not. Can we not use spark.kryo.classesToRegister spark conf property. This will enable us to only append our classes. We can keep a list of classes we intent to register in one place. Just read the conf property and append our list to it.

SachinJanani: @rishitesh Please correct me if i am wrong, as far as i know user can only specify registrator if he is creating the `SparkConf/SparkContext` by himself which we are not recommending right?. As mentioned in my comments in `LeadImpl.scala` if user is creating his own SparkContext then he has to register our registartor which he can extend to register his own classes or he can use `conf.registerKryoClasses`  method

rishitesh: Why user can not create a Spark conf ? I thought lead node has options for that. There must be.
It will be too much for user to extend our registrator. We will go by this route if no option left for us. Can we find out why we need to register our classes in kryo ? If ser/deser is in our control why we can't emulate SparkSQlSerializer, which just extends KryoSerializer with its own addition..  

For CachedBatch serialization we can surely extend SparkSQLSerializer. What's the deal with AQP related classes . Can we please find out ?

rishitesh: Please check https://github.com/EsotericSoftware/kryo#registration also, where different registration methods are explained. It would be good if we can keep look and feel of Spark as it is, and can work through our inner details in non invasive way. If not anyway we have your code to fallback on.

hbhanawat:
@rishim If we do not kryo register, the full class name will be written with the serialized data. This is unneccessary overhead. So, IMO, kryo.register should be called somehow. Emulating SparkSQlSerializer will be difficult because that would need changes in Spark so that it uses our version of serializer and not SparkSqlSerializer. 

So we are left with the option of either spark.kryo.classesToRegister or spark.kryo.registrator. I didn't like the option of spark.kryo.classesToRegister because in case of non embedded mode, the user will have to specify the list of classes that we would like to be kryo registered. So it is better to create a spark.kryo.registrator and ask user to specify spark.kryo.classesToRegister for whatever classes he wants to be registered. For advanced users, they will have to write their own registrator and aggregation or inheritance to include SnappyRegistrator. 

Btw, @SachinJanani I assume you have tested it in non embedded mode?

SachinJanani:
@rishitesh @hbhanawat @ymahajan As discussed incorporated review comments.following are the related PRs:
[PR-293](https://github.com/SnappyDataInc/snappydata/pull/293) ,[SnappySpark PR-37](https://github.com/SnappyDataInc/snappy-spark/pull/37),[snappy/aqp-70](https://github.com/SnappyDataInc/snappy-aqp/pull/70),[snappy/store](https://github.com/SnappyDataInc/snappy-store/pull/93)

sumwale:
@hbhanawat @SachinJanani We have another option that I am taking in Spark 2.0 merge. Own serializer implementation itself (for pooling SNAP-944) so user can specify that instead of KryoSerializer.

sumwale:
Custom kryo serialization has been implemented as part of SNAP-1136 which corrects all issues with using kryo for all of data, closures and netty so this PR is obsolete and can be closed.

SachinJanani:
Closing this PR as it is obsolete

PR #294, opened 2016-07-01
Added support for server and lead members HA tasks in framework. Also, added cluster start and stop scripts.

swatisawant:

  - Added support for server HA.
- Currently, support is added for only NICE_KILL as a stop mode.
- Added support to backup and restore the original servers conf file. This will be required at the end of the test for stopping all servers which have been started in the test.
 - Added cluster startup and stop scripts for both rowstore and snappy.
- Added respective bt files.
  - Added support for primary lead HA.
- Added sample tests to verify the server and lead HA.

## Patch testing

Created sample tests for snappy store HA and lead HA and verified the changes by running those tests.

## ReleaseNotes.txt changes

NA

## Other PRs 

No


dtests/src/test/java/io/snappydata/hydra/cluster/SnappyPrms.java
```diff
@@ -85,6 +85,13 @@
     public static Long useSplitMode;
 
     /**
+     * (boolean) - whether stop mode needs to be checked before deleting the config data if already exists.
+     * This is required in case user wants to start the cluster and then stop the same later on using different script.
+     * In this case, test should not delete the existing configuration data created by previous test.
+     */
+    public static Long isStopMode;
```
nthanvi: why this is needed? In any condition test should not delete any configuration data of old  test.

dtests/src/test/java/io/snappydata/hydra/cluster/SnappyTest.java
```diff
@@ -447,6 +457,13 @@ protected void writeConfigData(String fileName, String logDir) {
         }
     }
 
+    protected void writeNodeConfigData(String fileName, String nodeLogDir) {
+        String filePath = productConfDirPath + fileName;
+        File file = new File(filePath);
+        Set<String> fileContent = new LinkedHashSet<String>();
```
nthanvi: filecontent not used anywhere

dtests/src/test/java/io/snappydata/hydra/cluster/SnappyPrms.java
swatisawant: This is required. As per the scripts(start-servers.sh, start-locators.sh, start-leads.sh) requirement, the configuration data has to be available in respective conf files under product/conf directory. So the config data specific to each test is not actually created in test folder.
In my test run, it creates the required conf files specific to test configurations under the $product/conf directory(e.g. servers, locators and leads files). So, for each test there should be new conf files for different configurations.
If I don't delete the conf files then how would, I create the configurations specific to the current test?

Also, the test config data created by any test run should not be lying under the $product/conf directory even though I don't want to run another test, right?

PR #297, opened 2016-07-05
SNAP-656 Delink RDD partitions from buckets 

ymahajan:
- set default rdd partitions as number of cores in cluster 
- map buckets to target partitions using round robin assignment
- colocated tables should have same partitions and queries should work fine. 
- split mode support 

## ReleaseNotes.txt changes
yes

## Other PRs 
Store - https://github.com/SnappyDataInc/snappy-store/pull/85
Spark - https://github.com/SnappyDataInc/spark/pull/4
SnppyData - https://github.com/SnappyDataInc/snappydata/pull/297


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -90,7 +92,15 @@ class BaseColumnFormatRelation(
   @transient protected lazy val region = Misc.getRegionForTable(resolvedName,
     true).asInstanceOf[PartitionedRegion]
 
-  override lazy val numPartitions: Int = region.getTotalNumberOfBuckets
+  override lazy val numPartitions: Int = {
+//    val region = Misc.getRegionForTable(resolvedName, true).
+//        asInstanceOf[PartitionedRegion]
+//    region.getTotalNumberOfBuckets
+    val numCores = Runtime.getRuntime.availableProcessors()
```
rishitesh: This is evaluated at driver node. We need to consider sever nodes. Driver node num processors is not useful to us. Can you please see SchedulerBackend.defaultParallelism. That  takes total cores for slaves into consideration.
Catch however is spark.default.parallelism gets priority and if somebody configures bad we will suffer.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -339,6 +349,8 @@ class BaseColumnFormatRelation(
         val sql = s"CREATE TABLE $tableName $schemaExtensions " + " DISABLE CONCURRENCY CHECKS "
         logInfo(s"Applying DDL (url=${connProperties.url}; " +
             s"props=${connProperties.connProps}): $sql")
+        println(s"Applying DDL (url=${connProperties.url}; " +
```
rishitesh: Remove println

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatRelation.scala
```diff
@@ -138,11 +139,20 @@ class RowFormatRelation(
...
-      case pr: PartitionedRegion => pr.getTotalNumberOfBuckets
-      case _ => 1
+      case pr: PartitionedRegion =>
+        getNumPartitions
+      case _ =>
+        1
     }
   }
 
+  def getNumPartitions : Int = {
```
rishitesh: See comments above

core/src/test/scala/org/apache/spark/sql/NorthWindTest.scala
```diff
@@ -0,0 +1,1481 @@
...
+package org.apache.spark.sql
+
+import io.snappydata.SnappyFunSuite
+import org.apache.spark.Logging
+import org.apache.spark.sql.execution.aggregate.{SortBasedAggregate, TungstenAggregate}
+import org.apache.spark.sql.execution.joins._
+import org.apache.spark.sql.execution.{LocalTableScan, PartitionedPhysicalRDD, Project, Sort}
+import org.scalatest.{BeforeAndAfter, BeforeAndAfterAll}
+
+class NorthWindTest
```
rishitesh: Remove customer name from class.

core/src/test/scala/org/apache/spark/sql/NorthWindTest.scala
```diff
@@ -0,0 +1,1481 @@
...
+    assert(df.rdd.partitions.length == numPartitions,
+      "Mismatch got df.rdd.partitions.length ->" + df.rdd.partitions.length +
+        " but expected numPartitions ->" + numPartitions)
+  }
+
+  private def createAndLoadColocatedTables(snc: SnappyContext): Unit = {
+
+    val regions = snc.read.format("com.databricks.spark.csv")
+      .option("header", "true")
+      .load(getClass.getResource("/northwind/regions.csv").getPath)
```
rishitesh: remove specific names

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -98,25 +101,74 @@ object StoreUtils extends Logging {
...
+      val bucketSet = {
+        if (serverToBuckets.contains(owner)) serverToBuckets.get(owner).get
+        else new mutable.HashSet[Int]()
+      }
+      bucketSet += p
+      totalBuckets += p
+      serverToBuckets put(owner, bucketSet)
+    }
+    val numCores = Runtime.getRuntime.availableProcessors()
+    val numServers = GemFireXDUtils.getGfxdAdvisor.adviseDataStores(null).size()
```
rishitesh: This part wont be straight forward. If executor core is not set I think default is all the cores of a server. If different configuration machines are used our calculation will go for toss. e.g. if a server is with 8 cores and another with 16 cores, then we will loose processing power of some of the servers.

tests/common/src/main/resources/northwind/employees.csv
```diff
@@ -0,0 +1,10 @@
+EmployeeID,LastName,FirstName,Title,TitleOfCourtesy,BirthDate,HireDate,Address,City,Region,PostalCode,Country,HomePhone,Extension,Photo,Notes,ReportsTo,PhotoPath
```
rishitesh: Is it ok to use this data?

ymahajan: Yes, NorthWind Trader's schema is widely used in public domain, especially with the database research students. Also this data is shipped with Microsoft products.

core/src/test/scala/org/apache/spark/sql/NorthWindTest.scala
ymahajan: NorthWind Trader's is fictitious company name, not any real SnappyData customer's name.

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
ymahajan: We have assumed that, the cluster will be homogeneous with respect number of cores on each server. I will check if there is a way to handle heterogeneous case also.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
ymahajan: Removed println

PR #300, opened 2016-07-07
Reverse the order of stopping servers in ./sbin/snappy-stop-all.sh

rk-geode:

Currently the order of stopping servers is as specified in /conf/servers file. We use the same order during ./sbin/snappy-start-all.sh as well, as a result of this the last server stopped is also the last server started.
Changes made to stop the servers in reverse order as specified in /conf/servers.

## Patch testing
Verified changes with running snappy-start-all, snappy-stop-all and snappy-status-all .

## ReleaseNotes.txt changes

NA

## Other PRs 

No

### DIFF
--- a/cluster/sbin/snappy-nodes.sh
+++ b/cluster/sbin/snappy-nodes.sh
@@ -167,15 +167,30 @@ function startComponent() {
 }
 
 index=1
-
+declare -a arr
 if [ -n "${HOSTLIST}" ]; then
   while read slave || [[ -n "${slave}" ]]; do
     [[ -z "$(echo $slave | grep ^[^#])" ]] && continue
+    arr+=("${slave}");
     host="$(echo "$slave "| tr -s ' ' | cut -d ' ' -f1)"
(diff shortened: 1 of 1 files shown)


cluster/sbin/snappy-nodes.sh
```diff
@@ -167,15 +167,30 @@ function startComponent() {
...
   done < $HOSTLIST
+  if echo $"${@// /\\ }" | grep -wq "stop"; then
+    line=${#arr[@]}
+    if [ $((index-1)) -eq $line ]; then
+      for (( i=${#arr[@]}-1 ; i>=0 ; i-- )) ; do
+        ((index--))
+        CONF_ARG=${arr[$i]}
+        host="$(echo "$CONF_ARG "| tr -s ' ' | cut -d ' ' -f1)"
+        args="$(echo "$CONF_ARG "| tr -s ' ' | cut -d ' ' -f2-)"
+        startComponent "$@"
```
hbhanawat: startComponent looks odd here. Rename it to 'execute' or any other better name.

hbhanawat:
Looks good.

PR #301, opened 2016-07-08
[SNAP-822][SNAP-415] Changes related  to Insert into.. select from

SachinJanani:
-insert into select was failing if one of the table in query was column table this is due the change in the grammar which was causing queries releated to column table to be executed on store and were not routed to spark os change the grammar
 
## Patch testing
Added Unit tests to cover all the scenarios

## ReleaseNotes.txt changes

## Other PRs 
Yes
[Snappy-sotre-PR-84](https://github.com/SnappyDataInc/snappy-store/pull/84)


ymahajan:
put into is not supported/tested since column tables don't support updates yet.

ymahajan:
looks good for merge

PR #309, opened 2016-07-19
[SNAP 863] Initial snappydata interpreter for zeppelin and changes for starting interpreter along with lead

SachinJanani:
These PR contains the following changes:
* Zeppelin Snappydata and Snappydatasql interpreters
* Starting zepplein interpreter thread along with lead node
* Configurations related to zeppelin interpreters
* Configure Repl classloader for Dynamic class loading from zeppelin

## Patch testing
Manual tested interpreters from zeppelin.It works as required

## ReleaseNotes.txt changes
No .Document will be added in the related PR

## Other PRs 
No


cluster/src/main/java/org/apache/zeppelin/interpreter/SnappydataSqlZeppelinInterpreter.java
```diff
@@ -0,0 +1,235 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.apache.commons.lang.StringUtils.containsIgnoreCase;
+
+/**
+ * Snappydatasql interpreter used to connect snappydata cluster using jdbc for performing sql
+ * queries
+ * 
+ * @author sachin
```
rishitesh: remove author info.

cluster/src/main/java/org/apache/zeppelin/interpreter/SnappydataZeppelinInterpreter.java
```diff
@@ -0,0 +1,800 @@
+/*
```
rishitesh: Check the licensing info. I think both Apache and SnappyData license would be applicable here.

cluster/src/main/java/org/apache/zeppelin/interpreter/SnappydataZeppelinInterpreter.java
```diff
@@ -0,0 +1,800 @@
...
+import scala.tools.nsc.settings.MutableSettings;
+import scala.tools.nsc.settings.MutableSettings.BooleanSetting;
+import scala.tools.nsc.settings.MutableSettings.PathSetting;
+
+
+/**
+ * Most of the contents of this class is borrowed from Spark Interpreter in zeppelin
+ * https://github.com/apache/zeppelin/blob/master/spark/src/main/java/org/apache/zeppelin/spark/SparkInterpreter.java
+ */
+public class SnappydataZeppelinInterpreter extends Interpreter {
```
rishitesh: SnappyData is the convention instead of Snappydata.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -342,6 +354,23 @@ class LeadImpl extends ServerImpl with Lead with Logging {
...
+   * As discussed by default zeppelin interpreter will be enabled.User can disable it by
+   * setting "zeppelin.interpreter.enable" to false in leads conf file.User can also specify
+   * the port on which intrepreter should listen using  property zeppelin.interpreter.port
+   * @param bootProperties
+   */
+  private def checkAndStartZeppelinInterpreter(bootProperties: Properties): Unit = {
+    //As discussed ZeppelinRemoteInterpreter Server will be enabled by default.
+    if (bootProperties.getProperty(Constant.ENABLE_ZEPPELIN_INTERPRETER, "true").equals("true")) {
+      val port = bootProperties.getProperty(Constant.ZEPPELIN_INTERPRETER_PORT, "3768").toInt
+      val remoteInterpreterServer: SnappyRemoteInterpreterServer = new SnappyRemoteInterpreterServer(port)
```
rishitesh: Rather than calling it SnappyRemoteInterpreterServer can't we call it only SnappyInterpreterServer. As for lead node its not remote. Only for Zeppelin its remote.

cluster/src/main/resources/snappydata/interpreter-setting.json
```diff
@@ -0,0 +1,70 @@
+[
```
rishitesh: Can user change this file ? I guess not. Then why to keep in a Json file. Does Zeppelin mandates it ?

cluster/src/main/java/org/apache/zeppelin/interpreter/SnappydataSqlZeppelinInterpreter.java
```diff
@@ -0,0 +1,235 @@
...
+  public SnappydataSqlZeppelinInterpreter(Properties property) {
+    super(property);
+
+    new Thread() {
+      public void run() {
+        while (true) {
+          InterpreterContextRunner runner;
+          try {
+            runner = queue.take();
+            Thread.sleep(400);
```
rishitesh: Why this sleep is needed ?

cluster/src/main/java/org/apache/zeppelin/interpreter/SnappydataSqlZeppelinInterpreter.java
```diff
@@ -0,0 +1,235 @@
...
+          for (int i = 1; i < md.getColumnCount() + 1; i++) {
+            if (i > 1) {
+              msg.append(TAB);
+            }
+            msg.append(replaceReservedChars(isTableType, md.getColumnName(i)));
+          }
+          msg.append(NEWLINE);
+
+          int displayRowCount = 0;
+          while (resultSet.next() && displayRowCount < getMaxResult()) {
```
rishitesh: getMaxResult can be called once before the loop. It will avoid calling a  method multiple times.

cluster/build.gradle
```diff
@@ -203,3 +206,41 @@ shadowJar {
     )
   }
 }
+
+
+task generateSnappydataZeppelinInterpreterJar(type: Jar, dependsOn: compileScala) {
```
rishitesh: This task should be integrated with the task which generates our distribution. I don't remember exactly. Please check with Amogh.

PR #310, opened 2016-07-22
Snap 293

nthanvi:
1. Added support for dynamic Jar loading in snappy. The idea here is to support the existing snappy command for loading  the  jar file. For that lead node driver and executor class loader are modified in such a way that they look for the jar in the snappy-store if it is not found in regular class path.

2. Apart form above change there is another problem that once the job is submitted through the job server and if user make any changes to the jar related to submitted application and try to rerun again then executors are not able to pick the new jar, they pick the old jar as they are long running and  old jar is still in ClassPath. to fix this issue , a new DynamicURLClassLoader implementation  is done which maintains a Loader per Jar.

3. The current implementation of SnappySQLJOB and JavaSnappySQLJob.java is kind of inconsistent in sence of method to overload and return types.
Now only two methods are exposed to the user runSnappyJob and isValidJob  with a single return Type SnappyValidation so user has a small learning curve and limited entry points.

4. all the existing jobs are modified to use the new Method Implementations of SnappySQLJob and SnappyStreamingJob etc.

5. code formatting  for the existing jobs.

## Patch testing

dunit , junit , clean precheckin with store

## ReleaseNotes.txt changes

No. but will update the related *md files for same. 

## Other PRs 
https://github.com/SnappyDataInc/spark-jobserver/pull/1
https://github.com/SnappyDataInc/snappy-spark/pull/39
https://github.com/SnappyDataInc/snappy-store/pull/88


cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -0,0 +1,41 @@
...
+  private val overwriteFiles = env.conf.getBoolean("spark.files.overwrite", false)
+
+  override def createClassLoader(urls: Array[URL],
+      parentLoader: ClassLoader, userClassPathFirst: Boolean): MutableURLClassLoader = {
+    new DynamicURLClassLoader(urls, parentLoader, parentFirst = !userClassPathFirst , overwriteFiles)
+  }
+
+}
+
+
```
sumwale: This override can also be added for IsolatedClientLoader.classLoader. It is used for any external JDBC URLs apart from our gemxd drivers (e.g. if user wishes to use it to load data into gemxd). If it is not added there, we need docs on how user should go about doing it (perhaps JDBC drivers have to be in SPARK_DIST_CLASSPATH while other parts will be fine with dynamic install-jar).

sumwale:
@rishitesh Please review the Java job API changes in this.

cluster/src/main/java/org/apache/spark/streaming/JavaSnappyStreamingJob.java
```diff
@@ -18,28 +18,33 @@
...
+import spark.jobserver.SparkJobBase;
 import spark.jobserver.SparkJobValidation;
+import org.apache.spark.util.SnappyUtils;
+import org.apache.spark.util.Utils;
 
-public abstract class JavaSnappyStreamingJob implements SnappyStreamingJob {
+public abstract class JavaSnappyStreamingJob implements SparkJobBase {
 
-  abstract public  Object runJavaJob(JavaSnappyStreamingContext snc, Config jobConfig);
+  abstract public Object runSnappyJob(JavaSnappyStreamingContext snc, Config jobConfig);
```
sumwale: Why this rename? I suppose the idea was to emphasize that the code needs to use the Java wrapper JavaSnappyStreamingContext (and not the regular SnappyStreamingContext). @rishitesh your thoughts?

cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
nthanvi: I will remove the following line 
private val overwriteFiles = env.conf.getBoolean("spark.files.overwrite", false)

as it is not needed after the changes in the job server. The Job server always create a new jar file where we have added our own suffix in the name so now executor knows that the jar belongs to the same job and it can  reload it. executors can also copy this jar as it is a new jar for them.

cluster/src/main/java/org/apache/spark/streaming/JavaSnappyStreamingJob.java
```diff
@@ -18,28 +18,33 @@
...
       Config jobConfig);
 
   @Override
-  public Object runJob(Object sc, Config jobConfig) {
-    return runJavaJob(new JavaSnappyStreamingContext((SnappyStreamingContext)sc), jobConfig);
+  final public SparkJobValidation validate(Object sc, Config config) {
+    ClassLoader parentLoader = Utils.getContextOrSparkClassLoader();
+    ClassLoader currentLoader = SnappyUtils.getSnappyStoreContextLoader(parentLoader);
+    Thread.currentThread().setContextClassLoader(currentLoader);
+    return SnappyJobValidate.validate(isValidJob(new JavaSnappyStreamingContext((SnappyStreamingContext)sc), config));
```
sumwale: line looks too long; please break

cluster/src/main/scala/org/apache/spark/util/DynamicURLClassLoader.scala
```diff
@@ -0,0 +1,156 @@
...
+          .getOrElse(loadFromStore(className, false)
+              .getOrElse(loadFromParent(parent, className, true).get))
+    }
+
+    if (resolve) {
+      resolveClass(c)
+    }
+
+    c
+  }
```
sumwale: This seems to be using Spark loader first, then fallback to store loader. I think there will be less confusion if we use only ours (so job-server can do an install-jar for a new provided jar). Lets discuss.

Secondly why do we need to do explicit loadFromJars too? Will loadFromParent not suffice?

cluster/src/main/scala/org/apache/spark/util/DynamicURLClassLoader.scala
```diff
@@ -0,0 +1,156 @@
...
+      } catch {
+        case _: Throwable => {
+          // do nothing.
+        }
+      }
+    }
+
+    logInfo(" Adding new Loader with key:" + key + " with File: " + url.getPath )
+    urlList.put(key, loader)
+  }
```
sumwale: Can you explain what this override is doing?

cluster/src/main/scala/org/apache/spark/util/DynamicURLClassLoader.scala
```diff
@@ -0,0 +1,156 @@
...
+
+  private def getClassLoader(url: URL): MutableURLClassLoader = {
+    if (parentFirst) new MutableURLClassLoader(Array(url), parent)
+    else new ChildFirstURLClassLoader(Array(url), parent)
+  }
+
```
sumwale: Why double wrapping parent? Don't understand this additional URL handling here or above.

cluster/src/main/scala/org/apache/spark/util/DynamicURLClassLoader.scala
```diff
@@ -0,0 +1,156 @@
...
+      urlList.asScala.map(tuple => {
+        tuple._2.getResources(name).asScala
+            .filter(!parent.getResources(name).asScala.contains(_)).asJavaEnumeration
+      }).toArray
+
+    val parentResources = Array(parent.getResources(name).asInstanceOf[util.Enumeration[_]])
+
+    new CompoundEnumeration(childResources ++ parentResources)
+  }
+
```
sumwale: don't understand the purpose of this

cluster/src/main/scala/org/apache/spark/util/DynamicURLClassLoader.scala
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    c match {
+      case Some(_) => c
+      case None =>
+        if (throwException) throw new ClassNotFoundException(className)
+        else None
+    }
+
+  }
```
sumwale: Don't understand the purpose of this. Is this something to do with how job-server is loading and does it have its own classloader that this is trying to emulate (and add to)?

cluster/src/main/scala/org/apache/spark/util/SnappyUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+      override def loadClass(name: String): Class[_] = {
+        try {
+          parent.loadClass(name)
+        } catch {
+          case cnfe: ClassNotFoundException =>
+            Misc.getMemStore.getDatabase.getClassFactory.loadClassFromDB(name)
+        }
+      }
+    }
+
```
sumwale: This is all that should have been required in DynamicURLClassLoader too. This inconsistent behaviour of DynamicURLClassLoader vs these calls will probably cause trouble.

Also from what I saw the parent here will be DynamicURLClassLoader itself, no?

cluster/src/main/scala/org/apache/spark/util/SnappyUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+    val files2 = for ((childName, baseName) <- classNamesWithBase) yield {
+      TestUtils.createCompiledClass(childName, tempDir, toStringValue, baseName, classpathUrls)
+    }
+    val jarFile = if (jarName.isEmpty) {
+      new File(tempDir, "testJar-%s.jar".format(System.currentTimeMillis()))
+    }
+    else new File(tempDir, jarName.format(System.currentTimeMillis()))
+    TestUtils.createJar(files1 ++ files2, jarFile)
+  }
+
```
sumwale: Since this is being used only be tests, move this to io.snappydata.util.TestUtils

cluster/src/test/scala/io/snappydata/benchmark/snappy/TPCH_Snappy_Query.scala
```diff
@@ -1,123 +1,141 @@
...
+      return new SnappyJobInvalid("Specify Query number to be executed")
+    }
+
+    println(s"tempqueries : $tempqueries")
+
+    queries = tempqueries.split(",")
+
+    SnappyJobValid()
+  }
+}
```
sumwale: Check with Kishor about these large scale formatting changes. He may have many uncommitted local changes that will all conflict with above. I think it will be better to let him correct the formatting.

sumwale:
Haven't looked at the tests yet but looked through pretty much rest of the code. Please ensure we have tests for:
a) install/replace/drop from job-server (is drop possible from job-server?)
b) install/replace/drop using the usual GemXD procedures
c) install/replace/drop of JDBC drivers

We also need docs for how user should install/replace/drop jars as well as JDBC drivers which is common case users try to load import data into snappydata. If c) is not possible, then we need to document and test that too (SPARK_DIST_CLASSPATH)

cluster/src/main/java/org/apache/spark/streaming/JavaSnappyStreamingJob.java
rishitesh: It was more to do with type resolution in case of scala generic type. The API proposed is looking cleaner if its running all the tests. As use only have to deal with two methods. Only input types will change.

nthanvi:
Added another pull request after incorporating the review comments and 2.0 merge.
#337

For some open issues with PR open the Jira https://jira.snappydata.io/browse/SNAP-999?filter=-1

PR #311, opened 2016-07-23
[SNAP-933] use CREATE EXTERNAL TABLE for non-snappydata tables

sumwale:

The API already requires createExternalTable for non-snappydata tables
while for snappydata provided tables it requires createTable. This change
makes the DDL consistent with that convention adding CREATE EXTERNAL TABLE
for those.

Fixed the flag handling of onlyBuiltIn and onlyExternal to be consistent.

Also made the other table related DDLs consistent to not require TEMPORARY.

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


rishitesh:
Looks good Sumedh.

PR #312, opened 2016-07-26
The result of join of replicated tables should have a local join with any other table

hbhanawat:
The result of join of replicated tables should have a local join with any other table.

## Patch testing

Added a unit test that analyzes the plan to ensure that no sortMergeJoin is involved with multiple replicated tables and one partitioned table.


rishitesh:
Looks good for merge.

PR #317, opened 2016-07-28
[SNAP-950] Multiple optimizations seen during profiling

sumwale:

- CachedBatch creation: Use direct AbstractCompactExecRow.getAs* calls to get final java
  values/objects instead of expensive ExecRow.getColumn().get* calls that create intermediate DVDs
- CachedBatch creation: avoid switch on DataType+metadata for every row. Instead cache the end types
  and use a fast switch on ints (code generation might be overkill and may not buy much)
- Skip stats object (avoids its serialization/deserialization cost) for now in CachedBatch creation
  since it is not being used
- Use SparkShellRowRDD for split mode execution over row tables (currently only
    done for row buffer of a column table)
- Push down only those filters to row tables that can use indexes.
- Added support for push down of more filters like IN, AND, OR, LIKE for
  row tables (those that can use indexes)
- Spark row table iteration: use direct access to internal AbstractCompactExecRow
  where possible with getAs* calls

## Patch testing

precheckin and store precheckin

## ReleaseNotes.txt changes

NA

## Other PRs

Created over previous PR #311 for SNAP-933

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/CachedBatchCreator.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/CachedBatchCreator.scala
@@ -16,150 +16,32 @@
  */
 package org.apache.spark.sql.execution.columnar
 
-import java.sql.Types
 import java.util.UUID
 
 import scala.collection.mutable
-import scala.collection.mutable.ArrayBuffer
 
 import com.pivotal.gemfirexd.internal.engine.access.heap.MemHeapScanController
 import com.pivotal.gemfirexd.internal.engine.store.AbstractCompactExecRow
-import com.pivotal.gemfirexd.internal.iapi.sql.execute.ExecRow
 import com.pivotal.gemfirexd.internal.iapi.store.access.ScanController
 
-import org.apache.spark.sql.catalyst.InternalRow
-import org.apache.spark.sql.catalyst.expressions.{SpecificMutableRow, UnsafeArrayData, UnsafeMapData, UnsafeRow}
-import org.apache.spark.sql.catalyst.util.DateTimeUtils
-import org.apache.spark.sql.collection.UUIDRegionKey
-import org.apache.spark.sql.types._
-import org.apache.spark.unsafe.Platform
-import org.apache.spark.unsafe.types.UTF8String
+import org.apache.spark.sql.catalyst.expressions.SpecificMutableRow
+import org.apache.spark.sql.execution.CompactExecRowToMutableRow
+import org.apache.spark.sql.types.StructType
 
-class CachedBatchCreator(
+final class CachedBatchCreator(
     val tableName: String, // internal column table name
     val userTableName: String, // user given table name (row buffer)
-    val schema: StructType,
+    override val schema: StructType,
     val externalStore: ExternalStore,
     val columnBatchSize: Int,
-    val useCompression: Boolean) {
-
-  type JDBCConversion = (SpecificMutableRow, ExecRow, Int) => Unit
-
-  def createInternalRow(execRow: ExecRow,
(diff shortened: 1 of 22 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/CachedBatchCreator.scala
```diff
@@ -16,150 +16,32 @@
...
-      }
-      i = i + 1
-    }
-    mutableRow
-  }
+    val useCompression: Boolean) extends CompactExecRowToMutableRow {
 
   def createAndStoreBatch(sc: ScanController, row: AbstractCompactExecRow,
-      batchID: UUID, bucketID: Int): mutable.HashSet[Any] = {
+      batchID: UUID, bucketID: Int): mutable.HashSet[AnyRef] = {
```
rishitesh: This return type looks a bit redundant. As we are converting them again to Java types by JavaConversions, we can get the Java types at the first place. Can reduce one conversion.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -273,6 +276,56 @@ object ExternalStoreUtils {
       })
   }
```
rishitesh: This method can be made equal to BaseRelation.unhandledFilters. This wont give much benefit for column table, but for row table it will skip the Spark layer filter.

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatRelation.scala
```diff
@@ -64,28 +67,65 @@ class RowFormatRelation(
...
+      while (itr.hasNext) {
+        // first column of index has to be present in filter to be usable
+        val indexCols = itr.next().getIndexDescriptor.baseColumnPositions()
+        cols += baseColumns(indexCols(0))
+      }
+    }
+    cols
+  }
+
+  override def unhandledFilters(filters: Array[Filter]): Array[Filter] =
```
rishitesh: Ok, Ignore the earlier comment.

rishitesh: We don't handle unHandledPredicates in StoreDataSourceStrategy. Can be another follow up check in to this one. Will track that.

core/src/main/scala/org/apache/spark/sql/execution/columnar/CachedBatchCreator.scala
sumwale: done

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatRelation.scala
sumwale: done

sumwale:
@rishitesh, pushed changes for StoreDataSourceStrategy and HashSet. Please see.

sumwale:
@rishitesh merging now for move to spark 2.0 (will cause conflicts esp SNAP-933). Will incorporate any more review comments in 2.0 merge branch.

jramnara:
does this change provide for predicate pushdown to our external store? maybe just for Row tables?

rishitesh:
Yes for row store, and row buffer part of column tables. Also only those filters which can be satisfied by an index rather than a scan. Spark does the scan filter anyway.

sumwale:
@jramnara Yes only for row tables for now which can be satisfied by index (else Spark scan is more efficient). Also avoids double scan i.e. remove filters that have been successfully pushed down from Spark layer filters.

Column table filter pushdown will be handled separately.

PR #318, opened 2016-08-02
Snap 947

nthanvi:
disable the  concurrency checks for column table.

## Patch testing

added JUnit and precheckin

## ReleaseNotes.txt changes

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/94

### DIFF
--- a/core/src/main/scala/io/snappydata/util/ServiceUtils.scala
+++ b/core/src/main/scala/io/snappydata/util/ServiceUtils.scala
@@ -97,8 +97,8 @@ object ServiceUtils {
 
     "jdbc:" + Constant.SNAPPY_URL_PREFIX + (if (locatorUrl.contains(",")) {
       locatorUrl.substring(0, locatorUrl.indexOf(",")) +
-          ";secondary-locators=" + locatorUrl.substring(locatorUrl.indexOf(",") + 1)
-    } else locatorUrl)
+          "/;secondary-locators=" + locatorUrl.substring(locatorUrl.indexOf(",") + 1 ) + "/"
+    } else locatorUrl +"/")
   }
 
   def clearStaticArtifacts(): Unit = {
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
@@ -155,7 +155,7 @@ object ExternalStoreUtils {
         // Already connected to SnappyData in embedded mode.
         Constant.DEFAULT_EMBEDDED_URL + ";host-data=false;mcast-port=0"
       case SplitClusterMode(_, _) =>
-        ServiceUtils.getLocatorJDBCURL(sc) + "/route-query=false"
+        ServiceUtils.getLocatorJDBCURL(sc) + ";route-query=false"
       case ExternalEmbeddedMode(_, url) =>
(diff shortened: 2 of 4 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -155,7 +155,7 @@ object ExternalStoreUtils {
         // Already connected to SnappyData in embedded mode.
         Constant.DEFAULT_EMBEDDED_URL + ";host-data=false;mcast-port=0"
       case SplitClusterMode(_, _) =>
-        ServiceUtils.getLocatorJDBCURL(sc) + "/route-query=false"
+        ServiceUtils.getLocatorJDBCURL(sc) + ";route-query=false"
```
kneeraj: Why was this not showing error earlier. Any idea? The accepts url was returning ok for the first part of the url but the property route-query=false was not being set as intended. What problems surfaced due to this.

core/src/test/scala/org/apache/spark/sql/store/ColumnTableTest.scala
```diff
@@ -106,6 +106,37 @@ class ColumnTableTest
...
+    assert (Misc.getRegionForTable(table , true).getAttributes.getConcurrencyChecksEnabled == false)
+
+    snc.dropTable(table)
+
+    snc.sql(s"create table $table (col1 int) using row " +
+        s"options(PERSISTENT 'SYNCHRONOUS' , PARTITION_BY 'COL1')" )
+
+    assert (Misc.getRegionForTable(table , true).getAttributes.getConcurrencyChecksEnabled == true)
+
+    snc.dropTable(table)
```
kneeraj: Test is fine. Please add a dunit test with persistent and eviction configured and do some restarts of vms and check too. Add a test for split mode too. Just for sanity it will be good. If something else will be there related to restarts it will be caught in the unit testing itself.

core/src/main/scala/io/snappydata/util/ServiceUtils.scala
```diff
@@ -97,8 +97,8 @@ object ServiceUtils {
 
     "jdbc:" + Constant.SNAPPY_URL_PREFIX + (if (locatorUrl.contains(",")) {
       locatorUrl.substring(0, locatorUrl.indexOf(",")) +
-          ";secondary-locators=" + locatorUrl.substring(locatorUrl.indexOf(",") + 1)
-    } else locatorUrl)
+          "/;secondary-locators=" + locatorUrl.substring(locatorUrl.indexOf(",") + 1 ) + "/"
```
kneeraj: There is still some issue with multiple urls. Try to get into the same pr.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
nthanvi: The change is only applicable for split cluster mode. In case of secondary locators the url was not formed properly and was causing the issue.

core/src/main/scala/io/snappydata/util/ServiceUtils.scala
nthanvi: The issue with multiple locator is fixed. is there any other issue with multiple URL's that I am not aware of?

PR #323, opened 2016-08-08
[SNAP-961] Fix passing of some DDL extension clauses

sumwale:

- corrected formatting for PERSISTENT, OFFHEAP and SERVER_GROUPS in proper
  format as expected by snappydata-store
- split the PERSISTENT property into a new DISKSTORE which just specifies
  the diskstore name; former specifies async/sync persistence or whether it on/off
- don't access numPartitions in ColumnFormatRelation.destroy else it causes
  a RegionDestroyedException in case it is invoked to cleanup create table failure
- added dunit test in DDLRoutingDUnitTest to check for cases of exception for
  above clauses as well as when those are proper (no SERVER_GROUPS test yet)

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/commit/6207f0f407ea8a2fbfd67d9ce0da24f5ccd524df has the store side changes

### DIFF
--- a/cluster/src/dunit/scala/io/snappydata/cluster/DDLRoutingDUnitTest.scala
+++ b/cluster/src/dunit/scala/io/snappydata/cluster/DDLRoutingDUnitTest.scala
@@ -16,7 +16,7 @@
  */
 package io.snappydata.cluster
 
-import java.sql.{Connection, DriverManager}
+import java.sql.{Connection, DriverManager, SQLException}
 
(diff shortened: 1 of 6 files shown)


sumwale:
@kneeraj please also see the commit on store referenced above

kneeraj:
Looks good.

PR #331, opened 2016-08-17
Column Buffer scan returns UnsafeRows but still ConvertToUnsafe operator is applied

hbhanawat:

1. Added code to convert iterators on row buffers and row tables to return RDD[UnsafeRow] instead of RDD[SpecificMutableRow]. This was done by applying UnsafeProjection.
2. Removed unnneccessary conversion of iterator on column buffer from RDD[UnsafeRow] to RDD[Row].
3. Now since, row buffers, row tables and column buffers return RDD[UnsafeRow] there is no need to apply ConvertTounsafe operator. Removed code that calls toCatalystRDD.
4. Instead of PrunedFilteredScan, PrunedUnsafeFilteredScan has been introduced that returns RDD[InternalRow] and not RDD[Row] to avoid conversions. 

## Patch testing

Precheckin


core/src/main/scala/org/apache/spark/sql/execution/CompactExecRowToMutableRow.scala
```diff
@@ -221,6 +223,6 @@ trait CompactExecRowToMutableRow extends ResultWasNull {
       }
       i = i + 1
     }
-    mutableRow
+    unsafeproj.apply(mutableRow)
```
sumwale: This is unnecessary for the call from CachedBatchCreator. I will suggest this one always returning SpecificMutableRow while the child classes apply the unsafe projection if required (i.e. only CompactExecRowIteratorOnRS will while CachedBatchCreator will not).

Ideally this whole class can be changed to directly create UnsafeRow directly (preferably with CodeGeneration) instead of SpecificMutableRow, but since the main caller is RowFormatScanRDD its not that important (esp for TPCH) while it will probably not make a lot of difference for CachedBatchCreator once the change as I suggested is done.

core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCAppendableRelation.scala
```diff
@@ -105,13 +104,13 @@ case class JDBCAppendableRelation(
 
   // TODO: Suranjan currently doesn't apply any filters.
   // will see that later.
-  override def buildScan(requiredColumns: Array[String],
-      filters: Array[Filter]): RDD[Row] = {
+  override def buildUnsafeScan(requiredColumns: Array[String],
+                         filters: Array[Filter]): RDD[InternalRow] = {
```
sumwale: Formatting not proper (previous one was correct). Probably not using SnappyData's codeStyleSettings.xml, so correct as mentioned in build-instructions.md

core/src/main/scala/org/apache/spark/sql/execution/datasources/StoreDataSourceStrategy.scala
```diff
@@ -42,25 +42,10 @@ private[sql] object StoreDataSourceStrategy extends Strategy{
         filters,
         t.numPartitions,
         t.partitionColumns,
-        (a, f) => toCatalystRDD(l, a, t.buildScan(a.map(_.name).toArray, f))) :: Nil
-
+        (a, f) => t.buildUnsafeScan(a.map(_.name).toArray, f)) :: Nil
```
sumwale: Also add handling for just PrunedUnsafeFilteredScan in addition to PartitionedDataSourceScan. For testing I am using something like this in addition to these changes:

```scala
    case PhysicalOperation(projects, filters,
        l@LogicalRelation(t: PrunedUnsafeFilteredScan, _, _)) =>
      pruneFilterProject(
        l,
        projects,
        filters,
        0,
        Seq.empty[String],
        (a, f) => t.buildUnsafeScan(a.map(_.name).toArray, f)) :: Nil
```

then in pruneFilterProjectRaw add handling for numPartition == 0 like:

```scala
      val scan = if (numPartition > 0) {
        execution.PartitionedPhysicalRDD.createFromDataSource(
          projects.map(_.toAttribute),
          numPartition,
          joinedCols,
          scanBuilder(requestedColumns, candidatePredicates, pushedFilters),
          relation.relation)
      } else {
        execution.DataSourceScanExec.create(
          projects.map(_.toAttribute),
          scanBuilder(requestedColumns, candidatePredicates, pushedFilters),
          relation.relation, metadata, relation.metastoreTableIdentifier)
      }
```

I will attach my full patch to this file to JIRA.

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
```diff
@@ -75,7 +75,7 @@ private[sql] object PartitionedPhysicalRDD {
   }
 }
 
-trait PartitionedDataSourceScan extends PrunedFilteredScan {
+trait PartitionedDataSourceScan extends PrunedUnsafeFilteredScan {
```
sumwale: With these changes, the explicit UnsafeProjection in PartitionedPhysicalRDD.doExecute is unnecessary and should be removed. Without that removal you will not get the expected boost at all. For TPCH q1 that change is what helped most (~30% improvement) while these changes by themselves had minimal impact.

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatScanRDD.scala
```diff
@@ -411,7 +414,7 @@ final class InternalRowIteratorOnRS(conn: Connection,
       i = i + 1
     }
     moveNext()
-    mutableRow
+    unsafeproj.apply(mutableRow)
```
sumwale: As I mentioned above, ideally this can directly create an UnsafeRow with CodeGeneration, but since it is RowFormatScanRDD so low priority.

core/src/main/scala/org/apache/spark/sql/row/JDBCMutableRelation.scala
```diff
@@ -148,8 +149,8 @@ case class JDBCMutableRelation(
   final lazy val executorConnector = ExternalStoreUtils.getConnector(table,
     connProperties, forExecutor = true)
 
-  override def buildScan(requiredColumns: Array[String],
-      filters: Array[Filter]): RDD[Row] = {
+  override def buildUnsafeScan(requiredColumns: Array[String],
+                               filters: Array[Filter]): RDD[InternalRow] = {
```
sumwale: correct formatting

sumwale:
Overall looks good -- see some comments above esp about PartitionedPhysicalRDD. For RowFormat creating UnsafeRow directly is not that critical, but for samples it can be so for that PR I will open a separate once that handles it.

core/src/main/scala/org/apache/spark/sql/execution/CompactExecRowToMutableRow.scala
hbhanawat: Thanks for pointing this out. I didn't realize that CachedBatchCreator code is using this too. Fixed.

core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCAppendableRelation.scala
hbhanawat: Fixed

core/src/main/scala/org/apache/spark/sql/execution/datasources/StoreDataSourceStrategy.scala
hbhanawat: Are you doing it with your patch or do you want me to do it? Either way is fine with me.

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
hbhanawat: This came with 2.0 merge and hence didn't realize. Fixed it.

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatScanRDD.scala
hbhanawat: Yes, codegeneration is the way to go but I thought that directly creating an UnsafeRow would be equally slow.

core/src/main/scala/org/apache/spark/sql/execution/datasources/StoreDataSourceStrategy.scala
sumwale: I have attached it to the JIRA. Check it and add to this PR.

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatScanRDD.scala
sumwale: Above does two copies, once into SpecificMutableRow then to UnsafeRow which can be optimized (+ avoiding UTF8 conversion by using same format in GemXD as mentioned in TODO). Can be low priority since it is for row tables only which we will deprecate in long term.

PR #332, opened 2016-08-17
[SNAP-824] Changes for SNAP-824

dshirish:
Added parser support for CUBE, ROLLUP, GROUPING SETS

Will add these changes for SnappyAQPParser as well.

## Patch testing
Added a unit test


jramnara:
Shouldn't this change go into spark?

sumwale:
@jramnara Its present in the new Spark 2.0 parser. This PR adds support to snappydata parser.

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -162,6 +162,12 @@ abstract class SnappyDDLParser(session: SnappySession)
   final def WEEK: Rule0 = rule { intervalUnit(Consts.WEEK) }
   final def YEAR: Rule0 = rule { intervalUnit(Consts.YEAR) }
 
+  //cube, rollup, grouping sets
+  final def CUBE = rule { keyword(Consts.CUBE) }
+  final def ROLLUP = rule { keyword(Consts.ROLLUP) }
+  final def GROUPING = rule { keyword(Consts.GROUPING) }
+  final def SETS = rule { keyword(Consts.SETS) }
+
```
sumwale: Can you add the return type explicitly (Rule0) else recent IDEA complains for scalastyle (public methods should declare return types)?

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -348,6 +348,42 @@ class SnappyParser(session: SnappySession)
...
+    val bitmasks: Seq[Int] = groupingSets.map(set => set.foldLeft(0)((bitmap, col) => {
+      require(keyMap.contains(col), s"$col doesn't show up in the GROUP BY list")
+      bitmap | 1 << keyMap(col)
+    }))
+    GroupingSets(bitmasks, groupByExprs, child, aggregations)
+  }
+
+  protected final def groupingSetExpr: Rule1[Seq[Expression]] = rule {
+    '(' ~ ws ~ (expression + (',' ~ ws)) ~ ws ~ ')'  ~>  ((gs: Seq[Expression]) => gs) |
+        (expression + (',' ~ ws))  ~>  ((gs: Seq[Expression]) => gs) |
```
sumwale: The "~ ws" after the expression group (third one from start) is unnecessary. However you do need to add it after ')' instead.

The convention in all non-terminals is trailing "~ ws" so expression already has a trailing ws.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -348,6 +348,42 @@ class SnappyParser(session: SnappySession)
...
 
+  private def extractGroupingSet(
+      child: LogicalPlan,
+      aggregations: Seq[NamedExpression],
+      groupByExprs: Seq[Expression],
+      groupingSets: Seq[Seq[Expression]]): GroupingSets = {
+    val keyMap = groupByExprs.zipWithIndex.toMap
+    val bitmasks: Seq[Int] = groupingSets.map(set => set.foldLeft(0)((bitmap, col) => {
+      require(keyMap.contains(col), s"$col doesn't show up in the GROUP BY list")
+      bitmap | 1 << keyMap(col)
```
sumwale: Does "1 << " require a parenthesis due to operator precedence: bitmap | (1 << keyMap(col))

sumwale: AstBuilder.withAggregation seems to be following the other way. Start with 0xffff.. then it is using "bitmap & ~(1 << (numExpressions - 1 - index))". Shouldn't this be doing the same?

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -348,6 +348,42 @@ class SnappyParser(session: SnappySession)
...
+  protected final def groupingSetExpr: Rule1[Seq[Expression]] = rule {
+    '(' ~ ws ~ (expression + (',' ~ ws)) ~ ws ~ ')'  ~>  ((gs: Seq[Expression]) => gs) |
+        (expression + (',' ~ ws))  ~>  ((gs: Seq[Expression]) => gs) |
+        '(' ~ ws ~ ')'  ~>  (() => (Seq[Expression]()))
+  }
+
+  protected final def cubeRollUpGroupingSet: Rule1[(Seq[Seq[Expression]], Option[String])] = rule {
+    CUBE ~> (() => (Seq(Seq[Expression]()), Option("CUBE"))) |
+    ROLLUP ~> (() => (Seq(Seq[Expression]()),Option("ROLLUP"))) |
+    GROUPING ~ SETS ~ ('(' ~ ws ~ (groupingSetExpr ~ ws).+(',' ~ ws) ~ ws ~ ')' ~ ws)  ~>
```
sumwale: The second and fourth "~ ws" should be removed once the convention I mentioned above is followed. So you can be:
(groupingSetExpr + (',' ~ ws)) or groupingSetExpr.+(',' ~ ws) which is a bit consistent and cleaner

Perhaps we should refactor to add a separate def for (',' ~ ws) which is being used in many places.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -348,6 +348,42 @@ class SnappyParser(session: SnappySession)
...
+        ((gs: Seq[Seq[Expression]]) => (gs, Option("GROUPINGSETS")))
+  }
+
+  protected final def groupBy: Rule1[(Seq[Expression], Seq[Seq[Expression]],  Option[String])] = rule {
+    GROUP ~ BY ~ (expression + (',' ~ ws)) ~
+        (WITH ~ cubeRollUpGroupingSet).? ~>
```
sumwale: As per SqlBase.g4, the GROUPING SETS can appear without a WITH? The grammar there is confusing:

```
( WITH kind=ROLLUP | WITH kind=CUBE | kind=GROUPING SETS ...
```

This suggests there is a WITH for CUBE but not for GROUPING SETS?

sumwale: Confirmed in hive manual that GROUPING SETS does not have a WITH: https://cwiki.apache.org/confluence/display/Hive/Enhanced+Aggregation,+Cube,+Grouping+and+Rollup

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -348,6 +348,42 @@ class SnappyParser(session: SnappySession)
...
+        ((groupingExpr: Seq[Expression], crgs: Option[(Seq[Seq[Expression]], Option[String])]) =>
+        { val emptyCubeRollupGrSet = (Seq(Seq[Expression]()), Option("")) // if cube, rollup, GrSet is not used
```
sumwale: Due to bugs in even recent IDEA in type inference (due to which it showed most of rules in red), I ended up declaring types for (?, +, *) which are normally (Option, Seq, Seq) respectively, to be Any and then an explicit cast in body. Perhaps you can do the same to remove those annoying error markers in IDEA.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -587,9 +623,20 @@ class SnappyParser(session: SnappySession)
...
           .map(UnresolvedAlias(_, None))
-      val withProjection = g.asInstanceOf[Option[Seq[Expression]]]
-          .map(Aggregate(_, expressions, withFilter))
-          .getOrElse(Project(expressions, withFilter))
+      val gr = g.asInstanceOf[Option[(Seq[Expression], Seq[Seq[Expression]], Option[String])]]
+      val withProjection = gr.map(x => {
+        x._3.get match {
+          // group by cols with rollup
+          case "ROLLUP" => Aggregate(Seq(Rollup(x._1)), expressions, withFilter)
+          // group by cols with cube
```
sumwale: Instead of this matching etc, a cleaner approach is to have groupBy and cubeRollUpGroupingSet return a Rule1 of function (Seq[Expression], LogicalPlan) => LogicalPlan

So then groupBy can be like below when cubeRollUpGroupingSet likewise returns Rule1 of (Seq[Expression], Seq[Expression], LogicalPlan) => LogicalPlan:

```scala
protected final def groupBy: Rule1[(Seq[Expression], Seq[Seq[Expression]],  Option[String])] = rule {
  GROUP ~ BY ~ (expression + (',' ~ ws)) ~ cubeRollUpGroupingSet.? ~>{ (exprs: Any, g: Any) =>
    (expressions: Seq[Expression], plan: LogicalPlan) =>
      val groupByExprs = exprs.asInstanceOf[Seq[Expression]]
      g.asInstanceOf[Option[(Seq[Expression], Seq[Expression], LogicalPlan) => LogicalPlan] match {
        case Some(crgs) =>crgs(groupByExprs, expressions, plan)
        case None => Aggregate(groupByExprs, expressions, plan)
      }
  }
}
```

sumwale:
Overall looks good -- see some comments above. Haven't looked at the test closely.

My plan is to have a bunch of classes on our side extend Spark's SQL test suites overriding the context to be SnappyContext. So this will test with all the SQL that is tested for Spark.

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
dshirish: Done.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
dshirish: Done.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
dshirish: I had referred to HiveQl.scala from 1.6 code which was doing (bitmap | 1 << keyMap(col)).  

https://github.com/apache/spark/blob/branch-1.6/sql/hive/src/main/scala/org/apache/spark/sql/hive/HiveQl.scala#L488

The new 2.0 code in AstBuilder.withAggregation is setting the bits differently as you mentioned (did not find any bug filed for it). Have now changed the make it similar as 2.0 as this one appears to be correct. Let me know what you think.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
dshirish: Done. Removed unnecessary "~ ws" in this code for now.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
dshirish: Corrected, WITH clause now not required for GROUPING SETS

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
dshirish: Done. Changed the type to Any and then explicitly casting it here.

dshirish:
The test I have added just makes sure that the output SQL queries that use CUBE/ROLLUP/GROUPING SETS is same that of corresponding operation on DataFrame.  There does not seem to be good test coverage for for GROUPING SETS in Spark.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -348,39 +348,44 @@ class SnappyParser(session: SnappySession)
...
       require(keyMap.contains(col), s"$col doesn't show up in the GROUP BY list")
-      bitmap | 1 << keyMap(col)
+      bitmap & ~(1 << (numExpressions - 1 - keyMap(col)))
     }))
     GroupingSets(bitmasks, groupByExprs, child, aggregations)
   }
 
   protected final def groupingSetExpr: Rule1[Seq[Expression]] = rule {
-    '(' ~ ws ~ (expression + (',' ~ ws)) ~ ws ~ ')'  ~>  ((gs: Seq[Expression]) => gs) |
+    '(' ~ ws ~ (expression + (',' ~ ws)) ~ ')' ~ ws ~>  ((gs: Seq[Expression]) => gs) |
```
sumwale: The action ~> part in this looks redundant (pops and pushes the same) so can be removed.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -348,39 +348,44 @@ class SnappyParser(session: SnappySession)
...
         (expression + (',' ~ ws))  ~>  ((gs: Seq[Expression]) => gs) |
```
sumwale: Here too action ~> part is redundant.

sumwale:
Ok, looks good -- couple of minor additional comments about redundant action. If you think the Rule1[function] approach I suggested is worth it, then make that change else looks good for merge.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -348,6 +348,47 @@ class SnappyParser(session: SnappySession)
...
+        ) |
+    GROUPING ~ SETS ~ ('(' ~ ws ~ (groupingSetExpr + (',' ~ ws)) ~ ')' ~ ws)  ~>
+        ((gs: Seq[Seq[Expression]]) => (gs, Option("GROUPINGSETS")))
+  }
+
+  protected final def groupBy: Rule1[(Seq[Expression], Seq[Seq[Expression]],  Option[String])] = rule {
+    GROUP ~ BY ~ (expression + (',' ~ ws)) ~
+        (cubeRollUpGroupingSet).? ~>
+        ((groupingExpr: Any, crgs: Any) =>
+        { val emptyCubeRollupGrSet = (Seq(Seq[Expression]()), Option("")) // if cube, rollup, GrSet is not used
```
sumwale: Since you are using Option[] why not use None here? Else use just a string with empty string as above and remove Option here and in matching below.

dshirish:
Merged the changes. Will also add to AQP parser. Removed the redundant action ~> parts. Will try and incorporate the Rule1[function] approach as well.

PR #340, opened 2016-08-24
[SNAP-1002] Removing unwanted zeppelin interpreter code and loading interpreter u…

SachinJanani:
- Removed Zeppelin code from snappydata and moved it to separate repo
- Loading zeppelin interpreter using reflection

## Patch testing
Ran precheckin some test fails but not related to this PR

## ReleaseNotes.txt changes

(Does this change require an entry in ReleaseNotes.txt? No

## Other PRs 
Yes [zeppelin-interpreter- PR](https://github.com/SnappyDataInc/zeppelin-interpreter/pull/1)

…sing reflections


cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -131,18 +136,30 @@ class LeadImpl extends ServerImpl with Lead with Logging {
...
+           * interpreter will set some properties for classloader for repl
+           * which needs to be specified while creating sparkcontext in lead
+           */
+          logInfo("About to initialize SparkContext with SparkConf")
+          val method: Method = zeppelinIntpUtilClass.getMethod("initializeZeppelinReplAndGetConfig")
+          val obj: Object = method.invoke(null)
+          val props: Properties = obj.asInstanceOf[Properties]
+          props.asScala.foreach(kv => conf.set(kv._1, kv._2))
+        } catch {
+          /* [Sachin] So we need to log warning that
```
rishitesh: I think exiting will make sense. What is the default value of ENABLE_ZEPPELIN_INTERPRETER ?

SachinJanani: Default value of ENABLE_ZEPPELIN_INTERPRETER is false as sumedh suggested that zeppelin interpreter should only be enabled if user wants it otherwise it should be disabled.Ok will change the code to exit the LeadNode if interpreter jar is not found and if property is enabled.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -131,18 +136,30 @@ class LeadImpl extends ServerImpl with Lead with Logging {
...
-         * which needs to be specified while creating sparkcontext in lead
-         */
-        val props: Properties = ZeppelinIntpUtil.initializeZeppelinReplAndGetConfig()
-        props.asScala.foreach(kv => conf.set(kv._1, kv._2))
-      }
-      logInfo("About to initialize SparkContext with SparkConf=" + conf.toDebugString)
 
+        try {
+
+          var zeppelinIntpUtilClass = Class.forName("org.apache.zeppelin.interpreter.ZeppelinIntpUtil")
```
rishitesh: I am not sure if it would matter , but lets use org.apache.spark.util.Utils.classForName() method consistently across application.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -440,4 +477,6 @@ object LeadImpl {
   def clearInitializingSparkContext(): Unit = {
     startingContext.set(null)
   }
+
```
rishitesh: remove unnecessary lines

rishitesh:
Looks ok for merge. See some minor comments.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
SachinJanani: As ZeppelinIntpUtil class is used to call some static method but scala does not support to reflect static method I have used java reflections ( https://issues.scala-lang.org/browse/SI-6459 )

SachinJanani: sorry i looked into the org.apache.spark.util.Utils.classForName() and found it also use java reflections so will replace that

cluster/build.gradle
```diff
@@ -181,41 +181,5 @@ scalaTest {
...
-  include 'org/slf4j/**'
-  include 'jline/console/**'
-  include 'jline/internal/**'
-  include 'jline/**'
-  include 'org/apache/commons/lang/**'
-
-  include 'org/apache/zeppelin/interpreter/**'
- // exclude 'org/apache/zeppelin/interpreter/SnappydataSqlZeppelinInterpreter**'
-  archiveName 'snappydata-zeppelin-interpreter-'+version+'.jar'
-}
```
sumwale: There still are references to zeppelin-interpreter/spark/jdbc in this build file. Are they still required after these changes?

SachinJanani: No they are not needed now.Forgot to remove it from build.gradle.Will remove it in next checkin

PR #341, opened 2016-08-25
[SNAP-996] Fix for Bug SNAP-996

SachinJanani:
- Fix for Bug SNAP-996
- Fixed Failing DUint test in precheckin

## Patch testing
Ran precheckin and some test related to streaming fails

## ReleaseNotes.txt changes
No
## Other PRs 
NA


sumwale:
@SachinJanani sorry, didn't notice this JIRA. I created a new one here https://jira.snappydata.io/browse/SNAP-1007 with a PR which addresses the more general issue of reserved vs non-reserved keywords. I think marking limit as reserved is not a clean solution because the issue exists for many other non-reserved keywords (e.g. SORT/CLUSTER/DISTRIBUTE BY and others) and we cannot be marking all those as reserved.

cluster/src/dunit/scala/io/snappydata/externalstore/ColumnTableDUnitTest.scala
```diff
@@ -33,33 +33,49 @@ import org.apache.spark.sql.{Row, SaveMode, SnappyContext}
...
+  val currentLocatorPort = locPort
 
   def testTableCreation(): Unit = {
     startSparkJob()
   }
 
+  //After discussing with Namratam Commenting out this test as it causes other tests to fail
+/*
+
   def testTableCreationWithHA(): Unit = {
```
sumwale: Has this test been added recently? I didn't see failures due to this test previously.
And its Namrata not Namratam :-)

SachinJanani: This test fails intermittently .And if it fails all other subsequent Dunit test fails.I have raised a jira for this https://jira.snappydata.io/browse/SNAP-1006 .As per the stack trace it seems issue at store side.Sorry for the typo :)

SachinJanani:
Closing this PR as @sumwale  has already raised a PR addressing  more general issue of reserved vs non-reserved keywords

cluster/src/dunit/scala/io/snappydata/externalstore/ColumnTableDUnitTest.scala
sumwale: Ok, looks like a serious issue. Someone will need to look into it.

PR #343, opened 2016-08-25
[SNAP-1007] Add support for non-reserved keywords

sumwale:

As per SNAP-1007, this primarily adds support for non-reserved keywords which are not allowed in some places in the SQL grammar where otherwise any identifier is allowed.

 - marked all keywords as either reserved or non-reserved
 - added a "strictIdentifier" which can be neither of reserved or non-reserved;
   mark all aliases to use "strictIdentifier" (but identifier when with AS)
 - refactored to add "commaSep" rule for all uses of comma followed by whitespace
 - marked CAST, CURRENT_DATE/TIMESTAMP as reserved due to ambiguities otherwise
 - some formatting corrections

## Patch testing

precheckin (no new failures)

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
@@ -136,7 +136,7 @@ abstract class SnappyBaseParser(session: SnappySession) extends Parser {
   }
 
   /**
-   * A strictIdentifier is more restricted that an identifier in that neither
+   * A strictIdentifier is more restricted than an identifier in that neither
    * any of the SQL reserved keywords nor non-reserved keywords will be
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
```diff
@@ -127,6 +135,22 @@ abstract class SnappyBaseParser(session: SnappySession) extends Parser {
     }
   }
 
+  /**
+   * A strictIdentifier is more restricted that an identifier in that neither
```
dshirish: Typo in comment

dshirish:
Looks good to me.

PR #344, opened 2016-08-26
SNAP-926

hbhanawat:

copied the source of snappy examples to the distribution folder during the product build. Also, fixed the folder path for the python file.

## Patch testing

Tested the Getting started for the python to ensure that the path for the python file is correct.


kneeraj:
Looks good.

PR #345, opened 2016-08-26
[SNAP-1008] Pause zeppelin scala interpreter when critical up event i…

SachinJanani:
* This PR addresses the first part of SNAP-1008 used to monitor memory of lead node which is used to pause zeppelin interpreter

## Patch testing
Ran precheckin some tests fail which are not related to this PR

## ReleaseNotes.txt changes
No
## Other PRs 
N/A
…s triggered

### DIFF
--- a/cluster/src/main/java/io/snappydata/gemxd/MemoryNotificationFactory.java
+++ b/cluster/src/main/java/io/snappydata/gemxd/MemoryNotificationFactory.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2016 SnappyData, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+package io.snappydata.gemxd;
+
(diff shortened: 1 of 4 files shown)


cluster/src/main/java/io/snappydata/gemxd/MemoryNotificationListener.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package io.snappydata.gemxd;
+
+/**
+ * Created by sachin on 24/8/16.
+ */
+public interface MemoryNotificationListener {
+  public void notifyCriticalUp();
+  public void notifyEvictionUp();
+}
```
sumwale: Why a new interface MemoryNotificationListener? What's the issue with use ResourceListener directly (from wherever it is going to be)?

cluster/src/main/java/io/snappydata/gemxd/MemoryNotificationFactory.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  public static void notifyEvictionUp() {
+    for (MemoryNotificationListener mnl : notificationListeners) {
+      mnl.notifyEvictionUp();
+    }
+  }
+
+
+}
```
sumwale: See comment below. Both this class and LeadNodeMemoryListener looks redundant to me when ResourceListener can be used directly.

cluster/src/main/java/io/snappydata/gemxd/MemoryNotificationListener.java
SachinJanani: @sumwale  You mean implementing ResourceListener directly in zeppelin interpreter and adding it to ResourceManager ?

SachinJanani:
Deleting this PR as this code is redundant.Will only need changes in zeppelin-interpreter

cluster/src/main/java/io/snappydata/gemxd/MemoryNotificationListener.java
sumwale: yes

PR #350, opened 2016-08-30
Changes for SNAP-653

dshirish:
+ Registered SnappyHiveMetaStoreEventListener and process onDropTable callback in it (invoked when table dropped from Hive metastore). This callback will be invoked on either embedded snappy lead or external cluster driver (i.e depending on from where the table is dropped)

+ When a table is dropped in on external cluster driver send a message (SnappyRemoveCachedObjectsFunction) to all nodes in embedded cluster to do the clean up. Similarly send a message from embedded lead to external driver if table dropped in embedded mode

## Patch testing
Modified SplitClusterDUnitTest to assert that cached relation is removed from SnappyStoreHiveCatalog after a table drop from external table (in testColumnTableCreation())

Ran precheckin(known failures in streaming tests and ColumnTableDUnitTest#testTableCreationWithHA

## Other PRs 
Store side changes for this PR
https://github.com/SnappyDataInc/snappy-store/pull/102

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
@@ -493,7 +493,7 @@ object ExternalStoreUtils {
     }
   }
 
-  private def removeCachedObjects(table: String): () => Iterator[Unit] = () => {
+  def removeCachedObjects(table: String): () => Iterator[Unit] = () => {
     ConnectionPool.removePoolReference(table)
     CodeGeneration.removeCache(table)
     Iterator.empty
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
@@ -16,6 +16,7 @@
  */
 package org.apache.spark.sql.execution.columnar.impl
 
+import java.lang
 import java.util.{Collections, UUID}
 
 import scala.collection.concurrent.TrieMap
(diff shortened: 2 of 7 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
```diff
@@ -139,6 +141,27 @@ object StoreCallbacksImpl extends StoreCallbacks with Logging with Serializable
...
+      // cleanup invoked on embedded mode nodes
+      // from external cluster (in split mode) driver
+      ExternalStoreUtils.removeCachedObjects(table)
+      stores.remove(table)
+      // clean up cached hive relations on lead node
+      if (GemFireXDUtils.getGfxdAdvisor.getMyProfile.hasSparkURL) {
+        SnappyStoreHiveCatalog.registerRelationDestroy()
+      }
+    } else {
+      // clean up invoked on external cluster driver (in split mode)
```
kneeraj: Should we have any assertion that this is driver of split mode only?

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -493,7 +493,7 @@ object ExternalStoreUtils {
     }
   }
 
-  private def removeCachedObjects(table: String): () => Iterator[Unit] = () => {
+  def removeCachedObjects(table: String): () => Iterator[Unit] = () => {
     ConnectionPool.removePoolReference(table)
     CodeGeneration.removeCache(table)
```
kneeraj: For this cleanup may be we can get a handle of the generated code for batch insert for a particular table which is dropped later.

So create a table.
Do df insert.
Check whether code got generated for batch insert.
If yes get the execution key.
Drop the table.
Check in the generated code cache for this key. It should be clean.

core/src/main/scala/org/apache/spark/sql/hive/SnappyHiveMetaStoreEventListener.scala
```diff
@@ -0,0 +1,93 @@
...
+
+      case SplitClusterMode(_, _) =>
+        // table is dropped on external cluster so send message to
+        // embedded mode nodes
+        val otherMembers = GfxdMessage.getOtherMembers
+        if (otherMembers.size > 0) {
+          val table = getTableName(tableEvent)
+          log.debug(s"onDropTable invoked for $table")
+          val args = SnappyRemoveCachedObjectsFunction.
+              newArgs(table.toUpperCase, true)
```
kneeraj: Shirish himself pointed out to remove .uppercase from table.

kneeraj:
Looks good. Merge it. A couple of comments + add the test suggested for 1.0 at least.

PR #352, opened 2016-08-31
Added simplified APIs to obtain and create a connection in SnappyData…

rishitesh:
 Added simplified APIs to obtain and create a connection in SnappyData. These APIs can be used to get a pooled connection, non-pooled connection.
Also this APIs can be used in drivers and executors alike. All the complexities have been wrapped.

## Patch testing
Added tests and example code to test the same. 
## ReleaseNotes.txt changes
NA

## Other PRs 
NA

…. These APIs can be used to get a pooled connection, non-pooled connection.

Also this APIs can be used in drivers and executors alike. All the complexities have been wrapped.


core/src/main/scala/org/apache/spark/jdbc/ConnectionConf.scala
```diff
@@ -0,0 +1,120 @@
...
+
+import org.apache.spark.sql.SnappyContext
+import org.apache.spark.sql.execution.columnar.ExternalStoreUtils
+import org.apache.spark.sql.execution.columnar.ExternalStoreUtils.CaseInsensitiveMutableHashMap
+import org.apache.spark.sql.sources.ConnectionProperties
+import org.apache.spark.util.Utils
+
+import scala.collection.mutable
+
+class ConnectionConf(val connProps: ConnectionProperties) extends Serializable {
```
hbhanawat: I don't think there is any need for ConnectionProperties. This should be replaced with ConnectionConf everywhere including the JDBCSourceAsStore.

core/src/main/scala/org/apache/spark/jdbc/ConnectionConf.scala
```diff
@@ -0,0 +1,120 @@
...
+   * Sets a property for a pool.
+   * For detailed list for Hikari refer https://github.com/brettwooldridge/HikariCP/wiki/Configuration
+   * For detailed list for Tomcat refer https://people.apache.org/~fhanik/jdbc-pool/jdbc-pool.html
+   *
+   * Even if you don't mention any properties it defaults it to some sensible values
+   *
+   * @param prop property name
+   * @param value property value
+   */
+  def setPoolConf(prop: String, value: String) = {
```
hbhanawat: Somehow, adding pool props using setPoolConf and connection props using setConnectionProps looks cumbersome. Can we take the pool props as a single map, like we were doing previously. And a single setConf that would set the connection properties. @jramnara ?

core/src/test/scala/org/apache/spark/jdbc/ConnectionConfTest.scala
```diff
@@ -0,0 +1,161 @@
...
+  test("test serializibility") {
+    val data = Seq(Seq(1, 2, 3), Seq(7, 8, 9), Seq(9, 2, 3), Seq(4, 2, 3), Seq(5, 6, 7))
+    val rdd = sc.parallelize(data, data.length).map(s => new Data(s(0), s(1), s(2)))
+    val dataDF = snc.createDataFrame(rdd)
+
+    dataDF.write.format("row").mode(SaveMode.Append).saveAsTable("MY_SCHEMA.MY_TABLE")
+
+    val conf = new ConnectionConfBuilder(snc).build()
+
+    rdd.foreachPartition(d => {
```
hbhanawat: A dunit would be better to test this. Going forward if there are any multinode issues with this API, they won't be caught.

examples/src/main/scala/io/snappydata/examples/DataUpdateJob.scala
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.spark.jdbc.{ConnectionUtil, ConnectionConfBuilder}
+import org.apache.spark.sql.{SnappyJobValid, SaveMode, SnappyJobValidation, SnappyContext, SnappySQLJob}
+
+case class Data(col1: Int, col2: Int, col3: Int)
+
+/**
+ * An example to use ConnectionUtil to mutate data stored in SnappyData.
+ * 
+ */
+object DataUpdateJob extends SnappySQLJob {
```
hbhanawat: This is good. I think we all should make a practice to add examples like this.

core/src/test/scala/org/apache/spark/jdbc/ConnectionConfTest.scala
rishitesh: Sure will add one. I thought all example jobs get executed through quickStart. Apparently not.

core/src/main/scala/org/apache/spark/jdbc/ConnectionConf.scala
rishitesh: This can be taken in a separate PR. Those changes will unnecessary crowd this PR , which is for a specific purpose. Will file a JIRA ticket to track .

core/src/main/scala/org/apache/spark/jdbc/ConnectionConf.scala
rishitesh: We can add another two methods , which will take maps. And keep this methods as it is what say ?

rishitesh: Yeah that makes sense. Will change it.

PR #353, opened 2016-08-31
Snap 735: Supporting VARCHAR with size and processing STRING as VARCHAR(32762), by default.

ashetkar:
* SNAP-735: Supporting VARCHAR with size and processing STRING as VARCHAR(32762), by default.
* Provided SQL conf (spark.sql.stringAsClob) and query hint (--+ stringAsClob(1)) to force processing STRING as CLOB.
* Added a dunit test.

## Patch testing
precheckin is in-progress, new dunit test passes.

## ReleaseNotes.txt changes
NA

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/104


cluster/src/dunit/scala/io/snappydata/cluster/StringAsVarcharDUnitTest.scala
```diff
@@ -0,0 +1,183 @@
...
+ */
+class StringAsVarcharDUnitTest(val s: String)
+    extends ClusterManagerTestBase(s) with Logging {
+
+  val tableName1 = "colTab1"
+
+  val tableName2 = "colTab2"
+
+  override def tearDown2(): Unit = {
+    super.tearDown2()
```
kneeraj: Probabaly not required. Check and remove.

cluster/src/dunit/scala/io/snappydata/cluster/StringAsVarcharDUnitTest.scala
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  private def getANetConnection(netPort: Int,
+      useGemXDURL: Boolean = false): Connection = {
+    val driver = "com.pivotal.gemfirexd.jdbc.ClientDriver"
```
kneeraj: I think we should take this method in the base class and everyone should just use it. This code is repeated everywhere.

cluster/src/dunit/scala/io/snappydata/cluster/StringAsVarcharDUnitTest.scala
```diff
@@ -0,0 +1,183 @@
...
+
+    val s = conn.createStatement()
+    var (query, count, cols, tName) = join match {
+      case true => (s"select t1.col_int, t1.col_string, t2.col_varchar, t2.col_clob, t2.col_char " +
+          s"from $tableName1 t1, $tableName2 t2 where t1.col_int = t2.col_int", 2, 5, "T1")
+      case false => (s"select * from $tableName2", 2, 5, tableName2)
+    }
+    conf match {
+      case "none" =>
+      case "hint" => query += " --+ stringAsClob(1)"
```
kneeraj: Add a test for stringAsClob(0) so that it works like default is tested.

kneeraj:
Ideally we should have hints where in user can specify for specific columns. Blindly renedering them all as varchar(maxsize) or all as clobs may not be very useful. Lets discuss this.

cluster/src/dunit/scala/io/snappydata/cluster/StringAsVarcharDUnitTest.scala
```diff
@@ -0,0 +1,183 @@
...
+          ", col type " + md.getColumnTypeName(i) + ", col table name = " + md.getTableName(i))
+    }
+    assert(md.getColumnName(1).equals("COL_INT"))
+    assert(md.getColumnTypeName(1).equals("INTEGER"))
+
+    assert(md.getColumnName(2).equals("COL_STRING"))
+    assert(md.getColumnTypeName(2).equals(stringType))
+
+    assert(md.getColumnName(3).equals("COL_VARCHAR"))
+    assert(md.getColumnTypeName(3).equals("VARCHAR"))
```
kneeraj: Test the size also from the metadata. You can use:

ResultSetMetaData.getPrecision ( jdbc api )

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
```diff
@@ -237,7 +246,21 @@ class SparkSQLExecuteImpl(val sql: String,
       case t: DecimalType => (StoredFormatIds.SQL_DECIMAL_ID, t.precision, t.scale)
       case FloatType => (StoredFormatIds.SQL_REAL_ID, -1, -1)
       case DoubleType => (StoredFormatIds.SQL_DOUBLE_ID, -1, -1)
-      case StringType => (StoredFormatIds.SQL_CLOB_ID, -1, -1)
+      case s: StringType =>
+        if (snc.sessionState.conf.stringAsClob || stringAsClob) {
+            if (f.metadata.contains(Constant.CHAR_SIZE_NAME) &&
```
kneeraj: Please change the name 'CHAR_SIZE_NAME' to something more appropriate which conveys the intention.

Like: CHARTYPE_SIZE_PROP

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
```diff
@@ -237,7 +246,21 @@ class SparkSQLExecuteImpl(val sql: String,
       case t: DecimalType => (StoredFormatIds.SQL_DECIMAL_ID, t.precision, t.scale)
       case FloatType => (StoredFormatIds.SQL_REAL_ID, -1, -1)
       case DoubleType => (StoredFormatIds.SQL_DOUBLE_ID, -1, -1)
-      case StringType => (StoredFormatIds.SQL_CLOB_ID, -1, -1)
+      case s: StringType =>
+        if (snc.sessionState.conf.stringAsClob || stringAsClob) {
```
kneeraj: 'snc.sessionState.conf.stringAsClob' -- Put this in a session state constant as this is true for the entire session.

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
```diff
@@ -237,7 +246,21 @@ class SparkSQLExecuteImpl(val sql: String,
...
+            if (f.metadata.contains(Constant.CHAR_SIZE_NAME) &&
+                !f.metadata.getString(Constant.CHAR_TYPE_NAME).equals("STRING")) {
+              (StoredFormatIds.SQL_VARCHAR_ID,
```
kneeraj: I think we need a StoreFormatId mapping for constant width char type also. Before this change all CHAR or VARCHARs were going as CLOB. Now VARCHHAR will go as VARCHAR and even CHARS will go as VARCHARs with this code. Isn't it?

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
```diff
@@ -422,7 +445,7 @@ object SparkSQLExecuteImpl {
             dvd.setValue(in.readDouble())
           case StoredFormatIds.SQL_VARCHAR_ID =>
             // read the full length as an integer
-            val utfLen = in.readInt()
+            val utfLen = InternalDataSerializer.readSignedVL(in).toInt
```
kneeraj: We would need a corresponding case handling for CHAR type.

core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
```diff
@@ -189,7 +189,7 @@ abstract class SnappyBaseParser(session: SnappySession) extends Parser {
   }
 
   protected final def primitiveType: Rule1[DataType] = rule {
-    STRING ~> (() => StringType) |
+    STRING ~> (() => CharType(32672, baseType = "STRING")) |
```
kneeraj: Can we have a constant for this number as well.

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -464,8 +465,9 @@ abstract class SnappyDDLParser(session: SnappySession)
         t: DataType, notNull: Any, cm: Any) =>
       val builder = new MetadataBuilder()
       val (dataType, empty) = t match {
-        case CharType(size, fixed) =>
-          builder.putLong("size", size).putBoolean("fixed", fixed)
+        case CharType(size, baseType) =>
+          builder.putLong(Constant.CHAR_SIZE_NAME, size)
```
kneeraj: See above. If better property names can be used.

core/src/main/scala/org/apache/spark/sql/row/GemFireXDDialect.scala
```diff
@@ -116,12 +116,13 @@ abstract class GemFireXDBaseDialect extends JdbcExtendedDialect {
   override def getJDBCType(dt: DataType,
       md: Metadata): Option[JdbcType] = dt match {
     case StringType =>
-      if (md.contains("size")) {
-        if (md.contains("fixed") && md.getBoolean("fixed")) {
-          Some(JdbcType(s"CHAR(${md.getLong("size")})",
+      if (md.contains(Constant.CHAR_SIZE_NAME)) {
+        if (md.contains(Constant.CHAR_TYPE_NAME) &&
+            md.getString(Constant.CHAR_TYPE_NAME).equals("CHAR")) {
```
kneeraj: Since since dialect mapping is there then we can actually store things as VARCHAR ot CHAR as the case may be instead of CLOB. I am not sure though. Please check. But we can do after 0.6

kneeraj:
Looks good.

ashetkar:
@sumwale could you please now review the PR, as @kneeraj too contributed to this PR.

sumwale:
Haven't looked at the precise changes, but from usability and correctness perspective this has inherent problems. What if user declares the column explicitly as CLOB in CREATE TABLE? Then user will always have to provide a hint which will be both surprising for the user as well as lead to problems later. For example what if most of the time it works because strings fit in 32K so devs did not even notice that hint should have been added but then in production deployment a larger string comes along that fails at runtime? Users will need to upgrade and re-deploy the app?

To me it does not look like a shortcut like this will cut it. We will need to go all the way to tie up the DDL declaration with VARCHAR or CLOB in plan and then SparkSQLExecuteImpl needs to honour that. Firstly DDL declaration should lead to corresponding proper declaration in GemXD table creation for both row tables and row buffer of column table (is that fixed?). A rule in SnappyData which will scan the leaf STRING types, check if meta-data has a "size" then propagate to higher layers unless it hits an explicit CAST to some other type.

kneeraj:
>> What if user declares the column explicitly as CLOB in CREATE TABLE? 
This has been taken care of during the parsing phase itself. If the user has specified CLOB it will be rendered as CLOB. During parsing we update the size property of metadata and for clob it is undefined or -1, which is looked at before deciding it as CHAR, VARCHAR or CLOB. 

>> Then user will always have to provide a hint which will be both surprising for the user as well as lead to problems later.  For example what if most of the time it works because strings fit in 32K so devs did not even notice that hint should have been added but then in production deployment a larger string comes along that fails at runtime? Users will need to upgrade and re-deploy the app?
Not relevant. See above comment.

kneeraj:
**>> Firstly DDL declaration should lead to corresponding proper declaration in GemXD table creation for both row tables and row buffer of column table (is that fixed?)**

The changes has been done in the dialect classes in the mapping methods also as per the metadata properties. And it is working. @ashetkar can you please confirm this.

I think we are good now and go ahead. The implementation is as per what we discussed in huddle or otherwise. If we need to tighten it we can take it as an item for 1.0.

sumwale:
> This has been taken care of during the parsing phase itself. If the user has specified CLOB it will be rendered as CLOB. During parsing we update the size property of metadata and for clob it is undefined or -1, which is looked at before deciding it as CHAR, VARCHAR or CLOB. 

No, no, this works only for a flat "select * from table ..." kind of query. As I had mentioned in the discussions and looked up the code right before Amogh and yourself, the metadata does not get transferred by any operators other than the base AttributeReference we create from table. I had mentioned CAST as a problem case then but the code showed it will happen for most operators, if not all.

CAST is a problem even otherwise, as I had mentioned, because if user has specified a CAST(<...> as CLOB) or vice-versa then that would be the expected output type regardless of query hint or no hint. The CAST operator too does not handle metadata and just the type.

Secondly a quick look at the previous use of VARCHAR_ID shows that code there has been changed to convert to UTF8 explicitly. This has major performance implication. We need to use some other ID.

@soubhik-c do you have any inputs?

sumwale:
Actually my main concern is about the defaults. If no hint is specified then is the default VARCHAR(32K) as the PR description suggests, or is it what the result schema tells? Either case has problems as I mentioned, but perhaps for 0.6 we can stick with the second option.

kneeraj:
>> No, no, this works only for a flat "select * from table ..." kind of query.

Confusion here. We thought the metadata is carried all through. Just checked at some places in the code and it is not getting copied.

kneeraj:
**>> Actually my main concern is about the defaults. If no hint is specified then is the default VARCHAR(32K) as the PR description suggests, or is it what the result schema tells? Either case has problems as I mentioned, but perhaps for 0.6 we can stick with the second option.**

Ok, will do the changes as per second option.

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
```diff
@@ -352,7 +386,15 @@ object SparkSQLExecuteImpl {
           val sb = new StringBuilder()
           Utils.dataTypeStringBuilder(other, sb)(col)
           // write the full length as an integer
-          hdos.writeUTF(sb.toString(), true, false)
+          val utf8String = UTF8String.fromString(sb.toString())
```
sumwale: Can we avoid adding this unneeded conversion back and forth? This can have significant perf implications for queries returning decent number of rows. Did you try using some other typeId (other than VARCHAR/CHAR/CLOB)?

kneeraj: Why cant we write the string as simple string obtained from complex type string builder directly as VARCHAR and write it as writeUTF?

kneeraj: Ok. Got it. There are lots of them which will never be used and we just need it for identifying complex type.

PR #354, opened 2016-09-01
Snap 1006

hbhanawat:

 When we configure the test, we configure the initialcapacity of the region. But, this was happening only after the test vms were started for the first time. So, the first time initialcapacity is set as the default value but on recovery the test sets initial capacity as a different value. If the region attributes change on recovery, the recovered entries are inserted into the region map (otherwise the recovered entries are marked for the current region).

So the main issue is - When the recovered entries are inserted into the region map it tries to access the bucket region and fails because it has not been initialized. Not sure if it is a bug. I will open a JIRA after discussion.

For the test made the configuration consistent to fix the issue at hand.


kneeraj:
setUp method already has doCommonSetup. And is it not suppossed to get invoked before every test?

kneeraj:
Looks good.

We need not worry about incompatibility during restarts as user has no control to do it when the system is restarted. This can be closed as just a test issue.

PR #355, opened 2016-09-02
SNAP-1018 use install_jar_bytes/replace_jar_bytes

nthanvi:

SNAP-1018 use install_jar_bytes/replace_jar_bytes procedures to load the jar to the store so that jar could be loaded to all the servers in the cluster regardless of the local availability of jar.

## Patch testing

manual testing

## ReleaseNotes.txt changes

## Other PRs 
NO

SNAP-1018 use install_jar_bytes/replace_jar_bytes procedures to load the jar to the store so that jar could be loaded to all the servers in the cluster regardless of the local availability of jar.

### DIFF
--- a/cluster/src/dunit/scala/org/apache/spark/DynamicJarInstallationDUnitTest.scala
+++ b/cluster/src/dunit/scala/org/apache/spark/DynamicJarInstallationDUnitTest.scala
@@ -22,7 +22,7 @@ import java.io.File
 import java.net.URL
 import java.sql.{Connection, DriverManager}
 import _root_.io.snappydata.cluster.ClusterManagerTestBase
-import _root_.io.snappydata.test.dunit.AvailablePortHelper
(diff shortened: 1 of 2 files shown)


kneeraj:
Looks good. How have you tested this? Any test which will run with the suite? A test also to check the use of the same jar after restarts of nodes like lead and servers.

nthanvi:
I have done the manual testing for  scenarios involving the lease node  and @swatisawant  is working on a hydra for the added coverage.
The DUnit coverage that we  have  is covering the  jar install/replace and delete activities on driver node and executers but it  we not have coverage for lead (job submission / rest API) due to the limitation of current DUnit framework.
The support to test lead through rest api is planned for 1.0 under SNAP-999

kneeraj:
What about the restart part of lead and datastore. Hydra is different. These type of tests if not covered in test suites has shown problems later. So please add and then merge.

PR #356, opened 2016-09-02
CreateTableUsing might be called from createExternalTable. Earlier th…

rishitesh:
CreateTableUsing might be called from createExternalTable. Earlier the builtin flag used to be false. But now its true after Spark2.0.

## Patch testing
Ran quickstart tests
## ReleaseNotes.txt changes
NA
## Other PRs 

NA


sumwale:
looks good

PR #357, opened 2016-09-02
Snap 1024

rishitesh:
a) In Spark 1.6 temp tables were stored against LogicalPlan of the DataFrame. Our catalog makes several assumptions regarding this to consider it as a temp table e.g. while dropping table.

However Spark2.0 analyses the logical plan and converts it to a LogicalRelation before storing that into catalog. Hence the assumptions break.

b) Excluded Univocity Parsers from snappy-core, as this library is already in compile time dependency of spark.sql

## Patch testing

Add a test. Running precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -76,6 +76,7 @@ dependencies {
     exclude(group: 'org.scala-lang', module: 'scala-library')
     exclude(group: 'org.scala-lang', module: 'scala-reflect')
     exclude(group: 'org.scala-lang', module: 'scala-compiler')
+    exclude(group: 'com.univocity', module: 'univocity-parsers')
   }
 
   testCompile project(':dunit')
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -230,6 +230,9 @@ class SnappySession(@transient private val sc: SparkContext,
       case LogicalRelation(br, _, _) =>
         br match {
           case d: DestroyRelation => d.truncate()
+          case _ => if (!ignoreIfUnsupported) {
+            throw new AnalysisException(s"Table $tableIdent cannot be truncated")
+          }
         }
       case _ => if (!ignoreIfUnsupported) {
         throw new AnalysisException(s"Table $tableIdent cannot be truncated")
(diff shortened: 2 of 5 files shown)


core/build.gradle
```diff
@@ -76,6 +76,7 @@ dependencies {
     exclude(group: 'org.scala-lang', module: 'scala-library')
     exclude(group: 'org.scala-lang', module: 'scala-reflect')
     exclude(group: 'org.scala-lang', module: 'scala-compiler')
+    exclude(group: 'com.univocity', module: 'univocity-parsers')
```
kneeraj: Was this causing any trouble or just by observation you found out?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -907,15 +910,20 @@ class SnappySession(@transient private val sc: SparkContext,
           case _ => // ignore
         }
         cacheManager.uncacheQuery(Dataset.ofRows(this, plan))
-        sessionCatalog.unregisterDataSourceTable(tableIdent, Some(br))
+        if(sessionCatalog.isTemporaryTable(tableIdent)){ // This is due to temp table
```
kneeraj: So "table not found" kind of error was coming when trying to drop temp tables which was directly trying unregisterDataSourceTable?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -230,6 +230,9 @@ class SnappySession(@transient private val sc: SparkContext,
       case LogicalRelation(br, _, _) =>
         br match {
           case d: DestroyRelation => d.truncate()
+          case _ => if (!ignoreIfUnsupported) {
```
kneeraj: Why is this required? How is the case of "_" related to temp tables only.

kneeraj:
Looks ok to me except a couple of questions I have asked for my clarity.

core/build.gradle
rishitesh: This causes trouble if we use CSV reader in our core/example projects, due to version conflicts

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
rishitesh: Correct. As the plan is no more a LogicalPlan, no case was handling it.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
rishitesh: Its for cleanup of relations which supports truncate. It was causing a scala match error if we dont handle it.

PR #358, opened 2016-09-03
[SNAP-1025] Fix for SNAP-1025

SachinJanani:
*  Fix for SNAP-1025
*  Modified test to assert values returned for test

## Patch testing
* Ran precheckin and test pass

## Other PRs 
N/A


core/src/main/scala/org/apache/spark/sql/streaming/StreamSqlHelper.scala
```diff
@@ -68,7 +68,7 @@ object StreamSqlHelper {
     val encoder = ExpressionEncoder[A]()
     val schema = encoder.schema
     val logicalPlan = LogicalDStreamPlan(schema.toAttributes,
-      stream.map(encoder.toRow))(ssc)
+      stream.map(encoder.toRow(_).copy()))(ssc)
     new SchemaDStream(ssc, logicalPlan)
   }
```
sumwale: Doesn't the other one below also require it? Perhaps not exercised in any existing test.

sumwale:
@ymahajan yes we need to determine why copy is required here (apparently even 1.6 had this problem). Without copy just the base df.show (without join) itself has the problem though count is correct. Normally each partition iterator will process row by row unless an operator needs to store (e.g. sorting) which will make an explicit copy.

PR #361, opened 2016-09-06
downgrading derby version for tests.

sonalsagarwal:
Using derby 10.8.2.2 in tests rather than 10.12.1.1 .This is because the hydra tests are using the derby just for verification and the higher version has some issues with obtaining CLOBs due to which the tests are not working. 
Since only validation is required with derby results so it is ok to revert the version in tests.

## Patch testing
Ran few bts with the change/

## ReleaseNotes.txt changes
NA.

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/106

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -104,7 +104,7 @@ allprojects {
     scalatestVersion = '2.2.6'
     jettyVersion = '9.2.16.v20160414'
     guavaVersion = '14.0.1'
-    derbyVersion = '10.12.1.1'
+    derbyVersion = '10.8.2.2'
     pegdownVersion = '1.6.0'
     snappyStoreVersion = '1.5.1-SNAPSHOT'
     pulseVersion = '1.5.0'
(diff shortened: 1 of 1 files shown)


sumwale:
I don't think we should change the version in the top-level gradle build. Spark now depends on the newer derby version and could break for those using Spark features that depend on it.

Instead we should fix the tests to comply with the new behaviour. There is another option of having store depend on older derby version like in your other PR (only for testCompile/Runtime or tests project) and then exclude top-level newer derby when building the tests-sql shadowJar. I think its easier to just fix the tests and not go through the hassle of adjusting the build files to work correctly with different versions.

sonalsagarwal:
Testing a fix in the test itself. In that case, we need not change the derby version and above changes can be discarded.

sonalsagarwal:
Reverted the above changes and created new PR for new changes.

New pull request : https://github.com/SnappyDataInc/snappy-store/pull/114

PR #362, opened 2016-09-06
Hive is not getting closed , as it picks up HiveDB from ThreadLocal. …

rishitesh:
  Hive is not getting closed , as it picks up HiveDB from ThreadLocal. SparkListener is a separate thread and it does not clean up the current connection to metastore.
If application starts stops SparkContext then this results in CacheClosedException from stale connections.

Hence added clean-up code while creating a new client. 

## Patch testing

pre-checkin ran successfully

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


kneeraj:
Looks good.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -145,6 +145,8 @@ class SnappyStoreHiveCatalog(externalCatalog: ExternalCatalog,
    */
   protected def configure(): Map[String, String] = Map.empty
 
+
+
```
sumwale: unnecessary whitespace added

sumwale:
looks good

PR #363, opened 2016-09-07
Snap 69

sonalsagarwal:

In jobserver-defaults.conf, rootdir for spark/jobserver/filedao and spark/jobserver/datadao point to /tmp/spark-jobserver. 
For multi users, we cannot overwrite the existing directory created by other user. So, we change the rootdir for above and use gemxd tempdir by default for each lead.  

But as we are creating a new directory for each lead, we loose the previous state when it is restarted.
## Patch testing

Started the cluster to verify that we do not create a /tmp/spark-jobserver dir, and instead create a gemfirexdtemp_\* dir and spark-jobserver is created under it.
## ReleaseNotes.txt changes

NA.
## Other PRs

NA.


nthanvi:
Why not to use the same directory on restart as  lead-HA should be transparent to user. In this case the information regarding the executed jobs etc will be lost and when user will try to re-submit the existing job he will get failure like class do not exist etc.

PR #364, opened 2016-09-07
[SNAP-1036] optimized generated code iteration for row tables

sumwale:

Avoid creation of UnsafeRow copies in generated code for row table iterators.

- in embedded mode, don't push down any projections; push down filters when there are relevant indexes as before
- optimized iterators for: a) directly using Region iteration when no projections or filters, b) on CompactExecRows when filters, c) on ResultSet directly when projections
- adding a RowTableScan plan which uses the above classes to set column variables directly (iterator either returns CompactExecRows or a resultset which is iterated)
- fixed ResultSet iteration to keep two separate variables: doMove and hasNextValue

## Patch testing

precheckin with store ongoing

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/107

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/CompactExecRowToMutableRow.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/CompactExecRowToMutableRow.scala
@@ -31,23 +31,15 @@ import org.apache.spark.sql.types.{DataType, Decimal, DecimalType, StructType}
 import org.apache.spark.unsafe.Platform
 import org.apache.spark.unsafe.types.UTF8String
 
-trait CompactExecRowToMutableRow extends ResultWasNull {
+abstract class CompactExecRowToMutableRow extends ResultNullHolder {
 
   val schema: StructType
-
-  protected final var wasNull: Boolean = _
-
-  protected final lazy val defaultCal = new GregorianCalendar()
 
   protected final val dataTypes: ArrayBuffer[DataType] =
     new ArrayBuffer[DataType](schema.length)
 
   protected final val fieldTypes = StoreUtils.mapCatalystTypes(
     schema, dataTypes)
-
-  override final def setWasNull(): Unit = {
-    wasNull = true
-  }
 
   protected final def createInternalRow(execRow: AbstractCompactExecRow,
       mutableRow: SpecificMutableRow): InternalRow = {
@@ -222,5 +214,16 @@ trait CompactExecRowToMutableRow extends ResultWasNull {
       i = i + 1
     }
     mutableRow
+  }
+}
+
+class ResultNullHolder extends ResultWasNull {
+
+  final var wasNull: Boolean = _
+
+  final val defaultCal = new GregorianCalendar()
+
+  override final def setWasNull(): Unit = {
+    wasNull = true
   }
 }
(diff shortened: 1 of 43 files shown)


core/src/main/scala/org/apache/spark/sql/execution/CompactExecRowToMutableRow.scala
```diff
@@ -31,24 +31,16 @@ import org.apache.spark.sql.types.{DataType, Decimal, DecimalType, StructType}
 import org.apache.spark.unsafe.Platform
 import org.apache.spark.unsafe.types.UTF8String
 
-trait CompactExecRowToMutableRow extends ResultWasNull {
+abstract class CompactExecRowToMutableRow extends ResultNullHolder {
```
kneeraj: Why this "wasnull" interface required. Did not get this. Someone would toggle it to non null once the mutable row is created from the compactexecrow?

sumwale: The wasNull interface is part of ResultSet which is set only after a read of a column to tell whether previous read returned a null value or not. Required for primitive values at least that cannot be tested for nullability otherwise.

We carried over similar into AbstractCompactExecRow read methods which accepts a "ResultWasNull" object to set into it whether the previous read returned a null value. Normally EmbedResultSet implements that interface so that it can return appropriate value for subsequent wasNull() call. However, here we are reading the AbstractCompactExecRow directly hence creating a holder object that can be passed to its getAs* methods (either CompactExecRowToMutableRow or created directly in RowTableScan generated code when using raw region iterator).

sumwale: Also CompactExecRowToMutableRow is not used by row table scan at all. It is only used when inserting as of now. The generated code directly gets handle to AbstractCompactExecRow (either via direct region iterator or peeking inside EmbedResultSet when there are index filters but no projections), then invokes its required getAs* methods passing a common ResultNullHolder object to check whether the result was null or not (the setWasNull is ultimately invoked by RowFormatter is required).

core/src/main/scala/org/apache/spark/sql/execution/CompactExecRowToMutableRow.scala
```diff
@@ -224,3 +216,20 @@ trait CompactExecRowToMutableRow extends ResultWasNull {
     mutableRow
   }
 }
+
+class ResultNullHolder extends ResultWasNull {
+
+  final var wasNull: Boolean = _
+
+  final val defaultCal = new GregorianCalendar()
```
kneeraj: More than a review it will be a doubt clearing exercise :-) ... You can answer in brief?
Why this defaultCal object required in this class?

sumwale: DVD stores encoded date/time that needs a Calendar object to convert to Date object. CompactExecRow.getAsDate/getAsTimestamp require the Calendar for the same reason. Likewise, ResultSet.getDate also requires it. This object avoids creating for every column value and keeps reusing it after clearing.

sumwale:
merging to master; will take care of any additional comments subsequently

PR #365, opened 2016-09-07
[SNAP-1034] vectorized access for CachedBatch buffers

sumwale:

- Adding decoder classes for CachedBatch buffers for each of the supported encodings: Uncompressed, DictionaryEncoding, RunLengthEncoding, IntDeltaEncoding, LongDeltaEncoding, BooleanBitSetEncoding.
   These classes provide for native get* methods to extract values directly from the byte array (rather than using a ByteBuffer and extracting into UnsafeRow as the current Spark implementation does)
   Also provide equivalent next* methods to move to next value. This is required separately and cannot be combined with get* since for filters the get calls for filtered rows will be skipped (but next* has to be done for all columns regardless)
   Encoder methods have been added to base class and Uncompressed but not used yet.
- ColumnTableScan implements the plan to do code generation using above classes. It first iterates over the row buffer (with the code added for RowTableScan in SNAP-1036 then the CachedBatch, and then a loop from 1 to numRows in CachedBatch)
- ColumnTableScan also adds a separate Seq[RDD[InternalRow]] which can be used to pass in any additional RDDs that need to be iterated as normal UnsafeRows. ColumnTableScan will record the partition index beyond which such RDDs will be provided, and uses normal UnsafeRow iteration for those. This is currently used for reservoir RDD of sampling relation.
- A global constant flag in BaseColumnFormatRelation tells whether to use the old UnsafeRow iterator and ColumnTableScan respects it. This is used only for perf comparison and will be removed at some point when new impl is considered stable.

(temp commented out code in JDBCSourceAsColumnarStore for multiple iterations that will be removed subsequently)

Note that unlike the Spark one, the decoder takes care of endianness of the platform for uniformity since for snappydata column tables these CachedBatches can be persisted and potentially used on a different platform (a non-concern for Spark which only uses for temporary caching). Little-endian has been kept as default since the most common hardware namely intel is so. This should be fixed for the encoders too which will be done once support for encoding is also added via generated code and these new classes.

## Patch testing

precheckin

## ReleaseNotes.txt changes

Perhaps can be claimed as a feature over Spark in our docs and other material.

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/8


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -0,0 +1,278 @@
...
+import org.apache.spark.sql.types._
+
+/**
+ * Physical plan node for scanning data from a SnappyData column table RDD.
+ * If user knows that the data is partitioned across all nodes, this SparkPlan
+ * can be used to avoid expensive shuffle and Broadcast joins.
+ * This plan overrides outputPartitioning and makes it inline with the
+ * partitioning of the underlying DataSource.
+ */
+private[sql] final case class ColumnTableScan(
```
rishitesh: May be I am missing something, but could not see where this plan is getting invoked?

rishitesh:
I am assuming our store strategy will create the physical plan "ColumnTableScan". 
To understand more, we are still processing one cached batch at a time, but probably we wont need more than that.
At scan level we don't plan( and most likely should not )  to add any support of batching as Parquet file relation has. 
Most of the cost saving with this PR will be from native calls to byte arrays.

rishitesh:
Looks ok for merge

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
hbhanawat: Rishi, you need to look at https://github.com/SnappyDataInc/snappydata/pull/364 to understand the usage of ColumnTableScan.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -80,11 +81,13 @@ class BaseColumnFormatRelation(
     with PartitionedDataSourceScan
     with RowInsertableRelation {
 
+  val scanAsUnsafeRows = false
```
hbhanawat: Until the code the stabilizes, we should have a SQLConf parameter instead of a global constant to disable the vectorized approach. This would help in TPCH testing to avoid restarting of the servers if we want to try performance of other type of scan. However, for now, the scan is always set as cached batch scan.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -0,0 +1,278 @@
...
+        s"""
+            $rowInput = (partitionIndex < $internalRowPartitionIndex)
+                ? (scala.collection.Iterator)inputs[0].next() : inputs[0];
+        """)
+      ctx.addMutableState("scala.collection.Iterator",
+        colInput,
+        s"""
+            $colInput = (partitionIndex < $internalRowPartitionIndex)
+                ? (scala.collection.Iterator)inputs[0].next() : null;
+        """)
```
hbhanawat: As I understand, partitionIndex for the otherRDDs should be beyond the total number of rdd partitions of column table. Its quite novel (and may be performant) to bring the iteration of three rdds in a single function. However, can this dependency of having otherRDDs partitionindex greater than the column table be avoided. It will be difficult to understand and maintain this later.

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/ColumnEncoding.scala
```diff
@@ -0,0 +1,346 @@
...
+
+  def readMap(columnBytes: Array[Byte]): UnsafeMapData
+
+  def readStruct(columnBytes: Array[Byte], numFields: Int): UnsafeRow
+
+  def initializeEncoding(dataType: DataType, batchSize: Int): Array[Byte]
+
+  def writeBoolean(columnBytes: Array[Byte], value: Boolean): Unit
+
+  def writeByte(columnBytes: Array[Byte], value: Byte): Unit
```
hbhanawat: What are these write functions for? I thought we just have to read the encoding, we would still use Spark code to encode.

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/ColumnEncoding.scala
```diff
@@ -0,0 +1,346 @@
...
+    f.setAccessible(true)
+    f
+  }
+
+  private[columnar] val BITS_PER_LONG = 64
+
+  val littleEndian: Boolean = ByteOrder.nativeOrder == ByteOrder.LITTLE_ENDIAN
+
+  val allEncodings: Array[(DataType, Boolean) => ColumnEncoding] = Array(
+    createPassThrough,
```
hbhanawat: In split mode, when working with someother spark version can this mapping of typeId as index create a problem?

hbhanawat:
There are no tests for various encodings. I doubt that we have test coverage all the types in current tests. We need unit tests for this.

hbhanawat:
Vast code. Reviewed it at a high level.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
sumwale: Yes, sorry I tried to split my full set of changes into separate ones for row and column tables but not fully separate and plan changes are only in #364.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
sumwale: Actually I plan to drop this completely once I get precheckin fully clean and one set of TPCH run completed (hopefully by tomorrow). Hence not added SQLConf setting since it is very temporary mostly for my local testing (for TPCH we already have previous results to compare against).

sumwale:
@rishitesh In fact in future we may need to split the CachedBatches even further if memory pressure is high, since currently all the scanned columns for a CachedBatch have to be in memory. This will not work very well with concurrent queries (was already the case with previous code).

>  At scan level we don't plan( and most likely should not ) to add any support of batching as Parquet file relation has. 

That is exactly what is being done by creation of ColumnTableScan (see ExistingPlans). What Parquet does extra is expose a shell ColumnarBatch.Row for operators that need to use InternalRow (currently sorter, disk writer and for transferring from one stage to other). This avoids a copy into UnsafeRow and is something we can do. In fact we can do better by exposing an InternalRow for all types and not just primitives as ColumnarBatch.Row

>  Most of the cost saving with this PR will be from native calls to byte arrays. 

No, most of the saving is from avoiding the copy from CachedBatch to UnsafeRow and then reading from that again. Now it directly reads from CachedBatch byte arrays into local variables which is about as performant as it can get (apart from using enhanced processor SIMD instructions).

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
sumwale: This has been added primarily for sample RDD which has a reservoir RDD in addition to column table. The problem with avoiding it in code generation is that there is no easy way to combine non-code generated RDD with code generated one (unless we add our own replacement for WholeStageCodegenExec). The result will be fully non-code generated RDD for sampling which will be a complete killer.

> However, can this dependency of having otherRDDs partitionindex greater than the column table be avoided.

We can add our own flavour of UnionRDD for this which will always guarantee that. Right now I took the existing one since it assigns partition IDs in order (and hence the later ones in the Seq will always have higher partition index). Perhaps can be tracked as a JIRA if you think its important.

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/ColumnEncoding.scala
sumwale: My plan was to eventually replace the encoder too using this. Spark's one uses slow byte buffers, multiple levels of indirection and a non-optimal storage of nulls (as integer positions at the start) -- see my commented out code in NullableColumn in this file.

The bigger problem with Spark's encoding is what I have mentioned in the PR too. It always uses Platform.put methods which will use platform specific endianess. For caching purpose it does not matter, but for us these are persisted to disk too so if the files are moved to a different platform then things will completely fall apart. Hence in the read methods I have taken care of endianess and reversed bytes if required (prefers little-endian assuming intel platform is now more widespread than any other). This will be a blocker when we want to support other platforms like solaris, itanium etc.

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/ColumnEncoding.scala
sumwale: The typeId mapping matches what Spark has. Not clear what the question is?

sumwale:
@hbhanawat Yes, I will be extending Spark test suites to use SnappyContext. This is to be done for all kinds of query tests in Spark layer that I will take up once this is done.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
hbhanawat: Actually having a UnionRDD makes sense. I think we should track it.

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/ColumnEncoding.scala
hbhanawat: In split mode, if we are working with a different spark version that has changed the type Ids, this will go for a toss. A remote possibility though.

hbhanawat:
>> Yes, I will be extending Spark test suites to use SnappyContext. This is to be done for all kinds of query tests in Spark layer that I will take up once this is done.

Ok. Then this looks good for merge.

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/ColumnEncoding.scala
sumwale: In split mode if we are working with a different spark version, then many more serious issues can come up. We can support only the versions that are exactly compatible with what we are shipping with (i.e. 2.0.0 for now).

hbhanawat: Hmm. Haven't thought too much about it but since we are discussing, should we release Snappy with different versions of Spark in future. Otherwise, it would be very limiting. For e.g. when we release Snappy 1.0 with Spark 2.1, we should also release Snappy 1.0 with Spark 2.0. Something like what Spark does with hadoop versions.

sumwale: I don't know if we have any bandwidth to do such a thing. It could be done for patch version changes but any move to support something like 2.0 and 2.1 can take a lot of effort and maintaining different copies of code given that Spark interfaces change dramatically over minor revisions too (e.g. recall 1.4 to 1.5 or 1.5 to 1.6). Not to mention double testing/perf effort and so on.

Normally we expect people to be using our bundled spark distribution, but for those who cannot or are tied to specific distributions. You can bring this up in next huddle or status with Jags/Sudhir, but I don't think it is feasible.

jramnara: This is less important right now. We should just stick to the most stable and performant version of Spark. The product fully encapsulates Spark so it isn't the same as Spark working with an external Hadoop.

We may have to embark on becoming a layered product on top of Apache Spark which would drive this version support discussion. And, of course, once the product is GA, I would expect customers to be using different versions of Snappy and hence different versions of Spark.

PR #366, opened 2016-09-07
Excluded stock spark binaries while building cluster project.

hbhanawat:

Excluded stock spark binaries while building cluster project


ymahajan:
looks good, thanks for doing this.

sumwale:
should also be done for aqp/dtests/... for test runtime

PR #367, opened 2016-09-07
Excluded stock spark binaries while building dtest

hbhanawat:

Excluded stock spark binaries while building dtest


ymahajan:
looks good.

PR #369, opened 2016-09-09
SNAP-1035

ymahajan:

Disabled ColumnFormatRelation.flushRowBuffer for split mode, as no GemFireCache available on Executors

We recently added an optimization of flushing row buffer for the df.write.insertInto case, so that the remaining rows will also be flushed to column store. 
Since ColumnFormatRelation.flushLocalBuckets is executed on Executors, there won't be any GemFireCache available in split mode. 

## Patch testing

Added split mode dunit tests with df.insertInto using northwind schema 

## ReleaseNotes.txt changes

No change

## Other PRs 

No other PR

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
@@ -38,7 +38,7 @@ import org.apache.spark.sql.row.GemFireXDDialect
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.store.{CodeGeneration, StoreUtils}
 import org.apache.spark.sql.types.StructType
-import org.apache.spark.sql.{DataFrame, Row, SQLContext, SaveMode, SnappySession}
+import org.apache.spark.sql._
 
 /**
  * This class acts as a DataSource provider for column format tables provided Snappy.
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -192,7 +192,12 @@ class BaseColumnFormatRelation(
       truncate()
     }
     JdbcExtendedUtils.saveTable(data, table, schema, connProperties)
-    flushRowBuffer()
+
+    SnappyContext.getClusterMode(_context.sparkContext) match {
```
hbhanawat: Low priority comment: This can be moved to ExternalStoreUtils, just like isSplitOrLocalMode.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -192,7 +192,12 @@ class BaseColumnFormatRelation(
...
+      case SnappyEmbeddedMode(_, _) => flushRowBuffer()
```
hbhanawat: combine the two conditions in a single line.

hbhanawat:
Looks good.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -192,7 +192,12 @@ class BaseColumnFormatRelation(
...
+      case LocalMode(_, _) => flushRowBuffer()
+      case _ =>
+    }
```
sumwale: Strange, the same call has been there in ColumnFormatSamplingRelation for very long but similar issue never reported for sample. Due to that other call I would suggest that this check go inside flushRowBuffer() itself.

PR #372, opened 2016-09-13
Snap 1047

rishitesh:
Changed cached insert logic to be extendible . 
Updated PartitionRegionStats to reflect true row count.
This stats can be used by vsd , mbeans to reflect true row count.
## Patch testing

precheckin

## ReleaseNotes.txt changes
NA
## Other PRs 
yes. store related PR

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
@@ -169,7 +169,8 @@ class BaseColumnFormatRelation(
   override def cachedBatchAggregate(batch: CachedBatch): Unit = {
     // if number of rows are greater than columnBatchSize then store
     // otherwise store locally
-    if (batch.numRows >= Constant.COLUMN_MIN_BATCH_SIZE) {
+    if (batch.numRows >= Constant.COLUMN_MIN_BATCH_SIZE ||
+        java.lang.Boolean.getBoolean("forceFlush")) {
       externalStore.storeCachedBatch(ColumnFormatRelation.
           cachedBatchTableName(table), batch)
     } else {
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
@@ -17,6 +17,7 @@
 package org.apache.spark.sql.execution.columnar.impl
 
 import java.sql.{Connection, ResultSet, Statement}
+import java.util.UUID
 
 import com.gemstone.gemfire.internal.cache.{AbstractRegion, PartitionedRegion}
 import com.pivotal.gemfirexd.internal.engine.Misc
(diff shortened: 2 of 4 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -169,7 +169,8 @@ class BaseColumnFormatRelation(
   override def cachedBatchAggregate(batch: CachedBatch): Unit = {
     // if number of rows are greater than columnBatchSize then store
     // otherwise store locally
-    if (batch.numRows >= Constant.COLUMN_MIN_BATCH_SIZE) {
+    if (batch.numRows >= Constant.COLUMN_MIN_BATCH_SIZE ||
+        java.lang.Boolean.getBoolean("forceFlush")) {
```
kneeraj: Is this a new property you have added? How is this related to this change?

rishitesh: This is test flag. We need to test some cases where we intend to flush always.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -58,6 +59,28 @@ final class JDBCSourceAsColumnarStore(_connProperties: ConnectionProperties,
...
+      batchId: UUID, partitionId: Int): (Connection => Any) = {
+    {
+      (connection: Connection) => {
+        super.doInsert(tableName, batch, batchId, partitionId)(connection)
+        connectionType match {
+          case ConnectionType.Embedded =>
+            val resolvedName = ExternalStoreUtils.lookupName(tableName,
+              connection.getSchema)
+            val region = Misc.getRegionForTable(resolvedName, true)
+            region.asInstanceOf[AbstractRegion] match {
```
kneeraj: What is the plan for Split for later? Can you file a jira just for a place holder. Later we can update it with the approach.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
kneeraj: Just for testing?

rishitesh: Yes. Our scala tests uses a batch size of 3. if we use this flag it will always insert in cached batch table. Later we can validate number of rows which landed in column table.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
rishitesh: Yes sure. Will add a JIRA.

kneeraj:
Looks good.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
rishitesh: https://jira.snappydata.io/browse/SNAP-1048 added.

PR #374, opened 2016-09-14
Scheduling mode as fair 

hbhanawat:

Made the scheduling mode of the default pool as FAIR after a discussion with the team

## Patch testing

Did some manual testing. Ran the quick start successfully.


sumwale:
Looks good. Have you tested that this actually applies and shows as FAIR in the GUI (or elsewhere)?

hbhanawat:
Yes, it shows that in UI.

PR #380, opened 2016-09-24
Snap 730

soubhik-c:

Support global index and colocated joins/filter whenever possible
## Patch testing

precheckin and additional tests upto 3 way join. more join relations to be tested.
## ReleaseNotes.txt changes

Yes, listed this bug id.
## Other PRs

https://github.com/SnappyDataInc/snappy-store/pull/115

Distributed Index:

```
+ populate the index during creation of index.
+ WHERE or JOIN predicates uses distributed index if matches.
+ Added index -> index colocation and querying
+ more test cases added and separated out corner cases that needs to be
addressed.

* Adding index query hint

    + Now explicit index choice can be provided with query hint
    + Added tests for the same.

+ Now redundant predicates in where clause or join clause works fine.

+ Support index append while inserting into table.

    * When RowBuffer gets converted to CachedBatch, indexes are maintained
     now.

    * Renamed withIndex queryhint to simply 'index'.
    e.g. select * from tab --+index(XXX)

    * Maintaining index dependencies on each Executors and therefore renamed
     'stores' variable to 'executorCatalog' and the map now have a case
     class instead of a tuple.

    * Introduced indexCache as a separate generated code for index
      maintaenance whereby just 'stmt.setXXX()' and 'stmt.addBatch' happen
      one row at a time. This avoids row cloning obtained from RowBuffer
      using scan controller that reuses template row.

    * ExternalStore now acquires a longer duration connection. BaseTable
      CachedBatch creation and index preparedStatement.addBatch all happen
      under a single connection and hence can be transactionally controlled
      later.

    * ExecutorCatalog captures index schema separately in anticipation to
      column groups that will be introduced for an index.


+ Adding base table rowbuffer scan

    + normalizing the table schema to uppercase during as per sparkConf
    df.write.format("column")...saveAsTable("xxx")

      this is required otherwise create index and table schema's column name
      case mismatch results into exception.

    + refactored BaseColumnFormatRelation.buildUnsafeScan and splitted into
    buildRowBufferScan which is used by
    IndexColumnFormatRelation.buildBaseTableRowBufferScan

+ Introducing shuffleExchange on base table rowbuffer during index scan

+ Cleaning up ExternalStore.getConnectedExternalStore

    + This now fixes JDBCSourceAsColumnarStore storeCachedBatch and hence
    passes more tests.

+ more code cleanup and NorthWindTest is now passing.

+ Moving to optimizer rule instead of analysis

    + DI choice ideally is an optimization and should be applied along with
    other rules like joinOrder done by Hemant.

    + This will help in cost based optimization as well later.

    tests: precheckin

+ moving to more stricter partitioning column check

    right now partition_by can be expressed in terms of expression instead
    of just column names in column partitioning.

    gemxd partition by expr and spark expr function sets may or maynot match
    exactly. hence unless we have a more sophisticated expression matcher
    b/w gem and spark way, doing a strict colum name only check for picking
    up indexes.

    so now following example won't be supported,

       create index i1 (...) partition_by column( month(ts) )
       select * from tab where month(ts) == 1

+ supporting no index as query hint

    Per suggestion, now one can issue 'select * from tab --+index() ' to
    disable optimizer to pick up any index.

    added test to verify it.
```

======================== next major changes ==============================

  Handling n-way join first cut

General changes:
1. added lots of scaladoc information and code comments necessary
   to understand the reasoning and scenarios catered to.
2. moved much of the code RuleUtils. Added new subrule file capturing
   join order strategies.
3. introduced Replacement and ColocatedReplacements to capture
   colocation chains that simplifies the code.

===== detail change log ====================
- Added subrules for deciding join order and introduced various
- JoinOrderStrategies in query hints like

select \* from /_joinOrder=RWF,NCF_/ table

this will apply replicated tables with Filters at the leaf of left deep
join tree and then non-colocated tables with Filters and then largest
colocation chain. Its mandatory to mention all the join order
strategies.

for details see JOS and JoinOrderStrategy classes.
- By Default giving priority to max colocation chain, if same, pick based on max
  no. of  partitioning cols and then by index size.
- Index choice test with 3 or more tables have few conditions checked
  and few more to come.
- toString is now shortened for Replacement class.
- capturing the joinRefs cols incrementally as we go, because table
  might get replaced with an index and hence existing join keys won't get
  picked up just based on table to table reference accumulation.
- Corrected table to index AttributeReference mapping based on an array
  instead of relying on HashSet (i.e. AttributeReference equality).
  Reason being, AttributeReference equality checks for exprId which will
  be different when mapping table to index Column references. AttributeSet
  also is infeasible here because it will create a fresh set of
  AttributeReference objects that won't map to index relation schema.
- got rid of replaceWithIndex method and filter, join cases in the
  ResolveIndex#apply method. Its now more simpler in terms of handling
  multiple tables while we apply multiple transformations on plan.
  (retaing replaceWithIndex method which we will remove subsequently)
- Added more tests primarily picking up max partition cols index.
- Fixing more complex cases
- ResolveIndex rule is now before ReOrderJoin because it otherwise
  pushes down filters that anyway required to be pulled out for
  colocation testing. This simplifies our join order strategies.
- blowing up the join conditions for predicate that constitutes
  joinKeys. filtering out filter conditions because all the
  conditions are now in PartialPlan.conditions
- added pr -> rep -> pr based indirect colocation test
- Got rid of DefaultAction class instead moved much of the code to
  RuleUtils. Helps in debugging and avoids double evaluation.
- Made the join order strategies to include and exclude filters instead
  of *WithFilters classes whereever applicable. ApplyRest now addresses
  replicated and non-colocated without filter tables first before blind
  folding the remaining tables.
- Moved common foldLeft code inside PartialPlan that helps in
  simplifying foldLeft on various collections like
  PartialPlan.replicated etc.
- Moving CreateIndexTest to spark.sql.store so that more indepth
  verification of plans can be done.
- Improved readability of the code by changing to shorthand format of
  foldLeft

- Moving the shuffleExchange rdd generation to .execute phase instead of
logicalPlan generation time. This fixes baseTable rowBufferScan.

- Introducing a generic ZipPartitionScan that takes a base SparkPlan and
another SparkPlan while maintaining WholeStage generation on the base
plan and executing other SparkPlan. (With Sumedh's help)

- Simplified createFromDataSource logic whereby
IndexColumnFormatRelation is handled alongside Column and Row scans.
BaseTable RowBufferScan happens remotely using projection/filter push
down during 3 or more tables joins.

- Added tests for the same, added beforeEach and afterEach that clears
up indexes/tables irrespective of exceptions in the test. This helps
subsequent tests not failing. Added various debug flags in the test that
prints useful information during code compilation failures.

- Corrected the test to invoke df.collect and print validation failures
accurately.


core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
```diff
@@ -0,0 +1,410 @@
...
+  lazy val analyzer = snappySession.sessionState.analyzer
+
+  override def apply(plan: LogicalPlan): LogicalPlan = replaceWithIndex(plan, { optimizedPlan =>
+
+    optimizedPlan.flatMap {
+      case ExtractEquiJoinKeys(joinType, leftKeys, rightKeys, extraConditions, left, right) =>
+        (left, right) match {
+          case HasColocatedEntities(colocatedEntities, leftReplacements, rightReplacements) =>
+
+            val joinKeys = leftKeys.zip(rightKeys).flatMap { case (leftExp, rightExp) =>
```
hbhanawat: Why do we need this block? just the following thing should have done the trick: 

val joinKeys = leftKeys.zip(rightKeys)

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
```diff
@@ -0,0 +1,410 @@
...
+
+    val rightRegion = Misc.getRegionForTable(right.resolvedName, true)
+    val rightLeader = rightRegion.asInstanceOf[AbstractRegion] match {
+      case pr: PartitionedRegion => ColocationHelper.getLeaderRegionName(pr)
+    }
+
+    leftLeader.equals(rightLeader)
+  }
+
+  private def fetchIndexes2(plan: LogicalPlan): Seq[LogicalPlan] =
```
hbhanawat: Remove fetchIndexes2 code. Not required.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
```diff
@@ -0,0 +1,410 @@
...
+        * will output :
+        * (1, 4)
+        * (1, 5)
+        * (2, 4)
+        * (2, 5)
+        * (3, 4)
+        * (3, 5)
+        *
+        */
+      val leftRightEntityMapping = for {
```
hbhanawat: Why is this for loop when fetchIndexes is returning a single pair.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
```diff
@@ -0,0 +1,410 @@
...
+        (leftTable, leftIndexes) <- fetchIndexes(left)
+        (rightTable, rightIndexes) <- fetchIndexes(right)
+        leftSeq = Seq(leftTable) ++ leftIndexes
+        rightSeq = Seq(rightTable) ++ rightIndexes
+      } yield {
+        leftSeq.zip(Seq.fill(leftSeq.length)(rightSeq)) flatMap {
```
hbhanawat: Is this is creating a pairing between the two sides of the join with the first pair that of the table? Can the code be simplified?

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
```diff
@@ -0,0 +1,410 @@
...
+        val (leftTable, rightTable) = mappedElements(0) // first pairing is always (table, table)
+        for {
+
+          (leftPlan, rightPlan) <- mappedElements
+
+          leftRelation = unwrapBaseColumnRelation(leftPlan) if leftRelation.nonEmpty
+          rightRelation = unwrapBaseColumnRelation(rightPlan) if rightRelation.nonEmpty
+
+          if isColocated(leftRelation.get, rightRelation.get)
+        } yield {
```
hbhanawat: Since we are expecting only one left and right replacement, can we use a construct other than for-yield?

core/src/main/scala/org/apache/spark/sql/store/CodeGeneration.scala
```diff
@@ -260,6 +283,32 @@ object CodeGeneration extends Logging {
         "dialect")).asInstanceOf[GeneratedStatement]
   }
 
+
+  private[this] def compileGeneratedIndexUpdate(table: String,
```
hbhanawat: Why do we need the complexity of generated code when we are adding one row at a time? For table updates, multiple rows were being added in a single update command and iterating over them is faster with codegen. Wouldn't prepared statement have been enough in this case?  If I am not missing something basic, we should remove this whole code generation part for the index updates.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
```diff
@@ -0,0 +1,410 @@
...
+import org.apache.spark.sql.catalyst.plans.logical.{LogicalPlan, SubqueryAlias}
+import org.apache.spark.sql.catalyst.rules.Rule
+import org.apache.spark.sql.execution.columnar.impl.{BaseColumnFormatRelation, ColumnFormatRelation, IndexColumnFormatRelation}
+import org.apache.spark.sql.execution.datasources.LogicalRelation
+
+
+/**
+  * Replace table with index hint
+  */
+case class ResolveQueryHints(snappySession: SnappySession) extends Rule[LogicalPlan]
```
hbhanawat: What query hint do I have to give if I want table to not pick any index?

core/src/test/scala/io/snappydata/app/CreateIndexTest.scala
```diff
@@ -16,33 +16,104 @@
  */
 package io.snappydata.app
```
hbhanawat: No DUnits? I think we need few Dunits to test both the insertion and query validation.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -122,6 +123,28 @@ class SnappyParser(session: SnappySession)
...
+  private final def updatePerTableQueryHint(tableIdent: TableIdentifier,
+      optAlias: Option[String]) = {
+    val indexHint = queryHints.remove(QueryHint.Index.toString)
+    if (indexHint.nonEmpty) {
+      val table = optAlias match {
+        case Some(alias) => alias
+        case _ => tableIdent.unquotedString
+      }
+
+      queryHints.put(QueryHint.Index.toString + table, indexHint.get)
```
soubhik-c: queryHints gets cleared across multiple queries. If you mean multiple tables, it is per tableIdentifier.

select * from t1 /*++index(dd)*/ , t2 /*++index(d3)*/ where ....

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
soubhik-c: well generally either its some index or default behaviour of optimizer. We can add blank index name to have this feature. fairly easy to do if we need it.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
soubhik-c: to support simple unary functions like trim, month, nvl etc.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
soubhik-c: fetchIndexes is returning a collection (not single pair) and for comprehension is opening an iterator.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
soubhik-c: This is generating the search space for 2 tables and its indexes.
example: tab1 having 2 indexes defined and tab2 is having 3 indexes defined. as shown in the commented code it will generate all combinatorial pairs b/w the tables & indexes. So, this will generate a seq of pairs viz. (t1, t2), (t1, t2_i1), (t1, t2_i2), (t1, t2_i3), (t1_i1, t2), (t1_i2, t2). Rest of the code just eliminates infeasible join relations and comes up with colocated join pairs.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
soubhik-c: one left and right replacement for each element in mappedElements. Again keep in mind a <- col is like for (a : col) in java.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
soubhik-c: yup. :)

core/src/main/scala/org/apache/spark/sql/store/CodeGeneration.scala
soubhik-c: code generation is required to set each parameter of the table. like in citi, if table has 250 columns, setting each column of an index otherwise will incur lots of switch case runtime evaluation. this have direct bearing on performance of CachedBatch creation and IMO is required.

core/src/test/scala/io/snappydata/app/CreateIndexTest.scala
soubhik-c: right, I will be putting in dunits but before that have lots of outstanding items still to cover. for e.g. I am in the process of putting in replicated region and 3 or more N-way join index pickup.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
soubhik-c: I got mislead by the name partitioningColumn in ColumnFormatRelation populated during create table.
On second thought, I suppose we need to do a more exhaustive check of expression equivalence between gemxd style expr and spark style "catalyst.expressions.Expression". This code for now should expect AttributedReference and do a strict column name check. Will alter it accordingly.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
hbhanawat: It will be good to have to compare the index performance against the table performance.

core/src/main/scala/org/apache/spark/sql/store/CodeGeneration.scala
hbhanawat: + @sumwale 

There is no hot-loop in your code. Did you quantify the difference between the performance with code gen and without code -gen? From what I think, you would not get any benefit. Code generation adds lot of complexity and is a debugging challenge. We should not add code generation unless we are sure about the benefits.

soubhik-c: I am not sure what additional complexity you are talking about, it is reusing table generation code refactored.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
soubhik-c: right, but IMO we should do it once I add best index selection code for n-way join (which I am about to complete in couple of days).

core/src/main/scala/org/apache/spark/sql/execution/columnar/CachedBatchHolder.scala
```diff
@@ -16,11 +16,19 @@
  */
 package org.apache.spark.sql.execution.columnar
 
+import java.sql.PreparedStatement
+
+import scala.collection.mutable.ArrayBuffer
+
 import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.execution.columnar.impl.{ColumnFormatRelation, IndexColumnFormatRelation}
```
soubhik-c: it already showing up with Index*Rel in toString. Have fixed that in BaseColumnFormatRelation.toString

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
```diff
@@ -20,61 +20,76 @@ import java.lang
...
-  def registerExternalStoreAndSchema(context: SQLContext, tableName: String,
-      schema: StructType, externalStore: ExternalStore,
-      batchSize: Int, compress: Boolean): Unit = {
-    stores.synchronized {
-      stores.get(tableName) match {
-        case None => stores.put(tableName, (schema, externalStore))
-        case Some((previousSchema, _)) =>
-          if (previousSchema != schema) {
-            stores.put(tableName, (schema, externalStore))
+  def registerExternalStoreAndSchema(catalogEntry: ExecutorCatalogEntry): Unit = {
```
soubhik-c: right. I have just played along with @suranjan 's changes :).

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
```diff
@@ -0,0 +1,399 @@
+/*
```
soubhik-c: right. was feeling too lazy to rename .. will do that.

core/src/main/scala/org/apache/spark/sql/sources/SparkOptimizations.scala
```diff
@@ -0,0 +1,399 @@
...
+
+  private def unwrapBaseColumnRelation(
+      plan: LogicalPlan): Option[BaseColumnFormatRelation] = plan collectFirst {
+    case LogicalRelation(relation: BaseColumnFormatRelation, _, _) =>
+      relation
+    case SubqueryAlias(alias, LogicalRelation(relation: BaseColumnFormatRelation, _, _)) =>
+      relation
+  }
+
+  private def isColocated(left: BaseColumnFormatRelation, right: BaseColumnFormatRelation) = {
```
soubhik-c: not very clear about the last portion. Did you mean, create table ... partition_by `spark expr` ? Assuming two tables created over the api using same partitioner can take advantage of colocated join, this guy will simply think tables are non-colocated. If there are indexes defined by our way and those indexes on each table are declared as colocated, then yes, this will be replaced with indexes on both side. Isn't that we want ?

core/src/test/scala/io/snappydata/app/CreateIndexTest.scala
```diff
@@ -138,12 +586,62 @@ class CreateIndexTest extends SnappyFunSuite {
...
+        selectRes.show
+      }
+
+      selectRes
+    }
+
   }
+
+  private def validateIndex(index: Seq[String], tables: String*)(df: DataFrame): Unit = {
+    val (indexesMatched, indexesUnMatched) = df.queryExecution.optimizedPlan.collect {
```
soubhik-c: I suppose optimizedPlan will only have IndexColumnRel which gets translated into ColumnTableScan within SparkPlan and is hard to distinguish b/w table vs index scan. Yes I agree we are really avoiding exchange check should be there. One of the item in my todo list :)

core/src/test/scala/io/snappydata/app/CreateIndexTest.scala
```diff
@@ -138,12 +586,62 @@ class CreateIndexTest extends SnappyFunSuite {
...
+    }.partition(tab => tables.exists(t => tab.table.indexOf(t.toUpperCase) > 0))
+
+    if (tablesFound.size != tables.size) {
+      df.explain(true)
+      fail(s"Expected tables $tables not found. UnFound tables ${tablesNotFound} ")
+    }
+
+
+  }
+
```
soubhik-c: good point. noted but will do it later.

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -0,0 +1,772 @@
...
+      tablesAppeared.exists(t => t.indexOf(tab.toUpperCase) > 0))
+
+    val unexpected = tablesAppeared.filterNot(tab =>
+      tables.exists(t => tab.indexOf(t.toUpperCase) > 0))
+
+    if (tablesFound.size != tables.size) {
+      df.explain(true)
+      fail(s"Expected tables ${tables.mkString(",")} but not found " +
+          s"${tablesNotFound.mkString(",")} with additional tables ${unexpected.mkString(",")}")
+    }
```
ymahajan: @soubhik-c same strict validation needed as mentioned for indexes case above

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -0,0 +1,772 @@
...
+        s"and t1.col4 = xx.col5 ") {
+      // t1 -> t4, t2 -> t4
+      validateIndex(Seq.empty, table1, table2, table4)(_)
+    }
+
+    executeQ(s"select t1.col2, t2.col3 from $table1 t1, $table4 t4, $rtable5 t5, $table2 t2 " +
+        s"where t1.col4 = t4.col4 " +
+        s"and t1.col2 = t5.col2 and t1.col3 = t5.col3 " +
+        s"and t5.col2 = t2.col2 and t5.col3 = t2.col3 ") {
+      // t1 -> t4, t1 -> t5 -> t2
```
ymahajan: @soubhik-c  Doesn't this require queryHint "includeGeneratedPaths" or it is enabled by default ?

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
```diff
@@ -104,8 +110,21 @@ private[sql] object PartitionedPhysicalScan {
...
+          partitionColumns, relation, allFilters, schemaAttributes)
+        val table = i.getBaseTableRelation
+        val (a, f) = scanBuilderArgs
+        val baseTableRDD = table.buildRowBufferRDD(Array.empty,
+          a.map(_.name).toArray, f.toArray, false)
+        val rowBufferScan = RowTableScan(output, baseTableRDD, numBuckets,
+          Seq.empty, table)
+        val bufferExchange = ShuffleExchange(columnScan.outputPartitioning,
+          rowBufferScan)
+        ZipPartitionScan(columnScan, bufferExchange)
```
ymahajan: So rowBuffer will always need to do exchange to match the column table indexScan partitioning.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -58,22 +59,37 @@ class SnappySessionState(snappySession: SnappySession)
...
+  override lazy val optimizer: Optimizer = new SparkOptimizer(catalog, conf, experimentalMethods) {
+    override def batches: Seq[Batch] = {
+      implicit val ss = snappySession
+      var insertedSnappyOpts = 0
+      val modified = super.batches.map {
+        case batch if batch.name.equalsIgnoreCase("Operator Optimizations") =>
+          insertedSnappyOpts += 1
+          val (left, right) = batch.rules.splitAt(batch.rules.indexOf(ReorderJoin))
+          Batch(batch.name, batch.strategy, left ++ Some(ResolveIndex()) ++ right
+              : _*)
```
ymahajan: @soubhik-c Looks like ResolveIndex rule is now applied before ReorderJoin? to avoid filter pushdown before ResolveIndex?

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -0,0 +1,772 @@
...
+    indexesToDrop += index4
+
+    snContext.sql(s"create index $index31 on $table3 " +
+        s" (COL2, COL3) Options (colocate_with  '$index4')")
+    indexesToDrop += index31
+
+    val executeQ = QueryExecutor(snContext)
+
+    executeQ(s"select * from $table1 tab1 join $table3 tab2 on tab1.col1 = tab2.col1") {
+      validateIndex(Seq(index2), table3)(_)
```
soubhik-c: @ymahajan thanks for looking at the tests :) .. yes basis for index choice is most number of partitioning columns. Largest colocation chain overrides choice of index though.

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -0,0 +1,772 @@
...
+      // t1 -> t2, t1 -> t4
+      validateIndex(Seq(index1), table2, table4)(_)
+    }
+
+    executeQ(s"select t1.col2, t2.col3 from $table1 t1, $table4 t4, $table2 t2 " +
+        s"where t1.col2 = t2.col2 and t1.col3 = t2.col3 " +
+        s"and t1.col4 = t4.col4 ") {
+      // t1 -> t4, t1 -> t2
+      validateIndex(Seq(index1), table2, table4)(_)
+    }
```
soubhik-c: @ymahajan only one colocated join is possible for a given query. Picking up another global index instead of a table will still result in shuffle. So, I don't see any advantage of picking up additional index other than load balancing on the 3rd table scan. But as members are anyway limited and if indexes belong to different redundancy zone then probably this is worth it otherwise not.

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
soubhik-c: @ymahajan by default generatedPaths are disabled meaning if atleast one user provided join relation is found, generated path with replicated table in between is never attempted. Here its picking up index4 which is explicit relation b/w table1 and table4.

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -0,0 +1,772 @@
...
+  private def validateIndex(index: Seq[String], tables: String*)(df: DataFrame): Unit = {
+    val (indexesMatched, indexesUnMatched) = df.queryExecution.optimizedPlan.collect {
+      case l@LogicalRelation(idx: IndexColumnFormatRelation, _, _) => idx
+    }.partition(rel => index.exists(i => rel.table.indexOf(i.toUpperCase) > 0))
+
+    if (indexesMatched.size != index.size) {
+      df.explain(true)
+      fail(s"Expected index selection ${index.mkString(",")}, but found" +
+          s" ${indexesUnMatched.mkString(",")}")
+    }
```
soubhik-c: @ymahajan it is doing strict validation based on incoming indexes. here its printing actual names of the indexes matched.
```java
val (indexesMatched, indexesUnMatched) = df.queryExecution.optimizedPlan.collect {
+      case l@LogicalRelation(idx: IndexColumnFormatRelation, _, _) => idx
+    }.partition(rel => index.exists(i => rel.table.indexOf(i.toUpperCase) > 0))
+
```

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
soubhik-c: @ymahajan yeah :) atleast pushing the filters and predicates down to rowBuffers although non-pushable filters won't get in there :(.. Earlier it was pushing down filters that are non-pushable at ColumnBuffer level but now due to the way code warrants, i made it optional. Sumedh and me think that rowbuffer will be minimal and probably it is okay. If performance profiling shows otherwise we might need to push non-pushable fitlers and projection down there.

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -0,0 +1,772 @@
...
+
+    executeQ(s"select t1.col2, t2.col3 from $table1 t1 join $table2 t2 on t1.col2 = t2.col2 " +
+        s"where t1.col3 = t2.col3 ", true) {
+      validateIndex(Seq(index2), table2)(_)
+    }
+
+    executeQ(s"select t1.col2, t2.col3 from $table1 t1 join $table2 t2 on t1.col2 = t2.col2 " +
+        s"where t1.col2 = t2.col2 and t1.col3 = t2.col3 ") { df =>
+      // validateIndex(Seq(index2", table2"))(_)
```
soubhik-c: @ymahajan good catch. No here join conditions are repeating on col2 but that shouldn't stop picking up the index. Lost in translation.. will enable it and check.

soubhik-c:
precheckin is clean.

PR #383, opened 2016-09-27
[SNAP-1066] Fix for bug SNAP-1066

SachinJanani:
* Removed Databricks csv parser and added mapping for that to support backward compability

## Patch testing
Ran precheckin some tests failed but they were not related to this changes

## ReleaseNotes.txt changes

## Other PRs 
https://github.com/SnappyDataInc/spark/pull/12


ymahajan:
Looks good, this is discussed with @sumwale  also. testing done manually, observed only in sql script. covered in spark tests also. We should add a test using sql script in NorthWindTest so that these issues will be caught in precheckin itself

PR #387, opened 2016-10-04
Snap 1021

rishitesh:

Change : Added one ExternalCatalog to SnappyStore.

 We were already using a hive meta store as an external catalog. But in code,  we were not using it the SparkSession way. SparkSession has provision for an external catalog which is shared across all sessions and a session catalog which is session specific. But we were still using an InMemoryCatalog as External catalog. 

What was the problem :
 If not using an external catalog properly, which points to the underlying actual catalog , a lot of internal APIs as well as an external API Catalog.scala is broken for Snappy. To fix them a lot of code duplication was required in SnappyHiveStoreCatalog e.g. listTables, listFunctions etc. 

After the change: SnappyStoreHiveCatalog should only override those methods which are really needed and we are writing some custom code and changing behavior of SessionCatalog. 

## Patch testing

pre-checkin clean

## ReleaseNotes.txt changes
NA

## Other PRs 

AQP related changes
https://github.com/SnappyDataInc/snappy-aqp/pull/99


core/src/main/scala/org/apache/spark/sql/hive/HiveClientUtil.scala
```diff
@@ -0,0 +1,289 @@
...
+
+  val HIVE_METASTORE_VERSION = HiveUtils.HIVE_METASTORE_VERSION
+  val HIVE_METASTORE_JARS = HiveUtils.HIVE_METASTORE_JARS
+  val HIVE_METASTORE_SHARED_PREFIXES =
+    HiveUtils.HIVE_METASTORE_SHARED_PREFIXES
+  val HIVE_METASTORE_BARRIER_PREFIXES =
+    HiveUtils.HIVE_METASTORE_BARRIER_PREFIXES
+
+  val HIVE_PROVIDER = "spark.sql.sources.provider"
+  val HIVE_SCHEMA_NUMPARTS = "spark.sql.sources.schema.numParts"
```
rishitesh: Will clean the code once merged.

PR #388, opened 2016-10-05
SNAP-1113: Remove changes in primitiveType() for parsing STRING as CharType

ashetkar:
* Remove changes in primitiveType() which were parsing STRING as CharType instead of StringType.
* Instead do it in columnCharType() which affects only the snappy DDL path.

## Patch testing
* Ran precheckin -Pspark. No new failures.

## ReleaseNotes.txt changes

## Other PRs 
NA


sumwale:
Looks good for merge. Thanks.

PR #392, opened 2016-10-11
[SNAP-1083] Fixing multiple issues in RDD de-linking

sumwale:

There were multiple issues in RDD de-linking that are addressed here.

1. Partitioning assigned to row buffer could be different from that assigned to column store due to its dynamic nature (e.g. preferred node for bucket is different as returned by GemFire or executors change); in general partitioning needs to be aligned for entire collocation chain
2. numPartitions returned by PartitionedDataSourceScan could be different from the actual runtime partitioning determined for the RDD
3. In Spark's changes, it applies a targetPartitioning with numBuckets to create Exchange. This does not work since Exchange result cannot honour numBuckets (only numPartitions) and in general cannot be collocated with a table which is partitioned in similar manner. This bug was not caught previously since we never ran a test where number of table partitions (i.e. total number of cores) is same as number of shuffle partitions. This is done by default now (see 4). Taken care of in Spark layer PR.
4. Also set shuffle partitions as per total number of cores in the system. An extended SnappyConf now overrides SQLConf to do that.

Point 3 above means that we can no longer apply the optimization to shuffle only one table when there is no collocation by setting number of shuffle partitions same as number of buckets (and because our hashing is exactly compatible with Catalyst's). After thinking about it, the only viable solution seems to be to optimize partition creation etc to the extent that creating larger number of partitions has negligible effect on overall timings. I intend to do that separately in any case (1. use Kryo as closure serializer with dynamic registration of new classes automatically using a pooled kryo like in https://github.com/SnappyDataInc/snappy-spark/pull/42, 2. optimize ClosureCleaner)

- inherit the partitioning of column store to its row buffer and not determine
  it independently since "preferred node" for a bucket as returned by GemFire
  can be different for the two collocated regions
- take out numPartitions from Snappy BaseRelation implementations since it is
  determined dynamically at runtime by the data RDD and use numBuckets instead
  to determine the special replicated case
- change numPartitions in Snappy's PartitionedDataSource plan implementations
  to determine numPartitions from data RDD
- added generic handling of inheriting the partitioning as determined for a
  leader region down the collocation chain uniformly for all else collocated
  join can be buggy if mapping is determined for all independently (due to
      "preferred nodes" being different or executors changing at runtime)
- added a SnappyConf that extends SQLConf that also determines the shuffle
  partitions dynamically from available executor cores (if not set explicitly)
- updated NorthWind tests for above changes; enabled all disabled queries
  in both scalaTest and dunit since all queries now pass cleanly

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/15


ymahajan:
This will also fix - https://jira.snappydata.io/browse/SNAP-1026 
Q38 was disabled because of SNAP-1026

ymahajan:
@sumwale Will this optimization work for split mode also ?

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -96,8 +114,64 @@ class SnappySessionState(snappySession: SnappySession)
...
+   */
+  @volatile private[this] var dynamicShufflePartitions: Int = SQLConf
+      .SHUFFLE_PARTITIONS.defaultValue match {
+    case _ if session == null => -1
+    case Some(d) => if (super.numShufflePartitions == d) {
+      session.sparkContext.schedulerBackend.defaultParallelism()
+    } else -1
+    case None => session.sparkContext.schedulerBackend.defaultParallelism()
+  }
+
```
ymahajan: This takes care of https://jira.snappydata.io/browse/SNAP-955  also.

ymahajan:
Looked at the spark PR also, the changes look good.

ymahajan:
Swati verified this checkin by running northwind hydra on mulithost. 
1. with shuffledPartitions specified 
2. with default shuffledPartitions 
3. with shuffledPartitions as same as number of cores

sumwale:
> @sumwale Will this optimization work for split mode also?

No, haven't changed the split mode case. For that the changes to HashParitioning etc in Spark layer will need to be rethought.

PR #404, opened 2016-10-18
Snap 1087

hbhanawat:
This pull request maintains stats (which include lower bound, upper bound, null count etc)  for every column. And then uses the upper bound and lower bound values of columns to filter out the cached batches. This will be a perf enhancement for the queries which filters extensively. 

To maintain the stats, enabled the stat generation code that was already there and maintained stats as part of the cached batch. 

To filter out the stats, following things were done: 

1. Take the filter predicates and the stats schema from the relation to the scan level so that they can be utilized.  For this a class StatsPredicate was created. 
2. Compile the filter predicates to form a SpecificPredicate class on every executor. The filter predicate creation and compilation was picked up from InMemoryTableScanExec. 
3. During iteration make an UnsafeRow out of the stats bytes and then apply the predicate on this UnsafeRow.
4. Added stats for cached batches seen and skipped 

Currently this optimization does not apply for split mode, for sample tables and for off heap iterators. These will be fixed in next few PRs. 

Precheckin was run. Added a test case that reads the stats and verifies the number of cached batches scanned.


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -63,22 +65,109 @@ import org.apache.spark.{Dependency, Partition, RangeDependency, SparkContext, T
...
+      }
+    }
+
+    metricsCreatedBeforeInit = metricsCreatedBeforeInit ++
+        Map("cachedBatchesSeen" -> SQLMetrics.createMetric(sparkContext,
+      "cached batches seen"),
+    "cachedBatchesSkipped" -> SQLMetrics.createMetric(sparkContext,
+      "cached batches skipped by the predicate"))
+
+    new StatsPredicate(newPredicate,
```
rishitesh: Can you please rename it to StatsPredicateCompiler ? It will suggest we are passing a function rather than compiled predicates.

sumwale:
@hbhanawat Sorry didn't get time to look into this so far. My main concern is this "StatsPredicateCompiler" approach. It is very costly to ship across such objects which includes such closures and all -- I am having trouble due to this with my optimized SnappyHashAggregate where it is trying to ship all sorts of classes like CodegenContext, even maps and other stuff which is failing.

Secondly I don't understand why the need to do this at all. Why not simply generate the code fragment like the rest of the stuff in Codegeneration doProduce? Simply generate a new function or something in the same CodegenContext of ColumnTableScan. It will be much more efficient, will not require unnecessary serialization and the complexities associated. Also I have plan caching changes which too will break with this model completely that requires plan code generation on the fly.

sumwale:
@hbhanawat This also has problems with: a) will not help avoid faulting in overflowed columns in table scans, so there will be little advantage of this optimization when majority of data is on disk, b) will not work in split mode

As we had discussed, the simplest approach would have been to add separate columns for MAX/MIN columns so those could be passed in regular Seq[Filter] in the buildUnsafeScan and filter strings created for query like being done in row table relation. This would have worked just fine for first cut and addressed all of the above. To optimize it, we could then use single UnsafeRow column as now, then go to code generation for embedded mode and a function on GemXD layer to do the filter in split mode.

Anyway for now I will be fixing this to avoid shipping the heavy LogicalPlans, rather generated embedded generated code which is quite straightforward.

sumwale:
@hbhanawat @rishitesh @kneeraj  Have changed to use generated code instead here https://github.com/SnappyDataInc/snappydata/commit/984f808eccf867cec36ca7b737cd8dbbb9fe2c8f
Many of the plan related changes have been reverted since they were unnecessary. Also metrics related changes have been cleaned up to be straightforward like before.

The main change is in ColumnTableScan. I have re-used much of the code there, just used those Expressions to generate code for a function. Split mode still needs to be taken care of via a store layer function as noted in the TODO there. Support for new iterator by Neeraj will need to be integrated when it is available. The iterator can return something like a wrapper RowLocation or ExecRow instead of raw byte[], then stats column can be projected out first and then other required columns.

cluster/src/dunit/scala/org/apache/spark/sql/CachedBatchScanDUnitTest.scala
```diff
@@ -0,0 +1,124 @@
...
+    val seenid = a.head.accumulatorMetrics.filter(x =>
+    {x._2.name == "cached batches seen"}).head._1
+    val skippedid = a.head.accumulatorMetrics.filter(x =>
+    {x._2.name == "cached batches skipped by the predicate"}).head._1
+
+    (metricValues.filter(_._1 == seenid).head._2.toInt,
+        metricValues.filter(_._1 == skippedid).head._2.toInt)
+  }
+
+
```
sumwale: Any reason to have a dunit and not a scalatest? DUnits increase the precheckin runtime by a whole lot and should be avoided unless required. This case looks perfect white-box checks for a scalatest.

For example take a look at Spark test suite. When I had last checked, it had 5X or so more tests than snappy+gemxd+gem combined and still runs in under 2 hours (as compared to >12 hours for all combined for us). IMO we need to be careful about more test coverage and avoid dunits unless required. Likewise we have many tests, especially in AQP, that repeatedly load the same set of airline data taking a whole lot of time but hardly providing proportionate coverage. Had mentioned this to AQP team and they are tracking as an issue.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -63,22 +65,109 @@ import org.apache.spark.{Dependency, Partition, RangeDependency, SparkContext, T
...
+
+      case LessThanOrEqual(a: AttributeReference, l: Literal) => statsFor(a).lowerBound <= l
+      case LessThanOrEqual(l: Literal, a: AttributeReference) => l <= statsFor(a).upperBound
+
+      case GreaterThan(a: AttributeReference, l: Literal) => l < statsFor(a).upperBound
+      case GreaterThan(l: Literal, a: AttributeReference) => statsFor(a).lowerBound < l
+
+      case GreaterThanOrEqual(a: AttributeReference, l: Literal) => l <= statsFor(a).upperBound
+      case GreaterThanOrEqual(l: Literal, a: AttributeReference) => statsFor(a).lowerBound <= l
+
```
sumwale: StartsWith can also be added

cluster/src/dunit/scala/org/apache/spark/sql/CachedBatchScanDUnitTest.scala
```diff
@@ -0,0 +1,124 @@
...
+    assert(skipped2 > 0, "No Cached batches should have been scanned")
+
+    // ***Check for the case when some of the cached batches are scanned ****
+    previousExecutionIds = snc.sharedState.listener.executionIdToData.keySet
+
+    val df_someCachedBatchesScan = snc.sql(
+      "select AVG(ArrDelay) arrivalDelay, UniqueCarrier carrier " +
+          "from AIRLINE where ArrDelay < 20  " +
+          "group by UniqueCarrier order by arrivalDelay")
+
```
sumwale: This test adds checks for a few operators. Do you want to add for other operators too? Perhaps we can take from InMemory test suites which would have better coverage?

hbhanawat:
>> My main concern is this "StatsPredicateCompiler" approach. It is very costly to ship across such objects which includes such closures and all -- I am having trouble due to this with my optimized SnappyHashAggregate where it is trying to ship all sorts of classes like CodegenContext, even maps and other stuff which is failing.

CodegenContext should not have shipped as part of the Expressions. I am not very clear about the problem that you were facing but, from my perspective, I wanted to keep the StatPredicateCompiler at the iterator level so that when Neeraj's changes arrive, we could avoid reading the cached batches at the iterator level. Now since the generated code is handling the stats, it will be tricky. I would look more into your changes but I think we could have generated the code body in the driver side and then passed it as part of StatsPredicateCompiler and then compiled and executed the body in the executor side.

sumwale:
> CodegenContext should not have shipped as part of the Expressions.

The closure for predicate being shipped was from inside the ColumnTableScan so it took the entire physical plan hierarchy with it. For Localjoin class which called consume for code generation, the parent field would get set to ObjectHashAccessor which had CodegenContext etc. I had to explicitly clear the parent field using reflection to make it work.

> . Now since the generated code is handling the stats, it will be tricky.

Why so? It has generated a function which can be directly invoked wherever required.

>  I think we could have generated the code body in the driver side

This always happens in wholestage codegen. Code body is generated on driver side, shipped as part of WholestageCodegenRDD and compiled on executor. Why is this additional "StatsPredicateCompiler" required? As I mentioned, this way from InMemory scan is 1.6 model and not required (rather should be avoided) in 2.0 wholestage codegen model.

PR #411, opened 2016-10-24
a) Changed the position of task "docs" to b ein root project. As invo…

rishitesh:
a) Changed the position of the task "docs" to be in root project. As invoking it from cluster was not configuring other the projects before hand and the check hasPlugin('scala') was failing for all.
b) Removed Kafka projects from the document for the time being. As scaladoc depends on single classpath Kafka-0.8 and Kafka-0.10 was creating problems from each other.

c) Updated publish-site.sh to reflect scala 2.11 version. 

## Patch testing
Checked the doc manually. Ran gradle tasks product to verify for broken build. 

## ReleaseNotes.txt changes
NA
## Other PRs 

NA


dhavalstr:

rishitesh:
Run precheckin

rishitesh:
Run precheckin

rishitesh:
Run precheckin

kneeraj:
@rishitesh when this task was added inside tools ( then it was tools ) then though the task was there it use to go inside all the projects and check for scala plugin to decide whether to add in the source list. Bringing to root is ok but I don't think the place has anything to do with whether all the sub projects were getting picked up or not.

kneeraj:
Just one comment about moving the task. If that is not required we can revert it. In fact base is a better position but at that time the base build.gradle had no scala plugin so I chose tools. It would be good if we understand the reason of why it stopped working.

Change looks ok though.

rishitesh:
Run precheckin

rishitesh:
Run precheckin

rishitesh:
Run precheckin

build.gradle
```diff
@@ -1053,7 +1053,51 @@ distTar.mustRunAfter clean, cleanAll
...
+
+task deleteDocsDir(type: Delete) {
+  delete "${rootProject.buildDir}/docs"
+}
+
+task docs(type: ScalaDoc) {
+  apply plugin: 'scala'
+  dependsOn deleteDocsDir
+  Set<String> allSource = []
+  def docProjects = rootProject.subprojects.collectMany { project ->
```
rishitesh: Its a problem with gradle configuration. Looks like if not needed gradle does not configure  or initializes  a project. Hence hasPlugin('scala') fails and all scala files are skipped.

SachinJanani:
Run precheckin

rishitesh:
Merging as the failures were known errors.

PR #415, opened 2016-10-28
[SNAP-1147] Fix for SNAP-1147 to properly handle dropping of collocat…

SachinJanani:
- Fix for SNAP-1147 to properly handle dropping of collocated table by adding Dependent relation for column table involve in collocation
## Patch testing

Writted unit test and it passed successfully
## ReleaseNotes.txt changes

## Other PRs

NA

…ed table


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -972,16 +972,16 @@ class SnappySession(@transient private val sc: SparkContext,
...
         br match {
           case d: DestroyRelation => d.destroy(ifExists)
+            cacheManager.uncacheQuery(Dataset.ofRows(this, plan))
+            if (sessionCatalog.isTemporaryTable(tableIdent)) {
+              // This is due to temp table
+              // can be made from a backing relation like Parquet or Hadoop
+              sessionCatalog.unregisterTable(tableIdent)
+            } else {
+              sessionCatalog.unregisterDataSourceTable(tableIdent, Some(br))
+            }
```
sumwale: Can you tell why has this code block moved here? The idea of original was to able to drop normal EXTERNAL and TEMPORARY tables with the drop table command too. With this change it will no longer be possible.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -548,6 +548,12 @@ class ColumnFormatRelation(
         throw e
     }
   }
+
+  /** Base table of this relation. */
+  override def baseTable: Option[String] = _origOptions.get("COLOCATE_WITH")
```
sumwale: I think you need the CaseInsensitive map here for lookup. See parameters in createRelation in ColumnFormatRelation.scala.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -548,6 +548,12 @@ class ColumnFormatRelation(
...
+
+  /** Name of this relation in the catalog. */
+  override def name: String = _table
```
sumwale: The underscore was used to indicate a parameter which is not a field but just passed to constructor. If it is being used in body (and thus will become a member field) then better to remove underscore. Parent class has a public "table" val so better to use that.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -470,23 +469,21 @@ class ColumnFormatRelation(
...
     DependencyCatalog.getDependents(table)
 
-  override def recoverDependentsRelation(): Unit = {
-    var indexes: Array[String] = Array()
-    try {
-      val options = new CaseInsensitiveMutableHashMap(this.origOptions)
-      indexes = options(ExternalStoreUtils.INDEX_NAME).split(",")
-    } catch {
-      case e: NoSuchElementException =>
+  override def recoverDependentsRelation(properties: Map[String, String]): Unit = {
```
kneeraj: recoverDependentRelations is proper.

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatRelation.scala
```diff
@@ -214,6 +216,48 @@ class RowFormatRelation(
...
+      val dr = sncCatalog.lookupRelation(sncCatalog.newQualifiedTableName(rel)) match {
+        case LogicalRelation(r: DependentRelation, _, _) => r
+      }
+      addDependent(dr, sncCatalog)
+    })
+
+  }
 }
 
 final class DefaultSource extends MutableRelationProvider {
```
kneeraj: You would need to add this line:

val dependentRelations = parameters.remove(ExternalStoreUtils.DEPENDENT_RELATIONS)
( siimilar to ColumnFormatRelation. Please check )

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -166,6 +166,26 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
                   (JdbcExtendedUtils.ALLOW_EXISTING_PROPERTY -> "true")).resolveRelation()
         }
 
+
+
```
kneeraj: Unnecessary blank lines.

core/src/main/scala/org/apache/spark/sql/sources/RuleUtils.scala
```diff
@@ -38,16 +39,19 @@ import org.apache.spark.sql.sources.Entity.{INDEX, INDEX_RELATION, TABLE}
...
+    }
+  }
+
   def fetchIndexes(snappySession: SnappySession,
       table: LogicalPlan): Seq[(LogicalPlan, Seq[LogicalPlan])] = table.collect {
     case l@LogicalRelation(p: ParentRelation, _, _) =>
       val catalog = snappySession.sessionCatalog
-      (l.asInstanceOf[LogicalPlan], p.getDependents(catalog).map(idx =>
-        catalog.lookupRelation(catalog.newQualifiedTableName(idx))))
-    case s@SubqueryAlias(alias, child@LogicalRelation(p: ParentRelation, _, _)) =>
```
kneeraj: This was anyways useless as SubqueryAlias extends Logical plan?

kneeraj:
Overall looks good. Some minor comments. Take care and then merge.

core/src/main/scala/org/apache/spark/sql/sources/RuleUtils.scala
soubhik-c: No. It wasn't earlier when rule was getting applied at the analyzer level. Off late this rule is shifted to optimizer level and EliminateSubqueryAlias rule gets applied first and therefore now its not required to handle. Nothing to do with inheritence.

SachinJanani:
Done with the suggested changes in review and ran precheckin all test passed except scalastyle for spark which I will fix in other PR

PR #416, opened 2016-11-04
SNAP-982: Support persistent UDFs in SnappyContext

rishitesh:
The PR pertains to support for persistent function definition in SnappyStore. Rather than session specific UDFs , users can create functions which can be used across sessions, also after system restart.

Also, this PR handles a case where a UDF definition is changed, the new UDF definition is loaded by the classloader. For this, we need to reset the executor side classloader which is used for loading classes.

**This PR includes**

- DDL support to create/drop functions. (UDFS and UDAFS)
- Hive catalog integration for functions
- We will change the classloader from executor side whenever a function is dropped. This will ensure that correct version of the class is loaded. 

**Limitations**

-  Complex type like Map, Array as UDF input/output types is not tested yet.
- Float type UDFs not working in Row table. It works fine for column tables. (#Issue). We need to document it. Row table returns a Double type even if Float is mentioned in DDL. 
- Helper commands like listFunctions are not supported. 

## Patch testing
pre-checkin

## ReleaseNotes.txt changes

SNAP;982: SnappyStore now supports persistent UDF( User defined functions)  and UDAF(User Defined Aggregate Function). One can define a function and can use it across sessions. Also as the definition of the function is stored in a persistent catalog , it can be used after node restart as well.

## Other PRs 

- https://github.com/SnappyDataInc/snappy-store/pull/123
- https://github.com/SnappyDataInc/spark/pull/43
- https://github.com/SnappyDataInc/spark-jobserver/pull/7


core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
```diff
@@ -396,8 +404,6 @@ object SnappyParserConsts {
   final val EXTENDED = nonReservedKeyword("extended")
   final val EXTERNAL = nonReservedKeyword("external")
   final val FULL = nonReservedKeyword("full")
-  final val FUNCTION = nonReservedKeyword("function")
-  final val FUNCTIONS = nonReservedKeyword("functions")
```
rishitesh: This is a problem with identifier.
There is a syntax SHOW ALL FUNCTION where "ALL" is optional. So it can be SHOW FUNCTIONS <funcName> or SHOW ALL FUNCTIONS. In first case it fails and takes Key work FUNCTIONS as an identifier . Once made reserved keyword it works as identifiers ignores such literals.

core/src/main/scala/org/apache/spark/sql/internal/JarUtils.scala
```diff
@@ -0,0 +1,119 @@
...
+   */
+  def addToSparkJars(prefix: String, path: String) {
+
+    val callbacks = ToolsCallbackInit.toolsCallback
+    if (callbacks != null) {
+      val localName = path.split("/").last
+      val changedFileName = s"${prefix}-${localName}"
+
+      val changedFile = new File(workingDir, changedFileName)
+      if (!changedFile.exists()) {//After creation removeFromSparkJars() is the only place which can remove this jar
```
hbhanawat: As discussed, this would not work in cluster restart and lead HA scenarios. Please add a test for lead HA atleast.

core/src/main/scala/org/apache/spark/sql/internal/JarUtils.scala
```diff
@@ -0,0 +1,119 @@
...
+        jarFile.delete()
+      }
+
+      val keyToRemove = sparkContext.listJars().filter(getName(_) == getName(jarFile.getPath))
+      if (keyToRemove.nonEmpty) {
+        println(s"Removing ${path} from Spark Jars by Driver loader")
+        callbacks.removeAddedJar(sparkContext, keyToRemove.head)
+      }
+      //Remove the jar from all live executors.
+      org.apache.spark.sql.collection.Utils.mapExecutors(
```
rishitesh: If one user only wants to change the logic of UDF he will drop and create the function again. If no job other job has been executed between dropFunction and createFunction the new added function will not take effect.

core/src/main/scala/org/apache/spark/sql/internal/JarUtils.scala
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * An utility class to store jar file reference with their individual classloaders. This is reflect class changes at driver side.
+ * e.g. If an UDF definition changes the driver should pick up the correct UDF class.
+ * This class can not initialize itself after a driver failure. So the callers will have  to make sure that the classloader gets
+ * initialized after a driver startup. Usually it can be achieved by adding classloader at query time.
+ *
+ */
+class ContextJarUtils(sparkContext: SparkContext) extends Logging{
+
+  val workingDir = new File(System.getProperty("user.dir"))
```
rishitesh: I have modified the code to add a separate folder snappy-jars inside user directory.

rishitesh: This is a problem. Will discuss further how to avoid this. Probably we will need some kind of  region callback here.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -722,6 +732,150 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
     }
   }
 
+  def toUrl(resource: FunctionResource): URL = {
```
rishitesh: done

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -401,10 +469,10 @@ abstract class SnappyDDLParser(session: SnappySession)
   // SHOW FUNCTIONS func1;
   // SHOW FUNCTIONS `mydb.a`.`func1.aa`;
   protected def show: Rule1[LogicalPlan] = rule {
-    SHOW ~ TABLES ~ ((FROM | IN) ~ identifier).? ~> ((ident: Any) =>
+   SHOW ~ TABLES ~ ((FROM | IN) ~ identifier).? ~> ((ident: Any) =>
```
rishitesh: Will add it

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -338,6 +341,71 @@ abstract class SnappyDDLParser(session: SnappySession)
     }
   }
 
+  protected final def resourceType: Rule1[FunctionResource] = rule {
+    identifier ~ stringLiteral ~> { (rType: String, path: String) =>
+      val resourceType = rType.toLowerCase
+      resourceType match {
+        case "jar" | "file" | "archive" =>
```
rishitesh: Removed the other types

cluster/src/dunit/scala/org/apache/spark/DynamicJarInstallationDUnitTest.scala
```diff
@@ -201,7 +228,7 @@ object DynamicJarInstallationDUnitTest {
     val loader = Thread.currentThread().getContextClassLoader
     assert(loader != null)
     try {
-      val fakeClass = loader.loadClass(className).newInstance()
+      val fakeClass = Class.forName(className, false, loader).newInstance()
       assert(fakeClass != null)
       assert(fakeClass.toString.equals(version))
```
rishitesh: Will add the test case

cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -56,32 +57,52 @@ class SnappyExecutor(
...
+      val diffJars = addedJarFiles.diff(newJarFiles)
+      if (diffJars.size > 0) {
+        diffJars.foreach(classloader.removeURL)
+        logInfo("As some of the Jars have been deleted, setting up a new ClassLoader for subsequent Threads")
+        diffJars.foreach(d => println(d))
+        logInfo("job jars after delete in previous loader")
+        classloader.jobJars.keySet.foreach(println)
+        this.urlClassLoader = new SnappyMutableURLClassLoader(classloader.getURLs(),
+          classloader.getParent, classloader.jobJars)
+        this.replClassLoader = urlClassLoader
```
rishitesh: Agreed. Will change that.

PR #417, opened 2016-11-11
Snap 1154

sonalsagarwal:
- Removed version from the quickstart jar name - Currently we include the snappydata version in the qucikstart jar name, so after every release, it will become mandatory to modify the examples to point correct quickstart jar. By removing the version from jar name, the above overhead will be eliminated 
- Changed quickstart jar path and name of jar in the examples - The current path for the quickstart jar mentioned in the examples in incorrect, hence need to modify the path point to correct location.
- Added run commands to the examples, where it was missing - Adding run commands as guidelines to the users.
- Fixed few issues observed while running the examples.

## Patch testing
Ran examples individually.
Ran runQuickstart task.

## ReleaseNotes.txt changes

## Other PRs 
SNAP-1150 - Documentation changes required to be done as per the modified quickstart jar name.


rishitesh:
Looks ok.

PR #419, opened 2016-11-16
Snappy job server app APIs are still taking SnappyContext as in input…

rishitesh:
Snappy job server app APIs are still taking SnappyContext as in input parameter. Changed it to SnappySession.

Also added a constructor for SnappySession which takes SparkSession as an argument. 
Have not deprecated the earlier constructor which takes a SparkContext as an argument.
Plan is to make it private so all users can use is '

val snappy = new SnappySession(spark)

## Patch testing
Ran precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

na


cluster/src/main/scala/org/apache/spark/sql/SnappyContextFactory.scala
```diff
@@ -36,11 +36,12 @@ class SnappyContextFactory extends SparkContextFactory {
 
 object SnappyContextFactory {
 
-  private[this] val snappyContextLike =
-    SnappyContext(LeadImpl.getInitializingSparkContext)
+  private[this] val snappySession =
+    new SnappySession(LeadImpl.getInitializingSparkContext)
+
+  protected def newSession(): SnappySession with ContextLike =
+    new SnappySession(snappySession.sparkContext) with ContextLike {
```
hbhanawat: The spark state is not being shared. We need to share the state across multiple SnappySessions. I would recommend fixing the builder API if there is no quick fix for this issue.

rishitesh: This problem has been taken care by Spark itself and we don't see multiple tabs now.

cluster/src/main/scala/org/apache/spark/sql/SnappyContextFactory.scala
```diff
@@ -78,9 +78,9 @@ trait SnappySQLJob extends SparkJobBase {
     }
   }
 
-  def isValidJob(sc: SnappyContext, config: Config): SnappyJobValidation
+  def isValidJob(sc: SnappySession, config: Config): SnappyJobValidation
 
-  def runSnappyJob(sc: SnappyContext, jobConfig: Config): Any
+  def runSnappyJob(sc: SnappySession, jobConfig: Config): Any
 
 }
```
sumwale: Should this class be called SnappyContextFactory anymore or SnappySessionFactory?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -62,6 +62,10 @@ class SnappySession(@transient private val sc: SparkContext,
     this(sc, None)
   }
 
+  def this(spark: SparkSession) {
+    this(spark.sparkContext, None)
+  }
+
```
sumwale: I don't think this constructor makes sense. It gives the impression to users that this will somehow be tied to the passed SparkSession while this creates a new one. Let users pass SparkContext explicitly so the second constructor which is private[sql] should now be public.

cluster/src/main/scala/org/apache/spark/sql/SnappyContextFactory.scala
rishitesh: Yup. Will change it.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
rishitesh: hmmm. Agreed.

PR #421, opened 2016-11-18
Full resultSet validation in northWind tests

swatisawant:

- Added support for full resultset validation for northWind queries
- Modified the cross join queries to include actual column names instead of * in order to avoid the duplicate column name issue while writing the query results(df) to csv file
- Modified the fullResultSet validation code to avaoid the OOME in join queries due to df.collect
- Modified the suppliers.csv in order to make it work correctly for Q27 and 29 on spark side
- Added dataTypeConverter function for Double and bigDecimal
- Parameterized the fullResultSet validation in order to have some tests with full ResultSet Validation and some will just do numRows validation. This is required to reduce the regression time
- Fixed issue with order of resultset in case of partitioned row tables
- Added support for collecting the query results in a seperate folder in test directory
- Added support in framework to load snappyTests jar and storeTests jar in driver and executors classpath in order to use the hydra classes in snappy job and spark app
- Added support to create the golden files in first test and use the same in all rest of the tests. This will reduce the time required for creating the files in each test
- Enabled full resultSet validation in all northWind tests
- Fixed a issue with framework where in the jps command was also returning PID of MasterController, due to which the test was getting stopped with Hang Status after executing cleanUp task.
This issue was introduced recently.
- Increased the northWind test time in order to make sure that HA task is done atleast once in a test

## Patch testing
Verified the changes by running northWind.bt

## ReleaseNotes.txt changes
NA

## Other PRs 
No


dtests/src/test/java/io/snappydata/hydra/northwind/northWind.bt
```diff
@@ -5,6 +5,7 @@ io/snappydata/hydra/northwind/nwReplicatedRowTablesTest.conf
   D=worker workerHosts=3 workerVMsPerHost=1 workerThreadsPerVM=1
   dataFilesLocation="$GEMFIRE/../../../tests/common/src/main/resources/northwind/"
   tableType="ReplicatedRow"
+  fullResultSetValidation = true
```
ymahajan: why this configurable ? We should always validate full resultset, so it should be always true.

dtests/src/test/scala/io/snappydata/hydra/northwind/NWQueries.scala
```diff
@@ -286,21 +285,21 @@ object NWQueries {
...
     " ON Orders.OrderID = Order_Details.OrderID"
-  val Q52: String = "SELECT * FROM orders RIGHT JOIN order_details" +
+  val Q52: String = "SELECT orders.OrderID as OOID, CustomerID,EmployeeID,OrderDate,RequiredDate,ShippedDate,ShipVia,Freight,ShipName,ShipAddress,ShipCity,ShipRegion,ShipPostalCode,ShipCountry FROM orders RIGHT JOIN order_details" +
     " ON Orders.OrderID = Order_Details.OrderID"
-  val Q53: String = "SELECT * FROM orders FULL OUTER JOIN order_details" +
+  val Q53: String = "SELECT orders.OrderID as OOID, CustomerID,EmployeeID,OrderDate,RequiredDate,ShippedDate,ShipVia,Freight,ShipName,ShipAddress,ShipCity,ShipRegion,ShipPostalCode,ShipCountry FROM orders FULL OUTER JOIN order_details" +
     " ON Orders.OrderID = Order_Details.OrderID"
-  val Q54: String = "SELECT * FROM orders FULL JOIN order_details" +
+  val Q54: String = "SELECT orders.OrderID as OOID, CustomerID,EmployeeID,OrderDate,RequiredDate,ShippedDate,ShipVia,Freight,ShipName,ShipAddress,ShipCity,ShipRegion,ShipPostalCode,ShipCountry FROM orders FULL JOIN order_details" +
```
ymahajan: @swatisawant can you add these query changes from (Q46 to Q54) in NorthWindTest and NorthWindDUnit. Lets keep the unit tests and hydra tests are in sync.  just change it and include it in the same PR

ymahajan:
@swatisawant Since this northwind framework is discovering many bugs now, can we add more queries(search on google) and create distributed indexes in the tests.

ymahajan:
@swatisawant Can you please add the full result set validation in unit tests also. Let's talk on Monday and you can also take sachin's help on this.

sonalsagarwal:
Looks good to me.

dtests/src/test/java/io/snappydata/hydra/northwind/northWind.bt
swatisawant: We are planning to generate the large data for northWind tests and  while running tests with large data, the full resultSet validation shouldn't be done. That's the reason I have added this parameter here.
In this way we can use the same tests by just tweaking two parameters i.e. fullResultSetValidation and dataFilesLocation.

dtests/src/test/scala/io/snappydata/hydra/northwind/NWQueries.scala
swatisawant: Done.

PR #424, opened 2016-11-24
[SNAP-1180] Use ConfigEntry mechanism for SnappyData properties

sumwale:

- added SQLConfigEntry and convenience methods to: a) allow using from non-spark namespace
  since those are spark/sql private, b) avoid Builder pattern and just use SQLConfigEntry.apply
- changed io.snappydata.Property to use SQLConfigEntry and added documentation strings for all
- renamed "complexTypeAsClob" hint to "complexTypeAsJson"

## Patch testing

Will be tested as overall perf branch testing.

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


cluster/src/test/resources/log4j.properties
```diff
@@ -74,3 +74,5 @@ log4j.logger.org.apache.spark.scheduler.FairSchedulableBuilder=WARN
 log4j.logger.org.apache.spark.scheduler.cluster.CoarseGrainedSchedulerBackend$DriverEndpoint=WARN
 log4j.logger.org.apache.spark.storage.BlockManagerInfo=WARN
 log4j.logger.org.apache.spark.executor.SnappyExecutor=WARN
+
+# log4j.logger.org.apache.spark.sql.execution.WholeStageCodegenExec=DEBUG
```
kneeraj: Deliberate?

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitTest.scala
```diff
@@ -549,10 +549,10 @@ object SplitClusterDUnitTest extends SplitClusterDUnitTestObject {
 
     // also check access to complex types as string
     rs = stmt.executeQuery(
-      s"SELECT * FROM $tableName --+ complexTypeAsClob(1)")
+      s"SELECT * FROM $tableName --+ complexTypeAsJson(1)")
```
kneeraj: The hint is gone for clob which is a super type for JSON. So all complex type can be represented as JSON?

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -144,35 +151,65 @@ object Property extends Enumeration {
       if (altName == null) name else name + '/' + altName
   }
 
-  type Type = ValueAlt
+  case class SQLValue(name: String, configEntry: SQLConfigEntry)
```
kneeraj: Why were 2 types required? Val and SQLValue?

soubhik-c: SQLValue is used for SQLContext like columnBatchSize.

cluster/src/test/resources/log4j.properties
soubhik-c: yes

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -144,35 +151,65 @@ object Property extends Enumeration {
...
-    Constant.SPARK_PREFIX)
+    override def toString(): String = name
+  }
 
-  val McastPort = Val(s"${Constant.STORE_PROPERTY_PREFIX}mcast-port",
-    Constant.SPARK_PREFIX)
+  type Type = ValueAlt
 
-  val JobserverEnabled = Val(s"${Constant.JOBSERVER_PROPERTY_PREFIX}enabled")
+  type SQLType = SQLValue
```
soubhik-c: Please handle this case in LeadImpl snappyProperties variable. I am posting a modified code here because it won't let me comment on unmodified section of LeadImpl.

```scala
  private def extractPropertyKey(prop: String) = if (prop.startsWith(Constant
      .PROPERTY_PREFIX) && !prop.startsWith(Constant.STORE_PROPERTY_PREFIX)) {
    Some(prop.substring(Constant.PROPERTY_PREFIX.length))
  } else if (prop.startsWith(Constant.SPARK_SNAPPY_PREFIX) &&
      !prop.startsWith(Constant.SPARK_STORE_PREFIX)) {
    Some(prop.substring(Constant.SPARK_SNAPPY_PREFIX.length))
  } else {
    None
  }

  private[snappydata] val snappyProperties = Utils.getFields(Property).flatMap {
    case (_, propVal: Property.Type) =>
      extractPropertyKey(propVal.name)
    case (_, propVal: Property.SQLType) =>
      extractPropertyKey(propVal.name)
  }.toSet
```

cluster/src/test/resources/log4j.properties
sumwale: Yes, because this is required frequently while debugging so added a commented one for ease.

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitTest.scala
sumwale: Earlier code was converting to a custom string format in https://github.com/SnappyDataInc/snappydata/commit/20e8ef6ddcdb89742d4bff675b2dda1f72ca3990#diff-cc0d92e180612747c169d3aaceaacfa0L464 . This was not a good idea since user will not be able to parse the result easily. So that has been taken out and instead now using jackson for JSON output as is done by Spark json format handling.

sumwale:
> However I am not sure whether we would want jobserver default property to be anything other than 'true' and IMO should be left with hard coded instead of def val which somebody inadvertently might land up into false.

This is already in the codebase so can be taken out you think so. However, I have marked those properties with "isPublic=false" so won't be visible (and we need not document) and it just remains an internal property. This is the standard way in Spark to mark properties as internal or public, and former should be touched only if the user is absolutely sure about it.

core/src/main/scala/io/snappydata/Literals.scala
sumwale: Ok, missed this.

core/src/main/scala/io/snappydata/Literals.scala
sumwale: Unapply is used in ServiceUtils. In fact it will be better to use extractors instead of multiple prop.startsWith checks and I will be adding better extractors.

sumwale:
Changed to use extractors in LeadImpl property matching for both Spark and SQL properties.

Removed ValueAlt.apply() and instead changed to use name.

PR #426, opened 2016-11-25
[SNAP-1136] Pooled version of Kryo serializer which works for closures

sumwale:

- new PooledKryoSerializer that does pooling of Kryo objects (else performance is bad if
    new instance is created for every call which needs to register and walk tons of classes)
- has an overridden version for ASCII strings to fix (https://github.com/EsotericSoftware/kryo/issues/128);
  currently makes a copy but will be modified to use one extra byte to indicate end of string
- optimized external serializers for StructType, and Externalizable having readResolve() method;
  using latter for StorageLevel and BlockManagerId
- added optimized serialization for the closure used by SparkSQLExecuteImpl (now a proper class instead)
- fixed index column determination in RowFormatRelation (was off by 1 due to 0 based vs 1 based)

## Patch testing

precheckin with spark precheckin in progress

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/27


sumwale:
run precheckin

sumwale:
run precheckin

soubhik-c:
for my understanding, could you please tell how kryo serialization in core is compatible with stock spark where none of the kryo serialization in our fork is there yet.

sumwale:
In upstream Spark core kryo serialization has been disabled for closures (SPARK-12414, https://github.com/apache/spark/pull/11150) and is only for data. Our changes to snappy-spark make it usable for closures, as well as for Spark Netty messages which was not configurable earlier. Secondly the PooledKryoSerializer has the fix for string overwrite issue (https://github.com/EsotericSoftware/kryo/issues/128) and has been updated to 4.0.0 to get the fix for https://github.com/EsotericSoftware/kryo/pull/342 so it corrects those issues that can get hit. However, with only data serialization these issues are also not hit so this works with upstream Spark.

In addition this is not enabled by default when working with upstream spark. It gets set only in LeadImpl and ExecutorInitiator by default. Users can enable it for split mode too using the "spark.serializer" property but that has not been tested (though I believe it should work fine as per the comments above).

PR #429, opened 2016-11-28
Changes to render StringType as VARCHAR for tables created via API.

ashetkar:
* Changes to render StringType as VARCHAR for tables created via API.
* Tables created via DDL are already taken care of.

## Patch testing
* Added testcases in the revamped dunit test.

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
```diff
@@ -365,6 +365,49 @@ object Utils {
...
+  /**
+   * Utility function to return a metadata for a StructField of StringType, to ensure that the
+   * field is rendered as CLOB by SnappyStore.
+   * @param md
+   * @return
+   */
+  def stringMetadata(md: Metadata = Metadata.empty): Metadata = {
+    new MetadataBuilder().withMetadata(md).putString(Constant.CHAR_TYPE_BASE_PROP, "STRING").build()
+  }
+
```
dshirish: Utility methods are needed for CHAR and VARCHAR metadata to allow user to specify these datatypes thru Structype schema. By default, String will return as varchar and can be overridden using the already existing query hint or system var if user wants data in CLOB format. So stringMetadata method may not be needed as user will use query hint when he/she needs data in clob format.

dshirish: Can we call it Utils.clobMetadata if the intention is to allow user to create clob column(similar to charMetadata/varcharMetadata) .

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -87,6 +87,8 @@ object Constant {
 
   val MAX_VARCHAR_SIZE = 32672
 
+  val MAX_CHAR_SIZE = 254
```
sumwale: Is this a limit in Derby for CHAR?

core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
```diff
@@ -368,6 +368,52 @@ object Utils {
...
+  /**
+   * Utility function to return a metadata for a StructField of StringType, to ensure that the
+   * field is stored (and rendered) as VARCHAR by SnappyStore.
+   * @param md
+   * @param size
+   * @return
+   */
+  def varcharMetadata(md: Metadata = Metadata.empty, size: Int = Constant.MAX_VARCHAR_SIZE):
+  Metadata = {
+    if (size < 1 && size > Constant.MAX_VARCHAR_SIZE) {
```
sumwale: Should be "size < 1 || size > ..."

core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
```diff
@@ -368,6 +368,52 @@ object Utils {
...
+      throw new IllegalArgumentException(s"VARCHAR size should be between 0 " +
```
sumwale: In the message "VARCHAR size should be between 1 ..."

core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
```diff
@@ -368,6 +368,52 @@ object Utils {
...
   }
 
+  /**
+   * Utility function to return a metadata for a StructField of StringType, to ensure that the
+   * field is stored (and rendered) as VARCHAR by SnappyStore.
+   * @param md
+   * @param size
+   * @return
+   */
+  def varcharMetadata(md: Metadata = Metadata.empty, size: Int = Constant.MAX_VARCHAR_SIZE):
```
sumwale: Can you reverse the order of arguments here? Users will normally not want to provide any explicit Metadata argument but may certainly want to specify some other max length, so let the size argument be the first one (and if users leave out second one then it will use Metadata.empty as above).

sumwale: I think we should make these methods Java friendly too. So I will suggest overloads instead of default arguments:

```scala
def varcharMetadata(): Metadata = varcharMetadata(Constant.MAX_VARCHAR_SIZE)
def varcharMetadata(size: Int): Metadata = varcharMetadata(size, Metadata.empty)
def varcharMetadata(size: Int, md: Metadata): Metadata = ...
```

core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
```diff
@@ -368,6 +368,52 @@ object Utils {
...
+  }
+
+  /**
+   * Utility function to return a metadata for a StructField of StringType, to ensure that the
+   * field is stored (and rendered) as CHAR by SnappyStore.
+   * @param md
+   * @param size
+   * @return
+   */
+  def charMetadata(md: Metadata = Metadata.empty, size: Int = Constant.MAX_CHAR_SIZE): Metadata = {
```
sumwale: See above comments about varcharMetadata

core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
```diff
@@ -368,6 +368,52 @@ object Utils {
...
+    if (size < 1 && size > Constant.MAX_CHAR_SIZE) {
+      throw new IllegalArgumentException(s"CHAR size should be between 0 " +
+          s"and ${Constant.MAX_CHAR_SIZE}")
```
sumwale: should be "||" as above. Also exception message as for varcharMetadata.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -297,7 +297,18 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
             val builder = new MetadataBuilder
             builder.withMetadata(f.metadata).putString("name", name).build()
           } else {
-            f.metadata
+            f.dataType match {
+              case s: StringType =>
+                if (!f.metadata.contains(Constant.CHAR_TYPE_BASE_PROP)) {
+                  val builder = new MetadataBuilder
+                  builder.withMetadata(f.metadata).putString(Constant.CHAR_TYPE_BASE_PROP,
+                    "STRING").putLong(Constant.CHAR_TYPE_SIZE_PROP, Constant.MAX_VARCHAR_SIZE)
```
sumwale: What does the combination of "STRING" and MAX_VARCHAR_SIZE mean? Shouldn't there be no size specification if "STRING" is the type?

core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
sumwale: CLOB is not Spark terminology, so I think we should avoid using that in Spark APIs.

core/src/main/scala/io/snappydata/Literals.scala
ashetkar: Yes, taken from Limits.DB2_CHAR_MAXWIDTH

core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
ashetkar: So the same should apply for VARCHAR and CHAR (with varcharMetadata() and charMetadata())?

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
ashetkar: The check in SparkSQLExecuteImpl.getSQLType() depends on both these properties. For STRING it is ignored.

core/src/main/scala/org/apache/spark/sql/collection/Utils.scala
sumwale: Those are new types we have introduced in Spark SQL, so perhaps we can have those in API too?

ashetkar:
Run precheckin

ashetkar:
Run precheckin

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
ashetkar: FYI, This and other comments have been incorporated.

PR #430, opened 2016-11-28
Fix

soubhik-c:

In jobserver-defaults.conf, rootdir for spark/jobserver/filedao and spark/jobserver/datadao point to /tmp/spark-jobserver.
For multi users, we cannot overwrite the existing directory created by other user. So, we change the rootdir for above and use workingDir mentioned by "-dir" during startup. This addresses concerns like loosing job configuration across restarts etc.

## Patch testing

Added unit test verifying the same.

## ReleaseNotes.txt changes
If upgraded, jobserver configuration requires to be redone. As we are yet to release GA, this is not important to mention in relelase notes.

## Other PRs 

- redirecting rootdir from /tmp to "-dir" startup parameter via
gemfirexd.system.home variable set in the launcher

- non-overridable js props like sqldao.rootdir is now dynamically
computed using the above.

- added tests to verify /tmp is no more there in the js startup
settings.

### DIFF
--- a/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
@@ -30,7 +30,7 @@ import com.gemstone.gemfire.distributed.internal.locks.{DLockService, Distribute
 import com.gemstone.gemfire.internal.cache.GemFireCacheImpl
 import com.pivotal.gemfirexd.FabricService.State
 import com.pivotal.gemfirexd.internal.engine.distributed.utils.GemFireXDUtils
-import com.pivotal.gemfirexd.internal.engine.store.ServerGroupUtils
+import com.pivotal.gemfirexd.internal.engine.store.{GemFireStore, ServerGroupUtils}
 import com.pivotal.gemfirexd.{FabricService, NetworkInterface}
 import com.typesafe.config.{Config, ConfigFactory}
 import io.snappydata.util.ServiceUtils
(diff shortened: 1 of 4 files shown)


cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -384,6 +385,29 @@ class LeadImpl extends ServerImpl with Lead with Logging {
...
+      assert (value.indexOf("/tmp") >= 0)
+      dynamicOverrides.setProperty(key, value.replace("/tmp", newValue))
+    }
+
+    val workingDir = System.getProperty(
+      com.pivotal.gemfirexd.internal.iapi.reference.Property.SYSTEM_HOME_PROPERTY, ".")
+    val defaultConf = ConfigFactory.parseResources("jobserver-defaults.conf")
+
+    var key = "spark.jobserver.filedao.rootdir"
+    replace(key, defaultConf.getString(key), workingDir)
```
sumwale: Instead of hackish replace of "/tmp" can't we have a token say "${WORKING_HOME}" in jobserver-defaults.conf that gets substituted by the value of environment variable which we set appropriately?

sumwale:
Found this note about environment variable replacement: https://github.com/typesafehub/config/blob/master/HOCON.md#substitution-fallback-to-environment-variables

Or perhaps we can change "/tmp" to our custom token and change it in code like above so we don't depend on "/tmp" rather our specific unique token.

soubhik-c:
introduced <dir> custom tag in the conf files provided by us. Notice jobserver application.conf supplied with it still have hardcoded /tmp but that shouldn't come into play at all because we override the rootdir while configuring js. End user can configure rootdir but won't take effect.

PR #438, opened 2016-12-02
[SNAP-1191] Plan caching changes

sumwale:

Add plan caching and reuse of SparkPlan, RDD and PlanInfo. Currently it is exact string match and tokenization support for constants embedded in query string will be added later.

- Added a new CachedDataFrame which is created by CacheLoader on a cache miss. It provides a generic "collectWithHandler" method that allows collecting results using a custom handler (which can, for example, put into local BlockManager like done by SparkSQLExecuteImpl)
- Compress partition results like that done in SparkPlan.getByteArrayRdd in static CachedDataFrame.apply() and corresponding decodeUnsafeRows method to decode
- Pass through the above compressed byte result end-to-end from executor to driver to DRDA server. The server invokes decodeUnsafeRows avoiding any additional deserialization or other cost for the additional transition.
- For the case of "take" i.e. limit, the executeTake uses similar approach as by SparkPlan.executeTake. However, instead of decoding every row it will try to get in full compressed blocks as much as possible and decode only the remaining rows in last block (CachedDataFrame.takeRows). Thus collectWithHandler needs two handlers -- one for the partition-wise result, and other for decoding if required.
- The localBlockStoreResultHandler and localBlockStoreDecoder are the handlers used by SparkSQLExecuteImpl. So only for limit queries will there be any decoding of the residual rows on lead.
- Added support for sending number of processors as part of GemXD profile (number of executors is different and by default 2 times the processor count) and use that to set the shuffle partitions in SnappyConf. This reduces the default shuffle size to be only as much as number of processors.
- Send bucket set as bit set so that lookup is faster (an array lookup instead of a hash set). Seen in Citi profiling.
- JSON conversion for complex types for JDBC clients

JSON conversion has been reviewed earlier here 20e8ef6 but was on different branch and is present here since it is related to the changes in SparkSQLExecuteImpl so it can be skipped

Changes to SnappyStrategies are coming from SNAP-1194 but got mixed up a bit here due to the complex way I had to split up these changes into separate patches.

## Patch testing

precheckin

## ReleaseNotes.txt changes

Plan caching can be mentioned in release notes.

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/128


ymahajan:
@sumwale
1) Will this snappySession.planCache take care of queryHints as well? I mean query is same but the hint is different so they will be caches separately?
2) Also in case of streaming, CQs are executed repeatedly and they can utilize snappySession.planCache as it is?

sumwale:
> take care of queryHints as well

It caches against exact query string including any query hints in comments.

> CQs are executed repeatedly and they can utilize snappySession.planCache as it is

There was some trouble with having CQs use it so currently SnappyStreamingContext.sql() calls session.sqlUncached(). Will need to do some explicit cleanup in the cached plan to reuse it correctly.

sumwale:
Merged to master in dfc016d

PR #439, opened 2016-12-02
[SNAP-1193] Optimized Collect aggregate plan to avoid last step exchange

sumwale:

- New CollectAggregateExec plan that will take a last stage SnappyHashAggregateExec and execute it on driver itself. Right now only for non-groupBy/distinct cases due to concerns with potentially large evaluation but this must be revisited once Spark adds plan statistics.
- This lets code-generation happen as in SnappyHashAggregateExec and just switches the producer of rows to be the InputAdapter around the result of runJob.
- Disabled bucket-RDD delinking now that nearly all the changes required for SNAP-1135 are in and partition overhead is minimal. It helps with avoiding an exchange in case of non-collocated join when one of the tables is already partitioned on that column.

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -628,9 +628,9 @@ task product(type: Zip) {
 
     if (rootProject.hasProperty('hadoop-provided')) {
       releaseFile.append("SnappyData ${version}${gitRevision} " +
-              "built with Hadoop ${hadoopVersion}  but hadoop not bundled \n")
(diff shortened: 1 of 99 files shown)


sumwale:
closing this and will re-create against master

PR #442, opened 2016-12-04
incrementing the size of concurrent segment hashmap in the bulk chang…

ahshahid:
This is to fix bug in ConcurrentSegmentHashMap where the size of the map always returned 0, causing the new cache size of reservoir to increase unboundedly. 

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

…e value method


core/src/main/scala/org/apache/spark/sql/collection/ConcurrentSegmentedHashMap.scala
```diff
@@ -224,7 +225,10 @@ private[sql] class ConcurrentSegmentedHashMap[K, V, M <: SegmentMap[K, V] : Clas
             while (idx < nhashes) {
               added = seg.changeValue(keys(idx), bucketId(hashes(idx)), change, isLocal)
               if (added != null) {
-                if (added.booleanValue()) numAdded += 1
+                if (added.booleanValue()) {
+                  numAdded += 1
+                  this._size.incrementAndGet()
+                }
```
sumwale: This is taken care of by adding numValues at the end and numValue variable was precisely for that reason. This was kept so to avoid having to do a slightly expensive increment every time. An additional increment above will set the wrong size. This is for the non-region case only, right? For region case the size will be obtained from the bucket regions.

```scala
275:     }
276:     if (numAdded > 0) _size.addAndGet(numAdded)
```

The case that is probably not being handled is if there is an exception in between. Is that what you were looking at? For the exception case, it can be taken care of in the existing finally block:

```scala
            success = true
          } finally {
            if (lockedState) {
              lock.unlock()
            }
            if (!success) _size.addAndGet(numAdded)
          }
```

ahshahid:
I missed that code snippet. So looks like there is still a bug which causes
miss.  will debug further tomorrow. No I was not looking for exception
case. no it is for both the cases region as well as non region.

ahshahid:
ok. the bug is that boolean add is a var which relies only on the last
state & hence things do not get added.
will fix it

ahshahid:
sorry . jumped the gun. looked at wrong func.

ahshahid:
Looks like the issue is that iterator has so much data , that the
assignment to size does not happen at all. the loop of adding the data
keeps adding & accumulating , while the flush  continues to see 0 , as the
numAdded has not been assigned.
I am pushing in a change so as to add numAdded more frequently. I will look
into it further to see if it should be done atomically such that when the
flushing thread executes, it sees the accurate number.

sumwale:
Looks good now. One question:

> no it is for both the cases region as well as non region.

How is that the case? Won't the size be the bucket region size?

sumwale:
You could push the addition of numAdded further up into the finally{} block if required.

ahshahid:
The snippet of code is common to both reservoir & non reservoir. numStrata
would reflect the count of entries either in region or non region segment
of concurrrent segmented hashmap, right?.  if this explains, good, else I
need to understand the question with some more info.

ahshahid:
Pushed the change of doing addition in finally, but in two places. pls
review

core/src/main/scala/org/apache/spark/sql/collection/ConcurrentSegmentedHashMap.scala
```diff
@@ -227,38 +236,33 @@ private[sql] class ConcurrentSegmentedHashMap[K, V, M <: SegmentMap[K, V] : Clas
...
-                if (added.booleanValue()) numAdded += 1
+                if (added.booleanValue()) {
+                  numAdded += 1
+                }
                 idx += 1
               } else {
                 // indicates that loop must be broken immediately
                 // need to take the latest reference of segmnet
                 // after segmnetAbort is successful
+                addNumToSize()
```
sumwale: move it to below lockedState=false just from code logic perspective in case unlock throws an exception (e.g. interrupted)

ahshahid:
I think it should be above the unlock ( where it is right now) as  the
flush thread to proceed when acquiring lock , will correctly see the size ,
so that if they clear the strata , the size would still be correct & there
are no concurrent changes leading to inconsistency. , right?

sumwale:
> flush thread to proceed when acquiring lock , will correctly see the size , so that if they clear the strata

Ok makes sense.

PR #444, opened 2016-12-05
[SNAP-1199] Fix for SNAP-1199.Treat external table as external instea…

SachinJanani:
* Treat external table as external instead of Row table to avoid catalog inconsistency while catalog cleanup

## Patch testing
Tested manually 

## ReleaseNotes.txt changes

## Other PRs 
[Store PR-130](https://github.com/SnappyDataInc/snappy-store/pull/130)

…d of Row table

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
+++ b/core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
@@ -717,7 +717,7 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
       case x: IndexColumnFormatRelation => ExternalTableType.Index
       case x: JDBCAppendableRelation => ExternalTableType.Column
       case x: StreamPlan => ExternalTableType.Stream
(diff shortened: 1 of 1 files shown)


sumwale:
Looks good. I think you may need additions to SnappyHiveCatalog to handle this.

jramnara:
What is this fix for? "Treat external table as external instead of Row table " ?  Do we now have meta data in our catalog to view external tables also?

sumwale:
@jramnara We always had the meta-data and this helps in distinguishing them properly (i.e. CREATE EXTERNAL TABLE also get recreated on restart). This issue is related to visibility of these tables between split-cluster vs embedded but the same meta-data can also be used to show these on snappy-shell.

soubhik-c:
looks good. as we discussed lets mention about Sample table cleanup handling.

SachinJanani:
After discussing with shirish we have also handled cleanup for Sample table.This handling is done in SnappyHiveCatalog.getisTableInStoreDD

SachinJanani:
Showing external table in snappy-shell will be handled as a part of SNAP-1208

PR #446, opened 2016-12-06
ct test coverage

sonalsagarwal:
- Adding tests for ct useCase. Tests create tables with ct schema and run queries for the schema.
- Added tests for embedded mode, split mode, with and without server and lead HA.
- Tests are configured to create tables using either sql scripts, snappy jobs or spark app(in case of split mode)
- We can configure redundancy and persistence mode in the bt file.
- Each test runs queries using sql scripts, snappy job and sparkApp (in case of split mode). Added validation for queries executed in snappy job and spark app.
- FullResultSet validation changes are partial. hence removed now.

## Patch testing
- Ran the hydra tests and analyzed the logs.


sonalsagarwal:
I will incorporate the review comments later.

PR #447, opened 2016-12-06
[SNAP-1205]Avoid exchange when the table is partitioned with the join key

hbhanawat:

When a join result is joined with another table that is already partitioned on the join key, the exchange isn't avoided.  

The issue is with bucket linking we need to ensure that the  numBuckets of the two join children is same. But, when one join child is a join result of two tables, numBuckets is always 0 and hence the partitioning mismatch. 

Avoided the exchange by disabling setting of buckets for non delinking cases. 

## Patch testing

Running pre-checkin.


ymahajan:
Looks good, OrderlessHashPartitioning needs to consider numBuckets only when delinking is enabled

sumwale:
Looks good. Will be adding back de-linking for the case when its a single table or collocated/replicated table joins (i.e. no exchange) for best performance so in that case the "numBuckets" will have to be adjusted by the caller and not here.

PR #448, opened 2016-12-06
Fix for SNAP-1207

SachinJanani:
*  Fix for SNAP-1207
* This issue was intermitent which was caused when PutAllPRMessage was received by a member which has PartitionRegionPool but ThreadLocal was not set hence null check is safe here.

## Patch testing
Tested by executing unit tests which was failing previously multiple times using a script

## ReleaseNotes.txt changes

(Does this change require an entry in ReleaseNotes.txt? If yes, has it been added to it?) NA

## Other PRs 

(Does this change require changes in other projects- store, spark, spark-jobserver, aqp? Add the links of PR of the other subprojects that are related to this change) No

### DIFF
--- a/cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
+++ b/cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
@@ -77,8 +77,11 @@ class SnappyMutableURLClassLoader(urls: Array[URL], parent: ClassLoader)
   protected val jobJars = scala.collection.mutable.Map[String, URLClassLoader]()
 
   protected def getJobName: String = {
-    val jobFile = Executor.taskDeserializationProps.
-        get().getProperty(io.snappydata.Constant.JOB_SERVER_JAR_NAME, "")
+    var jobFile = ""
+    if (null != Executor.taskDeserializationProps.get()) {
+      jobFile = Executor.taskDeserializationProps.get().getProperty(io.snappydata.Constant
+          .JOB_SERVER_JAR_NAME, "")
+    }
     new File(jobFile).getName
   }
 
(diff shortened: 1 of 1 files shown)


cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -77,8 +77,11 @@ class SnappyMutableURLClassLoader(urls: Array[URL], parent: ClassLoader)
   protected val jobJars = scala.collection.mutable.Map[String, URLClassLoader]()
 
   protected def getJobName: String = {
-    val jobFile = Executor.taskDeserializationProps.
-        get().getProperty(io.snappydata.Constant.JOB_SERVER_JAR_NAME, "")
+    var jobFile = ""
+    if (null != Executor.taskDeserializationProps.get()) {
```
soubhik-c: we can avoid double lookup of Thread Local

soubhik-c:
Not sure just thread local handling is enough in this situation because looking at the stack trace, PutAllPR message is not expected to go for a remote put and fail in finding task properties (taskProps). For now this should be alright till custom UDF etc dependent on external jar isn't getting hit.

soubhik-c:
anyway for now its alright to merge.

sumwale:
Agree with @soubhik-c that this is a generic issue with using Gem put/putAll for column tables. As of SNAP-490 it will no longer be a problem for Dataset inserts (which will do an exchange if table is partitioned, and not do "residual" puts into row buffer unlike current approach) but will still be an issue for JDBC puts. However, for JDBC puts we will need to provide some mechanism for users to put Spark objects like user-defined types (UDTs). That is the real problem and this NPE just exposes that design problem. So we should track:

a) How to provide JDBC users ability to put complex UDT objects. Support for other complex types Array/Map/Struct is provided via ComplexTypeSerializer where we should also add JSON conversion support (latter has been added for JDBC gets in 20e8ef6). Same way for UDTs we should provide serializer and JSON support for both puts and gets (or discuss with Jags on any better approaches).

b) How to provide this information and/or required classes to ClusterCallbacksImpl which will need to create the column batches.

Both are related issues so can be tracked as a single JIRA or two separate ones blocking one another.

PR #449, opened 2016-12-06
Insert performance optimizations

sumwale:

- added support for writing encoder data in ColumnEncoding.scala (and implementations)
  with separate ColumnDecoder and ColumnEncoder interfaces
- added new ColumnInsertExec plan which is a code-generated Spark plan to invoke the above interface
- added support for direct RDD inserts into the table via SnappySession.sqlImplicits._ so users can do:
  rdd.insertInto("table") as long as schema matches; this is both more efficient
  and easier to use than converting to Dataset
- added an efficient code-generated EncoderScanExec now used for SnappySession.createDataset(RDD)
- all CachedBatch creation routes other than via StoreCallbacksImpl now use the new way
- added a more efficient bitset format for null values both for writing and reading
- right now only dictionary encoding and uncompressed are implemented; run-length will be added next
- added IF EXISTS support in truncate table
- renamed TPCETrade to TAQTest

Hardware: Dell Latitude E7450 (2 cores 2 threads)

ColumnCacheBenchmark with two long columns.
Master: local[*]
Size: 50M
Dataset: sparkSession.range(size).selectExpr("id", "floor(rand() * 10000) as k")

```
Java HotSpot(TM) 64-Bit Server VM 1.8.0_111-b14 on Linux 4.4.0-21-generic
Intel(R) Core(TM) i7-5600U CPU @ 2.60GHz
Cache random keys:                       Best/Avg Time(ms)    Rate(M/s)   Per Row(ns)   Relative
------------------------------------------------------------------------------------------------
cache = T compress = T                      12676 / 12927          3.9         253.5       1.0X
cache = T compress = F                        3651 / 3815         13.7          73.0       3.5X
cache = F snappy = T                           400 /  482        125.0           8.0      31.7X
```

Note that SnappyData column store now has default compression (dictionary, run length, boolean bitset) that cannot be disabled -- only increased compression levels will be added in future. There is no reason to disable these because they result in space savings as well as increased scan performance in many cases (or dictionary optimization for many queries) with little to no overhead for other cases.

TAQTest with multiple columns including strings.
Master: local[*]
Size: 34M
See quoteRDD in [TAQTest.benchmarkRandomizedKeys](https://github.com/SnappyDataInc/snappydata/blob/4fd659cee5c8a5984af91ade3bfd712085a9eccb/cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/TAQTest.scala#L297) for the Dataset.

```
Java HotSpot(TM) 64-Bit Server VM 1.8.0_111-b14 on Linux 4.4.0-21-generic
Intel(R) Core(TM) i7-5600U CPU @ 2.60GHz
Cache random data:                       Best/Avg Time(ms)    Rate(M/s)   Per Row(ns)   Relative
------------------------------------------------------------------------------------------------
Q1: cache = T                               28160 / 28457          1.2         828.2       1.0X
Q1: cache = F snappy = T                      2099 / 2285         16.2          61.7      13.4X
```

## Patch testing

In progress

## ReleaseNotes.txt changes

Add note about ability to do: RDD.insertInto(table) using SnappySession.sqlImplicits

## Other PRs 

NA


sumwale:
Optimizing sample table inserts is a todo which will be taken up later (difficult for 0.7 release).

Main changes are the new ColumnInsertExec and the additions to ColumnEncoding.scala, DictionaryEncoding.scala, Uncompressed.scala. The EncoderScanExec is relatively minor addition though it helps with performance quite a bit when source is a generated RDD.

core/src/main/scala/org/apache/spark/sql/sources/StoreStrategy.scala
```diff
@@ -52,7 +54,16 @@ object StoreStrategy extends Strategy {
     case drop: DropTable =>
       ExecutedCommandExec(drop) :: Nil
 
-    case DMLExternalTable(name, storeRelation: LogicalRelation, insertCommand) =>
+    case p: EncoderPlan[_] =>
+      val plan = p.asInstanceOf[EncoderPlan[Any]]
+      EncoderScanExec(plan.rdd.asInstanceOf[RDD[Any]],
+        plan.encoder, plan.isFlat, plan.output) :: Nil
+
+    case logical.InsertIntoTable(l@LogicalRelation(p: PlanInsertableRelation,
```
sumwale: "create as select" will go to ColumnFormatRelation.insert which has been changed to use dataFrame.write itself (https://github.com/SnappyDataInc/snappydata/pull/449/files#diff-f95903beca8b4604e25ee9f70dd382ceL203) so it will come back to this case.

hbhanawat: Wouldn't StoreStrategy change the CreateTableUsingAsSelect into ExecutedCommandExec which will not use the insert route?

sumwale: Maybe I am missing something but all that will ultimately go to SnappySession.createTable which uses the InsertableRelation.insert. There is no code we have added to insert into snappydata tables other than BaseRelation.insert override (and the new ColumnInsertExec added in this PR).

hbhanawat: Ok. I thought differently because there is a different case block for  "create as select" in StoreStrategy which creates CreateMetastoreTableUsingSelect. If you have tested it, its ok,

sumwale: Yes, CreateMetastoreTableUsingSelect calls SnappySession.createTable which in turn calls InsertableRelation.insert which is the actual code that does the insert. Same path was being followed for DataFrameWriter.insertInto too earlier but now that is changed to ColumnInsertExec for column tables (no change for row tables).

PR #450, opened 2016-12-06
Updated LocalJoin to handle cover colocated join cases as well.

hbhanawat:

1. Updated LocalJoin to handle cover colocated join cases as well.
2. SnappyStrategies now convert a non-replicated table equi join into a local join if the plan.statistics.sizeInBytes < 100 MB. This default size can be altered using snappydata.localhashjoinsize SQLConf property.
3. Build side and stream side produce the code inside the doProduce of the LocalJoin. To supply the iterator of the rdd of build-side to the produce of build side, slightly hackish way has been used.
4. For colocated join, there is no need to cache the map.
5. Specified the clustering of local join so that the exchange nodes are introduced when needed.
6. Since the inputRdds of LocalJoin returns stream side rdd, the shuffle dependencies of the build rdd are never materialized during the query execution. This always results in 0 rows. Fixed this by creating a DelegateRDD for the stream RDD that has the dependencies of the stream rdd and the build rdd.
7. Fixed few issues in generated code exposed by the Local join tables.

## Patch testing

Ran precheckin

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
@@ -17,7 +17,7 @@
 package org.apache.spark.sql
 
 import org.apache.spark.sql.catalyst.expressions.aggregate.{AggregateExpression, AggregateFunction, Final, ImperativeAggregate, Partial, PartialMerge}
-import org.apache.spark.sql.catalyst.expressions.{Alias, Expression, NamedExpression}
+import org.apache.spark.sql.catalyst.expressions.{Alias, Expression, NamedExpression, RowOrdering}
 import org.apache.spark.sql.catalyst.planning.{ExtractEquiJoinKeys, PhysicalAggregation, PhysicalOperation}
 import org.apache.spark.sql.catalyst.plans.logical.{Join, LogicalPlan, ReturnAnswer}
 import org.apache.spark.sql.catalyst.plans.{ExistenceJoin, Inner, JoinType, LeftAnti, LeftOuter, LeftSemi, RightOuter}
@@ -61,15 +61,33 @@ private[sql] trait SnappyStrategies {
   object LocalJoinStrategies extends Strategy {
     def apply(plan: LogicalPlan): Seq[SparkPlan] = plan match {
       case ExtractEquiJoinKeys(joinType, leftKeys, rightKeys, condition,
-      left, right) =>
-        if (canBuildRight(joinType) && canLocalJoin(right)) {
+      left, right) if canBuildRight(joinType) && canLocalJoin(right) =>
+        makeLocalHashJoin(leftKeys, rightKeys, left, right, condition,
+          joinType, joins.BuildRight, true)
+      case ExtractEquiJoinKeys(joinType, leftKeys, rightKeys, condition,
+      left, right) if canBuildLeft(joinType) && canLocalJoin(left) =>
           makeLocalHashJoin(leftKeys, rightKeys, left, right, condition,
-            joinType, joins.BuildRight)
-        } else if (canBuildLeft(joinType) && canLocalJoin(left)) {
-          makeLocalHashJoin(leftKeys, rightKeys, left, right, condition,
-            joinType, joins.BuildLeft)
-        } else Nil
+            joinType, joins.BuildLeft, true)
+      case ExtractEquiJoinKeys(joinType, leftKeys, rightKeys, condition, left, right)
+        if canBuildRight(joinType) && canBuildLocalHashMap(right) ||
+            !RowOrdering.isOrderable(leftKeys) =>
+        makeLocalHashJoin(leftKeys, rightKeys, left, right, condition,
+          joinType, joins.BuildRight, false)
+      case ExtractEquiJoinKeys(joinType, leftKeys, rightKeys, condition, left, right)
+        if canBuildLeft(joinType) && canBuildLocalHashMap(left) ||
+            !RowOrdering.isOrderable(leftKeys) =>
+        makeLocalHashJoin(leftKeys, rightKeys, left, right, condition,
+          joinType, joins.BuildLeft, false)
       case _ => Nil
+    }
+
+    /**
+     * Matches a plan whose size is small enough to build a hash table.
+     *
+     */
+    private def canBuildLocalHashMap(plan: LogicalPlan): Boolean = {
+      plan.statistics.sizeInBytes <
+          io.snappydata.Property.LocalHashJoinSize.configEntry.getConf[Long](conf)
     }
 
     private def canBuildRight(joinType: JoinType): Boolean = joinType match {
@@ -106,9 +124,10 @@ private[sql] trait SnappyStrategies {
         right: LogicalPlan,
         condition: Option[Expression],
         joinType: JoinType,
-        side: joins.BuildSide): Seq[SparkPlan] = {
+        side: joins.BuildSide,
+        replicatedTableJoin: Boolean): Seq[SparkPlan] = {
       joins.LocalJoin(leftKeys, rightKeys, side, condition,
-        joinType, planLater(left), planLater(right)) :: Nil
+        joinType, planLater(left), planLater(right), replicatedTableJoin) :: Nil
     }
   }
 
--- a/core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
@@ -36,13 +36,14 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.codegen.{CodegenContext, ExprCode, GenerateUnsafeProjection}
 import org.apache.spark.sql.catalyst.expressions.{AttributeSet, BindReferences, BoundReference, Expression, UnsafeRow}
 import org.apache.spark.sql.catalyst.plans._
-import org.apache.spark.sql.catalyst.plans.physical.{Distribution, Partitioning, UnspecifiedDistribution}
+import org.apache.spark.sql.catalyst.plans.physical.{ClusteredDistribution, Distribution, Partitioning, UnspecifiedDistribution}
 import org.apache.spark.sql.collection.Utils
-import org.apache.spark.sql.execution._
+import org.apache.spark.sql.execution.{r, _}
 import org.apache.spark.sql.execution.metric.SQLMetrics
 import org.apache.spark.sql.snappy._
 import org.apache.spark.sql.types.{LongType, StructType, TypeUtilities}
-import org.apache.spark.{Partition, SparkEnv, TaskContext}
+import org.apache.spark.sql.{SnappyAggregation, SnappySession}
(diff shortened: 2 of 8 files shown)


core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -112,22 +116,41 @@ case class LocalJoin(leftKeys: Seq[Expression],
...
+        buildDataSize += estimatedSize
+        context.taskMetrics().incPeakExecutionMemory(estimatedSize)
+        context.addTaskCompletionListener(_ => hashed.close())
+        join(itr, hashed, numOutputRows)
+      }
+    } else {
+      streamRDD.zipPartitions(buildRDD) { (streamIter, buildIter) =>
+        val hashed = buildHashedRelation(buildIter)
+        join(streamIter, hashed, numOutputRows)
+      }
```
sumwale: Actually in doExecute() I think we should just do a WholeStageCodegen(this).execute() since there is no case in which code generation of this plan should fail (apart from a bug that is). I am not 100% sure about this but we can check that after release (e.g. see BatchedDataSourceScanExec.doExecute). It makes things simpler for us with less surprises -- we are already quite a surprised bunch :-)

core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -112,22 +116,41 @@ case class LocalJoin(leftKeys: Seq[Expression],
...
-      buildDataSize += estimatedSize
-      context.taskMetrics().incPeakExecutionMemory(estimatedSize)
-      context.addTaskCompletionListener(_ => hashed.close())
-      join(itr, hashed, numOutputRows)
+    if (buildRDD.partitions.size == 1) {
+      materializeDependencies(buildRDD, new mutable.HashSet[RDD[_]]())
+      val schema = buildPlan.schema
+      streamRDD.mapPartitionsPreserveWithPartition { (context, split, itr) =>
+        val start = System.nanoTime()
+        val hashed = HashedRelationCache.get(schema, buildKeys, buildRDD,
```
sumwale: Have taken out this cache in SNAP-490 and will always create HashedRelation.

core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -137,8 +160,22 @@ case class LocalJoin(leftKeys: Seq[Expression],
   // return empty here as code of required variables is explicitly instantiated
   override def usedInputs: AttributeSet = AttributeSet.empty
 
-  override def inputRDDs(): Seq[RDD[InternalRow]] =
-    streamedPlan.asInstanceOf[CodegenSupport].inputRDDs()
+  lazy val buildCodegenRDDs = buildPlan.asInstanceOf[CodegenSupport].inputRDDs()
+
+  override def inputRDDs(): Seq[RDD[InternalRow]] = {
+    // If the build side has shuffle dependencies.
+    if (buildCodegenRDDs.length == 1 && buildCodegenRDDs(0).dependencies.size >= 1 &&
```
sumwale: Can there be a case where buildCodegenRDDs are more than 1? I think we should just walk through all the dependencies of all buildCodegenRDDs to be on the safe side because now this is used for all kinds of joins. Earlier it was always on a replicated table so it was known that only one RDD will be coming in.

core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -161,57 +198,81 @@ case class LocalJoin(leftKeys: Seq[Expression],
...
+         |$allIterators = inputs;
+         |inputs = new scala.collection.Iterator[$buildRDDs.length];
+         |$taskContextClass $contextName = $taskContextClass.get();
+         |for (int $indexVar = 0; $indexVar < $buildRDDs.length; $indexVar++) {
+         |  $partitionClass[] parts = $buildPartsVar[$indexVar];
+         |  // check for replicate table
+         |  if (parts.length == 1) {
+         |    inputs[$indexVar] = $buildRDDs[$indexVar].iterator(
+         |      parts[0], $contextName);
+         |  } else {
```
sumwale: Since you have the "replicatedTableJoin" flag, it may be better to create two different snippets based on that and not depend on parts.length (rather parts.length can have an assertion for that case)

core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -290,9 +362,11 @@ case class LocalJoin(leftKeys: Seq[Expression],
     val relationTerm = ctx.freshName("relation")
     val relationIsUnique = ctx.freshName("keyIsUnique")
     val buildRDDRef = ctx.addReferenceObj("buildRDD", buildRDD)
-    val buildPartRef = ctx.addReferenceObj("buildPartition", buildPartition)
-    ctx.addMutableState(classOf[HashedRelation].getName, relationTerm,
-      prepareHashedRelation(ctx, relationTerm, buildRDDRef, buildPartRef))
+    if (buildRDD.partitions.size == 1) {
```
sumwale: Use replicatedTableJoin flag. Though doesn't matter since the non-optimized versions are due tol be removed soon.

core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -335,10 +410,15 @@ case class LocalJoin(leftKeys: Seq[Expression],
...
-    mapAccessor.generateMapLookup(entryVar, localValueVar, keyIsUniqueTerm,
+
+    val mapAccesCode = mapAccessor.generateMapLookup(entryVar, localValueVar, keyIsUniqueTerm,
       numRowsTerm, nullMaskVars, initCode, checkCondition,
       streamSideKeys, streamKeyVars, buildKeyVars, buildVars, input,
       resultVars, dictionaryArrayTerm, joinType)
+    s"""
+       $evaluatedInputVars
+       $mapAccesCode
+     """
```
sumwale: Need to check any effect on dictionary optimization that avoids materializing the string key (unless its a miss in the current dictionary array). Hopefully JIT will optimize out unused variable. Can you check what numbers you see with TPCETrade test after increasing the sizes by 10X to 34M and 5M? Also enable the RDD delinking flag temporarily. Normally I get <200ms best case 160-170 sometimes 190. If there is any change, then let this change go in but we will need to track the issue.

core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
```diff
@@ -61,17 +61,35 @@ private[sql] trait SnappyStrategies {
...
+          joinType, joins.BuildRight, true)
+      case ExtractEquiJoinKeys(joinType, leftKeys, rightKeys, condition,
+      left, right) if canBuildLeft(joinType) && canLocalJoin(left) =>
           makeLocalHashJoin(leftKeys, rightKeys, left, right, condition,
-            joinType, joins.BuildLeft)
-        } else Nil
+            joinType, joins.BuildLeft, true)
+      case ExtractEquiJoinKeys(joinType, leftKeys, rightKeys, condition, left, right)
+        if canBuildRight(joinType) && canBuildLocalHashMap(right) ||
+            !RowOrdering.isOrderable(leftKeys) =>
```
hbhanawat: RowOrdering  checks if the keys can be sorted. For e.g. user objects may not be orderable. In such cases, we can only do hash joins. However, this check is picked up from SparkStrategies as is and its utility for us is debatable.

core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
```diff
@@ -61,17 +61,35 @@ private[sql] trait SnappyStrategies {
...
+          joinType, joins.BuildLeft, false)
       case _ => Nil
     }
 
+    /**
+     * Matches a plan whose size is small enough to build a hash table.
+     *
+     */
+    private def canBuildLocalHashMap(plan: LogicalPlan): Boolean = {
+      plan.statistics.sizeInBytes <
```
hbhanawat: Yes, we are using static for 0.7 deadline.

core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -112,22 +116,41 @@ case class LocalJoin(leftKeys: Seq[Expression],
...
-      val hashed = HashedRelationCache.get(schema, buildKeys, buildRDD,
-        buildPartition, context)
-      buildTime += (System.nanoTime() - start) / 1000000L
-      val estimatedSize = hashed.estimatedSize
-      buildDataSize += estimatedSize
-      context.taskMetrics().incPeakExecutionMemory(estimatedSize)
-      context.addTaskCompletionListener(_ => hashed.close())
-      join(itr, hashed, numOutputRows)
+    if (buildRDD.partitions.size == 1) {
+      materializeDependencies(buildRDD, new mutable.HashSet[RDD[_]]())
```
hbhanawat: Yes, you are correct. I tested materializeDependencies it does not work for the same reason mentioned by you.

core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
hbhanawat: :-) Will do this in a subsequent checkin.

core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
hbhanawat: Ok. Will report the details by tomorrow.

hbhanawat:
I agree with your comments. I will take them in a subsequent checkin. Merging this PR.

sumwale:
Lets also rename LocalJoin to HashJoin or HashJoinExec -- since they are not taken by Spark yet, so we can avoid the Snappy prefix.

core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
sumwale: Ok, the issue reported by Sonal is similar. Both this and that one are due to the same change. For dictionary optimization I had changed the "usedInputs" method to return empty set with the expectation that code will instantiate required variables instead of expanding everything upfront in the consume() call. I modified it back for both the classes and checked the perf of TPCETest Q1 is down by 15-20% (160ms vs 190ms best, 180ms vs 210ms average). While not huge, this is still significant.

The thing I tried (which worked in SnappyHashAggregateExec) is to explicitly invoke: evaluateRequiredVariables(consumer.output, input, consumer.references) before consume is invoked which calls the child consume methods. Likewise you can try invoking this (you will have to pass input) in each of the "gen*JoinCodes" method in ObjectHashMapAccessor after the continue due to join failure but before the consumeResult is emitted. So, for example, for inner join you can place this before or after the "declareLocalVars" in emitted code, same in outer join though it won't matter there since there is no case of continue, or before consumeResult in anti-join. Did a quick check with inner-join that Q1 perf is back to original level, so if this works for the queries you saw failures for then it will be the more performant option.

PR #451, opened 2016-12-07
Fix for SNAP-1214

SachinJanani:
After changes SNAP-1199 the test  

> "Catalog is repaired after restart of the cluster" 

in CatalogConsistency scalatest is no longer valid as in scalatest we are not starting cluster. These test is properly covered in 

> CatalogConsistencyDunit.testCatalogRepairedWhenLeadRestarted.

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


dshirish:
Looks good

PR #452, opened 2016-12-07
[SNAP-1215] Fix columndunit test which was failing in precheckin due …

SachinJanani:
* ColumnTableDUnitTest  was Failing in precheckin but passes when ran individual this was due to improper cleanup from previous test which was resetting batchsize to 10000 causing assertion in this test to fail 

## Patch testing
Executed precheckin and this test passed

## ReleaseNotes.txt changes

## Other PRs 

…to improper cleanup of previous test in precheckin but was passing individual


sumwale:
Don't understand how a previous test can effect this test. Now that I look at it, the SparkContext is created only once and used throughout unless stopSpark is done explicitly and a fresh SparkContext created. So setting this property in SparkContext should not be done at all and each test should set into SnappyContext as required (and clear it at the end). Will take care of that as part of SNAP-490 that changes the property to be size based in any case and this property will no longer be valid.

PR #453, opened 2016-12-07
Adding system flag Enable Experimental Features

soubhik-c:

Adding system flag Enable Experimental Features
    
    Under this flag experimental features should be released. Presently, its
    DistributedIndex which by default is disabled. It is right now a system
    property and sql conf property.
    
    precheckin: TPCHDUnitTest is failing consistently apart from
    CacheBa*DUnitTest and :runQuickstart. Rest everything is clean including
    a similar failure as :runQuickstart in AQPDUnit testQuickStart.

## Patch testing

CreateIndexTest now uses a mix of system property and sqlconf property.

## ReleaseNotes.txt changes

## Other PRs 


core/src/main/scala/org/apache/spark/sql/sources/SnappyOptimizations.scala
```diff
@@ -408,7 +409,13 @@ case class ResolveIndex(implicit val snappySession: SnappySession) extends Rule[
 
   override def apply(plan: LogicalPlan): LogicalPlan = {
     val joinOrderHints = JoinOrderStrategy.getJoinOrderHints
-
+    val enabled: Boolean = io.snappydata.Property.EnableExperimentalFeatures.
+        configEntry.getConf[Boolean](snappySession.snappyContext.conf) ||
+        java.lang.Boolean.getBoolean(io.snappydata.Property.EnableExperimentalFeatures
+            .configEntry.key)
```
sumwale: I think we should avoid this explicit system property otherwise it will become a general practise which is not a good idea. I think better option will be to allow alternate "spark." (like done in Property.ValueAlt but not there for SQLVal) and let SparkConf handle those properties for both embedded and split modes. Let me see adding alternative values like done in ValueAlt for SQLVal too (or if you can do it then will be good since I have lots of other stuff to complete urgently).

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -105,29 +110,30 @@ class CreateIndexTest extends SnappyFunSuite with BeforeAndAfterEach {
...
-      validateIndex(Seq(indexThree))(_)
+      CreateIndexTest.validateIndex(Seq(indexThree))(_)
     }
   }
 
   test("Test create Index on Column Table using Snappy API") {
     val tableName: String = "tcol1"
     val snContext = SnappyContext(sc)
     context.set(snContext)
+    System.setProperty(io.snappydata.Property.EnableExperimentalFeatures.configEntry.key, "true")
```
sumwale: Is this to test setting as a system property?

sumwale: The afterEach is also setting the conf property, and System property is not being cleared at the end of each test either. So if the intention was to test either of conf or System.property then this is not achieving it. Set either of the two ways in the test, then clear both at the end of each test.

sumwale:
@soubhik-c I have the changes to SQLVal ready. Will be checked in a couple of hours.

core/src/main/scala/org/apache/spark/sql/sources/SnappyOptimizations.scala
sumwale: Ok, I pushed changes in 12119c1. So you can create this with alternative spark suffix. Then use:

```scala
EnableExperimentalFeatures.get[Boolean](snappySession.snappyContext.conf)
```

Spark prefixed system properties (and also "snappydata." ones in embedded mode) will automatically be picked up by SparkConf and from there will be copied to SQLConf by default so separate system property handling is not required and will work in split mode too.

Not required specifically for this I suppose since this is a temporary property, but this can be the general way for snappydata properties.

sumwale:
See the last comment -- please merge master and update.

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -194,7 +194,13 @@ object Property extends Enumeration {
 
   val HashJoinSize: SQLType = SQLVal[Long](s"${Constant.PROPERTY_PREFIX}hashJoinSize",
     "The join would be converted into a hash join if the table is of size less" +
-        "than hashJoinSize. Default value is 100 MB.", Some(100L * 1024 * 1024))
+        "than localhashjoinsize. Default value is 100 MB.", Some(100*1024*1024))
```
sumwale: revert this (probably while resolving the conflict)

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -49,13 +50,16 @@ class CreateIndexTest extends SnappyFunSuite with BeforeAndAfterEach {
...
+    System.clearProperty(io.snappydata.Property.EnableExperimentalFeatures.configEntry.key)
     super.afterAll()
   }
 
   override def afterEach(): Unit = {
     try {
       val snContext = context.getAndSet(null)
       if (snContext != null) {
+        snContext.setConf(io.snappydata.Property.EnableExperimentalFeatures.configEntry.key,
+          io.snappydata.Property.EnableExperimentalFeatures.configEntry.defaultValueString)
```
sumwale: Just this should work, no? `snContext.setConf(io.snappydata.Property.EnableExperimentalFeatures.name, "false")`
I don't think explicit passing of defaultValueString is required.

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -69,6 +73,7 @@ class CreateIndexTest extends SnappyFunSuite with BeforeAndAfterEach {
   test("Test choice of index according to where predicate") {
     val tableName = "tabOne"
     val snContext = SnappyContext(sc)
+    snContext.setConf(io.snappydata.Property.EnableExperimentalFeatures.configEntry.key, "true")
```
sumwale: .name instead of .configEntry.key will be good. Lets use that so it becomes a general practise for future code.

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -229,6 +235,7 @@ class CreateIndexTest extends SnappyFunSuite with BeforeAndAfterEach {
   }
 
   test("Test two table joins") {
+    System.setProperty(io.snappydata.Property.EnableExperimentalFeatures.configEntry.key, "true")
```
sumwale: .name here too

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -340,6 +347,7 @@ class CreateIndexTest extends SnappyFunSuite with BeforeAndAfterEach {
   }
 
   test("Test two table joins corner cases") {
+    System.setProperty(io.snappydata.Property.EnableExperimentalFeatures.configEntry.key, "true")
```
sumwale: .name here

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -401,6 +409,7 @@ class CreateIndexTest extends SnappyFunSuite with BeforeAndAfterEach {
 
     val snContext = SnappyContext(sc)
     context.set(snContext)
+    snContext.setConf(io.snappydata.Property.EnableExperimentalFeatures.configEntry.key, "true")
```
sumwale: .name here

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -591,6 +600,7 @@ class CreateIndexTest extends SnappyFunSuite with BeforeAndAfterEach {
...
```
sumwale: .name here

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
```diff
@@ -52,10 +53,22 @@ class CreateIndexTest extends SnappyFunSuite with BeforeAndAfterEach {
     super.afterAll()
   }
 
+  override def beforeEach(): Unit = {
+    try {
+      val snContext = SnappyContext(sc)
+      snContext.setConf(io.snappydata.Property.EnableExperimentalFeatures.configEntry.key, "true")
```
sumwale: is just ".name" not working?

core/src/main/scala/io/snappydata/Literals.scala
soubhik-c: ah!.. yes. doing it.

core/src/test/scala/org/apache/spark/sql/store/CreateIndexTest.scala
soubhik-c: noticed just now. doing it.

PR #454, opened 2016-12-07
[SNAP-1218] Enable RDD-bucket de-linking for single table and replicated table joins

sumwale:

- added a LinkPartitionsToBuckets rule in optimizer phase to check for any non-replicated
  table joins in the plan tree
- use SnappySession contextObject to add the flag to set linkPartitionsToBuckets flag (thus disabling de-linking)
- use the above flag for both the row/column RDD partitions
  as well as plan's partitioning (OrderlessHashPartitioning)
- updated NorthWind tests for above changes; also adding validation for expected partitions in joins
  and a flag in NWQueries to just emit warning for partition expectation mismatch instead of failing

## Patch testing

precheckin (in progress)

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -123,6 +126,23 @@ class SnappySessionState(snappySession: SnappySession)
...
+   * be created for tables to be the same as number of buckets. This will avoid
+   * exchange on one side of a non-collocated join in most cases.
+   */
+  object LinkPartitionsToBuckets extends Rule[LogicalPlan] {
+    def apply(plan: LogicalPlan): LogicalPlan = {
+      plan.foreach {
+        case j: Join if !planner.asInstanceOf[SnappyStrategies]
+            .LocalJoinStrategies.isLocalJoin(j) =>
+          // disable for the entire query for consistency
+          snappySession.linkBucketsToPartitions(flag = true)
```
hbhanawat: After a query is done, who resets this flag to false?

core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
```diff
@@ -81,13 +81,20 @@ private[sql] trait SnappyStrategies {
       case _ => Nil
     }
 
+    def isLocalJoin(plan: LogicalPlan): Boolean = plan match {
+      case ExtractEquiJoinKeys(joinType, _, _, _, left, right) =>
+        (canBuildRight(joinType) && canLocalJoin(right)) ||
+            (canBuildLeft(joinType) && canLocalJoin(left))
```
hbhanawat: These conditions only check for replicated tables. canBuildLeft(joinType) && canBuildLocalHashMap(left) || canBuildRight(joinType) && canBuildLocalHashMap(right) checks for non-replicated table joins as well.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -123,6 +126,23 @@ class SnappySessionState(snappySession: SnappySession)
...
 
+  /**
+   * This rule sets the flag at query level to link the partitions to
+   * be created for tables to be the same as number of buckets. This will avoid
+   * exchange on one side of a non-collocated join in most cases.
+   */
+  object LinkPartitionsToBuckets extends Rule[LogicalPlan] {
+    def apply(plan: LogicalPlan): LogicalPlan = {
+      plan.foreach {
+        case j: Join if !planner.asInstanceOf[SnappyStrategies]
```
hbhanawat: Why only for localjoin? Even for sort merge join the exchange can be avoided if rdd delinking is disabled.

core/src/main/scala/org/apache/spark/sql/execution/aggregate/SnappyHashAggregateExec.scala
```diff
@@ -585,10 +585,10 @@ case class SnappyHashAggregateExec(
        |$bufferEval
        |
        |// common sub-expressions
+       |$inputCodes
```
hbhanawat: What is this change for? Some other bug fix?

tests/common/src/main/resources/TPCH/RESULT/Snappy_q10.out
```diff
@@ -98,106 +98,3 @@
 106,Customer#000000106,91726.7176,3288.4200,ARGENTINA,xGCOEAUjUNG,11-751-989-4627,lose slyly. ironic accounts along the evenly regular theodolites wake about the special, final gifts.
```
hbhanawat: Why are you updating the test data files?

build.gradle
```diff
@@ -1115,3 +1117,15 @@ task precheckin {
   dependsOn cleanAll, buildAll, checkAll, allReports, runQuickstart
   dependsOn ':snappy-spark:scalaStyle', docs
 }
+
+if (project.hasProperty('trackBuildTime') ) {
```
hbhanawat: A couple of lines for this change would be needed in your final commit.

sumwale:
@hbhanawat Somehow other commits to master got mixed in this PR. Let me try to merge again.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
sumwale: The condition is reverse: if join is a LocalJoin then don't LinkPartitionToBuckets (not de-link) i.e. RDD-delinking will be in force only for replicated table joins.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
sumwale: Uses the contextObjects in SnappySession which is cleared in SnappySession.clearQueryData() which is cleared both by parser and at end of plan execution.

sumwale: In other words: if plan caching is in effect then SnappySession.evaluatePlan will clear it. If not, then parser will clear it.

core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
sumwale: Yes that was the idea. The LocalHashMap can be built for exchange results too where we want to skip de-linking (or rather enforce linking by LinkPartitionToBuckets). We could add more precise checks later which check for collocated join cases too, but right now played it safe so for collocated joins the partitions used will be same as buckets.

tests/common/src/main/resources/TPCH/RESULT/Snappy_q10.out
sumwale: Again another commit 1483386 on master to fix failures in TPCHDUnitTest introduced by Kishor's changes of LIMIT additions to queries in 3de70ff. Fixing other people's stuff to get clean precheckins.

build.gradle
sumwale: Again from master b94947d which is Soubhik's change.

sumwale:
@soubhik-c

> actually IMO, SQLVal and Val should extend ConfigEntry

No they cannot since ConfigEntry is private[spark] and SQLConf itself is private[sql]. Hence the wrapper SQLConfigEntry with much cleaner functionality than provided by SQLConf.

> Right now user code uses only the string variant
> ideally it should be like

No, the latter one is private[sql] and not meant for users. In fact SQLConf itself is not visible so users cannot use it and have to embed the strings. For our case it is public so users can do:

```scala
snContext.setConf(io.snappydata.Property.EnableExperimentalFeatures.name, "true")
```

Compare to current one in SQLConf which users can actually use:

```scala
sqlContext.setConf(SQLConf.WHOLESTAGE_CODEGEN_ENABLED.key, "true")
```

Which is similar.

See the public type-safe usage:

```scala
Property.HashJoinSize.get[Long](conf)
```

which is not available in SQLConf (all private).

Or compare the declaration:

```scala
  val LocalHashJoinSize = SQLVal[Long](s"${Constant.PROPERTY_PREFIX}localhashjoinsize",
    "The join would be converted into a hash join if the table is of size less" +
        "than localhashjoinsize. Default value is 100 MB.", Some(100*1024*1024))
```

vs in SQLConf where explicit longConf/createWithDefault and unnecessary builder stuff is needed:

```scala
  val AUTO_BROADCASTJOIN_THRESHOLD = SQLConfigBuilder("spark.sql.autoBroadcastJoinThreshold")
    .doc("Configures the maximum size in bytes for a table that will be broadcast to all worker " +
      "nodes when performing a join.  By setting this value to -1 broadcasting can be disabled. " +
      "Note that currently statistics are only supported for Hive Metastore tables where the " +
      "command<code>ANALYZE TABLE &lt;tableName&gt; COMPUTE STATISTICS noscan</code> has been run.")
    .longConf
    .createWithDefault(10L * 1024 * 1024)
```

This is an improvement in my opinion. If you have suggestions to better it, then let me know.

soubhik-c:
> No they cannot since ConfigEntry is private[spark] and SQLConf itself is private[sql]. Hence the wrapper SQLConfigEntry with cleaner functionality than provided by SQLConf.

How about we move SQLVal and Val classes to ```org.apache.spark.sql``` & ```org.apache.spark``` respectively in our codebase and just return based on respective type.

 ```

SQLConfigBuilder("spark.snappydata.xxxx").fallbackConf(
   SQLConfigBuilder("snappydata.xxxx")
    .doc("......description.....")
    .longConf
    .createWithDefault(1000L)
   )
```

sumwale:
> How about we move SQLVal and Val classes to org.apache.spark.sql & org.apache.spark respectively in our codebase and just return based on respective type.

Users still cannot use the setConf taking ConfigEntry since it is private so what's the point? All Spark tests also use entry.key pattern (entry.name for us).

sumwale:
Merging now. Any changes to config discussed will have to be taken up later and by those who are handling general configuration properties ( @SachinJanani and @soubhik-c )

soubhik-c:
>Users still cannot use the setConf taking ConfigEntry since it is private so what's the point? All Spark tests also use entry.key pattern (entry.name for us).

Not sure why you say that because spark conf properties users are already able to use irrespective of its visibility.

sumwale:
@soubhik-c SQLConf class is private access. But SQLConf object is public. So users can refer to constants from SQLConf but they cannot use the private[sql] methods (like SQLContext.setConf that takes ConfigEntry).

Check out all uses in Spark tests e.g. check usages of SQLConf.WHOLESTAGE_CODEGEN_ENABLED (FilteredScanSuite, PrunedScanSuite etc) where each and every one calls ".key" for that reason.

soubhik-c:
Ah!.. finally got it, I was checking from a test which was package [sql] enabled.

PR #455, opened 2016-12-08
Check the size of left side before deciding the build side 

hbhanawat:
Currently if a hashmap can be build for the right side, it is marked as the build side without checking the size of the left side.

Checking the size of the left side before marking right as the build side.


sumwale:
looks good

PR #457, opened 2016-12-08
Snap 1219

suranjan:

ColumnFormatRelation creates table and buckets while its creation.
  For new session too, if table is already present, table
  creation is a no-op. However, creation of buckets may not be a no-op
  in certain scenarios. e.g Persistent tables with a node down case.
  Now before creating buckets checking if buckets were already created.
## Patch testing

Manually

## ReleaseNotes.txt changes

N/A

## Other PRs 

N/A


kneeraj:
Looks good. Run precheckin and merge.

PR #458, opened 2016-12-08
Handling Exchange via EnsureRequirements

soubhik-c:

+ Now instead of creating ShuffleExchange explicitly, using
requiredChildDistribution to indicate shuffle is required at this
operator.

+ Adding IndexColumnFormatRelation too in avoiding 1 to many partition
to bucket mapping in other words avoiding delinking of partition ->
buckets.

tests: Added new dunit which will be enhanced to cover more conditions
as covered in unit test

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/cluster/src/dunit/scala/io/snappydata/cluster/DistributedIndexDUnitTest.scala
+++ b/cluster/src/dunit/scala/io/snappydata/cluster/DistributedIndexDUnitTest.scala
@@ -95,8 +95,6 @@ class DistributedIndexDUnitTest(s: String) extends ClusterManagerTestBase(s) {
 //      CreateIndexTest.validateIndex(Seq.empty, tableName)(_)
 //    }
 
-    System.setProperty("LOG-NOW", "xxx")
-    getLogWriter.info("SB: About to execute queries")
     executeQ(s"select * from $tableName where col2 = 'bbb' and col3 = 'halo' ") {
       CreateIndexTest.validateIndex(Seq(indexTwo))(_)
     }
@@ -104,8 +102,6 @@ class DistributedIndexDUnitTest(s: String) extends ClusterManagerTestBase(s) {
     executeQ(s"select * from $tableName where col1 = 111 and col3 = 'halo' ") {
       CreateIndexTest.validateIndex(Seq(indexThree))(_)
     }
-    getLogWriter.info("SB: Done executing the queries")
-    System.clearProperty("LOG-NOW")
   }
 
 }
--- a/cluster/src/dunit/scala/io/snappydata/cluster/DistributedIndexDUnitTest.scala
+++ b/cluster/src/dunit/scala/io/snappydata/cluster/DistributedIndexDUnitTest.scala
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) 2016 SnappyData, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+package io.snappydata.cluster
+
+import scala.collection.mutable.ListBuffer
+
+import org.apache.spark.sql.internal.SQLConf
+import org.apache.spark.sql.store.CreateIndexTest
+import org.apache.spark.sql.{SaveMode, SnappyContext}
+
+/**
+ * Tests various distributed index related tests.
+ */
+class DistributedIndexDUnitTest(s: String) extends ClusterManagerTestBase(s) {
+
+  val tablesToDrop = new ListBuffer[String]
+  val indexesToDrop = new ListBuffer[String]
+  override def tearDown2(): Unit = {
+    try {
+      val snContext = SnappyContext(sc)
+      if (snContext != null) {
+        snContext.setConf(io.snappydata.Property.EnableExperimentalFeatures.name,
+          io.snappydata.Property.EnableExperimentalFeatures.configEntry.defaultValueString)
+        snContext.setConf(SQLConf.AUTO_BROADCASTJOIN_THRESHOLD.key,
+          SQLConf.AUTO_BROADCASTJOIN_THRESHOLD.defaultValue.get.toString)
+        indexesToDrop.reverse.foreach(i => snContext.sql(s"DROP INDEX $i "))
+        tablesToDrop.reverse.foreach(t => snContext.sql(s"DROP TABLE $t "))
+        indexesToDrop.clear()
+        tablesToDrop.clear()
+      }
+    } finally {
+      super.tearDown2()
+    }
+  }
+
+  def createBaseTable(snContext: SnappyContext, tableName: String): Unit = {
+    val props = Map(
+      "PARTITION_BY" -> "col1")
+    snContext.sql("drop table if exists " + tableName)
+
+    val data = Seq(Seq(111, "aaa", "hello"),
+      Seq(222, "bbb", "halo"),
+      Seq(333, "aaa", "hello"),
+      Seq(444, "bbb", "halo"),
+      Seq(555, "ccc", "halo"),
+      Seq(666, "ccc", "halo")
+    )
+
+    val rdd = sc.parallelize(data, data.length).map(s =>
+      new Data2(s(0).asInstanceOf[Int], s(1).asInstanceOf[String], s(2).asInstanceOf[String]))
+    val dataDF = snContext.createDataFrame(rdd)
+
+    dataDF.write.format("column").mode(SaveMode.Append).options(props).saveAsTable(tableName)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
```diff
@@ -129,11 +135,22 @@ private[sql] object PartitionedPhysicalScan {
...
+        val otherPartKeys = partitionColumns.map(_.transform {
+          case a: AttributeReference => rowBufferScan.output.find(resolveCol(_, a)).getOrElse {
+            throw new AnalysisException(s"RowBuffer output column $a not found in " +
+                s"${rowBufferScan.output.mkString(",")}")
+          }
+        })
+        assert(columnScan.outputPartitioning.satisfies(
+          ClusteredDistribution(columnScan.partitionColumns)))
+        ZipPartitionScan(columnScan, columnScan.partitionColumns,
+          rowBufferScan, otherPartKeys)
```
sumwale: As I have mentioned previously, can there be cases where the generated RDD does not have preferredLocations. Right now the executed RDD will be ultimately the inputRDDs() of the plan which I suppose should suffice with ZipPartitionScan?

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
```diff
@@ -168,15 +185,23 @@ trait PartitionedDataSourceScan extends PrunedUnsafeFilteredScan {
 }
 
 private[sql] final case class ZipPartitionScan(basePlan: SparkPlan with CodegenSupport,
-    otherPlan: SparkPlan, otherRDD: RDD[InternalRow] = null,
-    relation: PartitionedDataSourceScan = null) extends LeafExecNode with CodegenSupport {
+    basePartKeys: Seq[Expression],
+    otherPlan: SparkPlan,
+    otherPartKeys: Seq[Expression],
+    otherRDD: Option[RDD[InternalRow]] = None,
+    relation: Option[PartitionedDataSourceScan] = None) extends LeafExecNode with CodegenSupport {
```
sumwale: Shouldn't this be a BinaryExecNode to be clearer in intent?

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
```diff
@@ -168,15 +185,23 @@ trait PartitionedDataSourceScan extends PrunedUnsafeFilteredScan {
...
   private val inputCode = basePlan.asInstanceOf[CodegenSupport]
 
+  override def children: Seq[SparkPlan] = basePlan :: otherPlan :: Nil
+
+  override def requiredChildDistribution: Seq[Distribution] =
+    ClusteredDistribution(basePartKeys) :: ClusteredDistribution(otherPartKeys) :: Nil
+
   override def inputRDDs(): Seq[RDD[InternalRow]] =
-    inputCode.inputRDDs ++ Seq(Option(otherPlan).fold(otherRDD)(_.execute()))
+    inputCode.inputRDDs ++ Seq(Option(otherPlan).fold(otherRDD.get)(_.execute()))
```
sumwale: If you are always expecting Some(rdd) here, then why is Option+get being used at all?

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
```diff
@@ -186,7 +211,7 @@ private[sql] final case class ZipPartitionScan(basePlan: SparkPlan with CodegenS
     val row = ctx.freshName("row")
     val columnsInputEval = Option(otherPlan).getOrElse(basePlan).output.zipWithIndex.map { case
       (ref, ordinal) =>
-      val baseIndex = Option(otherPlan).fold(relation.schema.fieldIndex(ref.name))(_ => ordinal)
+      val baseIndex = Option(otherPlan).fold(relation.get.schema.fieldIndex(ref.name))(_ => ordinal)
```
sumwale: same here about using Option(x) if you are always expecting Some(x)

cluster/src/dunit/scala/io/snappydata/cluster/DistributedIndexDUnitTest.scala
```diff
@@ -0,0 +1,111 @@
...
+    val executeQ = CreateIndexTest.QueryExecutor(snContext)
+//    executeQ(s"select * from $tableName where col1 = 111") {
+//      CreateIndexTest.validateIndex(Seq(indexOne))(_)
+//    }
+
+//    executeQ(s"select * from $tableName where col2 = 'aaa' ") {
+//      CreateIndexTest.validateIndex(Seq.empty, tableName)(_)
+//    }
+
+    System.setProperty("LOG-NOW", "xxx")
```
sumwale: What's this?

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
```diff
@@ -168,15 +185,23 @@ trait PartitionedDataSourceScan extends PrunedUnsafeFilteredScan {
 }
 
 private[sql] final case class ZipPartitionScan(basePlan: SparkPlan with CodegenSupport,
```
sumwale: CodgenSupport extends SparkPlan so you don't need to do this and just CodegenSupport will suffice

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
sumwale: Where are otherRDD and relation arguments to ZipPartitionScan being set?

soubhik-c: I think I understood your question... generated RDD as in the second RDD right ?
I haven't followed the partitioning till preferredLocation of generatedRDD but I assumed if number of partitions remain as columnScan, we should be good around referredLocation. I have verified that only. InputRDDs second one adapts to the first one because of second one being rdd of ShuffleExchange. Right now RowTableScan RDDs don't have preferred location or partition columns mentioned that will fetch data from all over & that is the intent no ?

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
soubhik-c: Well, this otherRDD is what Vivek was using for zipping ReservoirRDD and then it got unused. Once we get this right, he plans to use it again (fix the group by issue).

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
soubhik-c: Same as above, Option.fold behaves differently than Col.fold, @vivekwiz find, I kind of liked this oneliner :)

cluster/src/dunit/scala/io/snappydata/cluster/DistributedIndexDUnitTest.scala
soubhik-c: will cleanup.. was logging conditionally.

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
soubhik-c: It will, later :)

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
soubhik-c: Aha, right.

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
soubhik-c: On second thought, LeafExecNode is more appropriate because it acts as a last node. BinaryExecNode is more like a binary operator which this isn't. What do you say ?

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
sumwale: I don't think Vivek requires this for sample tables either -- not for sample as region. That one will go in as a normal region RDD exactly like being done for index.

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
sumwale: I guess its a matter of what semantics one assumes for BinaryExecNode. To me it is a plan which consume two separate (but related) plans. But a minor thing, so keep whatever you prefer.

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
sumwale: Perhaps my concern here is unwarranted. WholestageCodegenRDD will use ZipPartitionPartition which preserves the partitioning as in the original RDDs so it should be fine for this case.

soubhik-c: thanks to point out that RDD[InternalRow] can be wrapped with RowDataSourceScanExec and plugged into ZipPartitionScan , removing otherRDD in ZPS
@vivekwiz it will be something like this now whenever you intend to use.
```
        ZipPartitionScan(columnScan, columnScan.partitionColumns,
          RowDataSourceScanExec(
            columnScan.output, reservoirRDD, relation, columnScan
                .outputPartitioning,
            Map.empty, None
          ),
          columnScan.partitionColumns
        )


```

soubhik-c:
run precheckin

PR #459, opened 2016-12-09
[SNAP-1225] Performance improvements for hash joins (and other fixes)

sumwale:

As seen in TPCH Q4 and others, building hash join map was taking a lot of time. The major contributors are:

1) Copying of UTF8String -- while this is okay for small maps, it causes a large overhead for larger ones
2) EXISTS(SELECT *) is not optimized by Spark so all columns of the table appear in projection and get populated in the map. Currently TPCH queries have been modified to use "SELECT 1" -- needs fix in Spark optimizer or addition to snappydata optimizer.
3) Multi-values array increments in steps of 1. Now using a 25% increment (with minimum as 4)

Apart from above, micro-benchmarked insert/get performance of a bunch of hash maps: Spark's OpenHashSet, Scala's HashSet, JDK's HashSet, FastUtil's ObjectOpenHashSet and compared to Snappy's implementation. The conclusion was that ours is nearly the best in both inserts and gets. Updated the hashing function from FastUtil to improve performance a bit.

With above changes, the time taken by Q4 at 1X scale on laptop went from ~1s to 250-280ms. Detailed set of changes below:

- avoid creating a copy of incoming string for join map creation when the incoming
  data is from a column or row table (and there is no intervening Exchange); use UTF8String
  as field type in the generated class instead of a byte[] (latter was used to reduce
      overhead but for this case there is no overhead)
- increment the multi-values for a key in the join map in increments of 25% rather than 1
- increment columnBatchesSeen in scan_nextBatch() even when there was no filter defined
- updated the TPCH queries to use "SELECT 1" in EXISTS clause rather than "SELECT *"
- enhanced TPCH column tables to create/load from parquet files automatically; creation
  is done (from CSV files) when the flag TPCHColumnPartitionedTable.CREATE_PARQUET is set;
  added arguments to TPCHDUnitTest.execute to allow specifying warmup/runs (e.g. when running
      TPCHSuite manually for performance)
- added a more efficient hashInt method to HashingUtil taken from fastutil (which in turn
    takes it from Koloboke) -- this gives ~10% improvement in gets in micro-benchmarks
- moving pre-evaluation of input variables to before consuming the result of join
  (see comments in https://github.com/SnappyDataInc/snappydata/pull/450)

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/ObjectHashMapAccessor.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/ObjectHashMapAccessor.scala
@@ -24,7 +24,7 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.codegen.{CodegenContext, ExprCode}
 import org.apache.spark.sql.catalyst.expressions.{Attribute, BindReferences, Expression, NamedExpression}
 import org.apache.spark.sql.catalyst.plans._
-import org.apache.spark.sql.execution.exchange.Exchange
+import org.apache.spark.sql.execution.joins.{HashJoin, SortMergeJoinExec}
 import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.Platform
 import org.apache.spark.unsafe.array.ByteArrayMethods
@@ -305,14 +305,21 @@ case class ObjectHashMapAccessor(@transient session: SnappySession,
     val updateMinMax = integralKeys.map { index =>
       s"$hashMapTerm.updateLimits(${keyVars(index).value}, $index);"
     }.mkString("\n")
-    // for SnappyData column or row tables there is no need to make
-    // a copy of non-primitive values into the map since they are immutable
-    var hasExchange = false
-    val doCopy = child.find {
+
+    // For SnappyData column or row tables there is no need to make
+    // a copy of non-primitive values into the map since they are immutable.
+    // Also checks for other common plans known to provide immutable objects.
+    // TODO: can be extended for more plans as per their behaviour.
+    def providesImmutableObjects(plan: SparkPlan): Boolean = plan match {
       case _: PartitionedPhysicalScan => true
-      case _: Exchange => hasExchange = true; false
+      case FilterExec(_, c) => providesImmutableObjects(c)
+      case ProjectExec(_, c) => providesImmutableObjects(c)
+      // joins will always have immutable objects regardless of their children
+      case _: HashJoin | _: SortMergeJoinExec => true
       case _ => false
-    }.isEmpty || hasExchange
+    }
+    val doCopy = !providesImmutableObjects(child)
+
     val multiValuesUpdateCode = if (valueClassName.isEmpty) {
       s"""
         // no value field, only count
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/apache/spark/sql/execution/ObjectHashMapAccessor.scala
```diff
@@ -1260,23 +1294,32 @@ case class ObjectHashMapAccessor(@transient session: SnappySession,
...
         if ($stringVar.getBaseOffset() == $platformClass.BYTE_ARRAY_OFFSET
-            && ($obj = $stringVar.getBaseObject()) instanceof byte[]
-            && ($bytes = (byte[])$obj).length == $stringVar.numBytes()) {
+            && ($bytes = (byte[])$stringVar.getBaseObject()).length == $stringVar.numBytes()) {
           $colVar = $bytes;
         } else {
           $colVar = $stringVar.getBytes();
         }"""
+    // multimap holds a reference to UTF8String itself
+    case StringType if doCopy =>
```
hbhanawat: Can there be a case when doCopy= false but multimap=true? If yes, would the default case ("$colVar = ${resultVar.value};") work for it?

core/src/main/scala/org/apache/spark/sql/execution/ObjectHashMapAccessor.scala
```diff
@@ -304,32 +305,42 @@ case class ObjectHashMapAccessor(@transient session: SnappySession,
     val updateMinMax = integralKeys.map { index =>
       s"$hashMapTerm.updateLimits(${keyVars(index).value}, $index);"
     }.mkString("\n")
+    // for SnappyData column or row tables there is no need to make
+    // a copy of non-primitive values into the map
+    var hasExchange = false
+    val doCopy = child.find {
```
hbhanawat: Your comment says that for column table and row tables doCopy should be false but it returns true. Also, I must be missing something but child.find finds any node that matches one of these three conditions. So, it will always match the last condition.

core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -411,14 +409,10 @@ case class LocalJoin(leftKeys: Seq[Expression],
     }
     val streamKeyVars = ctx.generateExpressions(streamKeys)
 
-    val mapAccesCode = mapAccessor.generateMapLookup(entryVar, localValueVar, keyIsUniqueTerm,
+    mapAccessor.generateMapLookup(entryVar, localValueVar, keyIsUniqueTerm,
```
sumwale: Yes, just pushed a fix whereby the checkCondition's code is separately emitted with rest of the inputCodes at the top.

core/src/main/scala/org/apache/spark/sql/execution/ObjectHashMapAccessor.scala
sumwale: I am quite sure that's not the case. The find above is a short-cut for:
```scala
child.find { p match {
  case ...
} }
```
which matches the first pattern and returns true/false. What this find will do is return the first PartitionedPhysicalPlan. If no such is found then "doCopy" should be true as noted in the comments (that strings are immutable for our tables but may not be for any others) hence the isEmpty.

The bug was in the exchange check. I have moved out the variable as an additional condition on top in last push -- please see.

However, I do agree this check is a fragile one and can lead to problems for more complicated joins that do not have an exchange. Let me think more about it how to make it tighter and deterministic.

PR #461, opened 2016-12-11
[SNAP-1234] Check for collocated join before checking for broadcast

sumwale:

- now added a "isCollocatedJoin" methd that will walk through the tree to check cases of
  collocated joins (including a collocated join result itself joining with another collocated table)
- also check for the sizes to check for the better side to broadcast
- changed Parquet writing in TPCHColumnPartitionedTable to use default compression since
  for data on NFS cases, disk+network is going to be the bottleneck; also initialize the CREATE_PARQUET
  flag based on test system property "snappydata.test.create_parquet"

## Patch testing

precheckin, TPCH load testing, new tests to check for collocated/broadcast/hashjoin to be added subsequently

## ReleaseNotes.txt changes

Notes about hashJoinSize and recommendations for broadcast and hashJoin limits

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/35 created for overflow related issue which is related to this but not a dependency

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
@@ -61,24 +61,48 @@ private[sql] trait SnappyStrategies {
 
   object LocalJoinStrategies extends Strategy {
     def apply(plan: LogicalPlan): Seq[SparkPlan] = plan match {
+
       case ExtractEquiJoinKeys(joinType, leftKeys, rightKeys, condition,
       left, right) if canBuildRight(joinType) && canLocalJoin(right) =>
         makeLocalHashJoin(leftKeys, rightKeys, left, right, condition,
-          joinType, joins.BuildRight, true)
+          joinType, joins.BuildRight, replicatedTableJoin = true)
       case ExtractEquiJoinKeys(joinType, leftKeys, rightKeys, condition,
       left, right) if canBuildLeft(joinType) && canLocalJoin(left) =>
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
```diff
@@ -134,6 +159,51 @@ private[sql] trait SnappyStrategies {
...
+        case PhysicalOperation(_, _, ExtractEquiJoinKeys(_, lKeys, rKeys,
+        _, left, right)) =>
+          // If join is a result of collocated join, then the result can
+          // also be a collocated join with other tables if compatible
+          getCollocatedPartitioning(left, lKeys, right, rKeys)
+
+        case _ => (Nil, Nil, -1)
+      }
+
+      val (leftCols, leftKeyOrder, leftNumPartitions) = getCompatiblePartitioning(
```
hbhanawat: Changes look good to me apart from one confusion. How is the case handled when there is a filter/project node between Join node and the PhysicalOperation of both the children. Wouldn't getCompatiblePartitioning return (Nil, Nil, -1) for both left and right. So the partitons and keyorder check would satisfy.

sumwale: Check PhysicalOperation.unapply. It breaks the projections and filters already and unwraps the underlying physical operator.

hbhanawat: Oh yeah. Thanks.

PR #463, opened 2016-12-12
[SNAP-1240] Snappy monitoring dashboard

snappy-sachin:

SnappyData Product Features :

Create a new Dashboard UI in Spark UI Application to summarise and display the overall SnappyData cluster Statistics.
1. Include cluster level statistics like Cluster Status, No of Members, Lead, Locators and Data Servers, Clients, tables, etc.
2. Members summary like Members status, Name, Host, CPU and Memory details, used and total memory, clients. etc.
3. Tables summary like Table Name, Storage Model, Distribution Type, Row Count, In Memory Size and Total Size, etc. 

## Patch testing

* Tested Manually

## ReleaseNotes.txt changes

## Other PRs 

* https://github.com/SnappyDataInc/spark/pull/36
* https://github.com/SnappyDataInc/snappy-store/pull/134


cluster/src/main/scala/org/apache/spark/ui/SnappyDashboardTab.scala
```diff
@@ -0,0 +1,52 @@
...
+
+package org.apache.spark.ui
+
+import scala.collection.mutable.ArrayBuffer
+
+import org.apache.spark.internal.Logging
+import org.apache.spark.ui.JettyUtils._
+
+/**
+ * Created by skapse on 18/10/16.
```
rishitesh: Remove author info

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1028,6 +1029,10 @@ object SnappyContext extends Logging {
     }
   }
 
+  private def updateUI(sc: SparkContext): Unit = {
+    ToolsCallbackInit.toolsCallback.updateUI(sc.ui)
+  }
+
   private def invokeServices(sc: SparkContext): Unit = {
     SnappyContext.getClusterMode(sc) match {
       case SnappyEmbeddedMode(_, _) =>
```
rishitesh: You should also think, how new pages can be made pluggable. If required propose a change request upstream Spark. So that we don't have keep everything in Spark.

core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -104,6 +114,51 @@ object SnappyTableStatsProviderService extends Logging {
       delay, delay)
   }
 
+
+  def getAggregatedMemberStatsOnDemand: Unit = {
```
rishitesh: This method should be inside a try.catch block. So that any exception won't show up.

core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -30,19 +31,24 @@ import com.gemstone.gemfire.internal.SystemTimer
...
 
-import org.apache.spark.sql.{SnappyContext, SnappySession}
 import org.apache.spark.sql.collection.Utils
+import org.apache.spark.sql.{SnappyContext, SnappySession}
 import org.apache.spark.{Logging, SparkContext}
 
 object SnappyTableStatsProviderService extends Logging {
 
-  @volatile private var tableSizeInfo = Map[String, SnappyRegionStats]()
+  @volatile
```
rishitesh: Why volatile if only accessed by one thread ?

rishitesh: Ok, there is a reader as well Ignore.

snappy-sachin:
Run precheckin

core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
snappy-sachin: OK

snappy-sachin:
Run precheckin

PR #464, opened 2016-12-12
Fixes NPE in select distinct query

soubhik-c:
NPE Fix:

+ The original query is no more using the same code generation logic
where NPE was getting hit but instead 'select distinct ShipCountry from
orders' was hitting it.

+ Handled null as suggested by Sumedh
+ Tested manually as it wasn't showing up in embed mode, will add a test
in SplitClusterDUnit covering this case.

+ refactored the NW schema a bit to reuse externally.

precheckin: No new failures.
(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
@@ -661,6 +661,7 @@ private[sql] final case class ColumnTableScan(
     val sqlType = Utils.getSQLDataType(attr.dataType)
     val jt = ctx.javaType(sqlType)
     var jtDecl = s"final $jt $col;"
+    val nullVar = ctx.freshName("nullVal")
     val colAssign = sqlType match {
       case DateType => s"$col = $decoder.readDate($buffer, $cursorVar);"
       case TimestampType =>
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -683,9 +684,14 @@ private[sql] final case class ColumnTableScan(
             """.stripMargin
         dictionaryAssignCode =
             s"$dictIndex = $decoder.readDictionaryIndex($buffer, $cursorVar);"
+        val nullCheckAddon =
+          if (notNullVar != null) s"if ($notNullVar < 0) $nullVar = $col == null;\n"
+          else ""
```
sumwale: This will leave an empty semi-colon since its both here and in DictionaryOptimizedAccessor. I guess clearer for code clarity vs in generated code.

soubhik-c: right. did it intentionally..

soubhik-c:
Tried a split mode test but it's not failing without the fix whereas manual testing shows previous revision does have this issue.

I guess SplitSnappyClusterDUnitTest is running embedded and split mode in the same jvms and not exactly doing things what is expected in spit mode. SplitClusterDUnitTest is doing the right thing but because it is in core, cannot share NorthWindDU* helper methods.
IMO, SplitClusterDUnitTest#beforeClass should be moved to SplitClusterDUnitTestBase and SplitSnappyClusterDUnitTest should extend from DistributedTestBase instead of ClusterManagerTestBase, although don't exactly understand why its done the way it is right now.

PR #467, opened 2016-12-14
[SNAP-1252][SNAP-1251] Avoid exchange when join columns are superset of partitioning

sumwale:

- LocalJoin now adjusts its requiredChildDistributions to match that of a child in case
  partial set of join keys match the partitioning of one or both children
- check for subset match order to see if the same can be applied to both children,
  else if both had different matches then choose the child with bigger size as per plan statististics
- also add join column aliases from projections, if any, to OrderlessHashPartitioning

TODO: add similar to SortMergeJoinExec (or its optimized version in SnappyData when it is added)

## Patch testing

precheckin; TPCH Q2, Q5, Q9 plans, NorthWind test

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/37

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
@@ -136,9 +136,9 @@ case class LocalJoin(leftKeys: Seq[Expression],
     }
 
   /**
-   * Return if the partitioning is a subset of given join keys, and if so
-   * then return the subset as well as the indices of subset keys in the
-   * join keys (in order)
+   * Optionally return result if partitioning is a subset of given join keys,
+   * and if so then return the subset as well as the indices of subset keys
+   * in the join keys (in order).
    */
   private def getSubsetAndIndices(part: Partitioning,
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/apache/spark/sql/execution/joins/LocalJoin.scala
```diff
@@ -86,10 +87,69 @@ case class LocalJoin(leftKeys: Seq[Expression],
...
+   * Return if the partitioning is a subset of given join keys, and if so
+   * then return the subset as well as the indices of subset keys in the
+   * join keys (in order)
+   */
+  private def getSubsetAndIndices(part: Partitioning,
+      keys: Seq[Expression]): Option[(Seq[Expression], Seq[Int])] = part match {
+    case e: Expression =>
+      val numColumns = e.children.length
+      if (keys.length > numColumns) {
+        keys.indices.combinations(numColumns).map(s => s.map(keys.apply) -> s)
```
hbhanawat: combinations.. nice API. Pretty interesting way of writing this.

PR #471, opened 2016-12-16
[SNAP-1256] Allow setting of default spark memory manager

hbhanawat:
Added UseSparkMemoryManager property to specify if spark's default memory manager is to be used. We cannot use spark.memory.manager property to specify this as driver would try to load the specified class. So we cannot pass a flag or the fully qualified name of UnifiedMemoryManager (since it is a private class).

### DIFF
--- a/cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
+++ b/cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
@@ -30,6 +30,7 @@ import com.gemstone.gemfire.internal.cache.GemFireCacheImpl
 import com.pivotal.gemfirexd.internal.engine.Misc
 import com.pivotal.gemfirexd.internal.engine.distributed.utils.GemFireXDUtils
 import com.pivotal.gemfirexd.internal.engine.store.ServerGroupUtils
+import io.snappydata.Property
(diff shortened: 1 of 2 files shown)


cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
```diff
@@ -167,10 +168,14 @@ object ExecutorInitiator extends Logging {
                     }
                     // TODO: Hemant: add executor specific properties from local
                     // TODO: conf to this conf that was received from driver.
-
-                    // If memory manager is not set, use Snappy unified memory manager
-                    driverConf.setIfMissing("spark.memory.manager",
-                      SNAPPY_MEMORY_MANAGER)
+                    if (Property.UseSparkMemoryManager.getOption(driverConf).exists(_.toBoolean)) {
```
sumwale: Instead of adding another property, won't it be better to add a token like "default" instead? So in SparkEnv where it checks for this property, it can do like:
```scala
conf.getOption("spark.memory.manager").filterNot(_.equalsIgnoreCase("default")).map(Utils.classForName(_)
...
```

hbhanawat:
I was shying away from changing the spark code but I now remember that spark.memory.manager is a property introduced by us. Will discard this and create a new PR.

PR #472, opened 2016-12-16
--TPCH Changes

KishorBachhav:

 --In Memsql,
     Order table's clustered columnstore key is chnaged from O_CUSTKEY to O_ORDERKEY
     LINEITEM table's clustered columnstore key is chnaged from O_CUSTKEY to O_ORDERKEY
     Q4 and Q21 is changed where l_orderkey is projected instead of *
 --In Snappy,
     Small change at all places in queryExecution, where plan generation condition is added before collecting result
     Q2,Q11,Q17,Q22 were broken in multiple parts. Now original queries are used
 --Script
   Instead of just copying out files, logs files will also be copied to output locations
   While starting lead, locator, server no need to have classpath parameter.

Ran:
  TPCH perf tests

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


hbhanawat:
Looks good. Please merge so that we can run the TPCH tests using these changes

PR #473, opened 2016-12-19
 [SNAP-1260] Miscellaneous plan optimizations

sumwale:

This completes most of the plan inefficiencies I was tracking after analysis of all the TPCH queries. There is still one issue left that shows up in Q2 where pre-broadcast "isCollocated" check does not recognize SnappyHashAggregate output as collocated and thus broadcast gets introduced which causes plan to break in couple of places (in fact whole Q2 query can be run collocated which happens if broadcast is disabled or small).

That change needs to go to the "isCollocated" check in SnappyStrategies.scala but did not do that because the proper way would be to do it after EnsureRequirements like being done in CollapseCollatedPlans instead of specialized checks. However, it is slightly complicated for this release hence skipped for now.

- enhanced DelegateRDD to allow passing explicit preferred locations; use this to wrap
  all children including exchange from the table locations for collocated joins
- added a CollapseCollatedPlans that currently checks if aggregates are on partitioning keys
  (or superset), then avoid exchange as well as collapse the "Partial"+"Final"
  into a single "Complete" aggregate.
- add alias handling in SnappyHashAggregateExec result expressions that may have been added
  implicitly or explicitly and be used in further joins etc
- corrected outputPartitioning for LocalJoin like in SortMergeJoinExec
- updated a number of TPCH queries to use a more optimal join order; updated
  remaining queries to use the original versions (which works for all queries)
- set shuffle partitions explicitly to the table partitions to avoid any exchange
  further down the joins when other operators are mixed in and may lead to intermediate
  results having default shuffle partitions
- removed obsolete and unused doProduce/doConsume (and deps) in LocalJoin

Manually confirmed that no REMOTE iteration logs are now seen with 100GB runs on 5 servers.

## Patch testing

precheckin, Northwind, confirmed TPCH plans

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


sumwale:
> case when someone explicitly sets sql.shuffle.partitions?

No for that both dynamic and exection values are set to -1 and are not used (and super method called).

The purpose of this number was to explicitly set the default shuffle partitions to be the same as number of buckets (the minimum for all the tables). Kept it as separate from other dynamic one since it takes the minimum. Normally the tables will have same number of buckets and in that case default will match buckets.

It helps to avoid exchange for cases where all children have to shuffle but later could be collocated with other children. For example, a non-collocated join or groupBy may lead to exchange of all children using default shuffle partitions. However, further down the child hierarchy this result itself may be collocated with a table but cannot avoid exchange if default shuffle partitions are different from table buckets. Didn't note down the precise queries but saw this in few TPCH plans.

sumwale:
>  (the minimum for all the tables)

Its maximum not minimum.

PR #474, opened 2016-12-19
Added full Result Set validation in NorthWindDUnitTest

swatisawant:

- Added full Result Set validation in NorthWindDUnitTest
- Commented out Q37 for now as its failing with SNAP-1163. Will enable it later on after adding workaround in test

## Patch testing

## ReleaseNotes.txt changes

NA

## Other PRs 

No


ymahajan:
@swatisawant Please add data validation in scalaTest as well.

cluster/src/dunit/scala/org/apache/spark/sql/NorthWindDUnitTest.scala
```diff
@@ -508,128 +532,293 @@ class NorthWindDUnitTest(s: String) extends ClusterManagerTestBase(s) {
...
+  protected def getTempDir(dirName: String): String = {
+    val log: File = new File(".")
+    var dest: String = null
+    dest = log.getCanonicalPath + File.separator + dirName
+    val tempDir: File = new File(dest)
+    if (!tempDir.exists) tempDir.mkdir()
+    return tempDir.getAbsolutePath
+  }
+
+  def assertQueryFullResultSet(snc: SnappyContext, sqlString: String, numRows: Int, queryNum: String, tableType: String, pw: PrintWriter, sqlContext: SQLContext): Any = {
```
sumwale: Break line. Please take care of scalaStyle in our sources too. Currently scalaStyle checks for snappydata modules have to be disabled precisely because a very large number of violations in our code.

cluster/src/dunit/scala/org/apache/spark/sql/NorthWindDUnitTest.scala
```diff
@@ -508,128 +532,293 @@ class NorthWindDUnitTest(s: String) extends ClusterManagerTestBase(s) {
...
+      .map(row => {
+        var str = ""
+        row.toSeq.foreach(e => {
+          if (e != null)
+            str = str + e.toString + ","
+          else
+            str = str + "NULL" + ","
+        })
+        str
+      }).write.format("org.apache.spark.sql.execution.datasources.csv.CSVFileFormat").option("header", false).save(dest)
```
sumwale: Please avoid CSV writer. From what I have seen it is really slow. All you need in this method is df.collect().mkString(",")

cluster/src/dunit/scala/org/apache/spark/sql/NorthWindDUnitTest.scala
```diff
@@ -508,128 +532,293 @@ class NorthWindDUnitTest(s: String) extends ClusterManagerTestBase(s) {
...
+  def writeToFile(df: DataFrame, dest: String, snc: SnappyContext): Unit = {
+    import snc.implicits._
+    df.map(dataTypeConverter)(RowEncoder(df.schema))
+      .map(row => {
+        var str = ""
+        row.toSeq.foreach(e => {
+          if (e != null)
+            str = str + e.toString + ","
+          else
+            str = str + "NULL" + ","
```
sumwale: Adding to string this way for each row is really slow. At the very least should use a StringBuilder. Have seen validation via app taking long time due to this. See comment below.

cluster/src/dunit/scala/org/apache/spark/sql/NorthWindDUnitTest.scala
swatisawant: df.collect() gives OutOfMemroyError for cross join queries which returns 1.8 million records and that's the reason I switched to CSV writer.

sumwale: Ok, please do change to use a StringBuilder above.

PR #475, opened 2016-12-19
Fixes for SNAP-1261 : Lead is alive but servers are dead. UI Dashboard fails to come up.

snappy-sachin:

* Dashboard works fine even if all data servers goes down.

## Patch testing

* Tested Manually.


suranjan:
Looks good.

PR #477, opened 2016-12-20
Snap 1032

rishitesh:

a) Added a new class SnappyTestRunner to easily add command based test cases. submitting jobs, querying from SnappyShell etc.
This class can be extended to write multiple use cases.

b) Scala tests for all the examples in new examples folder, using SnappyTestRunner. As its a scala test it can easily be tested from IDEA also.

c) Moved quickstart test code to this new mode. Removed quickstart test related code from root build.gradle. Those were cluttering the build file.

## Patch testing

Runnig precheckin , which should include snappydata-examples:scalaTest as well.

## ReleaseNotes.txt changes
NA . 

## Other PRs 
NA


examples/src/test/scala/io/snappydata/quickstart/ExampleTestSuite.scala
```diff
@@ -0,0 +1,109 @@
...
+
+  }
+
+  test("SynopsisDataExample") {
+    Job(s"$snappyExamples.SynopsisDataExample",
+      localLead, quickStartJar,
+      Seq(s"data_resource_folder=$snappyHome/quickstart/data"))
+  }
+
+}
```
dshirish: Python how-to example CreateTable.py needs to be added.

Since some of the examples can be run with run-example script and also as a job, may be we should run one or two of those those using both ways just to make sure run-example is not broken for those examples.

rishitesh: Sure will add that.

PR #482, opened 2016-12-21
Add support to include Java Flight Recorder (JFR) parameters and SHUTDOWN_WAIT_TIME_MS parameter

swatisawant:

- Added support to include Java Flight Recorder (JFR) parameters for collecting diagnostic and profiling data while launching server and leader members.
- Added support in framework to include gemfire.CacheServerLauncher.SHUTDOWN_WAIT_TIME_MS parameter while launching lead and server member.

## Patch testing

Verified the changes by running northwind bt

## ReleaseNotes.txt changes

NA

## Other PRs 

No


dtests/src/test/java/io/snappydata/hydra/cluster/SnappyTest.java
```diff
@@ -2285,7 +2287,8 @@ protected void killVM(String vmDir, String clientName, boolean isLead) {
         Log.getLogWriter().info(clientName + " stopped successfully...");
     }
 
-    protected void startVM(String vmDir, String clientName, boolean isLead) {
+    protected void startVM(String vmDir, String clientName, boolean isLead) throws InterruptedException {
+        Thread.sleep(100000);
```
suranjan: What is this change for?

suranjan: Rest looks ok.

swatisawant: We had a discussion about the SNAP-1177 and at that time, you had suggested to make sure of restarting the member after certain delay. That's the reason, I added this delay of 100 secs before calling start for the member which is getting recycled.

suranjan: If we use SHUTDOWN_WAIT_TIME then is it needed?

swatisawant: No. It will also increase my test time. Will remove it and merge the changes.

PR #487, opened 2017-01-03
Fix Backward compatibility with 2.0.0 Spark

hbhanawat:

Snappy 0.7 should work with Spark versions >= 2.0.0 and versions < = 2.0.2 in connector mode. This PR fixes the issues related to that. 
- Changed the spark version for snappy-core to 2.0.0. 
- Copied three classes from Spark to Snappy as there is no compatibility between 2.0.0 and 2.0.2 versions of those classes. RunCommand, DescribeTableCommand, ExecutedCommandExec. 
- Added reflection to create JacksonGenerator object as its apply method has changed between two versions. 
- Added reflection to find out the parameter 'child' for CreateTableUsingAsSelect. 

## Patch testing
 
Clean precheckin.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -37,7 +37,7 @@ import io.snappydata.Constant
 import org.apache.spark.annotation.{DeveloperApi, Experimental}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.scheduler.{SparkListener, SparkListenerApplicationEnd}
-import org.apache.spark.sql.backwardcomp.SnappyExecutedCommandExec
+import org.apache.spark.sql.backwardcomp.ExecutedCommand
 import org.apache.spark.sql.catalyst.analysis.EliminateSubqueryAliases
 import org.apache.spark.sql.catalyst.encoders.RowEncoder
 import org.apache.spark.sql.catalyst.expressions.codegen.CodegenContext
@@ -1528,9 +1528,7 @@ object SnappySession extends Logging {
       case plan => plan
     }
     val (cachedRDD, shuffleDeps, rddId, localCollect) = executedPlan match {
-      case _: ExecutedCommandExec =>
-        throw new EntryExistsException("uncached plan", df) // don't cache
-      case _: SnappyExecutedCommandExec =>
+      case _: ExecutedCommandExec | _: ExecutedCommand =>
         throw new EntryExistsException("uncached plan", df) // don't cache
       case plan: CollectAggregateExec =>
         (null, findShuffleDependencies(plan.childRDD).toArray,
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -37,6 +37,7 @@ import io.snappydata.Constant
 import org.apache.spark.annotation.{DeveloperApi, Experimental}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.scheduler.{SparkListener, SparkListenerApplicationEnd}
+import org.apache.spark.sql.backwardcomp.SnappyExecutedCommandExec
 import org.apache.spark.sql.catalyst.analysis.EliminateSubqueryAliases
 import org.apache.spark.sql.catalyst.encoders.RowEncoder
 import org.apache.spark.sql.catalyst.expressions.codegen.CodegenContext
@@ -1528,6 +1529,8 @@ object SnappySession extends Logging {
     }
     val (cachedRDD, shuffleDeps, rddId, localCollect) = executedPlan match {
       case _: ExecutedCommandExec =>
+        throw new EntryExistsException("uncached plan", df) // don't cache
+      case _: SnappyExecutedCommandExec =>
         throw new EntryExistsException("uncached plan", df) // don't cache
       case plan: CollectAggregateExec =>
         (null, findShuffleDependencies(plan.childRDD).toArray,
--- a/core/src/main/scala/org/apache/spark/sql/backwardcomp/logicalPlans.scala
+++ b/core/src/main/scala/org/apache/spark/sql/backwardcomp/logicalPlans.scala
@@ -1,0 +1,257 @@
+/*
+ * Copyright (c) 2016 SnappyData, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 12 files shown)


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1529,6 +1530,8 @@ object SnappySession extends Logging {
     val (cachedRDD, shuffleDeps, rddId, localCollect) = executedPlan match {
       case _: ExecutedCommandExec =>
         throw new EntryExistsException("uncached plan", df) // don't cache
+      case _: SnappyExecutedCommandExec =>
+        throw new EntryExistsException("uncached plan", df) // don't cache
```
sumwale: Can combine with ExecuteCommandExec case above with an '|'

core/src/main/scala/org/apache/spark/sql/backwardcomp/logicalPlans.scala
```diff
@@ -0,0 +1,257 @@
...
+ * **Copy of DescribeTableCommand to ensure a compatibility between
+ * Spark 2.0.0 and Snappy Spark 2.0.2
+ *
+ * Command that looks like
+ * {{{
+ *   DESCRIBE [EXTENDED|FORMATTED] table_name partitionSpec?;
+ * }}}
+ */
+case class SnappyDescribeTableCommand(
+    table: TableIdentifier,
```
sumwale: Can we avoid "Snappy" prefix for each of the new plan types? Say:

SnappyRunnableCommand => ExecuteCommand
SnappyDescribeTableCommand => DescribeTable

core/src/main/scala/org/apache/spark/sql/backwardcomp/logicalPlans.scala
```diff
@@ -0,0 +1,257 @@
...
+    }
+  }
+
+  private def describeSchema(schema: StructType, buffer: ArrayBuffer[Row]): Unit = {
+    schema.foreach { column =>
+      val comment =
+        if (column.metadata.contains("comment")) column.metadata.getString("comment") else null
+      append(buffer, column.name, column.dataType.simpleString, comment)
+    }
+  }
```
sumwale: Please file a JIRA to enable showing extended properties for SnappyData tables.

core/src/main/scala/org/apache/spark/sql/backwardcomp/sparkPlans.scala
```diff
@@ -0,0 +1,56 @@
...
+ */
+package org.apache.spark.sql.backwardcomp
+
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.catalyst.{CatalystTypeConverters, InternalRow}
+import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.catalyst.plans.QueryPlan
+import org.apache.spark.sql.execution.SparkPlan
+
+private[sql] case class SnappyExecutedCommandExec(cmd: SnappyRunnableCommand) extends SparkPlan {
```
sumwale: Too much Snappy in names. Say: SnappyExecutedCommandExec => ExecutedCommand

core/src/main/scala/org/apache/spark/sql/sources/StoreStrategy.scala
```diff
@@ -32,39 +32,52 @@ object StoreStrategy extends Strategy {
...
+      import scala.reflect.runtime.{universe => ru}
+      import ru._
+      val rm = runtimeMirror(this.getClass.getClassLoader)
+      val pr = rm.reflect(a)
+      // 2.0.0 parameter name: child
+      // 2.0.2 parameter name: query
+      // Use reflection to find the value of parameter
+      val childterm = typeOf[CreateTableUsingAsSelect].members.find(a =>
+        a.fullName.contains("query") || a.fullName.contains("child"))
+      val value = pr.reflectField(childterm.get.asTerm).get
```
sumwale: Instead of reflection a cleaner way would be using Product.productElement (see the patch I had sent):
```scala
    case c@CreateTableUsingAsSelect(tableIdent, provider, partitionCols,
    _, mode, opts, _) =>
      val query = c.productElement(6).asInstanceOf[LogicalPlan]
      // CreateTableUsingSelect is only invoked by DataFrameWriter etc
      // so that should support both builtin and external tables
      ExecuteCommand(CreateMetastoreTableUsingSelect(tableIdent, None,
        None, None, SnappyContext.getProvider(provider, onlyBuiltIn = false),
        temporary = false, partitionCols, mode, opts, query,
        isBuiltIn = false)) :: Nil
```

core/src/main/scala/org/apache/spark/sql/sources/StoreStrategy.scala
```diff
@@ -32,39 +32,52 @@ object StoreStrategy extends Strategy {
...
+      SnappyExecutedCommandExec(drop) :: Nil
 
     case DMLExternalTable(name, storeRelation: LogicalRelation, insertCommand) =>
-      ExecutedCommandExec(ExternalTableDMLCmd(storeRelation, insertCommand)) :: Nil
+      SnappyExecutedCommandExec(ExternalTableDMLCmd(storeRelation, insertCommand)) :: Nil
 
     case PutIntoTable(l@LogicalRelation(t: RowPutRelation, _, _), query) =>
-      ExecutedCommandExec(PutIntoDataSource(l, t, query)) :: Nil
+      SnappyExecutedCommandExec(PutIntoDataSource(l, t, query)) :: Nil
+    case r: SnappyDescribeTableCommand => SnappyExecutedCommandExec(r) :: Nil
```
sumwale: redundant

core/src/test/scala/org/apache/spark/sql/store/SnappyCatalogSuite.scala
```diff
@@ -413,7 +413,10 @@ abstract class CatalogTestUtils {
         CatalogColumn("a", "int"),
         CatalogColumn("b", "string")),
       partitionColumnNames = Seq("a", "b"),
-      bucketColumnNames = Seq("col1"))
+      bucketColumnNames = Seq("col1"),
+      // Pass the properties so that the hack applied in SnappyExternalCatalog.createTable
+      // takes affect
+      properties = Map("spark.sql.sources.provider" -> "column"))
```
sumwale: Why is this required now? What used to happen with 2.0.2 without this?

hbhanawat: Well, I was trying to fix the failures in SnappyCatalogSuite (both with 2.0.0 and 2.0.2) and thought it could be the reason. But it isn't it. But kept it here as it takes it to the correct block which should be tested.

core/src/main/scala/org/apache/spark/sql/backwardcomp/logicalPlans.scala
hbhanawat: SNAP-1294

core/src/main/scala/org/apache/spark/sql/backwardcomp/logicalPlans.scala
hbhanawat: I find Snappy prefix better as it conveys the class from which it was copied immediately. But, I believe it's a matter a perception. Going ahead with what you have suggested.

PR #488, opened 2017-01-04
Add support in framework for starting the starting the job server on random unique port while launching lead member

swatisawant:

- Added support for dynamically creating the random port for jobserver while staring the lead member 
- Added a workaround in north wind tests for SNAP-1210
- Modified existing tests to work with the new changes done for starting the job server on random port instead of 8090

## Patch testing

Verified the changes by running northWind.bt, sample.bt and installJar.bt

## ReleaseNotes.txt changes

NA

## Other PRs 
No


sonalsagarwal:
Looks good to me.

dtests/src/test/scala/io/snappydata/hydra/northwind/NWTestUtil.scala
```diff
@@ -103,6 +68,22 @@ object NWTestUtil {
...
+      .map(row => {
+        val sb = new StringBuilder
+        row.toSeq.foreach(e => {
+          if (e != null)
+            sb.append(e.toString).append(",")
+          else
+            sb.append("NULL").append(",")
+        })
+        sb.toString()
+      }).write.format("org.apache.spark.sql.execution.datasources.csv.CSVFileFormat").option(
```
SachinJanani: I think there was some issue with CSVFileFormat that you discussed and so in NorthWindDunitTest.writeToFile you have used PrintWriter for writing the records.

swatisawant: Yes, using CSVFileFormat makes it very slow which results in increased test time and that's the reason, Sumedh has changed this code in dunit. I will incorporate the same in northwind later on in other PR.

sumwale: @swatisawant @SachinJanani Note that there are two main changes for dunits: a) repartition(1).sortWithinPartitions helps run the query in parallel as we also discussed in the presentation so the same should be used in hydra, b) writing directly instead of CSV writer as you mention

swatisawant: Yes Sumedh,
I am going to add this in northWind tests soon. Just didn't want to delay this PR due to these changes in northWind tests.

PR #496, opened 2017-01-18
Fix for SNAP-1269 to use user specified schema if present in case of …

SachinJanani:
* Fix for SNAP-1269 to use user specified schema if present in case of create table as select otherwise schema of base table will be used

## Patch testing
* Written scala test and verified it 

## ReleaseNotes.txt changes

## Other PRs 
NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -1019,7 +1019,11 @@ class SnappySession(@transient private val sc: SparkContext,
               userSpecifiedSchema = userSpecifiedSchema,
               partitionColumns = partitionColumns,
               options = params).write(mode, data)
-            (r, Some(r.schema))
+            if (None != userSpecifiedSchema) {
+              (r, Some(userSpecifiedSchema.get))
+            } else {
+              (r, Some(r.schema))
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1019,7 +1019,11 @@ class SnappySession(@transient private val sc: SparkContext,
               userSpecifiedSchema = userSpecifiedSchema,
               partitionColumns = partitionColumns,
               options = params).write(mode, data)
-            (r, Some(r.schema))
+            if (None != userSpecifiedSchema) {
+              (r, Some(userSpecifiedSchema.get))
+            } else {
+              (r, Some(r.schema))
+            }
```
sumwale: @SachinJanani Can you explain why is this required (and even seem incorrect for some cases)? The DataSource.resolveRelation will lead to dataSource.createRelation with the userSpecifiedSchema so the schema of the underlying relation should be as provided. As far as I can see the schema of query is not used in this method anywhere.

Even otherwise if the schema of the returned relation is incorrect then that will lead to other problems and just changing the schema entry in Hive catalog will not be a proper fix.

Btw in other note, for something like above we should use scala-like way rather than java-like if-else etc: ```userSpecifiedSchema.getOrElse(Some(r.schema))```

SachinJanani: @sumwale  In this case DataSource.write is used to creating relation which is using schema of query and same schema was being added in catalog which was causing this issuse,hence i added this check.

sumwale: @SachinJanani Right, so invoke createRelation separately as being done in JdbcExtendedUtils.externalResolvedDataSource

PR #498, opened 2017-01-19
Incorporated Review comments related to SNAP-1269

SachinJanani:

* Incorporated Review comments based on previous PR related to SNAP-1269 

## Patch testing
* Unit test passed and precheckin underway

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -51,7 +51,7 @@ import org.apache.spark.sql.execution.columnar.impl.ColumnFormatRelation
 import org.apache.spark.sql.execution.columnar.{ExternalStoreUtils, InMemoryTableScanExec}
 import org.apache.spark.sql.execution.command.ExecutedCommandExec
 import org.apache.spark.sql.execution.datasources.jdbc.JdbcUtils
-import org.apache.spark.sql.execution.datasources.{DataSource, LogicalRelation}
+import org.apache.spark.sql.execution.datasources.{CaseInsensitiveMap, DataSource, LogicalRelation}
 import org.apache.spark.sql.hive.{QualifiedTableName, SnappyStoreHiveCatalog}
 import org.apache.spark.sql.internal.{PreprocessTableInsertOrPut, SnappySessionState, SnappySharedState}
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1014,16 +1014,19 @@ class SnappySession(@transient private val sc: SparkContext,
...
-            if (None != userSpecifiedSchema) {
-              (r, Some(userSpecifiedSchema.get))
-            } else {
-              (r, Some(r.schema))
+              options = params)
+            val clazz: Class[_] = dataSource.providingClass
+            val r = clazz.newInstance() match {
+              case schemaRelationProvider: SchemaRelationProvider =>
+                schemaRelationProvider.createRelation(sqlContext, new CaseInsensitiveMap(params)
+                  , userSpecifiedSchema.getOrElse(data.schema))
```
sumwale: What happens for other cases that do not match SchemaRelationProvider (e.g. column tables)? It will fail with MatchError.

I think there are two options here:

a) Apply new schema to the DataFrame if the two don't match:
```scala
data = userSpecifiedSchema match {
  case Some(s) if (s != data.schema) => session.internalCreateDataFrame(data.queryExecution.toRdd, s)
  case _ => data
}
```
Then go ahead with normal DataSource.write call which will create with updated schema.

b) Alternatively use the SchemaRelationProvider to just do the create, and then let the write go ahead as above. In this case the creation and write will not be atomic in the sense that if write fails then table will not be automatically destroyed (e.g. see last lines in ColumnFormatRelation). This can be handled by checking for "r" against DestroyRelation and invoking its destroy if write fails.

Also discuss with Hemant since he is changing the order of hive catalog entry vs table creation+write.

SachinJanani: Thanks sumedh .I think for Column tables the ColumnarRelationProvider is used which implement SchemaRelationProvider so that will match here but will throw MatchError in other cases. I think option 'a' looks more safe.Will discuss once with Hemant about the ordering

SachinJanani:
Discussed with @hbhanawat about this seems that he has already added a fix (something similar to option 'a') in his PR hence closing this PR

PR #500, opened 2017-01-19
Executorcatalog

hbhanawat:

Before this checkin, the table meta data information on executors was supplied by the StoreInitRDD. This is not correct because if a table is created from SparkApp and then data is inserted into it, it fails to age the data to column buffer as the table's metadata is not available. Also, in HA conditions, this could create problems. The table metadata is anyway available on the executors inside hive catalog. Hence added code to read the hive meta data from the catalog on the executor itself. 

Secondly, cached batch size and usecompression were read from sqlconf. This is incorrect because the behaviour is confusing when the data is being inserted irom a spark app. Hence made these two parameters part of table metadata. They can be specified as part of the table properties. cached batch size if not specified as table properties, we try to read it from spark conf. This was done to make the exisiting test cases work. If it is not part of spark conf then default of 10000  is being used. 

After these two changes, store init rdd was not needed for snappydata data core but it was needed for aqp. So moved it to aqp. 

When a table is created using spark way of saveAsTable instead of createTable, the dd entry was created, data was inserted and then hive catalog entry is created. So, during data insertion the hive meta data was not found on the executors. To fix this , moved the hive catalog registration after the relation is created inside the createRelation method. Also since, createRelation can be called multiple times for a single table, avoided registration in hive catalog multiple times. 

Fixed the tests to set the cached batch sizes in the sparkconf instead of sqlconf. 

## Patch testing

Clean precheckin. Will be adding a test to verify the hive meta data on executors in embedded and connector modes. 

## Other PRs 

https://github.com/SnappyDataInc/snappy-aqp/pull/131
https://github.com/SnappyDataInc/snappy-store/pull/146


core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -152,20 +160,24 @@ object ExternalStoreUtils {
...
-        ServiceUtils.getLocatorJDBCURL(sc) + ";route-query=false"
-      case ExternalEmbeddedMode(_, url) =>
-        Constant.DEFAULT_EMBEDDED_URL + ";host-data=false;" + url
-      case LocalMode(_, url) =>
-        Constant.DEFAULT_EMBEDDED_URL + ';' + url
-      case ExternalClusterMode(_, url) =>
-        throw new AnalysisException("Option 'url' not specified for cluster " +
-            url)
+  def defaultStoreURL(sparkContext: Option[SparkContext]): String = {
+    sparkContext match {
```
rishitesh: What is scenario for SparkContext None ?

cluster/src/dunit/scala/org/apache/spark/sql/CachedBatchScanDUnitTest.scala
```diff
@@ -32,10 +32,9 @@ class CachedBatchScanDUnitTest(s: String) extends ClusterManagerTestBase(s) {
         "UniqueCarrier CHAR(6) NOT NULL"
 
     // reduce the batch size to ensure that multiple are created
-    snc.sql("set spark.sql.inMemoryColumnarStorage.batchSize = 4")
 
     snc.sql(s"create table if not exists airline ($ddlStr) " +
-        s" using column options (Buckets '2')")
+        s" using column options (Buckets '2', COLUMN_BATCH_SIZE '4')")
```
rishitesh: Please open a doc issue for the attribute list to be updated

rishitesh:
While I agree with the approach for reading metastore in executor side , I am not comfortable with moving the registerDataSource/unRegisterDataSource to actual data source classes.This is making an already complex method like createTable() more complex. Also implementing a new data source will be more complex. The writer has to take care of register/unregister and all the exceptional cases. Can't it be handled at framework level. Probably by puttinig in our own strategy to handle "CreateTableUsingAsSelect" .

ymahajan:
There are bunch of changes for catalog in Spark2.1 merge, we need to reconcile with their changes.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
hbhanawat: When this is called on the executor.

hbhanawat:
@rishitesh 

>> I am not comfortable with moving the registerDataSource/unRegisterDataSource to actual data source classes.This is making an already complex method like createTable() more complex. Also implementing a new data source will be more complex. The writer has to take care of register/unregister and all the exceptional cases. 

register/unregister are adding and deleting the information to hive catalog that will be used by the executors. So, I think it gives more flexibility to add relation specific information to the catalog. However, I have not removed the register/unregister functions from the createTable API. They will be called for any new relation that is added. See the usage of SnappyContext.internalTableSources. 


>> Can't it be handled at framework level. Probably by puttinig in our own strategy to handle "CreateTableUsingAsSelect" .

The problem is DataSource API is doing the insert.

PR #504, opened 2017-01-27
changed snappy-shell to snappy

hbhanawat:

snappy-shell is currently being used to start lead,locators and server and it is also used as a sql shell. But its name does not indicate that it is a sql shell. 

So, renamed snappy-shell to snappy. And created a new script snappy-sql that is just a wrapper over 'snappy' shell. So now our documents would tell that to start lead, locator and servers, use bin/snappy. While to start sql shell, run bin/snappy-sql. However, bin/snappy and bin/snappy-sql can be both be used to start components and start the sql shell. 

## Patch testing
manual testing and ran ./gradlew :snappy-examples:scalaTest 
 
## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/148


binsoo:
why? snappy_home is get spark_home

PR #507, opened 2017-02-08
initial implementation of UMM.

rishitesh:
Added a UnifieidMemory manager which accounts for SnappyData artifacts like table, index etc.

All the code which add/delete/update to the RegionMap needs accounting in terms of memory usage. However, memory needs to be acquired before adding to a region map.Most codes are in GemXD layer and can be seen in the linked PR.

Added UI fields to show memory manager status.

## Patch testing

Pre-checkin/Multiple scala and dunit tests added

## ReleaseNotes.txt changes

NA. It will require a doc change though guiding users how to set UMM prameters

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/150


hbhanawat:
Run precheckin

hbhanawat:
Run precheckin

PR #514, opened 2017-02-21
[Snap-1286] Changes for thin client smart connector

dshirish:
Initial implementation for thin client Smart connector

A user can use thin client connector by using conf property "snappydata.Cluster.URL". For example
```
    val spark: SparkSession = SparkSession
        .builder
        .appName("SmartConnectorExample")
       .config("snappydata.Cluster.URL", "localhost:1527")
        .getOrCreate
    val snSession = new SnappySession(spark.sparkContext)

```
 Added a new cluster mode ThinClientConnectorMode in SnappyContext for this purpose to set for thin client Smart connector .  Added a new class SnappyConnectorCatalog (that extends the existing SnappyStoreHiveCatalog) and overrides cachedDataSourceTables . This is used when cluster mode is  ThinClientConnectorMode. This will get Hive table object and associated metadata- bucket count, partitioning columns, index columns, bucket to server or replica to server mapping by executing a system proc sys.GET_TABLE_METADATA. For SmartConnectorRowRDD / SmartConnectorColumnRDD, partitioning info is picked from the SnappyConnectorCatalog.cachedDataSourceTables (earlier peer connector could directly get this from region). Similar changes done for other info such as bucket count, index columns etc.

Added a new helper class SmartConnectorHelper,  that is used to forward operations such as- table create/drop, index create/drop, getting metadata etc to embedded cluster.  Currently in various call such as SnappySession.createTable checking for the cluster mode and if it is ThinClientConnectorMode forwarding the calls to SnappyConnectorCatalog to be executed using system procedures.

Added ThinConnectorSplitClusterDUnitTest and ThinConnectorSnappyDUnitTest that extend existing SplitClusterDUnitTest and SplitSnappyClusterDUnitTest. Added tests for CTAS, index create/drop  thru  connector.

PR for store side changes at-
https://github.com/SnappyDataInc/snappy-store/pull/151

## Patch testing

## ReleaseNotes.txt changes
Release notes and docs both will need update

## Other PRs 


kneeraj:
I think the two pending items stated in the description related to udfs and versioning is also done. If so please edit it.

dshirish:
Updated the description. Version check is done in SET_BUCKETS_FOR_LOCAL_EXECUTION but will change to use io.snappydata.thrift.internal.ClientStatement to send version as done in the commented out code part io.snappydata.impl.SparkShellRDDHelper#executeQuery on master  (currently that part is commented on master due to some bug). This will ensure that version part works for replicated table.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
```diff
@@ -174,6 +180,94 @@ object StoreCallbacksImpl extends StoreCallbacks with Logging with Serializable
     SnappyStoreHiveCatalog.registerRelationDestroy()
   }
 
+  override def performConnectorOp(ctx: Object): Unit = {
+
+    val context = ctx.asInstanceOf[LeadNodeSmartConnectorOpContext]
+
+    context.getType match {
+      case LeadNodeSmartConnectorOpContext.OpType.CREATE_TABLE =>
```
rishitesh: Can we refactor these cases to methods. If we add further operations like (alter table etc ) this method will become big.

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatRelation.scala
```diff
@@ -82,32 +83,19 @@ class RowFormatRelation(
...
-          cols += baseColumns(pkCols(0) - 1)
-        }
-      }
+    clusterMode match {
+      case ThinClientConnectorMode(_, _) =>
+        cols ++= relInfo.indexCols
+
+      case _ =>
+        val indexCols = new Array[String](1)
+        GfxdSystemProcedures.getIndexColumns(indexCols, region)
```
rishitesh: In embedded mode , we can not the earlier approach to get indexcols?

core/src/main/scala/org/apache/spark/sql/SmartConnectorHelper.scala
```diff
@@ -0,0 +1,314 @@
...
+      tableIdent: QualifiedTableName,
+      provider: String,
+      userSpecifiedSchema: Option[StructType],
+      schemaDDL: Option[String],
+      partitionColumns: Array[String],
+      mode: SaveMode,
+      options: Map[String, String],
+      query: LogicalPlan,
+      isBuiltIn: Boolean): LogicalPlan = {
+    throw new AnalysisException("Not yet implemented")
```
dshirish: Not needed method will remove it.

core/src/main/scala/org/apache/spark/sql/SmartConnectorHelper.scala
```diff
@@ -0,0 +1,314 @@
...
+    getMetaDataStmt.registerOutParameter(8, java.sql.Types.VARCHAR) /*primary key columns*/
+    getMetaDataStmt.execute
+  }
+
+  def executeCreateUDFStatement(db: String, functionName: String,
+      className: String, funcResources: Array[(String, String)]): Unit = {
+    createUDFStmt.setString(1, db)
+    createUDFStmt.setString(2, functionName)
+    createUDFStmt.setString(3, className)
+    createUDFStmt.setBlob(4, SmartConnectorHelper.getBlob(funcResources, conn))
```
dshirish: Done. Used string for path.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -232,6 +232,7 @@ private[sql] final case class ColumnTableScan(
 
   private lazy val otherRDDsPartitionIndex = rdd.getNumPartitions
 
+
```
dshirish: Ok

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
dshirish: Yes, on my todo list. Will refactor possibly after merge

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatRelation.scala
dshirish: The approach is as earlier. Just moved the code to one common method than can be used both for connector or embedded mode.

core/src/main/scala/org/apache/spark/sql/SmartConnectorHelper.scala
```diff
@@ -0,0 +1,314 @@
...
+    createSnappyIdxStmt.setString(1, indexIdent.schemaName + "." + indexIdent.table)
+    createSnappyIdxStmt.setString(2, tableIdent.schemaName + "." + tableIdent.table)
+    createSnappyIdxStmt.setBlob(3, SmartConnectorHelper.getBlob(indexColumns, conn))
+    createSnappyIdxStmt.setBlob(4, SmartConnectorHelper.getBlob(options, conn))
+    createSnappyIdxStmt.execute()
+  }
+
+  def dropIndex(indexName: QualifiedTableName, ifExists: Boolean): Unit = {
+    runStmtWithExceptionHandling(executeDropIndexStmt(indexName, ifExists))
+    SnappyStoreHiveCatalog.registerRelationDestroy()
```
dshirish: As discussed this is just incrementing the relation destroy version.

PR #515, opened 2017-02-22
Hydra test coverage for distributed indexing

swatisawant:

 - Added support for new parameters(resultCollection, warmUpIterations, actualRuns) in  distibuted join test
 - Added a Snappy job which will do the result validation as well as performance benchmark for TPCH queries with and without DI
- Added a hydra test for testing distributed indexing feature
- Modified the test conf to include instll-jar task for loading required jars prior to running the job
- Fixed an issue where in install jar was not working when the jar containing the job class was already added in classpath of server
- Fixed an issue recently introduced due to modification done in TPCH_Snappy class for query numbers
- Set spark.sql.crossJoin.enabled = true property required for executing few cross join queries
- Added seperate tests for result validation and performance comparision for TPCH queries with and without DI
- Added support for redirecting performance comparision related logs to user specified output file
- Fixed issue(SNAP-1290) where in hydra takes double the time to execute a job

## Patch testing

Verified the changes by running distIndex.bt

## ReleaseNotes.txt changes

NA

## Other PRs 

No


spillai14:
Changes related to SNAP-1290 looks good.

sonalsagarwal:
Looks good.

soubhik-c:
LGTM, only thing is we should add verification code that ensures expected table/index is really being picked up. Towards that one can reuse CreateIndexTest.verifyIndex function which takes a df and expected list of indexes and tables.

PR #517, opened 2017-02-28
Snap 1346

kneeraj:

Tokenization work. Brief description of the changes.
- Added code in SnappyParser to identify literals in WHERE clause of select queries only.
- Tracked the literal number of the literals and replaced the literal with a ParamLiteral object in the logical plan which tells position as well as encapsulates the exact literal and its type.
- Caching the generated plan against (session, logical plan) instead of (session, sqlText)
The equality and hashcode of ParamLiteral is written such that they ignore exact literal value and just uses the position determined in the parser and the data type of the literal.
- Kept a handle of the references object in a new CachedPlanHelperExec plan. Injected this plan just below the WholeStageCodegenExec plan.
- The doProduce of the ParamLiteral injects LiteralValue ( new holder type object ) in the references of the GeneratedIterator. The LiteralValue keeps the position and the datatype also along the value.
- On re-execution the cache is looked up and if an exisiting plan is obtained then the constant values from the new logical plan is substitued in the LiteralValue's value field only.

## Patch testing

Ran precheckin. Almost clean. Known failures are coming. After Sumedh's checkin they will be good. Verified the perf and working through a couple of unit tests addded in CoulmnTableTest. Adding more tests.

## ReleaseNotes.txt changes

## Other PRs 


kneeraj:
Kishore will notify you as soon as this is done. Yes I am adding more tests and a test hook in snappysession to peek into the exact cache. BTW that test was for performance verification.

core/src/main/scala/org/apache/spark/sql/CachedPlanKey.scala
```diff
@@ -0,0 +1,86 @@
+/*
```
vivekwiz: Seems this file is redundant.

kneeraj: Yes. Using the CachedKey case class which I have defined in the SnappySession only. Will remove this.

core/src/test/scala/org/apache/spark/sql/store/TokenizationTest.scala
```diff
@@ -0,0 +1,252 @@
...
+    val x =((((((ints, longs).zipped.toArray, floats).zipped.toArray,
+      decimals).zipped.toArray, strs).zipped.toArray, dates).zipped.toArray, tstmps).zipped.toArray
+    val data = x map { case ((((((i, l), f), d), s), dt), ts) =>
+      (i, l, f, d, s, dt, ts)
+    }
+
+    val rdd = sc.parallelize(data, data.length)
+      .map(s => (s._1, s._2, s._3, s._4, s._5, s._6, s._7))
+    val dataDF = snc.createDataFrame(rdd)
+
```
soubhik-c: I think the test is not verifying tokenization of all data types because date is represented as Int ? ts is represented as Long ? Also, the code can be much simpler like 
```java
// code is compatible with jdk8 & above
    val data = (0 to numRows).zipWithIndex.map { case (_, i) =>
      val ins = Instant.now()
      (i, 1L * 1000 * i, 0.1f * i, 0.1d * i, s"abc$i",
          ins.toEpochMilli + i, ins.toString)
    }

    val rdd = sc.parallelize(data, data.length)
      .map(s => (s._1, s._2, s._3, s._4, s._5, s._6, s._7))
    val dataDF = snc.createDataFrame(rdd)

    snc.sql(s"Drop Table if exists $name")
    snc.sql(s"Create Table $name (a INT, b Long, c Float, d Double, s String, dt DATE, ts TIMESTAMP)"
        + "using column options()")
```
note: dt and ts types in the table is changed and insert ops is expected to auto typecast the input values.
I haven't considered timezone implications in the above code.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -663,7 +673,7 @@ class SnappyParser(session: SnappySession)
     SELECT ~ (DISTINCT ~> trueFn).? ~
     (namedExpression + commaSep) ~
     (FROM ~ ws ~ relations).? ~
-    (WHERE ~ expression).? ~
+    (WHERE ~ (test(setTokenize) ~ expression ~ test(setNoTokenize))).? ~
```
kneeraj: LGTM - I think means "Looks good to merge" :-)

Initially tokenization was working on all parts of the query, I mean the parsing was all proper and all the literals were getting properly captured. Generated code was also not the problem. The problem was coming in the analyzer phase where it was not able to understand the new node type and expected very definite things against certain clauses. Limit anything other than integer against a limit clause will fail. Since we did not want to change the analyzer phase code so reduced the scope to WHERE clause of queries -- nested to any level ( should cover the majority use cases )

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -742,9 +752,38 @@ class SnappyParser(session: SnappySession)
...
+
+  private def setNoSelect: Boolean = {
+    isselect = false
+    true
+  }
+
   override protected def start: Rule1[LogicalPlan] = rule {
-    query.named("select") | insert | put | dmlOperation | ctes |
-        ddl | set | cache | uncache | desc
+    (test(setSelect) ~ query.named("select")) | (test(setNoSelect) ~ (insert | put | dmlOperation | ctes |
```
kneeraj: We just need to align the options in the start method of of the parser. We can do it easily as far as parsing is considered. Tokenizing put and other dmls will, as you say, might come later if at all.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1613,9 +1618,47 @@ object SnappySession extends Logging {
...
+  class CachedKey(val session: SnappySession, val lp: LogicalPlan, val sqlText: String, val hintHashcode: Int) {
+
+    override def hashCode(): Int = {
+      (session, lp, hintHashcode).hashCode()
+    }
+
+    override def equals(obj: Any): Boolean = {
+      obj match {
+        case x: CachedKey => {
+          (x.session, x.lp, hintHashcode).equals(session, lp, hintHashcode)
```
kneeraj: yes. Will change. But an innoucuous mistake ..

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1613,9 +1618,47 @@ object SnappySession extends Logging {
     }
   }
 
+  class CachedKey(val session: SnappySession, val lp: LogicalPlan, val sqlText: String, val hintHashcode: Int) {
```
kneeraj: Will fix it.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1659,7 +1704,7 @@ object SnappySession extends Logging {
     val iter = planCache.asMap().keySet().iterator()
     while (iter.hasNext) {
       val item = iter.next()
-      if (item._1.id == sessionId) {
+      if (item.asInstanceOf[CachedKey].session.id == sessionId) {
```
kneeraj: Yes.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
sumwale: @kneeraj I still do think that some of the literals in projections should be replaced. Can you tell the analyzer exception with that in more detail? Perhaps we will have to make it selective as mentioned by Soubhik.

@soubhik-c Constants in casts are digits and not literals. I think substr should be fine with replacement since it already accepts a generic expression (and ParamLiteral is just another type of it). Can you list down other cases you remember from gemxd work?

@hbhanawat and others: this was the primary reason I was not inclined to throw away our parser and switch to Spark's parser. However, it also means that it will be much more difficult to move to Spark's parser once this is merged. To do the move we will need to rewrite the tokenization perhaps with hand-crafted lexer/parser as was done for gemxd (and which caused lots of trouble since it had very simplistic state machine for semantics). Without that the performance of Spark's parser is not upto mark (in the range of ~0.2-0.4ms from what I remember as opposed to ~0.02-0.05ms for ours). We will have to compare that work and the additional DDL etc clauses vs keeping track of features being added to Spark's parser and add those to ours.

sumwale: Instead of "test" for setting the values, a cleaner way will be to change those methods to be Rule0 that will do the setting in its action like:

```scala
protected final def enableTokenization: Rule0 = rule {
  MATCH ~> {() => ...}
}
```

Then you can just use those as regular rules without the artificial test().

sumwale:
Some quick initial comments. Will go through more later today.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
sumwale: GROUP BY is not important, but tokenization for HAVING will be preferable

sumwale: Perhaps all-caps names for the rules like TOKENIZE_START and TOKENIZE_END to make them stand out.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
sumwale: Good to see Soubhik and others being forced to be careful about formatting by Idea :-)

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1629,9 +1637,47 @@ object SnappySession extends Logging {
...
+        case e: Exists =>
+          e.copy(exprId = ExprId(0))
+        case l: ListQuery =>
+          l.copy(exprId = ExprId(0))
+        case p: PredicateSubquery =>
+          p.copy(exprId = ExprId(0))
+        case a: AttributeReference =>
+          AttributeReference(a.name, a.dataType, a.nullable)(exprId = ExprId(0))
+        case a: Alias =>
+          Alias(a.child, a.name)(exprId = ExprId(0))
```
sumwale: add AggregateExpression

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
kneeraj: Yes this is clean. I was not happy with this way either but didn't know about MATCH special rule.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: yes.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1605,6 +1608,8 @@ object SnappySession extends Logging {
     CacheBuilder.newBuilder().maximumSize(300).build(loader)
   }
 
+  def getPlanCache = planCache
```
kneeraj: This is a test hook for cache content validations. User should never use it.

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
```diff
@@ -0,0 +1,168 @@
...
+
+import java.util.Objects
+
+import com.esotericsoftware.kryo.io.{Input, Output}
+import com.esotericsoftware.kryo.{Kryo, KryoSerializable}
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.codegen.{CodegenContext, ExprCode}
+import org.apache.spark.sql.types._
+
+case class ParamLiteral(l: Literal, pos: Int) extends LeafExpression {
```
soubhik-c: I think this should inherit from Literal instaed of LeafExpression so that rest of the spark code doing case match on Literal won't stop matching. One such sideeffect I am observing with this change is filter push down. I suppose we don't have enough test validating filters are practically pushed down.

effected code snippet
```scala
DataSourceStrategy.scala
protected[sql] def translateFilter(predicate: Expression): Option[Filter] = {
    predicate match {
      case expressions.EqualTo(a: Attribute, Literal(v, t)) =>
        Some(sources.EqualTo(a.name, convertToScala(v, t)))
      case expressions.EqualTo(Literal(v, t), a: Attribute) =>
        Some(sources.EqualTo(a.name, convertToScala(v, t)))
 . . . 
```
used in StoreDataSourceStragy.scala
```scala
  private def pruneFilterProjectRaw(
      relation: LogicalRelation,
      projects: Seq[NamedExpression],
      filterPredicates: Seq[Expression],
      numBuckets: Int,
      partitionColumns: Seq[String],
      scanBuilder: (Seq[Attribute], Seq[Expression], Seq[Filter]) =>
          (RDD[Any], Seq[RDD[InternalRow]])) = {
 . . . 
    val (unhandledPredicates, pushedFilters) =
      selectFilters(relation.relation, candidatePredicates)
 . . . 
```
Above code now no more matches the constant in the where clause for a query like below
```sql
    snc.sql("select * from orders where o_orderkey = 1 ").collect().foreach(println)
    snc.sql("select * from orders where o_orderkey = 2 ").collect().foreach(println)
    snc.sql("select * from orders where o_orderkey = 3 ").collect().foreach(println)
```

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
soubhik-c: Doesn't matter, with proper codestyle settings this will flag with red color in the editor.

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
```diff
@@ -0,0 +1,168 @@
...
+        val valueTerm = ctx.freshName("value")
+        ev.copy(
+          s"""
+             |final boolean $isNull = $valueRef.value() == null;
+             |final short $valueTerm = $isNull ? ${ctx.defaultValue(dataType)}
+             |    : ((Short)$valueRef.value()).shortValue();
+           """.stripMargin, isNull, valueTerm)
+      case IntegerType | DateType =>
+        val isNull = ctx.freshName("isNull")
+        assert(value.isInstanceOf[Int], s"unexpected type $dataType instead of DateType or IntegerType")
```
soubhik-c: can be rewritten as
```scala
      case t@(IntegerType | DateType) =>
        val isNull = ctx.freshName("isNull")
        assert(value.isInstanceOf[Int], s"unexpected type $dataType instead of $t")
```

PR #522, opened 2017-03-04
If the ColumnBatchBuffer created is for sample table insertion, then …

ahshahid:
…skipping the code of intermediate CachedBatch creation during processing of the rows, due to size limit violation
From the code of sampling , it appears to me that at the end of  folding of each segment of the reservoir, a new column batch is being created.  So the only thing needed is that during the processing of the segment , a new CachedBatch should not get created. For that setting a Flag in ColumnBatchBuffer which if true, will by pass the code of storage of the StoreColumnBatch on violation of the size . This flag is set to true in the ColumnFormatSamplingRelation.
 
The pull request in AQP is 
[AQP-273](https://github.com/SnappyDataInc/snappy-aqp/pull/140)

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
@@ -61,6 +61,8 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
 
   @transient private var batchBucketIdTerm: Option[String] = None
 
+  @transient private var isSampleDataTerm: Option[String] = None
+
   def columnBatchSize: Int = batchParams._1
 
   def columnMaxDeltaRows: Int = batchParams._2
@@ -110,6 +112,7 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
              |}
           """.stripMargin)
         batchBucketIdTerm = Some(c.bucketIdTerm)
+        isSampleDataTerm = Some(c.isSampleDataTerm)
       case _ =>
     }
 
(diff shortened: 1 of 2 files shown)


ahshahid:
The alternative fix  is to just set the column batch size to -1 for sampled relation as Sumedh suggested.
The patch for that would be 
[patch.txt](https://github.com/SnappyDataInc/snappydata/files/818337/patch.txt)

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -264,7 +267,8 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
     }
     storeColumnBatchArgs = s"$batchSizeTerm, ${batchFunctionCall.toString()}"
     s"""
-       |if ($columnBatchSize > 0 && ($batchSizeTerm & $checkMask) == 0 &&
+       |if ( ${isSampleDataTerm.map(x => s"!$x && ").getOrElse("")}$columnBatchSize > 0 &&
```
sumwale: Can you split out this separately for clarity:

```
val checkReservoirMid = isSampleDataTerm.map(x => s"!$x && ").getOrElse("")

if (${checkReservoirMid}$columnBatchSize > 0 &&
```

sumwale:
@ahshahid Looks like you reverted the change with the last commit. Do you want to close this PR then? I think your previous solution is superior to change columnBatchSize=-1 for reasons we discussed on the email (ColumnBatch can become too big and will not even break across buckets etc).

ahshahid:
There will be no change in ColumnInsertExec etc . The only change will be in ColumnSamplingFormatRelation to pass batch size as -1

ahshahid:
I verified the code of append & also test, a cached batch in any case is getting created for each strata & it remains so even if batch size is -1

sumwale:
Ok, then this PR is not required and can be closed.

PR #532, opened 2017-03-11
[SNAP-1388] Update and delete plans with implementation for row tables

sumwale:

PR still in testing but created for early review.

- New Update and Delete LogicalPlans which are introduced by the parser for
  corresponding "update ... where <query>" and "delete ... where <query>" plans
- New MutableRelation interface that returns:
  - the additional key columns to be projected for firing the underlying update/delete
    which will use just the provided key columns that should be unique for each qualified row
  - SparkPlans for the update/delete operations
- implementation of the above for row tables in base JDBCMutableRelation
- some refactoring in TableInsertExec to split out CodegenSupportOnExecutor

TODO:
 - pass through the same connection as used for scan to update too if possible
 - API support equivalent to that in parser
 - new tests for some complex update/delete operations with subqueries etc which fail in GemXD

Was also adding the "INSERT INTO VALUES ..." support to parser but unfortunately the plans have
only been added to Spark 2.0.2 and do not exist in 2.0.0. If we decide on dropping support for 2.0.0
then can be enabled and not have to go through GemXD (with support for generic expressions apart
from VALUES)

## Patch testing

ongoing

## ReleaseNotes.txt changes

Document Update/Delete support in Spark allowing for the full range of SQL supported by Spark itself
(row tables for a start but soon for column tables)

## Other PRs 

NA


core/src/main/scala/org/apache/spark/sql/execution/row/RowDeleteExec.scala
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.spark.sql.sources.ConnectionProperties
+import org.apache.spark.sql.store.CodeGeneration
+import org.apache.spark.sql.types.{StructField, StructType}
+
+/**
+ * Generated code plan for deletes in a row table.
+ */
+case class RowDeleteExec(child: SparkPlan, resolvedName: String,
+    tableSchema: StructType, keyColumns: Seq[Attribute],
+    connProps: ConnectionProperties, onExecutor: Boolean)
```
hbhanawat: When will the RowDeleteExec and RowUpdateExec be called on executor? Do you plan to wire bulk updates and deletes in gemxd side to these?

core/src/main/scala/org/apache/spark/sql/execution/row/RowDeleteExec.scala
```diff
@@ -0,0 +1,103 @@
...
+    ctx.addMutableState("long", result, s"$result = -1L;")
+    stmt = ctx.freshName("statement")
+    batchCount = ctx.freshName("batchCount")
+    val childProduce = doChildProduce(ctx)
+    s"""
+       |if ($result >= 0L) return;
+       |$initCode
+       |try {
+       |  final java.sql.PreparedStatement $stmt = $connTerm.prepareStatement(
+       |    "${s"DELETE FROM $resolvedName WHERE $whereClauseString"}");
```
hbhanawat: I don't think I understand this code. The whereClauseString is just equality comparison. Why is a prepared statement with batching needed? 

I mean we are delegating deletes and updates to gemxd. I don't see any scan happening in snappy side of the code. What is the role of generated code? Same for RowDeleteExec.

core/src/main/scala/org/apache/spark/sql/execution/row/RowDeleteExec.scala
sumwale: Already wired -- see SnappyParser changes which allow execution of "update ...", "delete ..." statements directly using Snappy plan. Earlier these would just form a statement on lead and execute via JDBC for row tables (and fail for others).

core/src/main/scala/org/apache/spark/sql/execution/row/RowDeleteExec.scala
sumwale: > I don't think I understand this code. The whereClauseString is just equality comparison. Why is a prepared statement with batching needed?

Its equality comparison on only the key columns which is row-by-row updates. Given any arbitrary bulk delete/update string, the Snappy parser will transform it to a select+RowUpdateExec/DeleteExec plan where the update/delete forms row-wise updates using the primary key columns for each row as returned by the select plan (i.e. its child).

> I mean we are delegating deletes and updates to gemxd.

Changed to PK operations on gemxd. We could do direct region put/delete too if required but this was a bit easier.

> I don't see any scan happening in snappy side of the code. What is the role of generated code? Same for RowDeleteExec.

Scan is the child of this plan. SnappyParser will transform and "update ... query" into a RowUpdateExec plan  on top of query plan like usual. Same for RowDeleteExec. For column tables, there will be new implementations of these two plan in terms of what outlined in the spec.

We also need to add equivalent API support but need to discuss well first. I didn't think it important for GA so skipped it for now. Its not easy to express it in terms of DataFrameWriter -- perhaps introducing new "UpdateRow" and "DeletedRow" then a Dataset of those, and new writers like DataUpdateFrameWriter, DataDeleteFrameWriter?

core/src/main/scala/org/apache/spark/sql/execution/row/RowDeleteExec.scala
hbhanawat: That I got. What I don't understand is the use of CodegenSupportOnExecutor. I could see onExecutor being passed as false always. You used this for ColumnInsertExec to do generate code on executor for column batch insertion. Why is it needed here.

core/src/main/scala/org/apache/spark/sql/execution/row/RowDeleteExec.scala
hbhanawat: I will take a re-look with this information. 

Regarding the API, I don't think I have any great ideas. Using DataFrameWriter is going to be tricky - what would be the save mode, what if user persists it to a different data source. Can't we just have a Snappy specific API with no relation to other data sources?

core/src/main/scala/org/apache/spark/sql/execution/row/RowDeleteExec.scala
sumwale: Oh, missed the question. Added the "onExecutor" flag so that it can be used to generate local plans in future. For example smart connector mode or JDBC single-hop can push down filters which can be applied using such plans especially for column tables.

core/src/main/scala/org/apache/spark/sql/execution/row/RowDeleteExec.scala
sumwale: To clarify: a simple update/delete API should not be difficult and can be provided in an extension of Dataset or in SnappySession as an operator like: `Dataset.update(exprs: (String, String)*)` much like `Dataset.agg`. So then it can be used like other Dataset APIs e.g: `ds.select(..).filter(...).update("date" -> "CURRENT_DATE", "counter" -> "counter + 1)`
For GA this might suffice.

The use-case Soubhik and myself have discussed is to be able to bulk import operations from external source like CDC. Each row in the CDC view will provide information on the key, update/delete operation, changes etc. If we could provide an API similar to a writer for such cases then it will be great else it has to form update/delete operations row by row. In general, support for batch updates/deletes. JDBC uses the batch statement/preparedStatement API but something similar to those will be quite odd in the Spark APIs to be forming strings for each row.

rishitesh: Update & delete are kind of final things one can do with a dataset, just like insertInto(). Hence IMO these APIs should be part of DataFrameWriter or similar classes. I liked the idea of DataFrameUpdateWriter or DataFrameDeleteWriter.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -184,10 +184,44 @@ class SnappySessionState(snappySession: SnappySession)
...
+
+    private def getKeyAttributes(table: LogicalPlan,
+        child: LogicalPlan, plan: LogicalPlan): Seq[NamedExpression] = {
+      val keyColumns = table.collectFirst {
+        case LogicalRelation(mutable: MutableRelation, _, _) =>
+          val ks = mutable.getKeyColumns
+          if (ks.isEmpty) throw new AnalysisException(
+            s"Empty key columns for update/delete on $mutable")
+          ks
+      }.getOrElse(throw new AnalysisException(
```
rishitesh: Can you please elaborate the usage of keyColumns in the scheme of things. Is it that we would support bulk update & delete only for tables having a primary key? Or I am missing something?

core/src/main/scala/org/apache/spark/sql/execution/row/RowDeleteExec.scala
sumwale: Are you saying something like df.write.update()? What does "updating" a DataFrame into a table mean? An insertInto or putInto of a DataFrame makes sense but not update. This makes sense as a Dataset API.

A deleteFrom like what has been added might make sense saying "that delete all rows of given Dataset from target table". However, as of now it is not of much use except for CDC kind of scenario since it requires all the columns of source to match target. This is an unrealistic expectation and needs to be fixed to allow for differing columns and a way to provide column mapping from source to target.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
sumwale: "Key" means a unique way to quickly get to the required rows.

For row tables it will be either the primary key, or the internal generated long (a ROWID) -- essentially the corresponding region key, or a "RowLocation" if you may. We cannot be having a case where every row update/delete leads to another full table scan. The idea being that during the corresponding query the rows have already been located and there is a unique handle to the row which delete/update can use to quickly get to the row. The current "deleteFrom" implementation suffers from this weakness and will be unusable when table has no primary key.

Likewise for column tables, the "key" will be the batchUUID+position to uniquely locate the row to be updated/deleted. In case there is no exchange/broadcast in the query portion, then this can be further optimized to directly pick up the handle of the DeltaColumn being currently scanned.

core/src/main/scala/org/apache/spark/sql/execution/row/RowDeleteExec.scala
sumwale: To add to the "deleteFrom" comment:

You are probably confused between two completely different use-cases. One is where one would want to say "delete all rows from target table matching given input obtained from some source". This is deleteFrom which is more for CDC-like cases. The more common cases are different:

"update all rows in the table matching ABC criteria to change to XYZ"

"delete all rows from a table matching ABC criteria"

I don't see any clean way to express 1 using writer-like APIs. One could possibly express 2 as: snappy.table("table1").filter(ABC)....write.deleteFrom("table1") but this is:

a) quite convoluted

b) is in-congruent with the corresponding SQL: delete from table1 where ABC

c) is very inefficient when table has no primary key

d) is inefficient even if there is a primary key since the corresponding deletes are row-wise that has to check all the columns of the table like: "delete from table1 where col1=? and col2=? ..."

rishitesh: @sumwale 
The use cases mentioned by you  are
"update all rows in the table matching ABC criteria to change to XYZ"
"delete all rows from a table matching ABC criteria"

We already have SnappySession.update() & delete(). IMO those APIs should be enough for such use cases. ( we may change their underlying execution mechanism) 

What we were lacking is if each row's criteria is different from each other e.g. CDC case or to generalize, delete using a dataset as criteria. This also has been addressed to some extent by deleteFrom() even if it is inefficient.  

Regarding inefficiencies  won't that be documentation issue to advise users to use 
dataSet.select("primary_key_col").write.deleteFrom() rather than using the complete dataset. It would be same if someone fires delete from table where col1 = col2 = ....etc..

sumwale: > We already have SnappySession.update() & delete(). IMO those APIs should be enough for such use cases. ( we may change their underlying execution mechanism)

No those are just hacks that were added in demo release for row tables. They are just a partial break-down of SQL strings and an incomplete one at that since it takes only a WHERE clause and not a general query (e.g. sub-query or whatever all of which is cleanly done with this approach). It is completely out of place with the rest of Dataset API. This is the correct way to do it and those old APIs will be deprecated/removed.

> What we were lacking is if each row's criteria is different from each other e.g. CDC case or to generalize, delete using a dataset as criteria.

Yes, mentioned that for deleteFrom. Can you clarify how you are thinking similar can apply to update? I had earlier proposed an Dataset[UpdatableRow] kind of API for applying updates but now I think that the only case where this is useful is a PUT kind of scenario where only some columns are being put so should be subsumed into the putInto API.

> Regarding inefficiencies won't that be documentation issue to advise users to use
dataSet.select("primary_key_col")

Sure for row tables but doesn't apply for column tables. Maybe when we have indexes. Even so that is different from the use of keyColumns above which is an internal SparkPlan-level requirement that says: "update/delete is only possible if the plan knows how to quickly get to a row that was selected from itself"

sumwale:
closing this since it has been subsumed by https://github.com/SnappyDataInc/snappydata/pull/747

PR #544, opened 2017-03-22
possible fix for SNAP-1414. passing a reasonable initial size for the…

ahshahid:
… encoder term

I do not fully understand the code , but looks like the initial buffer size of the encoder is determined using columnBatchSize, which we are now passing -1 for sample creation case, which causes index out of bounds. So the change is to use default column batch size to determine defaultBatchTermSize, in case column batch size is -1.

Reviving this PR. As mentioned before, I do not understand the code,  but looks like given an initial size,  it is able to expand, so initial size need not be exact ( which will be the case of sample based buffers which do not have a defined batch size ), so using default column batch size in that case should suffice.
Sumedh/ Vivek pls review the change to bring closure ( one way or the other)

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -156,7 +157,8 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
        |  int $defaultRowSize = 0;
        |  ${declarations.mkString("\n")}
        |  $defaultBatchSizeTerm = Math.max(
-       |    (${math.abs(columnBatchSize)} - 8) / $defaultRowSize, 16);
+       |    (${math.abs( if (columnBatchSize > 0) columnBatchSize else Property.ColumnBatchSize.
+            defaultValue.get)} - 8) / $defaultRowSize, 16);
```
vivekwiz: Can you please verify from where columnBatchSize is being set to -1?

Two possible places are in ColumnFormatSamplingRelation:
      val buffer = batchCreator.createColumnBatchBuffer(
        // externalTableMetaData.columnBatchSize,
        // Fix for AQP-273, for sample table, the strata needs to remain as a single entry
        -1,

If so, good to change that code.

vivekwiz: Had a relook at your comments and code and this is what I think:

Lets pass an additional variable defaultBatchSizeTerm right through ColumnFormatSamplingRelation to ColumnInsertExec.

ahshahid:
I am a bit reluctant passing any additional parameter, as I do not
understand that code thoroughly. My suggestion is that I will checkin the
code with the current change, & let you in conjunction with Sumedh change
it further if you want to. As batch size term is calc. inside based on some
logic, passing it from outside may not look nice.

Regards
Asif

vivekwiz:
You may choose to go ahead with what seems best for you considering all options.

PR #550, opened 2017-03-31
Snap1323

vivekwiz:
Support prepared statement from JDBC route

## Patch testing
Precheckin

## ReleaseNotes.txt changes
No?

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/160


cluster/src/main/scala/io/snappydata/gemxd/SparkSQLPrepareImpl.scala
```diff
@@ -0,0 +1,226 @@
...
+  def allParamLiterals(plan: LogicalPlan, result: mutable.HashSet[ParamLiteral]): Unit = {
+    def allParams(plan: LogicalPlan): LogicalPlan = plan transformAllExpressions {
+      case bl@BinaryComparison(left: Expression, ParamLiteral(Row(pos: Int), NullType, 0)) =>
+        addParamLiteral(pos, left.dataType, left.nullable, result)
+        bl
+      case blc@BinaryComparison(left: Expression,
+      Cast(ParamLiteral(Row(pos: Int), NullType, 0), _)) =>
+        addParamLiteral(pos, left.dataType, left.nullable, result)
+        blc
+      case ble@BinaryComparison(left: Expression, CaseWhen(branches, elseValue)) =>
```
soubhik-c: I think this should be handled more generically instead of just CaseWhen. One should use the pattern used in other spark rules like

plan transform {

   case q: LogicalPlan => q transformAllExpressions {
    ....
}
}

Also, we should be able to inspect children of any expression and find out the paramliteral.

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLPrepareImpl.scala
```diff
@@ -0,0 +1,226 @@
...
+        addParamLiteral(pos, castType, false, result)
+        c
+      case cc@Cast(CaseWhen(branches, elseValue), castType: DataType) =>
+        handleCase(branches, elseValue, castType, false, result)
+        cc
+    }
+    handleSubQuery(allParams(plan), allParams)
+  }
+
+  def handleSubQuery(plan: LogicalPlan,
```
soubhik-c: Any level of nested subquery will automatically get handled with the pattern mentioned above.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -297,6 +297,17 @@ class SnappySessionState(snappySession: SnappySession)
 
   def getTablePartitions(region: CacheDistributionAdvisee): Array[Partition] =
     StoreUtils.getPartitionsReplicatedTable(snappySession, region)
+
+  var isPreparePhase: Boolean = false
```
soubhik-c: Is there a need of separate variable that too var ? I suppose it can be a method returning "pvs != null" to indicate perpare phase. Moreover, consider using Option[ParamterValueSet] instead of var.

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLPrepareImpl.scala
```diff
@@ -0,0 +1,226 @@
...
+import org.apache.spark.sql.types._
+import org.apache.spark.util.SnappyUtils
+
+
+class SparkSQLPrepareImpl(val sql: String,
+    val schema: String,
+    val ctx: LeadNodeExecutionContext,
+    senderVersion: Version) extends SparkSQLExecute with Logging {
+
+  if (Thread.currentThread().getContextClassLoader != null) {
```
soubhik-c: Could you explain what is this and why its required ?

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLPrepareImpl.scala
```diff
@@ -0,0 +1,226 @@
...
+import org.apache.spark.Logging
+import org.apache.spark.sql.{Row, SnappySession}
+import org.apache.spark.sql.catalyst.expressions.{BinaryComparison, CaseWhen, Cast, Exists, Expression, Like, ListQuery, ParamLiteral, PredicateSubquery, ScalarSubquery, SubqueryExpression}
+import org.apache.spark.sql.catalyst.plans.logical.LogicalPlan
+import org.apache.spark.sql.internal.SnappySessionState
+import org.apache.spark.sql.types._
+import org.apache.spark.util.SnappyUtils
+
+
+class SparkSQLPrepareImpl(val sql: String,
```
vivekwiz: This was done to avoid following line in SparkSQLExecuteImpl:
  private[this] val df = session.sql(sql)

Care has been taken to reuse some other methods of SparkSQLExecuteImpl by moving them to an Object.

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLPrepareImpl.scala
vivekwiz: Need to update same for SparkSQLPrepareImpl. Copy paste issue.

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLPrepareImpl.scala
vivekwiz: Will try to write new rules as suggested.

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLPrepareImpl.scala
vivekwiz: Needed with above rules, but will see if required or not with updated rules.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
vivekwiz: Will change. I thought readability is better this way.

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLPrepareImpl.scala
vivekwiz: Re-verified. Its needed. Please see corresponding piece of code in SparkSQLExecuteImpl

core/src/main/scala/org/apache/spark/sql/execution/CachedPlanHelperExec.scala
```diff
@@ -116,7 +123,33 @@ object CachedPlanHelperExec extends Logging {
       newpls: mutable.ArrayBuffer[ParamLiteral]): Unit = {
     literals.foreach { case lv@LiteralValue(_, _, p) =>
       lv.value = newpls.find(_.pos == p).get.value
-      val y = newpls.find(_.pos == p).get.value
     }
   }
+
+  def setValue(dvd: DataValueDescriptor): Any = dvd match {
```
vivekwiz: Renamed to getValue. 

I looked into getObject but that would lead to unnecessary object creation, so in my opinion current implementation is preferable.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
vivekwiz: Changed to Option[ParamterValueSet] however could not remove var. Also both the parameters are needed since execution case do not always have pvs but only in prep stmt case.

PR #554, opened 2017-04-07
[SNAP-1435] Added support for off-heap in SnappyMemoryManager.

rishitesh:
a) Added support for off-heap in SnappyMemoryManager.
A new config parameter MEMORY_SIZE is proposed in this change. This will driver the off heap memory size for SnappyData. 
User will give this value to configure its off heap requirement along with heap_size parameter.
When region data will be stored in off-heap main concern of users should be MEMORY_SIZE only. Heap will only be used for temp structures in very few places.

This change also has taken care of code which can be used to account off heap memory with the UMM. 
One should review the related store changes before going through this PR.

b) Formatting changes

## Patch testing
pre-checkin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/161


cluster/src/main/scala/org/apache/spark/memory/SnappyStorageEvictor.scala
```diff
@@ -66,7 +66,11 @@ class SnappyStorageEvictor extends Logging{
   }
 
   @throws(classOf[Exception])
-  def evictRegionData(bytesRequired: Long): Long = {
+  def evictRegionData(bytesRequired: Long, memoryMode: MemoryMode): Long = {
+    // @TODO use when regions can be stored in offheap
+    if (memoryMode.equals(MemoryMode.OFF_HEAP)){
+      return 0L
```
sumwale: Does this mean off-heap will skip the evictor? I guess this needs to implement the separate eviction logic for off-heap as we discussed.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -325,29 +337,17 @@ class SnappyUnifiedMemoryManager private[memory](
 
   override def dropStorageMemoryForObject(name: String,
                                           memoryMode: MemoryMode,
-                                          ignoreNumBytes: Long): Long =
-    synchronized {
```
sumwale: synchronized not required? And the executionPool, storagePool determination not required? Didn't understand this removal of ON_HEAP/OFF_HEAP check.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -385,6 +385,15 @@ private object SnappyUnifiedMemoryManager extends Logging{
     testCallbacks.map(l => l.onPositiveMemoryIncreaseDueToEviction(objectName, bytes))
   }
 
+  def getMemorySize(conf: SparkConf): Long = {
+    val cache = Misc.getGemFireCacheNoThrow
+    if (cache ne null) {
+      OffHeapStorage.parseOffHeapMemorySize(
+        Misc.getGemFireCache.getDistributedSystem.getConfig.getMemorySize)
```
sumwale: use "cache" already obtained above

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -385,6 +385,15 @@ private object SnappyUnifiedMemoryManager extends Logging{
...
+    } else { // for local mode testing
+      conf.getLong("snappydata.store.memory.size", 0L)
```
sumwale: When can this happen? Local mode also will boot GemFire. I think it should be taken out.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -397,16 +406,16 @@ private object SnappyUnifiedMemoryManager extends Logging{
     val minSystemMemory = (reservedMemory * 1.5).ceil.toLong
     if (systemMemory < minSystemMemory) {
       throw new IllegalArgumentException(s"System memory $systemMemory must " +
-          s"be at least $minSystemMemory. Please increase heap size using the --driver-memory " +
-          s"option or spark.driver.memory in Spark configuration.")
+        s"be at least $minSystemMemory. Please increase heap size using the --driver-memory " +
+        s"option or spark.driver.memory in Spark configuration.")
```
sumwale: This needs to be changed to say "-heap-size" or "-memory-size" for both leads and servers below.

cluster/src/main/scala/org/apache/spark/memory/SnappyStorageEvictor.scala
rishitesh: Correct. That's why have kept a TODO.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
rishitesh: In Local mode we create SparkEnv first , which will create the memory manager. Then we boot GemXD. Hence it will throw an exception

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
sumwale: @rishitesh Can you add this long value as a field+getter in GemFireCacheImpl constructor itself? Then it can be used both here as well as inside GemFireCacheImpl for initializing BufferAllocator used for off-heap allocation (in SNAP-1454)

sumwale:
@rishitesh Looks good overall -- see remaining comments. Eviction support for off-heap will have to be tracked separately.

sumwale:
@rishitesh Eviction support for off-heap (column tables only for now) is being taken care of in SNAP-1454 as discussed.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
sumwale: @rishitesh Please take note of the messages. We need to correct these to mention "heap-size/memory-size" for cluster mode and the other properties in local mode.

PR #557, opened 2017-04-10
Snap 1346 bugfix

kneeraj:
Some bug fixes in Tokenization

- Sub plans of type Broadcast hash joins were getting materialized during preparation of the full plan. So these kind of plans needed to be dealt differently. Now all these broadcast variables are being tracked and reprepared on subsequent executions with different constants.
- Problems with views fixed.
- Multiple wholestages inside the execution plan had a bug with Tokenization code. 

## Patch testing

precheckin

## ReleaseNotes.txt changes

## Other PRs 


soubhik-c:
LGTM. Only concern is a plan within broadcast variable is getting compiled again undermining the purpose of tokenization.

kneeraj:
Only concern is a plan within broadcast variable is getting compiled again undermining the purpose of tokenization.

The way broadcast is handled I found it difficult to reuse the compiled plan for that. So from the big tree only the broadcast variable will be reprepared so not completely ruining the purpose. We can revisit, if required to make it more optimized.

Tokenization looks useful mostly for oltp class queries to me and in my opinion we should not even strive to make all kinds of queries work with tokenization. Not only it will save insignificant amount than total cost of execution of an olap query but also reduce the chances of bug for complex queries.

BTW thanks for verifying the broadcast part with the tweak in the test.

soubhik-c:
Well I too had been contemplating as we discussed on supporting tokenization for oltp class of queries alone.

Now the thing is if an oltp query involves a join (distributed) then broadcast or exchange is expected to be super fast (having sufficient predicates reducing the search space) and therefore avoiding codegen completely is warranted. IMHO, present thing addresses the issue partly and should be looked at immediately.

hbhanawat:
>>we should not even strive to make all kinds of queries work with tokenization

What all queries would not work with tokenization? We started the tokenization work for TPCH benchmarking. In my discussion with @sudsmenon, he told that we need a TPCH report with the standard procedure and hence the tokenization.  So, I believe support for OLAP is required. 

Broadcast is a tricky thing because it is kind of result caching. So I am not sure how the tokenization should handle that. @jramnara has indicated that we should not cache broadcast variables so probably you are good there but not sure if he meant only testing it out.

soubhik-c:
>> @jramnara has indicated that we should not cache broadcast variables so probably you are good there but not sure if he meant only testing it out.
Just to add invalidating b.variable is okay, avoiding regeneration of underlying plan (code gen) is another point.

kneeraj:
>> What all queries would not work with tokenization? We started the tokenization work for TPCH benchmarking. In my discussion with @sudsmenon, he told that we need a TPCH report with the standard procedure and hence the tokenization. So, I believe support for OLAP is required.

All the queries where the optimizer changes the logical plan where it removes the constants from expressions and replaces with a new expression containing new constants then there won't be enev an opportunity to replace those unless we go to a level where we really go into the spark code change the transformations during optimizations to track how the constants were changed and put.

A lot of queries have been checked to see that it works. Like simple where clause, sub-queries, queries having broadcast etc. But a small tweak in the query can lead to an optimization sneak in which will make the injection of just the new constants difficult. 
Also tokenization should help only those queries where execution time is comparable or much less than the preparation time. For queries where this is opposite you will hardly save anything by having a cached plan.

>> Broadcast is a tricky thing because it is kind of result caching. So I am not sure how the tokenization should handle that. @jramnara has indicated that we should not cache broadcast variables so probably you are good there but not sure if he meant only testing it out.

Please have a look at CachedPlanHelperExec.doProduce and cacheddataframe.reprepareBroadcast for the same.

I need to have a cleaner for the previously broadcasted thing which is still pending.

hbhanawat:
We need to tell customers which queries would use tokenization and which won't. Right now, the answer is based on internal technical details and the scope is not clear to me. But, the exact nature of tokenization needs to be documented for the customers if there are quite few types of queries that won't use tokenization.

kneeraj:
>> We need to tell customers which queries would use tokenization and which won't.

Why is that. This is totally internal and transparent to the user. We are not going to tell anything to the customer. It is akin to some internal perf optimization.

hbhanawat:
Plan caching is something users are aware of. Why shouldn't we talk about it as a perf optimization that we have over and above Spark? Secondly, it will be good to minimize randomness from a developer's perspective. Saying that it may or may not use tokenization doesn't help. It may be difficult to tell exactly but a guideline would be good enough

kneeraj:
>> Plan caching is something users are aware of.
Are you talking about internal developers. Eternal users as far as I know will hardly bother as far as they are getting correct and performant results. For developers we can provide a trace for this. Categorizing queries for which will work and which will not work with Tokenization is not useful in my opinion. If it is required for sure then effort can be put to find out this and document this.

sumwale:
> Tokenization looks useful mostly for oltp class queries to me and in my opinion we should not even strive to make all kinds of queries work with tokenization.

To clarify this for all: tokenization is useful, actually required, not only for OLTP queries but for all kinds of queries. We are skipping it for some cases only to reduce the current work, but eventually would need it for all cases. There are two primary objectives with tokenization:

- Plan re-use and thus avoid re-creating the plans. This is a minor goal and useful only for OLTP class of queries.
- Drastically reduce the amount of code to be compiled. This is actually the major objective and why it is required for all cases, not just OLTP. In turn it effects queries in multiple ways:
  - The hit when new code has to be compiled all the time is large. Once the CodeGeneration cache (size 300) starts overflowing, then every query will need to be recompiled. The compilation overhead itself is the small portion. The big portion is the JIT kicking in and having to JIT the code. One can easily see it even in TPCH queries which take 2-3X more in first run in many queries but this hit will be even more with concurrent queries. But that itself has big trouble without tokenization -- see next point.
  - With concurrent queries (even 2-3), the CPU caches will easily start overflowing big time with a large number of plans required to be compiled all the time. Once that happens JIT will stop compilations and fall back to interpreted mode. The hit of that is huge: 10X or more (though even that can relatively not large if query itself is very expensive taking minutes/hours but those are not our primary target queries).

So this tokenization work is very critical and required for all scenarios, not just OLTP. About documenting what will not be tokenized: I think it is only a temporary set of limitations since we want to tokenize wherever possible due to the points mentioned above. So documenting this moving set of limitations can be avoided for now.

kneeraj:
The frequency/concurrency and the number of expensive queries ( olap ) the system will receive is not going to be big. Secondly changing data and all will warrant cache invalidation. So I am not quite sure how much actual gain are we talking about which will practically come in use.

jramnara:
Haven't read the whole thread, but, we need to keep the following in mind:

- We are chasing use cases where performance is, in general, very
important. It is a key differentiation we are touting. For instance, a
streaming application may decide to run multiple aggregations each time a
window tumbles (once a second).
- Broadly categorizing in terms of olap and oltp is not sufficient to
describe the importance of this feature. A olap query that aggregates
"measures" along specific dimensions (i.e. filters) is far more common that
aggregations over the entire data set. And, it is common to find hundreds
of dimensions (wide tables). i.e. non-repeating olap queries are common.
Imagine a bunch of Tableau worksheets, and a bunch of users - they all will
be often generating different queries.
In general, it is highly unlikely apps will repeat the same query over and
over again. Esp. makes little sense to do when it is expensive.
Essentially, to me, without this "tokenization" work, our optimization work
in the practical world, could be incomplete to many.
To be clear, I think, it is crucial we try to make sure this optimization
is applicable to the widest range of queries possible.


-----
Jags
SnappyData blog <http://www.snappydata.io/blog>
Download binary, source <https://github.com/SnappyDataInc/snappydata>

sumwale:
@kneeraj 

> The frequency/concurrency and the number of expensive queries ( olap ) the system will receive is not going to be big. Secondly changing data and all will warrant cache invalidation.

Even if system is running only 2-3 olap query types, nearly all the executions of those *will have different constant values*. In fact we are banking on the fact that the number of query types is small enough most of the time else even tokenization will not be enough to reduce the amount of compiled code.

Also see Jags' comment about this myth of OLTP vs OLAP. There is no such differentiation in practise in most cases, and more so for the ones we are targeting.

>  So I am not quite sure how much actual gain are we talking about which will practically come in use.

Huge and absolutely essential to have it, so be sure about it :-) Without it many of our gains will evaporate. Even take the way TPCH is actually supposed to be run: consecutive runs with different token values. That alone busts most of our performance claims completely. And it is just a small sample of what will happen in real use-cases.

sumwale:
> Secondly changing data and all will warrant cache invalidation.

That's a valid point and again underlies the importance of tokenization work.

Firstly with tokenization the hit of that is drastically reduced since one has to build up the cache only for different query types and not individual queries. As Jags mentioned there is hardly a case for users repeating same (expensive OLAP) queries with same constants, and the normal case will be queries repeated with different values. What that means is that tokenization can ensure that code compilation is not done repeatedly (avoiding the compiler and JIT hit etc as mentioned) and CPU instruction caches remain in check.

Secondly in many cases we are not expecting the eventual plan to change, so the hit is *only of plan re-creation*. The code compilation cache will get re-used and 90% or more of the hit is avoided. While this holds without tokenization too, that case will already be busted in most cases since one will hardly find the compiled plan for that in the cache as mentioned in the point above.

Lastly this also means we have to tighten up the invalidation by tracking exactly the effected plans and only invalidate those like in Derby but even more fine-grained. However, that is not high-priority since the hit of just the plan-recreation is not large (as per the point above) and this will mostly effect only OLTP class of queries for sometime -- like PK/index lookup ones. We will eventually come to that.

jramnara:
Just so I am clear, a compiled plan should not be invalidated for
updates/deletes in most situations. Is this correct?

-----
Jags
SnappyData blog <http://www.snappydata.io/blog>
Download binary, source <https://github.com/SnappyDataInc/snappydata>

sumwale:
> a compiled plan should not be invalidated for updates/deletes in most situations

We invalidate the plan when the size of table changes by more than 10% because that can lead to major change in plan (e.g. broadcast to exchange or vice-versa). So it should be infrequent.

Even then only the plan cache is invalidated and not the compiled code cache. So if the new plan is the same as old one and leads to the same generated code, then the previously compiled class will get re-used. The overhead of new plan creation is fixed and small (max few millis).

PR #558, opened 2017-04-11
[SNAP-1361] Added support for schema name in udf while querying e.g. select app.u…

rishitesh:

Added support for schema name in udf while querying e.g. select app.udfname(col_name) from table.

## Patch testing

pre-checkin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
@@ -604,16 +604,10 @@ class SnappyParser(session: SnappySession)
           (altPart, elsePart).asInstanceOf[WhenElseType])
   }
 
-  protected final def optionalIdentifier: Rule1[Option[String]] = rule {
-    ('.' ~ identifier) ~> ((x: String) => Some(x)) |
-    MATCH ~> (() => None)
-  }
-
   protected final def primary: Rule1[Expression] = rule {
     identifier ~ (
-      optionalIdentifier ~ '(' ~ ws ~ (
+      ('.' ~ identifier).? ~ '(' ~ ws ~ (
         '*' ~ ws ~ ')' ~ ws ~> ((n1: String, n2: Option[String]) =>
-          //assert(n2.equals(None), s"Builtin functions $n1 cannot have fully qualified name ")
           if (n1.equalsIgnoreCase("COUNT")) {
             AggregateExpression(Count(Literal(1, IntegerType)),
               mode = Complete, isDistinct = false)
--- a/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
@@ -22,7 +22,7 @@ import io.snappydata.QueryHint
 import org.parboiled2._
 import shapeless.{::, HNil}
 import org.apache.spark.sql.SnappyParserConsts.{falseFn, plusOrMinus, trueFn}
-import org.apache.spark.sql.catalyst.TableIdentifier
+import org.apache.spark.sql.catalyst.{FunctionIdentifier, TableIdentifier}
 import org.apache.spark.sql.catalyst.analysis._
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.expressions.aggregate.{AggregateExpression, Complete, Count}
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -606,33 +606,33 @@ class SnappyParser(session: SnappySession)
...
-                UnresolvedFunction(udfName, exprs, isDistinct = true)
-              }
-              w.asInstanceOf[Option[WindowSpec]] match {
-                case None => function
-                case Some(spec: WindowSpecDefinition) =>
-                  WindowExpression(function, spec)
-                case Some(ref: WindowSpecReference) =>
-                  UnresolvedWindowExpression(function, ref)
-              }
+      ('.' ~ identifier).? ~ '(' ~ ws ~ (
```
sumwale: What is this trying to do? An identifier preceeded by a '.'?

The qualified name case is already taken care of below in the '|' case: '.' ~ ws ~ (...) that creates an UnresolvedAttribute. If that does not work for UDFs then add whatever code is required there. Ideally UnresolvedAttribute should get resolved to a UDF during some analysis phase.

rishitesh: This change is to enable a query to pass by keeping schema name for UDFs e.g. select app.udf1(some_column) from table.
We could add in the place mentioned by you but that will duplicate the code above for UDFs.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -606,33 +606,33 @@ class SnappyParser(session: SnappySession)
...
+              case None => function
+              case Some(spec: WindowSpecDefinition) =>
+                WindowExpression(function, spec)
+              case Some(ref: WindowSpecReference) =>
+                UnresolvedWindowExpression(function, ref)
+            }
+          }
         ) |
         '.' ~ ws ~ (
             identifier. +('.' ~ ws) ~> ((i1: String, rest: Any) =>
```
sumwale: After looking at the parsing looks, change it as follows:

Rename "tableIdentifier" and "functionIdentifier" as "qualifiedIdentifier" and replace all usages of the two by it.

Change "idenfier" in the rule above to "qualifiedIdentifier" and then the parsing rule will look like:

```
qualifiedIdentifier ~ (
  '(' ~ ws ~ (
    ...
  ) |
  '.' ~ ws '*' ~> UnresolvedStar(...) |
  MATCH ~> UnresolvedAttribute.quoted _
...
)
```

i.e. like before just that top-level "identifier" becomes "qualifiedIdentifier" and the additional rule of UnresolvedAttribute gets subsumed in the MATCH case.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -642,10 +642,10 @@ class SnappyParser(session: SnappySession)
...
         MATCH ~> UnresolvedAttribute.quoted _
     ) |
-    '{' ~ FN ~ ws ~ identifier ~ '(' ~ (expression * commaSep) ~ ')' ~ ws ~ '}' ~ ws ~> {
-      (f: Any, e: Any) =>
-        f.asInstanceOf[String] match {
-          case f if f.equalsIgnoreCase("TIMESTAMPADD") =>
+    '{' ~ FN ~ ws ~ functionIdentifier ~ '(' ~ (expression * commaSep) ~ ')' ~ ws ~ '}' ~ ws ~> {
+      (f: FunctionIdentifier, e: Any) =>
+        f match {
+          case f if f.funcName.equalsIgnoreCase("TIMESTAMPADD") =>
```
sumwale: @soubhik-c Why is this special treatment of TIMESTAMPADD and SQL_TSI_DAY? Can't it be handled like a regular function call with any checks for the arguments inside the implementation of TimestampAdd. I guess you will need to add a new function which will internally call DateAdd but that is much better than cluttering the parser. In future too if we add more then folks will simply clutter the parser even more instead of cleanly handling it in a separate function.

The COUNT case is the one and only special case which can accept COUNT(*) arguments while no other function can, hence the special handling for COUNT else ideally even that should be avoided.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
rishitesh: Agreed. Will change that.

PR #559, opened 2017-04-12
Snap 1015 new

sonalsagarwal:
- Enabling FullResultSet Validation for ct tests.
- Refactoring test code to have generic methods in a sperate test class.
- Renamed order_details table to orders_details, to avoid name clash with other schema.
-  Modifying the sql Scripts to use user defined schema instead of inferring schema, to avoid inferring String to clob.

## Patch testing
Ran ct.bt


sumwale:
@kneeraj Once we implement the change to fix the CLOB vs VARCHAR confusion once for all, then these changes can be rolled back (to also validate) @sonalsagarwal @swatisawant

PR #561, opened 2017-04-14
Use sparkcontext default parallelism value for BlockAndExecutorId

vjr:
Use SparkContext default parallelism value for BlockAndExecutorId instead of totalCores and getNumProcessors.

Creating this PR more for discussion rather than actual merge since I'm nowhere near familiar with the Snappy codebase.

While running the "Getting Started" example option 3 "20x faster than Spark" I noticed that the Snappy portion was not scaling in performance when going from an 8 core CPU with SMT/hyperthreading disabled to enabling SMT/hyperthreading for a total of 16 logical processors.

In the first case (8 core CPU with hyperthreading disabled) the spark portion was running in about 2900 ms and the snappy portion in about 210 ms for an approximately 15x improvement, but in the second case (8 core CPU hyperthreading enabled for total of 16 logical CPUs) while the spark example improved by about 25 percent to about 2100 ms but the snappy portion remained at around 210 ms with the original code in branch-0.8.

With this change the snappy portion also improved by about 20 percent to about 170 ms benchmark average.

The CPU is an AMD Ryzen 7 1700 eight core with "SMT" a.k.a "Hyper Threading" for a total of 16 logical processing cores.


dhavalstr:

sumwale:
@vjr Can you note the exact changes in number of tasks being created and scheduled with and without this patch? That should be visible in the Spark GUI (port 5050 in cluster mode and 4040 in local mode). It should give us a better idea of what is happening and why defaultParallelism() is giving better numbers.

The reason for not using defaultParallelism(), esp in cluster mode, is that that will be based on the processors on the driver node which can be vastly different from the number of processors on the executors. That is why the onExecutorAdded looks at the number of actual processors from the profile of the new executor node. The reason for using profile information of exact number of processors rather than the configured "cores" on the executor is deliberate:

a) The default for "cores" on executors is (2*number of processors) -- search for spark.executor.cores in ExecutorInitiator.

b) This is a choice to allow for better CPU utilization in case of some of the tasks being IO bound. The other main reason is to achieve some level of concurrency in task executions. So we keep (2*processors) as total available "cores" on an executor while try to schedule atmost "processors" tasks per query/job so that the remaining "processors" cores are seen as free by the Spark scheduler.

Note that "cores" is a logical entity that the backend tells Spark scheduler for the available capacity on number of tasks that can be scheduled and each "core" will correspond to a thread in the execution pool on executor.

The above needs to be better discussed and more precise for concurrent tasks so that one large task does not hog all the cores not allowing new jobs/queries to be even scheduled. One of the ways discussed has been to create separate pools (as provided by [Spark](https://spark.apache.org/docs/latest/job-scheduling.html#default-behavior-of-pools)) limiting the number of cores between jobs especially low-latency ones.
The other idea that has been discussed is adding our own low-latency scheduler for point queries etc that will bypass the scheduler completely and directly spawn a pruned task.

Btw, the usual reason why simple queries do not scale on a single larger machine (but do on multiple smaller machines) is largely due to hardware limitations of CPU-RAM bus -- we saturate the capacity pretty quickly so throwing more processors does not help rather can be detrimental. This can be easily seen on a single machine by going from local[1] to local[2] etc where the performance improvement is nowhere close to being linear for such simple queries.

The RAM waits cannot even be interrupted by the OS so that capacity goes waste, and the normal CPU monitoring tools will keep showing the CPU as fully busy (as it is not waiting for any interruptible IO event). In such cases the capacity better be utilized for other concurrent queries which may actually be more IO intensive with EXCHANGE operators and all. Above is an item to be investigated more thoroughly for NUMA architectures in near future.


So much for theory. For for the purpose of fixing the performance behaviour seen, I think we need some numbers on how many tasks are being scheduled in the two cases to understand it better and come to a conclusion.

vjr:
@sumwale i might have been mistaken here. the GUI shows the same number of tasks for the insert and collect jobs when running only the snappy portion of the example. also getting about the same perf of 215 ms with and without the code change so not sure at the moment if i had tweaked or hardcoded values in other places in the code. will try to see how i observed the improvement earlier down to 170 ms - pretty sure i'd seen it in multiple runs of the example earlier.

sumwale:
@vjr It could have been multiple runs. Usually the time goes down with more runs of the same query. Can also try to hardcode the number of processors to smaller and larger values to see which number gives the best results in your setup.

sumwale:
@vjr closing this issue based on above. If there is more data then reopen.

PR #563, opened 2017-04-15
 [SNAP-1427] Reduce garbage in column allocators and dictionaries

sumwale:

Kishor's changes to TPCH:
- add parameters for persistence and redundancy to TPCH tables
- support for above in the createAndLoadTables script
- add a DataValidation job for TPCH query result validation

(Kishor, please add more details if required or if anything is missing)

The primary changes have been to use DirectByteBuffers for ColumnEncoders and a new dictionary
implementation that is fully off-heap including all key/value shell objects. The only objects now
used are top-level dictionary and buffer ones (<10 objects in total). Some major performance issues
with redundancy have also been addressed.

- Replaced ColumnData structure with ByteBuffer and added two flavours of allocators
  for those: HeapBufferAllocator and DirectBufferAllocator where latter is now used for expanding data
- Use HeapBufferAllocator for final data trimming (if required) in embedded/local mode
  so region can store it directly without any additional copying (this should change
      to the store setting when off-heap is enabled in column storage layer)
- Replaced mutable HashMap with TIntObjectHashMap in JDBCSourceAsStore iterator
- Adding a fully serialized ByteBufferHashMap for both heap and direct ByteBuffers
  - this uses fixed sized seriaized keys while variable key or value portions
    need to be put in "valueData" buffer
  - key also stores hash for first comparison which should filter out 99% of bucket clashes
  - re-uses ByteBuffer data but uses duplicate() methods to not hold on ByteBuffer object for too long
- Always prefer primaries for inserts (fixes most of the performance issue with redundancy)
- Increased default SnappyDataService interval to 20s. This should be reverted once the service
  itself is changed to be efficient instead of full table scan like now.
- Fixed an issue with complex type handling exposed after switching to DirectByteBuffers:
  - absolute position of individual fields was being passed in by generated code but this
    causes trouble if the buffer if reallocated to expand it due to being full during encoding;
    did not cause trouble with heap buffers since their absolute position remains the same
  - added a detailed comment to ColumnEncoder.initializeComplexType on the complex-type
    encoding layout

## Patch testing

Precheckin with store. Extensive TPCH testing.

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/162


core/src/main/scala/org/apache/spark/sql/execution/ObjectHashMapAccessor.scala
```diff
@@ -1311,10 +1309,9 @@ case class ObjectHashMapAccessor(@transient session: SnappySession,
     case StringType if !multiMap =>
       // copy just reference of the object if underlying byte[] is immutable
       val stringVar = resultVar.value
-      val platformClass = classOf[Platform].getName
       val bytes = ctx.freshName("stringBytes")
       s"""byte[] $bytes;
-        if ($stringVar.getBaseOffset() == $platformClass.BYTE_ARRAY_OFFSET
+        if ($stringVar.getBaseOffset() == Platform.BYTE_ARRAY_OFFSET
```
rishitesh: Would not it cause problem  ? Actual byte[] will be accessed in different JVM and driver is creating the code. There will be a high chance of mismatch as drivers will be generally will be less than 32G.

sumwale: No the method is still part of generated code (i.e. it is *not* ${Platform.BYTE_ARRAY_OFFSET}"). The only change here is that explicit qualified platformClass is unnecessary since it in the default imports of Spark's CodeGenerator.

core/src/main/scala/org/apache/spark/sql/execution/CachedPlanHelperExec.scala
```diff
@@ -56,12 +54,12 @@ case class CachedPlanHelperExec(childPlan: CodegenSupport, @transient session: S
       case None => session.addContextObject(REFERENCES_KEY,
         ArrayBuffer[ArrayBuffer[Any]](ctx.references))
     }
-    // keep a map of the first broadcasthashjoinexec plan and the corresponding ref array
+    // keep a map of the first Hroadcasthashjoinexec plan and the corresponding ref array
```
sumwale: correcting

sumwale:
@rishitesh Can you see my reply to your request for change, and remove it if the explanation sounds fine?

core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCSourceAsStore.scala
```diff
@@ -184,20 +185,21 @@ final class ColumnBatchIteratorOnRS(conn: Connection,
...
-  val ps = conn.prepareStatement(fetchColQuery)
-  var colBuffers: Option[scala.collection.mutable.HashMap[Int, (ByteBuffer, Blob)]] = None
+  val ps: PreparedStatement = conn.prepareStatement(fetchColQuery)
+  var colBuffers: Option[TIntObjectHashMap] = None
+
   def getColumnLob(bufferPosition: Int): ByteBuffer = {
     colBuffers match {
-      case Some(map) => map(bufferPosition)._1
+      case Some(map) =>
+        map.get(bufferPosition).asInstanceOf[(ByteBuffer, Blob)]._1
```
hbhanawat: TIntObjectHashMap is making the code complex and needs some asInstanceOf calls. Given that, the number of entries in the Hash map is going to be not more than the number of columns, would the performance difference be too much?

core/src/main/scala/org/apache/spark/sql/execution/ByteBufferHashMap.scala
```diff
@@ -0,0 +1,268 @@
...
+ *     header      body    padding
+ * }}}
+ * If key has variable length data, then it should be appended to the
+ * value data. The offset+hash code is read as a single long where LSB
+ * is used for hash code while MSB is used for offset, so the two can
+ * be reverse in actual memory layout on big-endian machines. Since there
+ * is no disk storage of this map so no attempt is made to have consistent
+ * endianness and thus the actual memory layout of the data.
+ */
+final class ByteBufferHashMap(initialCapacity: Int, val loadFactor: Double,
```
hbhanawat: For now, this is a map for only string values. Right? Can you add the documentation on how key data and valuedata are being used finally by the encoders. It took me some time to realize that keyData is ignored and then valueData is the only thing that is encoded. Also, explaining that the reading of value data happens in the DictionaryEncoderBase would be helpful.

sumwale: Right now used only by dictionary encoder, but same will also be used instead of ObjectHashSet for off-heap support and reducing the objects (which should fix GC issues with TPCH Q21 when using hashjoin among others).

Will add more javadocs.

core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCSourceAsStore.scala
sumwale: Agree that TIntObjectHashMap is not nice to use. As part of SNAP-1454 have added fastutil reference for use elsewhere (which is type safe) and we should make a point to use that every where. Can also gradually transition trove maps to it and take out trove like done in Geode.

The main reason for using the open hash maps is much lesser garbage more than performance (though fastutil maps are also quite a bit faster than JDK/Scala maps). Though it could be true that garbage here may not be a contributor but to be on the safe side we should just make that a standard practice and lessen our overall GC troubles.

Will change this to use fastutil and add the reference in this PR itself.

PR #564, opened 2017-04-17
[SNAP-1384] : Column Table Inserts can fail if generated code is big

rishitesh:
Modified ColumnInsertExec to handle wide schema to 1012 columns.

* Have separated out the methods handling wide schema so as not to impact performance of existing code and narrow schema
* Chunked the code to different methods to make java function size well withing 64K

## Patch testing

pre-checkin yet to run

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
@@ -516,6 +516,7 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
       genMultipleStatsMethods(ctx, "writeStats", statsCode, statsSchema, stats)
 
     ctx.INPUT_ROW = statsRowTerm
+    ctx.currentVars = null
     val statsEv = GenerateUnsafeProjection.createCode(ctx, statsExprs)
     val statsRow = statsEv.value
 
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
@@ -59,6 +59,8 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
   @transient private val MAX_CURSOR_DECLARATIONS = 30
   @transient private var cursorsArrayTerm: String = _
   @transient private var cursorsArrayCreate: String = _
+  @transient private var encoderArrayTerm: String = _
+  @transient private var cursorArrayTerm: String = _
 
   @transient private[sql] var batchIdRef = -1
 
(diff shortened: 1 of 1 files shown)


sumwale:
@rishitesh Sorry for somewhat late feedback, but it will be better to refactor the new methods into a separate class. This will avoid complicating this class. Two ways: a) just as a utility object class, b) as a new plan class which either extends ColumnInsertExec or else the two share common code via object class or a common trait which both extend

The second option seems cleaner to me, but whatever you prefer.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -78,7 +82,134 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
     else ("1", "0x0")
   }
 
+  def loop(code: String, numTimes: Int): String = {
+    s"""
+      for(int i=0;i< $numTimes; i++){
+          $code
+      }""".stripMargin
```
sumwale: there is no margin(|) here to strip

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -78,7 +82,134 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
...
+  }
```
sumwale: there is no margin(|) here to strip

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -78,7 +82,134 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
...
+    val encodingClass = classOf[ColumnEncoding].getName
+    val numInsertedRowsMetric = if (onExecutor) null
+    else metricTerm(ctx, "numInsertedRows")
+    schemaTerm = ctx.addReferenceObj("schema", relationSchema,
+      classOf[StructType].getName)
+
+    val schemaLength = relationSchema.length
+    encoderArrayTerm = ctx.freshName("encoderArray")
+    cursorArrayTerm = ctx.freshName("cursorArray")
+    encoderCursorTerms = relationSchema.map { _ =>
```
sumwale: do you need the separate terms too when there are the two arrays?

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -78,7 +82,134 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
...
+        batchBucketIdTerm = Some(c.bucketIdTerm)
+      case _ =>
+    }
+
+    val initEncoderCode =
+      s"""
+         |this.$encoderArrayTerm[i] = $encodingClass$$.MODULE$$.getColumnEncoder(
+         |           |  $schemaTerm.fields()[i]);
+       """.stripMargin
+
```
sumwale: Much of the code above this can be refactored and shared

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -78,7 +82,134 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
...
+    val initEncoderArray = loop(initEncoderCode, relationSchema.length)
+    val iniCursorCode = s"this.$cursorArrayTerm[i] = 0L;"
+    val iniCursorArray = loop(iniCursorCode, relationSchema.length)
+
+    ctx.addMutableState("org.apache.spark.sql.execution.columnar.encoding.ColumnEncoder[]",
```
sumwale: use embedded class variable rather than hardcoded class path:

```scala
  val encoderClass = classOf[ColumnEncoder].getName
  ... 
  ctx.addMutableState(s"$encoderClass[]", ...)
```

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -78,7 +82,134 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
...
+    }
+
+    val initEncoderCode =
+      s"""
+         |this.$encoderArrayTerm[i] = $encodingClass$$.MODULE$$.getColumnEncoder(
+         |           |  $schemaTerm.fields()[i]);
+       """.stripMargin
+
+    val initEncoderArray = loop(initEncoderCode, relationSchema.length)
+    val iniCursorCode = s"this.$cursorArrayTerm[i] = 0L;"
```
sumwale: Is the initialization to 0 of any use? It will be initialized with the "finish" of encoder in any case, no?

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -78,7 +82,134 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
...
+    val iniCursorArray = loop(iniCursorCode, relationSchema.length)
+
+    ctx.addMutableState("org.apache.spark.sql.execution.columnar.encoding.ColumnEncoder[]",
+      encoderArrayTerm,
+      s"""
+         |this.$encoderArrayTerm =
+         | new org.apache.spark.sql.execution.columnar.encoding.ColumnEncoder[$schemaLength];
```
sumwale: see comment about class above

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -78,7 +82,134 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
...
+         |$initEncoderArray
+        """.stripMargin)
+
+    ctx.addMutableState("long[]", cursorArrayTerm,
+      s"""
+         |this.$cursorArrayTerm = new long[$schemaLength];
+         |$iniCursorArray
```
sumwale: JVM guarantees 0 initialization of arrays so iniCursorArray is of no use.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
sumwale: To be paranoid, the loop code can be a separate method itself for many of the calls I saw below. But will need the variables to be either members or as arguments to the method. So you can have two versions of loop() method -- one which creates in a separate method where this is simple enough to do, and this one.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -78,7 +82,134 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
...
+       |  $declarations
+       |  $defaultBatchSizeTerm = Math.max(
+       |    (${math.abs(columnBatchSize)} - 8) / $defaultRowSize, 16);
+       |  // ceil to nearest multiple of $checkFrequency since size is checked
+       |  // every $checkFrequency rows
+       |  $defaultBatchSizeTerm = ((($defaultBatchSizeTerm - 1) / $checkFrequency) + 1)
+       |      * $checkFrequency;
+       |  $initEncoders
+       |  $childProduce
+       |}
```
sumwale: Can this if block be a separate method?

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -78,7 +82,134 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
...
+       |if ($batchSizeTerm > 0) {
+       |  $storeColumnBatch(-1, $storeColumnBatchArgs);
+       |  $batchSizeTerm = 0;
+       |}
+       |${
+      if (numInsertedRowsMetric eq null) ""
+      else s"$numInsertedRowsMetric.${metricAdd(numInsertions)};"
+    }
```
sumwale: ${} formatting has broken in the 4 lines above. You will need to correct manually since IDEA auto-formatter has trouble with this (and probably introduced these incorrect breaks).

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
rishitesh: Will change.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
rishitesh: Yes. Will remove it.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
rishitesh: Corrected the formatting.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
rishitesh: Corrected it.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
rishitesh: Removed it

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
rishitesh: Corrected

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
rishitesh: No, not needed. Will remove them.

PR #565, opened 2017-04-17
Snap 1210 Fix NullPointerException caused when writing dataframe containing timestamp column to csv files

SachinJanani:
* [SNAP-1210] Fix NullPointerException caused when writing dataframe containing timestamp column to csv files. The issue was caused by the transient field in FastDatePrinter which was not getting serialized.

## Patch testing
* Added unit test which is passing and also ran precheckin

## ReleaseNotes.txt changes
NA

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/171


sumwale:
@SachinJanani @dshirish 

This is turning out to be a bigger problem and can be with user classes/dependencies too. See SNAP-1395 and comments there. I think its time to fix this in a more generic way. We need to inspect the class for any specific readResolve/writeReplace/readObject/writeObject methods and then register that class using JavaSerializer if so. So something like below:

- Add a new kryo SerializerFactory implementation (e.g. see kryo's inbuilt ReflectionSerializerFactory) that takes care of above methods.
- In the makeSerializer, if the class implements Serializable then inspect it to find any readResolve/writeReplace/readObject/writeObject methods.
- If that is the case, then use KryoJavaSerializer. **Also invoke kryo.register** to specifically register that class to use KryoJavaSerializer. This will avoid future reflections for that class.
- Else fallback to kryo's default FieldSerializer.
- Register this as the default factory with kryo.setDefaultSerializer(factory).

If someone has a better idea than above to solve this in a generic way (or to improve it), then please comment.

SachinJanani:
@sumwale Implemented SerializerFactory as suggested but it will not work for FastDatePrinter.As FastDatePrinter implements Serializable but does not has either of  readResolve/writeReplace/readObject/writeObject so FieldSerializer is used for FastDatePrinter.Since it has few transient fields which was causing SNAP-1210 and are not serialized by the FieldSerializer in the kryo version that we have.This issue is solved in latest kryo library which has [setSerializeTransient](https://github.com/EsotericSoftware/kryo/blob/master/src/com/esotericsoftware/kryo/serializers/FieldSerializer.java#L470) method in FieldSerializer.So for now explicitly set JavaSerializer for FastDatePrinter.Also tested manually SNAP-1395 with this changes and the job using elasticsearch works fine.

core/src/main/scala/org/apache/spark/serializer/SnappyKryoSerializerFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+import com.esotericsoftware.kryo.factories.SerializerFactory
+import com.esotericsoftware.kryo.serializers.{JavaSerializer => KryoJavaSerializer, FieldSerializer =>
+KryoFieldSerializer}
+import com.esotericsoftware.kryo.{Serializer => KryoClassSerializer}
+import org.apache.spark.Logging
+
+/**
+ * This serializer factory will instantiates new serializers of a given class via reflection. If the class implements
+ * Serializable and has either of writeReplace/readResolve/readObject/writeObject then it will set JavaSerializer for
+ * that class else FieldSerailizer will be used
```
sumwale: Please limit the line size to max 100. Use latest IDEA (> 14) which will mark these scalaStyle errors (turn that on in inspection settings if not). We need to be turning on scalaStyle by default on our modules too but was not done due to Rishi's objection of number of pending merges which will get conflicts.

sumwale: "will instantiate"

core/src/main/scala/org/apache/spark/serializer/SnappyKryoSerializerFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+ *
+ */
+class SnappyKryoSerializerFactory extends SerializerFactory {
+
+  override def makeSerializer(kryo: Kryo, `type`: Class[_]): KryoClassSerializer[_] = {
+    if(isJavaSerializerRequired(`type`)) return new KryoJavaSerializer()
```
sumwale: add space after if() as per the scalaStyle

sumwale: Any specific reason to use ticked reserved name "type". Just use non-reserved name.

core/src/main/scala/org/apache/spark/serializer/SnappyKryoSerializerFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+    // TODO: Set serialize transient to true when new version of kryo will be used
+    return fieldSerializer;
+  }
+
+  private def isJavaSerializerRequired(clazz: Class[_]): Boolean = {
+    //if (`type`.isAssignableFrom(JavaSerializable)) {
+
+    if (classOf[JavaSerializable].isAssignableFrom(clazz)) {
+      return (hasInheritableReadWriteMethods(clazz, "writeReplace") || hasInheritableReadWriteMethods(clazz, "readResolve")
+        || hasInheritableReadWriteMethods(clazz, "readObject") || hasInheritableReadWriteMethods(clazz, "writeObject"))
```
sumwale: break the lines on the "||" to keep within 100

core/src/main/scala/org/apache/spark/serializer/SnappyKryoSerializerFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+    var current: Class[_] = `type`
+
+      while (current != null && method == null) {
+        try {
+          method = current.getDeclaredMethod(methodName)
+        }
+        catch {
+          case ex: NoSuchMethodException => {
+            current = current.getSuperclass
+          }
```
sumwale: NativeCalls in store has similar methods: getAnyMethod and getAnyField. Perhaps that can be moved in a common utility class (e.g. ClientSharedUtils) and used in both places.

sumwale:
How is it possible? If it does not have any of those methods then java serialization should also fail, no??

core/src/main/scala/org/apache/spark/serializer/SnappyKryoSerializerFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  }
+
+  private def hasInheritableReadWriteMethods(`type`: Class[_], methodName: String): Boolean = {
+    var method: Method = null
+    var current: Class[_] = `type`
+
+      while (current != null && method == null) {
+        try {
+          method = current.getDeclaredMethod(methodName)
```
sumwale: This is incorrect. You also need to pass parameterTypes to getDeclaredMethod. The writeReplace/readResolve methods don't take parameters but readObject/writeObject methods do take ObjectInputStream/ObjectOutputStream arguments (kryo's ExternalizableSerializable only looks at first two which is why you were probably confused).

Checking FastDatePrinter code, it does have readObject method as expected otherwise java serialization will also fail. That's where the transient fields are initialized: http://grepcode.com/file/repo1.maven.org/maven2/org.apache.commons/commons-lang3/3.4/org/apache/commons/lang3/time/FastDatePrinter.java#FastDatePrinter.readObject%28java.io.ObjectInputStream%29

sumwale: Another option could be to loop through getDeclaredMethods() searching on name, but that's only approximate check and parameterTypes should be provided.

cluster/src/dunit/scala/io/snappydata/externalstore/ColumnTableDUnitTest.scala
```diff
@@ -631,6 +631,18 @@ class ColumnTableDUnitTest(s: String) extends ClusterManagerTestBase(s) {
     assert(snc.sql("select count(*) from airline").count()>0)
     snc.sql("drop table airline")
   }
+
+  def testSNAP1210(): Unit = {
+    val snc = org.apache.spark.sql.SnappyContext(sc)
+
+    snc.sql(s"create table t1 using com.databricks.spark.csv options(path '${(getClass.getResource("/northwind/orders"+
+      ".csv").getPath)}', header 'true', inferschema 'true')")
```
sumwale: break the line

core/src/main/scala/org/apache/spark/serializer/PooledKryoSerializer.scala
```diff
@@ -127,9 +129,9 @@ final class PooledKryoSerializer(conf: SparkConf)
       BlockManagerId.getCachedBlockManagerId))
     kryo.register(classOf[StorageLevel], new ExternalizableResolverSerializer(
       StorageLevel.getCachedStorageLevel))
-    kryo.register(classOf[SLF4JLocationAwareLog], new KryoJavaSerializer())
-    kryo.register(classOf[Log4jLoggerAdapter], new KryoJavaSerializer())
-    kryo.register(classOf[NOPLogger], new KryoJavaSerializer())
+    //Explicitly register FastDatePrinter because it has transient fields which are not serialized by the
+    // FieldSerializer in the current version of kryo that we have
+    kryo.register(classOf[FastDatePrinter], new KryoJavaSerializer())
```
sumwale: should not be required once method lookup is corrected as mentioned

core/src/main/scala/org/apache/spark/serializer/SnappyKryoSerializerFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+ * Serializable and has either of writeReplace/readResolve/readObject/writeObject then it will set JavaSerializer for
+ * that class else FieldSerailizer will be used
+ *
+ */
+class SnappyKryoSerializerFactory extends SerializerFactory {
+
+  override def makeSerializer(kryo: Kryo, `type`: Class[_]): KryoClassSerializer[_] = {
+    if(isJavaSerializerRequired(`type`)) return new KryoJavaSerializer()
+    val fieldSerializer: KryoFieldSerializer[Nothing] = new KryoFieldSerializer(kryo, `type`)
+    // TODO: Set serialize transient to true when new version of kryo will be used
```
sumwale: Remove this TODO since java serialization also mandates that transient fields should not be serialized.

core/src/main/scala/org/apache/spark/serializer/SnappyKryoSerializerFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+    if (classOf[JavaSerializable].isAssignableFrom(clazz)) {
+      return (hasInheritableReadWriteMethods(clazz, "writeReplace") || hasInheritableReadWriteMethods(clazz, "readResolve")
+        || hasInheritableReadWriteMethods(clazz, "readObject") || hasInheritableReadWriteMethods(clazz, "writeObject"))
+    } else {
+      return false
+    }
+
+  }
+
+  private def hasInheritableReadWriteMethods(`type`: Class[_], methodName: String): Boolean = {
```
sumwale: Use a non-reserved name instead of back-ticked reserved "type". Have found this facility of back-ticking useless and confusing at best.

core/src/main/scala/org/apache/spark/serializer/SnappyKryoSerializerFactory.scala
```diff
@@ -0,0 +1,72 @@
...
+        try {
+          method = current.getDeclaredMethod(methodName)
+        }
+        catch {
+          case ex: NoSuchMethodException => {
+            current = current.getSuperclass
+          }
+        }
+      }
+    return ((method != null) && (method.getReturnType eq classOf[AnyRef]))
```
sumwale: This check is not correct for writeObject and readObject which return void. You can pass returnType too as an argument to this method to be completely sure (and can add to NativeCalls methods with null indicating skipping the check) -- void return type is Void.Type

SachinJanani:
@sumwale  Done with the required changes also moved methods from NativeCalls to ClientSharedUtils.Here is the pull request for store : https://github.com/SnappyDataInc/snappy-store/pull/171

sumwale:
Thanks, good for merge.

cluster/src/dunit/scala/io/snappydata/externalstore/ColumnTableDUnitTest.scala
```diff
@@ -631,6 +631,19 @@ class ColumnTableDUnitTest(s: String) extends ClusterManagerTestBase(s) {
...
+  def testSNAP1210(): Unit = {
+    val snc = org.apache.spark.sql.SnappyContext(sc)
+
+    snc.sql(s"create table t1 using com.databricks.spark.csv options(path " +
+      s"'${(getClass.getResource("/northwind/orders.csv").getPath)}', header 'true', " +
+      s"inferschema 'true')")
+    snc.sql("select * from t1").printSchema()
+    snc.sql("select * from t1").show
+    snc.sql("select * from t1").write.csv("/tmp/" + System.currentTimeMillis())
+    assert(snc.sql("select count(*) from t1").count() > 0)
```
sumwale: you may want to remove the csv file at the end of test else files will keep building in /tmp

PR #567, opened 2017-04-19
Snap 1459

hbhanawat:

For queries that have a join and an aggregate - During code generation, SnappyHashAggregate creates a class for the KeyBuffer. Then HashJoin creates a class for LocalMap. 

For queries such as 

select ORIGIN
from airline a, airlineref ar
where a.uniquecarrier = ar.code
GROUP BY origin

The types of class created by HashJoin i.e. LocalMap, is same as the types of the class created by hash aggregate i.e. KeyBuffer, because the query only has the group by column and no aggregates. And hence the LocalMap class is not registered and the already registered class of SnappyHashAggregate is returned. This is a problem because the structure of these classes is different.

To overcome this issue, registering the classes with a multimap parameter which differentiates between the hashjoin and hashaggregate.

## Patch testing

precheckin. Failing query passed. 

## ReleaseNotes.txt changes

## Other PRs 


hbhanawat:
Merged this as the patch has already been reviewed by @sumwale on branch-jj-0.8.1.

PR #569, opened 2017-04-20
[GITHUB-534] Match against variable name for dictionary optimization

### LINKED ISSUE #534: distinct function Nested queries problem
distinct function Nested queries problem  have a prolem，Whether it is bug？
I have a table.
CREATE TABLE app.yes_with(
  device_id  VARCHAR(200) , 
  device_brand  VARCHAR(200), 
  device_model  VARCHAR(200), 
  system_version  VARCHAR(200), 
  system_name  VARCHAR(200) ,
  oem_system_version  VARCHAR(200) , 
  oem_system_name  VARCHAR(200) ,
  sample_app_version  VARCHAR(200), 
  app_key  VARCHAR(200), 
  app_channel  VARCHAR(200) , 
  sample_user_id  VARCHAR(200) , 
  sample_package_name  VARCHAR(200) , 
  sample_app_name  VARCHAR(200) , 
  sdk_version  VARCHAR(200) , 
  api_level  VARCHAR(200), 
  mobile_corporation  VARCHAR(200) , 
  phone_num  VARCHAR(200) , 
  ip  VARCHAR(200) , 
  mac_address  VARCHAR(200) , 
  page_url  VARCHAR(200) , 
  page_title  VARCHAR(200), 
  package_name  VARCHAR(200) , 
  page_tag  VARCHAR(200) ,
  open_time bigint , 
  close_time bigint , 
  from_page_name  VARCHAR(200), 
  to_page_name  VARCHAR(200),
 day  VARCHAR(200) , 
 page  VARCHAR(200) )
  USING COLUMN OPTIONS (PERSISTENT 'asynchronous',PARTITION_BY  '**device_id**',BUCKETS '800');

The following statement is correct：
select a.device_brand,count(distinct a.**device_model**) from no_with a,yes_with b where a.device_brand=b.device_brand group by a.device_brand limit 10;

The following sentence error：
select sdk_version, count(distinct **device_id**) from (select sdk_version,device_id from  APP.YES_WITH group by sdk_version,device_id) a group by sdk_version ;

	at Remote Member 'isunode011(19649)<v16>:11310' in org.apache.spark.sql.execution.SparkPlan.executeQuery(SparkPlan.scala:133)
	at Remote Member 'isunode011(19649)<v16>:11310' in org.apache.spark.sql.execution.SparkPlan.execute(SparkPlan.scala:114)
	at Remote Member 'isunode011(19649)<v16>:11310' in org.apache.spark.sql.execution.aggregate.SnappyHashAggregateExec.doExecute(SnappyHashAggregateExec.scala:212)
	at Remote Member 'isunode011(19649)<v16>:11310' in org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$1.apply(SparkPlan.scala:115)
	at Remote Member 'isunode011(19649)<v16>:11310' in org.apache.spark.sql.execution.SparkPlan$$anonfun$execute$1.apply(SparkPlan.scala:115)
	at Remote Member 'isunode011(19649)<v16>:11310' in org.apache.spark.sql.execution.SparkPlan$$anonfun$executeQuery$1.apply(SparkPlan.scala:136)
Caused by: java.lang.StackOverflowError

sumwale:
Failing query:

select sdk_version, count(distinct device_id) from (select sdk_version,
    device_id from YES_WITH group by sdk_version, device_id) a group by sdk_version

(this fails only if the table YES_WITH is partitioned on device_id when EXCHANGE is avoided)

Fixes #534. The issue is due to hashjoin/aggregate referring to dictionary
variables that are not in its scope. Ideally the information about dictionary
variables should be transmitted through additional fields ExprCode() but
an extension to that does not work since Expression.genCode creates new
ExprCode objects without copying the incoming ones if possible
(in particular the simple BoundReference also ends up losing the
   original extension class object)

So the code resorts to using context objects via SnappySession. The lookup
key in current code consists of AttributeReference match but that breaks
down in this case since the AttributeReference refers to the table scan
column but actually comes from inner GROUP BY grouping key which
cannot come directly from table scan.

- now using lookup against the incoming variable in consume(..input)
- to allow initializing the dictionary array before ColumnBatch iteration,
  the batchConsume now creates a function to initialize it which returns null;
  the doConsume(..input) code overwrites this function with actual array
  if possible (i.e. when the key column comes from dictionary)
- for clarity removing old ExprCodeEx class and changing to DictionaryCode
  while hash variable for a set of keys is tracked separately
- added test of the reported query in QueryTest

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/cluster/src/test/scala/io/snappydata/QueryTest.scala
+++ b/cluster/src/test/scala/io/snappydata/QueryTest.scala
@@ -19,8 +19,9 @@ package io.snappydata
 
 import scala.collection.JavaConverters._
 
+import org.apache.spark.sql.execution.benchmark.ColumnCacheBenchmark
 import org.apache.spark.sql.internal.SQLConf
-import org.apache.spark.sql.{SnappyContext, SparkSession}
+import org.apache.spark.sql.{Row, SnappyContext, SparkSession}
 
 class QueryTest extends SnappyFunSuite {
 
@@ -89,5 +90,22 @@ class QueryTest extends SnappyFunSuite {
     if (result1 != result2) {
       fail(s"Expected result: $result2\nGot: $result1")
     }
+  }
+
+  /**
+   * Distinct query failure in code generation reported on github
+   * (https://github.com/SnappyDataInc/snappydata/issues/534)
+   */
+  test("GITHUB-534") {
+    val session = SnappyContext(sc).snappySession
+    session.sql("CREATE TABLE yes_with(device_id VARCHAR(200), " +
+        "sdk_version VARCHAR(200)) USING COLUMN OPTIONS(PARTITION_BY 'device_id')")
+    session.insert("yes_with", Row("id1", "v1"), Row("id1", "v2"),
+      Row("id2", "v1"), Row("id2", "v1"), Row("id2", "v3"))
+    val r = session.sql("select sdk_version, count(distinct device_id) from (" +
+        "select sdk_version,device_id from YES_WITH group by sdk_version, " +
+        "device_id) a group by sdk_version")
+    ColumnCacheBenchmark.collect(r,
+      Seq(Row("v1", 2), Row("v2", 1), Row("v3", 1)))
   }
 }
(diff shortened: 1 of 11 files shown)


soubhik-c:
LGTM, as you explained ExprCode cannot be used. Similar issue is with Literal handling too whereby explicit usage of class to construct object instead of .copy methods restricts us from extending and plugging code. So, using session to share seems to be the only way for now. Will go through the code more indepth and propose any other idea strikes me.

PR #572, opened 2017-04-23
Test for SNAP-1367

SachinJanani:
* Changes snappy parser to have column name to start with UNDERSCORE(_) or alphabhet
* Scala Test for SNAP-1367

## Patch testing
Ran precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/167


core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
```diff
@@ -116,7 +116,7 @@ abstract class SnappyBaseParser(session: SnappySession) extends Parser {
   protected def start: Rule1[LogicalPlan]
 
   protected final def identifier: Rule1[String] = rule {
-    atomic(capture(CharPredicate.Alpha ~ Consts.identifier.*)) ~
+    atomic(capture((CharPredicate("_") | CharPredicate.Alpha) ~ Consts.identifier.*)) ~
```
sumwale: For a single char just '_' without CharPredicate is more efficient. However, for best performance it is better to declare a new CharPredicate in SnappyParserConsts that adds '_' to Alpha like Consts.identifier: CharPredicate.Alpha ++ CharPredicate('_')

PR #576, opened 2017-04-27
hydra smoke bt addition to precheckin

swatisawant:

 - Added smoke.bt as a part of precheckin
- Moved SnappyTestRunner to the core repository
- Added selective set of northWind queries to be run in smoke bt
- Adding parameter to run smoke bt by default in precheckin
- Modified the smoke bt to use long running test framework in order to reduce the time required for cluster startup and stop in each test in smoke bt.
- Now for smoke bt, embedded mode and smart connector mode cluster will be started in first test and rest all tests will use the same cluster.
- Modified the framework to include split mode support in long running test scenario
- Fixed issues in long Running test framework
- Added end to end result validation for smoke bt
- Reduced number of workers to restrict it to 1 per host
- Modified the test code to have optional numRowsValidation
- Added support for fullResutsetValidation in case of user specified data volume
- Using default northWind data for smoke bt for now due to GC issues and Low Memory Exception with using 1GB of data in cluster configuration of 2 servers(4GB each) and redundancy set to 1
- Will change smoke bt to use 1 GB of data on SNAP-1427
- scala style formatting changes in all test files

## Patch testing

Verified the changes by running smoke bt and prechekin

## ReleaseNotes.txt changes

NA

## Other PRs 

No


build.gradle
```diff
@@ -840,10 +840,15 @@ task cleanAll {
 }
 task buildAll {
   dependsOn getTasksByName('assemble', true).collect { it.path }
+  dependsOn getTasksByName('product', true).collect { it.path }
```
rishitesh: Why this is needed ?

build.gradle
```diff
@@ -840,10 +840,15 @@ task cleanAll {
...
   dependsOn getTasksByName('testClasses', true).collect { it.path }
   dependsOn distTar, distZip
   mustRunAfter cleanAll
 }
+task buildDtests {
+  delete "${snappyProductDir}/dtests/build-artifacts/scala-2.11/"
```
rishitesh: path can be changed as 
 "${snappyProductDir}/dtests/build-artifacts/scala-${scalaBinaryVersion}/"

build.gradle
```diff
@@ -1007,6 +1012,10 @@ task publishDocs(type:Exec) {
 task precheckin {
   dependsOn cleanAll, buildAll, checkAll, allReports
   dependsOn ':snappy-spark:scalaStyle', docs
+  if (!project.hasProperty('smoke.skip')) {
```
rishitesh: smoke test will run by default ?

dtests/src/test/scala/io/snappydata/hydra/SnappyHydraRunner.scala
```diff
@@ -0,0 +1,91 @@
...
+    val c11 = "grep -v Q36"
+    val command2 = c9 #| c10 #| c11
+    try {
+      val output2: String = command2.!!
+      throw new Exception(s"Result mismatch observed in smoke bt run:\n " + output2 +
+          "\nCheck the logs at " + logDir + " for more details.")
+    }
+    catch {
+      case r: java.lang.RuntimeException =>
+        println("smoke bt run is successful.")
```
rishitesh: Why RunTimeException marks the BT as successful ?

dtests/src/test/scala/io/snappydata/hydra/SnappyHydraRunner.scala
```diff
@@ -0,0 +1,91 @@
...
+    if (logDir.exists) {
+      FileUtils.deleteDirectory(logDir)
+    }
+    /* val command: String =  s"$SNAPPYDATA_SOURCE_DIR/store/tests/core/src/main/java/bin/sample" +
+      s"-runbt.sh $logDir $SNAPPYDATA_SOURCE_DIR -d false io/snappydata/hydra/nwSmoke.bt" */
+    val command: String = s"$SNAPPYDATA_SOURCE_DIR/dtests/src/test/java/io/snappydata/hydra" +
+        s"/smoke.sh $SNAPPYDATA_SOURCE_DIR $logDir"
+    val (out, err) = executeProcess("smokeBT", command)
+
+    import scala.sys.process._
```
rishitesh: move the import to class level. If other tests added we need to reimport again.

build.gradle
swatisawant: Running all tasks from precheckin separately works fine(e.g. ./gradlew cleanAll, buildAll, checkAll), but running the same as ./gradlew precheckin doesn't work in case of smoke bt run.  
In this case, adding product target dependency in buildAll resolved above issue.
Here, I tried to creating a separate copy task and then explicitly calling it , but it didn't work. So as a quick solution, I added above dependency.

build.gradle
swatisawant: Yes, smoke bt will run by default.

dtests/src/test/scala/io/snappydata/hydra/SnappyHydraRunner.scala
swatisawant: For the grep command I used for searching the unexpected exceptions, when there are no exceptions observed(command returns no output) then the RunTimeException is thrown.

rishitesh: Does it contain some specific message, which can be trapped? My worry is if ur command fails due to some other reason also, then we will mark the BT successful.

dtests/src/test/java/io/snappydata/hydra/cluster/startDualModeCluster.bt
```diff
@@ -0,0 +1,5 @@
+io/snappydata/hydra/cluster/startDualModeCluster.conf
```
sonalsagarwal: The name startDualModeCluster.conf is already used for starting snappy cluster and smart mode cluster and purpose for both is different. We should think on renaming one of these as it may be confusing later, when we refactor the tests.

swatisawant:
There will be only one conf file going forward. And the one exists in northWind package will be removed later on, but just not to mix refactoring changes in this PR, I have kept that part untouched.
Going forward there will be just different local.conf for using this bt for long running cluster and normal cluster.

dtests/src/test/scala/io/snappydata/hydra/LoadCSVDataJob.scala
```diff
@@ -21,8 +21,8 @@ import org.apache.spark.sql.{SnappySQLJob, _}
 import spark.jobserver.{SparkJobValid, SparkJobValidation}
 
 /**
- * Created by swati on 3/8/16.
- */
+  * Created by swati on 3/8/16.
```
sonalsagarwal: Observed that you have removed this comment, may be this was missed.

PR #578, opened 2017-05-02
[Snap 1496] Wide table scan for column tables fails due to 64K limitation of JVM

rishitesh:
Issue: During table scan, if the number of columns projected is big then generated code generates a huge processNext() method.

The avoid that we have chunked the different parts of the scan code if the number of columns exceeds 30. 
Also the PR disables Snappy store optimization such as SnappyHashAggregates & HashJoin and fallbacks to Spark plans

## Patch testing

pre-checkin

## ReleaseNotes.txt changes

NA

## Other PRs 
NA


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -360,6 +358,8 @@ class SnappySession(@transient private val sc: SparkContext,
     SnappySession.clearSessionCache(id)
   }
 
+  private[sql] var disableStoreOptimizations : Boolean = false
```
hbhanawat: Will be good to have this as part of SnappySessionState.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1793,14 +1793,23 @@ object SnappySession extends Logging {
...
           (null, null)
         } else {
-          evaluatePlan(df, session, key.sqlText, key)
+          try {
+            evaluatePlan(df, session, key.sqlText, key)
+          } catch {
+            case e: CodeCompileException => {
+              session.disableStoreOptimizations = true
+              val df = session.executeSQL(key.sqlText)
+              evaluatePlan(df, session, key.sqlText, key)
```
hbhanawat: Any specific reason to not make disableStoreOptimizations as false after evaluatePlan, may be in a try finally block?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1818,6 +1827,8 @@ object SnappySession extends Logging {
     }
   }
 
+  val SNAPPY_STORE_OPTIMIZATION_DISABLED = "false"
```
hbhanawat: Unused.

core/src/main/scala/org/apache/spark/sql/execution/CachedPlanHelperExec.scala
```diff
@@ -97,7 +98,19 @@ case class CachedPlanHelperExec(childPlan: CodegenSupport, @transient session: S
...
+    // This does not handle nested aggregates or joins. Should we handle ?
+    val repeatSeq = childPlan.collect {
+      case p: NonRecursivePlans => !p.producedForPlanInstance
+      case _ => true
+    }
+    val shouldExecute = repeatSeq.reduce((a , b) => a && b)
+    if (shouldExecute) {
+      childPlan.execute()
+    } else {
+      throw new CodeCompileException("Could should not have reached here")
```
hbhanawat: Fix the exception string.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1793,14 +1793,23 @@ object SnappySession extends Logging {
...
         // since plan can change once caching is done (due to size stats)
         if (plan.find(_.isInstanceOf[InMemoryTableScanExec]).isDefined) {
           (null, null)
         } else {
-          evaluatePlan(df, session, key.sqlText, key)
+          try {
+            evaluatePlan(df, session, key.sqlText, key)
+          } catch {
+            case e: CodeCompileException => {
+              session.disableStoreOptimizations = true
```
hbhanawat: A debug log message would be good here.

core/src/main/scala/org/apache/spark/sql/execution/CachedPlanHelperExec.scala
```diff
@@ -97,7 +98,19 @@ case class CachedPlanHelperExec(childPlan: CodegenSupport, @transient session: S
   override def doConsume(ctx: CodegenContext, input: Seq[ExprCode], row: ExprCode): String =
     parent.doConsume(ctx, input, row)
 
-  override protected def doExecute(): RDD[InternalRow] = childPlan.execute()
+  override protected def doExecute(): RDD[InternalRow] = {
+    // This does not handle nested aggregates or joins. Should we handle ?
+    val repeatSeq = childPlan.collect {
+      case p: NonRecursivePlans => !p.producedForPlanInstance
```
hbhanawat: Why do we need methods/vars in NonRecursivePlans? Any Node that implements NonRecursivePlans we, return false. What is the use of calling startProducing at the beginning of doProduce when we know that producedForPlanInstance is always be going to be true for this node.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -428,21 +494,71 @@ private[sql] final case class ColumnTableScan(
...
+      if (isWideSchema) {
+        if (bufferInitCode.length > 1024) {
+          bufferInitCodeBlocks.append(bufferInitCode.toString())
+          bufferInitCode.clear()
+        }
+
+        bufferInitCode.append(
+          s"""
+          final $decoderClass $decoderVar = $decoder;
+          $bufferVar = ($buffer == null || $buffer.isDirect()) ? null
```
hbhanawat: This code can be refactored to declare bufferVar of non-wide case before the code block. Rest of the part should remain same.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -41,7 +39,7 @@ import org.apache.spark.sql.catalyst.rules.Rule
 import org.apache.spark.sql.collection.Utils
 import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.columnar.impl.IndexColumnFormatRelation
-import org.apache.spark.sql.execution.datasources.{DataSourceAnalysis, FindDataSourceTable, HadoopFsRelation, LogicalRelation, ResolveDataSource, StoreDataSourceStrategy}
+import org.apache.spark.sql.execution.datasources._
```
hbhanawat: I believe you don't have any changes to this class. You can leave it as it is

core/src/main/scala/org/apache/spark/sql/internal/CodeCompileException.scala
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+package org.apache.spark.sql.internal
+
+
+class CodeCompileException(msg : String) extends Exception(msg) {}
```
rishitesh: Ok. Will change it.

core/src/main/scala/org/apache/spark/sql/execution/CachedPlanHelperExec.scala
rishitesh: If I won't check for produced for plan instance it would again execute the same plan. Note that  CachedPlanHelperExec child is the plan itself. It's basically to break the recursion.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
rishitesh: I was getting some weird errors in finally block. Have not investigated further.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
rishitesh: I will be chunking the whole string into a method in case of wide-schema. note the code where I clear bufferInitCode ( bufferInitCode.clear). Moving code outside the block will break the local variables of the created code.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
rishitesh: Will do that.

core/src/main/scala/org/apache/spark/sql/execution/CachedPlanHelperExec.scala
rishitesh: Will change it

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
rishitesh: Will remove it.

core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
```diff
@@ -301,7 +317,7 @@ private[sql] object JoinStrategy {
  *
  * Adapted from Spark's Aggregation strategy.
  */
-object SnappyAggregation extends Strategy {
+private[sql] object SnappyAggregationStrategy extends Strategy {
```
rishitesh: The problem was for the session object, which is not available in SnappyAggregationStrategy. SnappyStrategies is constructed with a DefaultPlanner , hence could get hold of the session object easily. Not so with SnappyAggregationStrategy. Did not want to do a lot of refactoring to get the flag.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
rishitesh: Added a proper debug message.

rishitesh:
Merging this change.
Tests failed 
a) ColumnTableDunitTests ( passed on rerun)
b) ExecutorMessageDUnitTest  ( passed on rerun)
c) ClosedFormAQPDataFrameAPIPart1Test  ( passed on rerun)

PR #580, opened 2017-05-03
SNAP-1481 : Spark UI SQL tab needs to show the SQL query

snappy-sachin:

- UI SQL Tab now displays SQL Query string instead of CachedDataFrame handler in Description column.
- Queries executed from Snappy shell and spark shell will be displayed as actual query string under Description column.
- Queries executed using API will display CachedDataFrame handler description as per previous behavior.

## Patch testing

- Tested manually

## ReleaseNotes.txt changes

- SQL Tab on UI, Description column now displays actual SQL Query string executed instead of handler description text.

## Other PRs 

- None.

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -7,6 +7,12 @@
 # ADD THE JIRA TICKET ID, IF APPLICABLE.                                                           #
 ####################################################################################################
 
+Release 0.9
+
+- New Features/Fixed Issues
+
+  [SNAP-1481] SQL Tab on UI, Discrption column now displays actual SQL Query string executed inseted of handler 
+  description text.
 
 Release 0.8
 
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -11,8 +11,8 @@ Release 0.9
 
 - New Features/Fixed Issues
 
-  [SNAP-1481] SQL Tab on UI, Discrption column now displays actual SQL Query string executed inseted of handler 
-  description text.
+  [SNAP-1481] SQL Tab on UI, Description column now displays actual SQL Query string executed
+  instead of handler description text.
 
 Release 0.8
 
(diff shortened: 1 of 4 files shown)


sumwale:
@kneeraj @rishitesh I had added the string in CachedDataFrame but the problem with ParamLiteral is that it will show only the first string which was cached. Can you help @snappy-sachin to pass through the current string to CachedDataFrame.collectWithHandler? I think what needs to be done is make the queryString in CachedDataFrame a "var" and then change it with the current sqlText before returning it in SnappySession.getPlan

ReleaseNotes.txt
```diff
@@ -7,6 +7,12 @@
 # ADD THE JIRA TICKET ID, IF APPLICABLE.                                                           #
 ####################################################################################################
 
+Release 0.9
+
+- New Features/Fixed Issues
+
+  [SNAP-1481] SQL Tab on UI, Discrption column now displays actual SQL Query string executed inseted of handler
```
sumwale: typo "inseted"

snappy-sachin: corrected.. Thanks Sumedh

sumwale:
For now we can have manual checking will be fine. But a unit test will be good especially for the ParamLiteral case. Register a listener on SparkListenerBus in the test for SparkListenerSQLExecutionStart and verify that strings are as expected (both short and long forms). @kneeraj

ReleaseNotes.txt
ashetkar: Please break the line after 100 char as has been followed.
Typo 'Discrption'

snappy-sachin:
Run precheckin

snappy-sachin:
Ran precheckin and no related failure. 
Merging..

sumwale:
@kneeraj @snappy-sachin @rishitesh 

PLEASE DO NOT DO MERGES OR APPROVE WITHOUT RESPONSES TO THE COMMENTS. Usually reviewers won't have time to come back to check if no responses are provided to those, and merges done without addressing them.

There were two important points in the comments:

a) It will not work correctly for tokenized parameters. This was a MUST FIX before merge.

b) Unit test.

Now that merge is done, track above two in a separate ticket (or responses on why these points can be ignored).

snappy-sachin:
@sumwale , 
I have created a task (_SNAP-1532_) for adding unit test but forgot to add that in this PR while merging.
I just have added new task(_SNAP-1539_) for making the queryString in CachedDataFrame a "var" and then change it with the current sqlText before returning it in SnappySession.getPlan.

PR #581, opened 2017-05-04
SNAP-1351

hbhanawat:

When tasks are killed, the task threads are getting interrupted. This causes issues with snappy because snappy may be writing to an oplog and it generates a DiskAccessException. This DAE ends up closing the underlying regions.

Snappy threads cannot be interrupted.

## Patch testing

Tried out the scenario mentioned in the bug. Works fine. Precheckin. 

## Other PRs 
https://github.com/SnappyDataInc/spark-jobserver/pull/9

### DIFF
--- a/cluster/src/main/scala/org/apache/spark/executor/SnappyCoarseGrainedExecutorBackend.scala
+++ b/cluster/src/main/scala/org/apache/spark/executor/SnappyCoarseGrainedExecutorBackend.scala
@@ -92,7 +92,7 @@ class SnappyCoarseGrainedExecutorBackend(
     if (executor != null) {
       // kill all the running tasks
       // InterruptThread is set as true.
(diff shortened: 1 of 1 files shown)


soubhik-c:
Okay

cluster/src/main/scala/org/apache/spark/executor/SnappyCoarseGrainedExecutorBackend.scala
```diff
@@ -92,7 +92,7 @@ class SnappyCoarseGrainedExecutorBackend(
     if (executor != null) {
       // kill all the running tasks
       // InterruptThread is set as true.
```
sumwale: can you change this comment and also add the explanation as in the PR description?

PR #585, opened 2017-05-10
Added a check to get buckets from initialized members only.

rishitesh:
We follow two separate code path for getting PR buckets and their live owners. One is for linked partition case, another for unlinked one. 

Added a check to get buckets from initialized members only. If it does not get one, it would create a bucket for read, initialize it and return that node. 

## Patch testing
pre-checkin @TODO

## ReleaseNotes.txt changes

NA

## Other PRs 
NA


kneeraj:
forWrite is a boolean which has been added and that is why probably the conflict. You might need to reconcile your changes and then looking at the changes will make sense.

PR #586, opened 2017-05-10
[SNAP-1344] As streaming jobs are recurring jobs, the earlier mechanism of remov…

rishitesh:

As streaming jobs are recurring jobs, the earlier mechanism of removing dependent jar files were broken.
 This change will ensure we maintain a list of jars in the context itself and will remove the jars when the streaming context is stopped.
 When streaming context is stopped we can safely remove those jars.

## Patch testing

pre-checkin (todo)

## Other PRs 
https://github.com/SnappyDataInc/spark-jobserver/pull/10


SachinJanani:
Run precheckin

SachinJanani:
Run precheckin

cluster/src/main/scala/org/apache/spark/util/SnappyUtils.scala
```diff
@@ -50,6 +50,19 @@ object SnappyUtils {
...
   }
 
+  def removeJobJar(sc: SparkContext, jarName : String): Unit = {
+    def getName(path: String): String = new File(path).getName
+    val keyToRemove = sc.listJars().filter(getName(_) == getName(jarName))
+    if (keyToRemove.nonEmpty) {
+      val callbacks = ToolsCallbackInit.toolsCallback
+      //@TODO This is a temp workaround to fix SNAP-1133. sc.addedJar should be directly be accessible from here.
+      //May be due to scala version mismatch.
+      if(callbacks != null){
```
sumwale: formatting

cluster/src/main/scala/org/apache/spark/sql/streaming/SnappyStreamingContextFactory.scala
```diff
@@ -61,13 +56,18 @@ class SnappyStreamingContextFactory extends SparkContextFactory {
     new SnappyStreamingContext(LeadImpl.getInitializingSparkContext,
       Milliseconds(interval)) with ContextLike {
 
+      private val addedJars = scala.collection.mutable.ArrayBuffer.empty[String]
```
sumwale: don't need synchronized access?

rishitesh: No, at a time only one StreamingContext will be active.

cluster/src/main/scala/org/apache/spark/util/SnappyUtils.scala
rishitesh: Yes. That's the reason why I have added jar remove code in streaming context stop() method of SnappyStreamingContext , which we are extending in case of jobs.

sumwale: Ok, above comment was about formatting: lack of spaces after if and before brace. Also after comment slashes and limit to < 100 chars as per scala-style.

cluster/src/main/scala/org/apache/spark/sql/streaming/SnappyStreamingContextFactory.scala
sumwale: But addJobJar can be called by multiple threads no? Also didn't see any caller for addJobJar.

rishitesh: Hmm. You mean to see two threads trying submitting two streaming jobs at a time. Yes , probably we should guard that condition.
addJobJar will be getting called from JobServer. Please see the related PR  https://github.com/SnappyDataInc/spark-jobserver/pull/10

sumwale: @rishitesh If you agree that concurrent access is possible, then this should certainly be made thread-safe (synchronized or a concurrent map).

rishitesh: @sumwale I checked the code. This is valid only for streaming context. The jars being added here only concern when we submit a streaming jobs. Hence one streaming context. That means it is just a object level attribute which is set when the  context is created. There is no other code path where this jars are added. 
Other context will have its own list. Ultimately one context will fail saying only one streaming contetxt is allowed.  Hence we can do away with the sync block.

sumwale: @rishitesh Does this hold for the new Spark streaming too? There is no single streaming context, rather new source/sink registered as required. Perhaps it requires a new job interface, but if addJobJar is used there too then it will cause trouble (which will certainly be hard to detect for someone just using the method). Considering that not-so-future case this should be made thread-safe assuming it can be trouble then.
@ymahajan

sumwale: @rishitesh Also it is not clear to me how a single StreamingContext obviates the need to make this thread-safe. Can't multiple concurrent job submissions use the same StreamingContext end up adding jars, then a later one actually starts the context?

rishitesh: New Spark streaming will add additional complexities. Its a good point you raised. Structured streaming does not require an explicit streaming context to be created. A simple SparkSession ( and hence SnappySession) is enough to start a streaming job.

Theoretically it can be as easy as 
````object SomeStreamJob extends SnappySQLJob {
override def runSnappyJob(snc: SnappySession, jobConfig: Config): Any = {
val lines = snc.readStream
      .format("socket")
      .option("host", host)
      .option("port", port)
      .load().as[String]
}`

Should we explicitly stop starting streaming jobs in this fashion or we should support it. Supporting this would means out job jar clean up code needs more work. We can not clean a job jar ( from executors)  if its a streaming job . Then the question remains how to handle same class multiple versions ( for multiple stream jobs) ? Also multiple stream sessions can run concurrently. 

Also another problem is how to stop a running query. It is a query.stop and not a global one like StreaminngContext.stop(). That means query reference needs to be stored in job server against some id and that id needs to be returned ( shown in UI) to users. At some later point of time one can come and stop a query. Need to see if this is being worked upon with job-server. If this can be achieved class related problem can be taken care of.

rishitesh: @sumwale Regarding the later question,  multiple concurrent streaming jobs will create multiple SnappyStreaming Context objects just like multiple normal jobs will create multiple SnappySession objects. Each job ( hence each object)  will add their respective dependent URLs in their objects.

sumwale: @rishitesh I am confused on this. Isn't SnappyStreamingContext a singleton?

rishitesh: Not really, its usage we have suggested makes it singleton like. You can always do new SnappyStreamingContext.  Job-server ( our changes) usage this to create s new SnappyStreamingContext every time a stream job is submitted.  Only when one starts a streaming context it validates if other context is running or not.

sumwale: @rishitesh Is it so and do our docs say so about SnappyStreamingContext that only one thread can access it at a time? Do our docs say that only one streaming job can be submitted at a time?

rishitesh: @sumwale From our docs 
To start another streaming job with a new streaming context, you need to first stop the currently running streaming job, followed by its streaming context.

Probably we need to be more explicit.

sumwale: @rishitesh I think there is some confusion here. I am not asking if two StreamingContexts can be started which obviously cannot be due to Spark limitation. The question is: do we restrict users from submitting tasks to the same StreamingContext (e.g. gotten with getOrCreate) from different jobs? Are we documenting that do not attempt to submit tasks against the same context from two different jobs using REST/JDBC API?

PR #588, opened 2017-05-11
[SNAP-1542] Ensure that the size of buffer is at least the minimum required

sumwale:

- Reused buffer size should be at least the minimum required else create new one
- Also use repartition(1) in test result collection instead of coalesce(1)
so that the scan is parallel. Reduces its running time by an order of magnitude.

## Patch testing

hydra test mentioned in SNAP-1542 and precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/ColumnEncoding.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/ColumnEncoding.scala
@@ -310,7 +310,7 @@ trait ColumnEncoder extends ColumnEncoding {
     }
     if ((columnData eq null) || (columnData.limit() < (baseSize + defSize))) {
       var initByteSize = 0L
-      if (reuseUsedSize > baseSize) {
+      if (reuseUsedSize > baseSize + defSize) {
         initByteSize = reuseUsedSize
       } else {
         initByteSize = defSize.toLong * initSize + baseSize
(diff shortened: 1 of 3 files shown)


sumwale:
@swatisawant please see the test change

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/ColumnEncoding.scala
```diff
@@ -304,15 +304,16 @@ trait ColumnEncoder extends ColumnEncoding {
...
+    if ((columnData eq null) || (columnData.limit() < (baseSize + defSize))) {
+      var initByteSize = 0L
+      if (reuseUsedSize > baseSize + defSize) {
         initByteSize = reuseUsedSize
       } else {
-        initByteSize = defSize.toLong * initSize + numNullBytes
-        if (withHeader) {
-          initByteSize += 8L /* typeId + nullsSize */
-        }
+        initByteSize = defSize.toLong * initSize + baseSize
```
hbhanawat: Not sure what this is doing.

sumwale: The baseSize already has the "numNullBytes" + header of 8 bytes, so removed from here and added baseSize.

PR #590, opened 2017-05-12
Snap 1409

kneeraj:
Tokenization fixes:
Includes
- Fix for stale / wrong results from coming due to tokenization. Queries with scalar subqueries or like etc. which consume the constant provided to materialize a part of the query are identified and are not attempted for Tokenization. As a catch all we just compare whether the number of constants to be replaced after plan generation is same as what the parser thought was there. If there is a mismatch it does not tries Tokenization of such queries. 
- Plans with broadcast are not attempted for Tokenization at all. Will be taken later.

## Patch testing

New dunit added in TPCHDUnit test for tokenization in both embedded and spli cluster mode.

## ReleaseNotes.txt changes

Re-use of generated plan is attempted if the same query is fired with just different constant values.

## Other PRs 

None.

PR of the other subprojects that are related to this change)


cluster/src/test/scala/io/snappydata/cluster/QueryRoutingSingleNodeSuite.scala
```diff
@@ -239,8 +278,9 @@ class QueryRoutingSingleNodeSuite extends SnappyFunSuite with BeforeAndAfterAll
       // println("network server started")
       insertRows(tableName1, 1000, serverHostPort)
       insertRows(tableName2, 1000, serverHostPort)
-      query1(tableName1, tableName2, serverHostPort)
-      query2(tableName1, tableName2, serverHostPort)
+      // query1(tableName1, tableName2, serverHostPort)
+      (0 to 5).foreach(i => query2snc(tableName1, tableName2, serverHostPort, i))
+      // query2(tableName1, tableName2, serverHostPort)
```
vivekwiz: I could not recall but is query1 and query2 is disabled for Shoubhik's checkin?

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -49,26 +52,38 @@ import org.apache.spark.unsafe.Platform
 import org.apache.spark.util.CallSite
 import org.apache.spark.{Logging, NarrowDependency, ShuffleDependency, SparkContext, SparkEnv, SparkException, TaskContext}
 
-class CachedDataFrame(df: Dataset[Row], val queryString: String,
+class CachedDataFrame(df: Dataset[Row], var queryString: String,
```
vivekwiz: I did not see why queryString should be var?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1731,6 +1733,8 @@ object SnappySession extends Logging {
         ArrayBuffer[Any]]](CachedPlanHelperExec.BROADCASTS_KEY).getOrElse(
       mutable.Map.empty[BroadcastHashJoinExec, ArrayBuffer[Any]])
 
+    logDebug(s"all bc plans = ${allbroadcastplans} ... size = ${allbroadcastplans.size}")
+
```
vivekwiz: Need this log?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1775,6 +1783,44 @@ object SnappySession extends Logging {
 
       allLiterals.foreach(_.collectedForPlanCaching = true)
     }
+
+    logDebug(s"qe.executedPlan = ${df.queryExecution.executedPlan}")
```
vivekwiz: Need this log?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1739,8 +1743,8 @@ object SnappySession extends Logging {
             plan.childRDD.id, true)
       case _: LocalTableScanExec =>
         (null, Array.empty[Int], -1, false) // cache plan but no cached RDD
-      case _ if allbroadcastplans.nonEmpty =>
-        (null, Array.empty[Int], -1, false) // cache plan but no cached RDD
+      // case _ if allbroadcastplans.nonEmpty =>
+        // (null, Array.empty[Int], -1, false) // cache plan but no cached RDD
```
vivekwiz: Can we add comment if you want to keep it disabled?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1775,6 +1783,44 @@ object SnappySession extends Logging {
...
+                  s" or plans with broadcast hashjoins")
+              key.invalidatePlan()
+            }
+          })
+        }
+      }
+    }
+
+    if (key != null && key.valid) {
+      logDebug(s"Plan caching will be used for sql ${key.sqlText}")
```
vivekwiz: log?

core/src/main/scala/org/apache/spark/sql/execution/CachedPlanHelperExec.scala
```diff
@@ -64,7 +64,9 @@ case class CachedPlanHelperExec(childPlan: CodegenSupport, @transient session: S
     var nextStageStarted = false
     var alreadyGotBroadcastNode = false
     childPlan transformDown {
-      case bchj: BroadcastHashJoinExec =>
+      case bchj: BroadcastHashJoinExec => {
+        logDebug(s"Got a bchj = ${bchj} and nextstagestarted = ${nextStageStarted}")
+
```
vivekwiz: log?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -516,6 +516,8 @@ class SmartConnectorRowRDD(_session: SnappySession,
       ps.close()
     }
     val sqlText = s"SELECT $columnList FROM $resolvedName$filterWhereClause"
+     println(s"KN: pushed SQL to GemXD: $sqlText, " +
```
vivekwiz: remove this

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -182,6 +201,21 @@ class CachedDataFrame(df: Dataset[Row], val queryString: String,
     }
   }
 
+  private[sql] def waitForLastShuffleCleanup(): Unit = {
```
vivekwiz: I dont see this as used

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -254,6 +288,9 @@ class CachedDataFrame(df: Dataset[Row], val queryString: String,
               results(index) = resultHandler(index, r._1))
           results.iterator
       }
+      // submit cleanup job for shuffle output
+      // clearCachedShuffleDeps(sc)
+
```
vivekwiz: remove above lines

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -272,18 +309,24 @@ class CachedDataFrame(df: Dataset[Row], val queryString: String,
...
+            x
         }
+        println(s"newbchj = ${newbchj}")
         val tmpCtx = new CodegenContext
         val parameterType = tmpCtx.getClass
         val method = newbchj.getClass.getDeclaredMethod("prepareBroadcast", parameterType)
         method.setAccessible(true)
         val bc = method.invoke(newbchj, tmpCtx)
-        refs(broadcastIndex) = bc.asInstanceOf[(Broadcast[_], String)]._1
+        println(s"replacing bc var = ${refs(broadcastIndex)} with " +
```
vivekwiz: remove printf

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1763,6 +1767,10 @@ object SnappySession extends Logging {
         (rdd, findShuffleDependencies(rdd).toArray, rdd.id, false)
     }
 
+    val allallbroadcastplans = session.getContextObject[mutable.Map[BroadcastHashJoinExec,
```
vivekwiz: I dont see as used

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: Will be required later when broadcast plans will also be working with Tokenization.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: Ok.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: Yes it was being usd. Will keep it around as in the next phase of Tokenization this might be taken up.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: Yes this helps in analysis. Anyways it is in debug.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: Could not understand what is being pointed out here.

core/src/main/scala/org/apache/spark/sql/execution/CachedPlanHelperExec.scala
kneeraj: Again. What exactly is the issue.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
kneeraj: yes. slipped from my check while cleanup.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
vivekwiz: Is this log required?

kneeraj: Right now using it to really see whether as per expectation plan caching will be used or not. May be we can remove later.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
kneeraj: This entire method I am planning to make either no-op or not call from session.getPlan for now.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
kneeraj: This is because on the UI always the actual string should b edisplayed even if the cached plan is being used. The cached plan made could have been due to different set of constants but the current execution should reflect the actual current query executed.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
kneeraj: This was added to try waiting for cleanup of previous run of the cached plan so that it does not interferes with the current execution. The current execution should not wait for the cleanup synchronously.

vivekwiz:
Please include commented test of PreparedQueryRoutingSingleNodeSuite (search for 1478)

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -49,26 +52,38 @@ import org.apache.spark.unsafe.Platform
...
 
   /**
-   * Return true if [[collectWithHandler]] supports partition-wise separate
-   * result handling by default, else result handler is invoked for a
-   * single partition result.
-   */
+    * Return true if [[collectWithHandler]] supports partition-wise separate
+    * result handling by default, else result handler is invoked for a
+    * single partition result.
+    */
```
sumwale: Check your fomatting settings -- I think original one was correct as per Spark and in codeStyleSettings.xml. Change IDEA Settings->Editor->Code Style to use "Project" scheme, close IDEA, copy codeStyleSettings.xml to .idea, then start.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -117,9 +136,9 @@ class CachedDataFrame(df: Dataset[Row], val queryString: String,
 
 
   /**
-   * Wrap a Dataset action to track the QueryExecution and time cost,
-   * then report to the user-registered callback functions.
-   */
+    * Wrap a Dataset action to track the QueryExecution and time cost,
+    * then report to the user-registered callback functions.
+    */
```
sumwale: check formatting

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -272,18 +309,24 @@ class CachedDataFrame(df: Dataset[Row], val queryString: String,
       newpls: mutable.ArrayBuffer[ParamLiteral]): Unit = {
     if (allbcplans.nonEmpty && !firstAccess) {
       allbcplans.foreach { case (bchj, refs) =>
+        println(s"Repreparing for bcplan = ${bchj} with new pls = ${newpls.toSet}")
```
sumwale: remove println (perhaps logDebug?)

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -272,18 +309,24 @@ class CachedDataFrame(df: Dataset[Row], val queryString: String,
...
         val broadcastIndex = refs.indexWhere(_.isInstanceOf[Broadcast[_]])
         val newbchj = bchj.transformAllExpressions {
           case pl@ParamLiteral(_, _, p) =>
             val np = newpls.find(_.pos == p).getOrElse(pl)
-            ParamLiteral(np.value, np.dataType, p)
+            val x = ParamLiteral(np.value, np.dataType, p)
+            x.considerUnequal = true
+            x
         }
+        println(s"newbchj = ${newbchj}")
```
sumwale: remove println

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1854,13 +1900,10 @@ object SnappySession extends Logging {
         case a: Alias =>
           Alias(a.child, a.name)(exprId = ExprId(0))
         case l@ListQuery(query, _) =>
-          val xx = l.copy(query =
-              query.transform(transformExprID),
+          l.copy(query = query.transformAllExpressions(normalizeExprIds),
```
hbhanawat: We don't have to apply this rule to the children of ListQuery? listQuery as I can see have child expressions.

core/src/main/scala/org/apache/spark/sql/execution/CachedPlanHelperExec.scala
```diff
@@ -86,7 +88,17 @@ case class CachedPlanHelperExec(childPlan: CodegenSupport, @transient session: S
           }
           alreadyGotBroadcastNode = true
         }
+        bchj transformAllExpressions {
+          case pl: ParamLiteral =>
+            session.getContextObject[mutable.Map[BroadcastHashJoinExec, ArrayBuffer[Any]]](
```
hbhanawat: If we can setting the CachedPlanHelperExec.NOCACHING_KEY in the code above (line 77), we don't need this check?

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -49,26 +52,38 @@ import org.apache.spark.unsafe.Platform
...
+    * single partition result.
+    */
   def hasPartitionWiseHandling: Boolean = cachedRDD ne null
 
   private lazy val boundEnc = exprEnc.resolveAndBind(logicalPlan.output,
     sparkSession.sessionState.analyzer)
 
   private lazy val queryExecutionString = queryExecution.toString()
-  private lazy val queryPlanInfo = PartitionedPhysicalScan.getSparkPlanInfo(
-    queryExecution.executedPlan)
```
sumwale: Also make queryExecutionString as a def since the plan output will also need to be changed with different literals.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -272,18 +309,24 @@ class CachedDataFrame(df: Dataset[Row], val queryString: String,
...
     if (allbcplans.nonEmpty && !firstAccess) {
       allbcplans.foreach { case (bchj, refs) =>
+        println(s"Repreparing for bcplan = ${bchj} with new pls = ${newpls.toSet}")
         val broadcastIndex = refs.indexWhere(_.isInstanceOf[Broadcast[_]])
         val newbchj = bchj.transformAllExpressions {
           case pl@ParamLiteral(_, _, p) =>
             val np = newpls.find(_.pos == p).getOrElse(pl)
-            ParamLiteral(np.value, np.dataType, p)
+            val x = ParamLiteral(np.value, np.dataType, p)
+            x.considerUnequal = true
```
sumwale: This is to force mismatch of bchj plan if there are tokens, right?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1717,12 +1719,12 @@ object SnappySession extends Logging {
       val nocaching = session.getContextObject[Boolean](
         CachedPlanHelperExec.NOCACHING_KEY).getOrElse(false)
       if (nocaching) {
-        key.invalidatePlan()
+        key.invalidatePlan
```
sumwale: Revert. Method calls returning Unit should have parens as per scalaStyle guidelines (not enforced though).

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1717,12 +1719,12 @@ object SnappySession extends Logging {
...
       }
       else {
         val params1 = getAllParamLiterals(executedPlan)
-        if (!params1.sameElements(key.pls)) {
-          key.invalidatePlan()
+        if (!(params1.sameElements(key.pls))) {
+          key.invalidatePlan
```
sumwale: Why additional parens? Also add back parens to invalidatePlan.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
sumwale: So now you are using bchj mismatch instead of above to avoid caching? Does the bchj mismatch approach still result in caching albeit separate for each token value like was done without tokenization changes?

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
```diff
@@ -61,10 +64,12 @@ class ParamLiteral(_value: Any, _dataType: DataType, val pos: Int)
     31 * (31 * Objects.hashCode(dataType)) + Objects.hashCode(pos)
   }
 
+  var considerUnequal = false
+
   override def equals(obj: Any): Boolean = obj match {
     case a: AnyRef if this eq a => true
     case pl: ParamLiteral =>
-      pl.dataType == dataType && pl.pos == pos
+      pl.dataType == dataType && pl.pos == pos && !(pl.considerUnequal || considerUnequal)
```
sumwale: I think this condition means that if either of the ParamLiterals are marked as "unequal" i.e. participant in BCHJ params, then force them to be unequal. This will ensure same plan is not used for two different tokens in BCHJ (which probably also means that each of those will be cached separately like without tokenization for which I have asked clarification from @kneeraj)

cluster/src/test/scala/io/snappydata/cluster/QueryRoutingSingleNodeSuite.scala
kneeraj: No. We just modified the test to run in loop and check the plan caching for proper working again and again.
The failure which was coming due to the pruning should  be a separate test and let us identify and pass it on to Shoubhik for testing.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
kneeraj: Ok.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -77,7 +92,11 @@ class CachedDataFrame(df: Dataset[Row], val queryString: String,
         case Some(cleaner) =>
           var i = 0
           while (i < numShuffleDeps) {
-            cleaner.doCleanupShuffle(shuffleDependencies(i), blocking = false)
+            val shuffleDependency = shuffleDependencies(i)
+            cleaner.doCleanupShuffle(shuffleDependency, blocking = true)
```
kneeraj: Ok.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
kneeraj: Right now not attempting caching for broadcast so I am completely disabling this method.

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
kneeraj: As explained somewhere above, this was added to enforce inequality between two paramliterals even though their position and datatype are same. And yes this was added for transforming the broadcast sub plan and reprepare the broadcast variables. Better would be to replace the paramliterlas with a new literal object itself, which I am doing for injecting new constants for UI plan display.. So I am removing this.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
kneeraj: ok.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
kneeraj: Cleanup of previous execution artifact as communicated will come subsequently. Leaving the code related to those now as is so that tried options are there. Just for convinience.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1775,6 +1783,44 @@ object SnappySession extends Logging {
...
+    // This part is the defensive coding for all those cases where Tokenization
+    // support is not smart enough to deal with cases where the execution plan
+    // is modified in such a way that we cannot track those constants which
+    // need to be replaced in the subsequent execution
+    if (key != null) {
+      val nocaching = session.getContextObject[Boolean](
+        CachedPlanHelperExec.NOCACHING_KEY).getOrElse(false)
+      if (nocaching /* || allallbroadcastplans.nonEmpty */) {
+        logDebug(s"Invalidating the key because explicit nocaching")
+        key.invalidatePlan()
```
kneeraj: Well it depends on when the broadcast variable is prepared. In some cases a broadcast variable is materialized quite early. A separate context and code generated happens for that. In some cases it is part of the top level plan.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1775,6 +1783,44 @@ object SnappySession extends Logging {
...
+      }
+      else {
+        val params1 = getAllParamLiterals(executedPlan)
+        if (allLiterals.length != params1.length || !params1.sameElements(key.pls)) {
```
kneeraj: This has been added as a catchall for all such situations where there is a mismatch between the expected number of to be replaced constants and the actual number. Above it is being called only if 'nocaching' is false. I thnk we can leave it and revisit it once more and more types will be qualified and implemented for plan caching.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1775,6 +1783,44 @@ object SnappySession extends Logging {
...
+          logDebug(s"Invalidating the key because nocaching " +
+              s"allLiterals.length = ${allLiterals.length}," +
+              s" params1.length = ${params1.length} and key.pls = ${key.pls.length}")
+          key.invalidatePlan()
+        }
+        else if (params1.length != 0 ) {
+          params1.foreach(p => {
+            if (!allLiterals.exists(_.position == p.pos)) {
```
kneeraj: Again not only the length, verify the position as well. Yes should cange the log statement.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: Yes we do. In fact that was leading to a lot of problems. List elements can themselves be expressions (ParamLiterals)

core/src/main/scala/org/apache/spark/sql/execution/CachedPlanHelperExec.scala
kneeraj: No here it is like, if bchj is there but if it is not having ParamLiterals or constants then it is fine.

cluster/src/test/scala/io/snappydata/cluster/QueryRoutingSingleNodeSuite.scala
vivekwiz: Please enable query2

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
vivekwiz: Ok

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
vivekwiz: ok

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
vivekwiz: ok

core/src/main/scala/org/apache/spark/sql/execution/CachedPlanHelperExec.scala
vivekwiz: ok

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
vivekwiz: ok

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
vivekwiz: ok

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
vivekwiz: ok

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
sumwale: Ok for this commit. But if we are not going to address tokenization of bchj for 1.0 then perhaps we could add back caching with Literal for bchj as before. This can be done by changing back ParamLiteral to the Literal inside the broadcast plan sub-trees (which is equivalent of turning off tokenization for the selected portion of query). Can you track this for GA (either add tokenization for bchj which is preferred or else if too much work then change ParamLiteral to Literal instead of skipping caching entirely)?

cluster/src/test/scala/io/snappydata/cluster/QueryRoutingSingleNodeSuite.scala
vivekwiz: Please also enable one commented test (out of two) of PreparedQueryRoutingSingleNodeSuite (search for 1478)

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
sumwale: Still comment out the println because it breaks scalastyle. Too many reds in IDEA now in our code. Will be turning on scalaStyle by default for core, cluster, aqp soon.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
sumwale: Comment out the println as above or replace with logDebug. We should make it a practice to use logDebug's for all such cases which can be enabled selectively for the class when required in log4j.properties

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
sumwale: Right have mentioned the same above which should be done if we are not going to tackle tokenization for bchj soon.

cluster/src/test/scala/io/snappydata/cluster/QueryRoutingSingleNodeSuite.scala
kneeraj: Done.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: Created a ticket. https://jira.snappydata.io/browse/SNAP-1599

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
kneeraj: Done. And created a ticket SNAP-1601 also to clean this asynchronously.

PR #591, opened 2017-05-14
[SNAP-1555] filter out the DDL options in connection properties

sumwale:

- skip DDL options in ExternalStoreUtils when creating the set of connection properties
- corrected remaining instances of column batch size settings in tests to mean byte size (instead of number of rows)

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


sumwale:
run precheckin

PR #604, opened 2017-05-22
Snap 1518

dshirish:
Snap-1518: snappy-start-all does not start lead in large cluster

From log excerpt,  when lead was started a statement executed during Hive metsatore initialization failed as the region on which it executed was not yet initialized.
Unable to repro this issues even when just locator and lead is started (appears to be a timing issue)

As a defensive fix, in SnappyHiveCatalog.HMSQuery#initHMC function now retrying initialization of Hive metsatore if it fails due to SQLState.NO_DATASTORE_FOUND.

Let me know what you think.

## Patch testing
precheckin to be run prior to merge.  Opening the PR to get comments.

## ReleaseNotes.txt changes
NA
## Other PRs 
NA


core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
```diff
@@ -385,11 +385,44 @@ private void initHMC() {
...
+            noDataStoreFound = (t instanceof SQLException) &&
+                SQLState.NO_DATASTORE_FOUND.startsWith(
+                    ((SQLException)t).getSQLState());
+            t = t.getCause();
+          }
+          // wait for some time and retry if no data store found error
+          // is thrown due to region not being initialized
+          if (count < numRetries && noDataStoreFound) {
+            try {
+              Misc.getI18NLogWriter().warning(LocalizedStrings.DEBUG,
```
sumwale: slf4j logger should be used as standard for snappydata. This allows us to control logging module-wise easily in log4j.properties (plan is also to add a procedure/SET to change dynamically).

dshirish: @sumwale
Tried using following code but this doesn't log anything in the logs (even at warn and error levels). @hbhanawat  is looking into that part as this appears to be after effect SNAP-1420. Will change the code to use slf4j once that is resolved.
```
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
...
...
logger = LoggerFactory.getLogger(SnappyHiveCatalog.class);
...
...
 logger.error("SnappyHiveCatalog.HMSQuery.initHMC: No datastore found " +
                  "while initializing Hive metastore client. " +
                  "Will retry initialization after 500 milliseconds. " +
                  "Exception received is " + t);
         if (logger.isDebugEnabled()) {
            logger.error("Exception stacktrace:", ex);
         }

sumwale: Ok, we will need to make a sweep of snappy related logging later in any case.

PR #605, opened 2017-05-22
Fixing FileNotFoundException occured while running sparkApp for query validation

sonalsagarwal:
- Modifying test code to use SnappyTestUtils.writeToFile method for writing query result to csv.

- Changed default persistence mode to 'sync'.

## Patch testing
Ran ct.bt

### DIFF
--- a/dtests/src/test/scala/io/snappydata/hydra/TestUtil.scala
+++ b/dtests/src/test/scala/io/snappydata/hydra/TestUtil.scala
@@ -86,14 +86,14 @@ object TestUtil {
     var hasValidationFailed = validationFailed
 
     val snappyQueryFileName = s"Snappy_${queryNum}"
-    val snappyDest: String = getQueryResultDir("snappyQueryFiles") + File.separator +
+    val snappyDest: String = SnappyTestUtils.getTempDir("snappyQueryFiles") + File.separator +
         snappyQueryFileName
     pw.println(snappyDest)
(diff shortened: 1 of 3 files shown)


dtests/src/test/scala/io/snappydata/hydra/TestUtil.scala
```diff
@@ -179,7 +182,8 @@ object TestUtil {
       val logDir = log.listFiles.filter(_.getName.equals("snappyleader.log"))
       if (!logDir.isEmpty) {
         val leaderLogFile: File = logDir.iterator.next()
-        if (leaderLogFile.exists()) dest = dirString + File.separator + ".." + File.separator + ".." + File.separator + dirName
+        if (leaderLogFile.exists())
```
swatisawant: Use the SnappyTestUtils.getTempDir() method instead of this one as both are same.

sonalsagarwal: Done

sonalsagarwal:
The methods are not exactly same.The logic is same, but the behavior is differnt, this has exception handling for few things, and the validation continues. Agree with duplication of code, but this can be done when we do refactoring of test code, to completely avoid duplicate classes.

swatisawant:
Yes, you can take this up in next PR while refactoring the code. But, just make sure of modifying existing methods to support additional flags required for validation.

PR #606, opened 2017-05-23
Snap 1436

suranjan:

Test changes when memory is accounted while oldEntry is stored.
We wait till the oldEntry is cleared before asserting if the memory is same.

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


cluster/src/test/scala/org/apache/spark/memory/SnappyMemoryAccountingSuite.scala
```diff
@@ -440,6 +441,8 @@ class SnappyMemoryAccountingSuite extends MemoryFunSuite {
     snSession.insert("t1", row)
     assert(SparkEnv.get.memoryManager.storageMemoryUsed > 0) // borrowed from execution memory
     snSession.delete("t1", "col1=1")
+    // we need to wait for atleast OLD_ENTRIES_CLEANER_TIME_INTERVAL
+    Thread.sleep(Misc.getGemFireCache.getOldEntryRemovalPerid)
```
rishitesh: Cn we use waitForCriteria() here ?

PR #607, opened 2017-05-23
Fixed setting null txState as activeTxState

suranjan:

check for null before setting active tx state
## Patch testing

precheckin

## ReleaseNotes.txt changes

## Other PRs 


sumwale:
@suranjan 
Is this to inherit existing TXState from a previous connection/operation? Can you add more in PR description why this is required and if this addresses any open issues?

PR #608, opened 2017-05-24
Snap 1523

rishitesh:

This is a forked branch from SNAP-1454. Once 1454 is merged difference should be lot less.

a) Earlier we used a temp memory manager to record memory usage while GemXD is booting up. That caused problem if the data stored is more than memory allocated.
   As temp memory manager did not have any limits it used to allow recovering more than what can be accomodated. 
   Now changed temp managers to proper SnappyUnifiedMemoryManager , which will have all the memory limits in place. It can stop recovery after the threshold has been breached.

b) Recovery thread will now handle LowMemoryException from UMM. If it gets a LME it stops recovery process.

c) Alligned max storage fraction with eviction-heap-percentage. Could see some stable run by this.

d) No-Op MemoryManager now will delegate to Spark UnifiedMemoryManager. This is required for connector mode memory consumption by SnappyData operators.

## Patch testing

TODO precheckin

## ReleaseNotes.txt changes

NA


cluster/src/main/scala/org/apache/spark/memory/SnappyMemoryUtils.scala
```diff
@@ -26,16 +25,16 @@ object SnappyMemoryUtils {
     *
     * @return
     */
-  def isCriticalUp(accountedMemory: Long): Boolean =
-    Option(GemFireCacheImpl.getInstance).exists { cache =>
-      cache.getResourceManager.getHeapMonitor.failMemoryRequest(accountedMemory)
-    }
+  def isCriticalUp(): Boolean =
+    Option(GemFireStore.getBootingInstance).exists(g => g.thresholdListener.isCritical)
```
sumwale: _.thresholdListener.isCritical

cluster/src/main/scala/org/apache/spark/memory/SnappyMemoryUtils.scala
```diff
@@ -26,16 +25,16 @@ object SnappyMemoryUtils {
...
-   * Checks whether GemFire eviction threshold is breached
-   * @return
-   */
-  def isEvictionUp: Boolean = {
+    * Checks whether GemFire eviction threshold is breached
+    *
+    * @return
+    */
+  def isEvictionUp: Boolean =
     Option(GemFireStore.getBootingInstance).exists(g => g.thresholdListener.isEviction)
```
sumwale: _.thresholdListener.isEviction

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -71,54 +74,77 @@ class SnappyUnifiedMemoryManager private[memory](
...
   private val onHeapStorageRegionSize = onHeapStorageMemoryPool.poolSize
 
-  private val maxHeapStorageSize = (maxHeapMemory *
-      conf.getDouble("spark.memory.storageMaxFraction", 0.9)).toLong
-  // TODO: [sumedh] Not being used?
-  val maxExecutionSize: Long = (maxHeapMemory * 0.75).toLong
+  val evictionFraction = SnappyUnifiedMemoryManager.getStorageEvictionPercent(conf)
+
+  private val maxHeapStorageSize =
+    (maxHeapMemory * evictionFraction * 0.01).toLong
```
sumwale: change the name to evictionPercent

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -71,54 +74,77 @@ class SnappyUnifiedMemoryManager private[memory](
...
-        // transfer the memory map from tempMemoryManager on first use
-        MemoryManagerCallback.tempMemoryManager.memoryForObject.map { entry =>
-          val (objectName, mode) = entry._1
-          acquireStorageMemoryForObject(objectName,
-            MemoryManagerCallback.storageBlockId, entry._2, mode, null,
-            shouldEvict = false)
-        }
-        MemoryManagerCallback.tempMemoryManager.memoryForObject.clear()
-        _memoryForObjectMap
+      if (memoryMap eq null) synchronized {
```
sumwale: double synchronized?

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -456,14 +481,21 @@ class SnappyUnifiedMemoryManager private[memory](
...
       if (!enoughMemory) {
 
         // return immediately for OFF_HEAP with shouldEvict=false
-        if (offHeapNoEvict) return false
+        if (offHeapNoEvict || tempManager) return false
 
-        if (!offHeap && SnappyMemoryUtils.isCriticalUp(getStoragePoolMemoryUsed(
-          MemoryMode.ON_HEAP) + getExecutionPoolUsedMemory(MemoryMode.ON_HEAP))) {
+        // @TODO Commenting the below code as it causes strange behaviour. Memory
+        // allocation fails here , but somehow the query does not get cancelled.
```
sumwale: Is this TODO valid? See no commented out code.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -535,28 +568,34 @@ class SnappyUnifiedMemoryManager private[memory](
   override def releaseStorageMemoryForObject(objectName: String,
                                              numBytes: Long,
                                              memoryMode: MemoryMode): Unit = synchronized {
-    logDebug(s"releasing [SNAP] memory for $objectName = $numBytes")
-    if (memoryForObject.addTo(objectName, -numBytes) >= 0L) {
+    logDebug(s"releasing $managerId memory for $objectName = $numBytes")
+    if (memoryForObject.addTo(objectName -> memoryMode, -numBytes) != 0L) {
```
sumwale: Isn't >= 0 somewhat safer just in case there is an accounting mistake and results goes into -ve!

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -535,28 +568,34 @@ class SnappyUnifiedMemoryManager private[memory](
...
       memoryForObject.removeLong(objectName)
     }
   }
 
-  override def releaseStorageMemory(numBytes: Long, memoryMode: MemoryMode): Unit =
+  override def releaseStorageMemory(numBytes: Long, memoryMode: MemoryMode): Unit = {
     releaseStorageMemoryForObject(SPARK_CACHE, numBytes, memoryMode)
+    logDebug(s"releasing $managerId memory for $SPARK_CACHE $numBytes")
+    if (memoryForObject.containsKey(SPARK_CACHE)) {
+      memoryForObject.addTo(SPARK_CACHE -> memoryMode, -numBytes)
```
sumwale: Key in containsKey is incorrect -- should be a pair with memoryMode. Better to create the key once and use in both.

A more efficient version that will avoid two key lookups for most cases:

```scala
val key = SPARK_CACHE -> memoryMode
if (memoryForObject.addTo(key, -numBytes) != 0L) {
  super.releaseStorageMemory(numBytes, memoryMode)
} else {
  // key was not present so remove
  memoryForObject.removeLong(key)
}
```

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -217,18 +243,19 @@ class SnappyUnifiedMemoryManager private[memory](
...
-          val prev = memoryForObject.addTo(fromOwner, -totalSize)
+          val prev = memoryForObject.addTo(fromOwner -> mode, -totalSize)
           if (prev >= totalSize) {
-            memoryForObject.addTo(toOwner, totalSize)
+            memoryForObject.addTo(toOwner -> mode, totalSize)
           } else {
             // something went wrong with size accounting
-            memoryForObject.addTo(fromOwner, totalSize)
+            memoryForObject.addTo(fromOwner -> mode, totalSize)
             throw new IllegalStateException(
```
sumwale: create key pair once instead of every time

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -535,28 +568,34 @@ class SnappyUnifiedMemoryManager private[memory](
...
+    val bytesToBeFreed = memoryForObject.getLong(name -> memoryMode)
     val numBytes = Math.max(0, bytesToBeFreed - ignoreNumBytes)
-    logDebug(s"Dropping memory for $name = $numBytes (registered=$bytesToBeFreed)")
+    logDebug(s"Dropping $managerId memory for $name = $numBytes (registered=$bytesToBeFreed)")
 
     if (numBytes > 0) {
       super.releaseStorageMemory(numBytes, memoryMode)
-      memoryForObject.removeLong(name)
+      memoryForObject.removeLong(name -> memoryMode)
     }
```
sumwale: can create key pair once

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -580,6 +619,8 @@ object SnappyUnifiedMemoryManager extends Logging {
       math.max(getMaxHeapMemory / 20, 500L * 1024L * 1024L))
   }
 
+  private val DEFAULT_EVICTION_PERCENT = 80f
```
sumwale: I think we can increase these defaults: critical=95%, eviction=90%

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -635,6 +676,23 @@ object SnappyUnifiedMemoryManager extends Logging {
...
+    val cache = Misc.getGemFireCacheNoThrow
+    if (cache ne null) {
+      val thresholds = cache.getResourceManager.getHeapMonitor.getThresholds
+      if (thresholds.getEvictionThreshold > 0.1f) {
+        cache.getResourceManager.getHeapMonitor.getThresholds.getEvictionThreshold
+      } else {
+        DEFAULT_EVICTION_PERCENT
+      }
+    } else {
+      conf.getDouble("spark.testing.maxStorageFraction", DEFAULT_EVICTION_PERCENT).toFloat
```
sumwale: Property is a fraction which is getting mixed with percent. Use one consistent name to avoid confusion in this class -- perhaps fraction throughout since that is what Spark uses.

cluster/src/test/scala/io/snappydata/QueryTest.scala
```diff
@@ -99,21 +98,4 @@ class QueryTest extends SnappyFunSuite {
...
-    session.sql("CREATE TABLE yes_with(device_id VARCHAR(200), " +
-        "sdk_version VARCHAR(200)) USING COLUMN OPTIONS(PARTITION_BY 'device_id')")
-    session.insert("yes_with", Row("id1", "v1"), Row("id1", "v2"),
-      Row("id2", "v1"), Row("id2", "v1"), Row("id2", "v3"))
-    val r = session.sql("select sdk_version, count(distinct device_id) from (" +
-        "select sdk_version,device_id from YES_WITH group by sdk_version, " +
-        "device_id) a group by sdk_version")
-    ColumnCacheBenchmark.collect(r,
-      Seq(Row("v1", 2), Row("v2", 1), Row("v3", 1)))
-  }
```
sumwale: Why this change? Perhaps in resolving merge conflict -- revert.

core/src/main/scala/org/apache/spark/memory/StoreUnifiedManager.scala
```diff
@@ -18,136 +18,86 @@ package org.apache.spark.memory
...
-    }
-    memoryForObject(key) += numBytes
-    true
+                                             blockId: BlockId,
+                                             numBytes: Long,
+                                             memoryMode: MemoryMode,
+                                             buffer: UMMMemoryTracker,
+                                             shouldEvict: Boolean): Boolean = {
+    logDebug(s"Acquiring DefaultManager meemory for $objectName $numBytes")
+    SparkEnv.get.memoryManager.acquireStorageMemory(blockId, numBytes, memoryMode)
```
sumwale: Correct the formatting in this class. The previous indentation was correct. Perhaps you need to set the codeStyleSettings.xml again with "Project" type in IDEA.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -490,17 +534,18 @@ object CachedDataFrame
     if (dataLen == 0) return Iterator.empty
 
     val codec = CompressionCodec.createCodec(SparkEnv.get.conf)
-    val input = new Input(data, offset, dataLen)
```
sumwale: Do you need allocateExecutionMemory calls for input data bytes too?

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -399,15 +407,51 @@ object CachedDataFrame
...
+        finalBuffer.flip
+        memSize = finalBuffer.limit()
+        // Ask UMM before getting the array to heap.
+        // Taking unroll memory as this memory is cleaned up on task completion.
+        // On connector mode also this should account to the overall memory usage.
+        // We will ensure that sufficient memory is available by reserving
+        // two times as Kryo serialization will expand its buffer accordingly.
+
+        if (!MemoryManagerCallback.memoryManager.
+          acquireStorageMemoryForObject(
```
sumwale: Why not allocate in execution memory?

core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
```diff
@@ -294,7 +294,7 @@ object SnappyParserConsts {
   final val arithmeticOperator = CharPredicate('*', '/', '%', '&', '|', '^')
   final val exponent: CharPredicate = CharPredicate('e', 'E')
   final val numeric: CharPredicate = CharPredicate.Digit ++
-      CharPredicate('.')
+      CharPredicate('.') ++ exponent
```
sumwale: revert

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
rishitesh: Yes. Will change.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
rishitesh: Not valid. Will remove.

cluster/src/test/scala/io/snappydata/QueryTest.scala
rishitesh: Yes. Merge issues.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
rishitesh: Input is at driver side. We have not tested driver side memory manager much. I will file a ticket and check if we need some more changes at driver side.

core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
rishitesh: Yes.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
rishitesh: Off-heap eviction at 90% seems fair, but IMO heap eviction at 90% causes a lot of activity during query execution, eviction and all creates a lot of garbage and it quickly reaches critical limit. I would suggest lets keep eviction at 80 and once we get more confidence we can change the limit.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
rishitesh: Will change

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
rishitesh: Hmm, yes, probably that would be better.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
rishitesh: Hmm , yes.. Will change.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
rishitesh: not needed

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
rishitesh: Problem was, I wanted a memory allocation which should not evict data anymore. Right now our UMM acquireExecutionMemory() does not have that flag. I will have to change that before I can do this.

sumwale: Ok can let it remain storage for now but I think problem will be that under usual circumstances storage will always be about full and we don't want to evict data from here either.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
sumwale: My experience has been that GC always cleans up stuff before reaching 95%. So we can at least increase the critical to that especially if heap size is greater than say 4g (on Gem side too). Lets test more for eviction.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -71,31 +74,47 @@ class SnappyUnifiedMemoryManager private[memory](
 
   private val onHeapStorageRegionSize = onHeapStorageMemoryPool.poolSize
 
-  private val maxHeapStorageSize = (maxHeapMemory *
-      conf.getDouble("spark.memory.storageMaxFraction", 0.9)).toLong
-  // TODO: [sumedh] Not being used?
-  val maxExecutionSize: Long = (maxHeapMemory * 0.75).toLong
+  val evictionFraction = SnappyUnifiedMemoryManager.getStorageEvictionPercent(conf)
```
sumwale: This should really be a fraction so multiply with 0.01 right here

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -580,6 +618,8 @@ object SnappyUnifiedMemoryManager extends Logging {
       math.max(getMaxHeapMemory / 20, 500L * 1024L * 1024L))
   }
 
+  private val DEFAULT_EVICTION_FRACTION = 80f
```
sumwale: This should be a fraction so 0.8

core/src/main/scala/org/apache/spark/memory/StoreUnifiedManager.scala
```diff
@@ -18,135 +18,85 @@ package org.apache.spark.memory
...
 
   override def acquireStorageMemoryForObject(objectName: String,
       blockId: BlockId,
       numBytes: Long,
       memoryMode: MemoryMode,
       buffer: UMMMemoryTracker,
       shouldEvict: Boolean): Boolean = {
-    logDebug(s"Acquiring mem [NO-OP] for $objectName $numBytes")
-    true
+    logDebug(s"Acquiring DefaultManager meemory for $objectName $numBytes")
```
sumwale: typo: memory

core/src/main/scala/org/apache/spark/memory/StoreUnifiedManager.scala
```diff
@@ -18,135 +18,85 @@ package org.apache.spark.memory
...
+    SparkEnv.get.memoryManager.acquireStorageMemory(blockId, numBytes, memoryMode)
```
sumwale: There might be some race conditions here that can cause NPE during startup/shutdown? Perhaps better to have SparkEnv.get.memoryManager as a val in the class.

sumwale:
@rishitesh
Apart from some minor comments, the main remaining concern is whether eviction=false with acquireStorage memory will cause trouble. In many use-cases memory is almost full and this combination will likely fail frequently. I think that combination itself must be removed and instead all calling points should use eviction=false with acquireExecution memory.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -682,7 +675,7 @@ object SnappyUnifiedMemoryManager extends Logging {
     if (cache ne null) {
       val thresholds = cache.getResourceManager.getHeapMonitor.getThresholds
       if (thresholds.getEvictionThreshold > 0.1f) {
-        cache.getResourceManager.getHeapMonitor.getThresholds.getEvictionThreshold
+        (cache.getResourceManager.getHeapMonitor.getThresholds.getEvictionThreshold * 0.01).toFloat
```
sumwale: change the method name to "getStorageEvictionFraction"

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -562,26 +562,20 @@ class SnappyUnifiedMemoryManager private[memory](
...
-    if (memoryForObject.addTo(key, -numBytes) >= 0L) {
-      super.releaseStorageMemory(numBytes, memoryMode)
-    } else {
-      // objectName was not present
-      memoryForObject.removeLong(key)
+    super.releaseStorageMemory(numBytes, memoryMode)
+    if (memoryForObject.containsKey(key)) {
+      if (memoryForObject.addTo(key, -numBytes) == numBytes) {
+        memoryForObject.removeLong(key)
+      }
```
sumwale: I think original code was correct and more efficient. It avoids two lookup into the map for most cases. Something like below with your updated logic to always invoke super:

```scala
  super.releaseStorageMemory(numBytes, memoryMode)
  val originalNumBytes = memoryForObject.addTo(key, -numBytes)
  if (memoryForObject.addTo(key, -numBytes) <= numBytes) {
    // key not present or has gone down to zero
    memoryForObject.removeLong(key)
  }
```

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
sumwale: rename getStorageEvictionPercent to getStorageEvictionFraction

PR #609, opened 2017-05-24
CachedPlanHelper is now removed from the plan information for UI.

kneeraj:

CachedPlanHelperExec spark plan which has been introduced to facilitate Tokenization use to also show up in the dashboard UI. This is not useful to the user and so removing it from the dispaly by calling a simple transform.

## Patch testing

Manual.

## ReleaseNotes.txt changes

No.

## Other PRs 

None.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
+++ b/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
@@ -72,7 +72,7 @@ class CachedDataFrame(df: Dataset[Row], var queryString: String,
   def queryExecutionString: String = queryExecution.toString()
 
   def queryPlanInfo: SparkPlanInfo = PartitionedPhysicalScan.getSparkPlanInfo(
-    plan = queryExecution.executedPlan transformAllExpressions {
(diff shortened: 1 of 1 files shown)


kneeraj:
@rishitesh  it won't be a problem as this is just for the UI we are transforming and a new plan tree object is returned. The executedPLan is never touched.

sumwale:
As a general comment we may need to see how much overhead does these additional plan transformations have for low-latency queries.

kneeraj:
>> As a general comment we may need to see how much overhead does these additional plan transformations have for low-latency queries.

True. This was going in my mind too. May be we need some discussion here as to how to minimize the effect on low latency queries or should be handled differently.

So for now should I go ahead and merge?

sumwale:
Fine for merge, but we may need to note it down somewhere. We will be coming to low latency query performance soon.

sumwale:
Btw, the PlanInfo caching was done because I did see overhead of PlanInfo creation for low latency queries as part of general optimizations that we done in SNAP-1191. We need to circle back on it and see how much it effects overall performance.

PR #610, opened 2017-05-25
Relocating Guava package

hbhanawat:

Guava package is creating problems when running with 3rd party jars for example cassandra connector. Relocating the guava jars in our core package to overcome this issue

## Patch testing

Manual testing and precheckin .


core/build.gradle
```diff
@@ -201,6 +201,8 @@ shadowJar {
   relocate 'org.apache.spark.unsafe', 'io.snappydata.org.apache.spark.unsafe'
   // relocate fastutil for possible conflicts with user dependencies
   relocate 'it.unimi.dsi.fastutil', 'io.snappydata.it.unimi.dsi.fastutil'
+  // relocate the guava's com.google packages
+  relocate 'com.google', 'io.snappydata.com.google'
```
sumwale: guava classes seem to be using "com.google.common" so perhaps better to use the full prefix

hbhanawat: Done.

PR #614, opened 2017-05-26
SNAP-1615

vivekwiz:
Currently if column being aggregated has a NULL value while grouping on a string column, the grouping row itself produce a new row with Null column.

To fix same, as discussed with Sumedh, while scanning row table for string column now look at actual value to decide whether its Null or not. Earlier it was looking at a default value.

[Sumedh] To clarify above: there are two sets of decoders in our column table scan (actually three but ignore the 3rd). One is on column batches and second on row buffer. First one will tell nullability of column value right away but the latter one uses JDBC API so it can return proper "wasNull()" only after the value has been read. So the combined nullability handling there returns a tri-state with the third indicating that "don't know: needs to be gotten after reading the value".

The bug was that with dictionary optimization (which does not apply to row buffer data), the actual value was not being read for the third "don't know" state rather only the dictionary index was being read (that returns -1 for row buffer data).

## Patch testing
Precheckin. Added test.

## ReleaseNotes.txt changes
No

## Other PRs 
No


cluster/src/test/scala/io/snappydata/cluster/QueryRoutingSingleNodeSuite.scala
```diff
@@ -231,12 +237,16 @@ class QueryRoutingSingleNodeSuite extends SnappyFunSuite with BeforeAndAfterAll
         s") " +
         s" limit 20" +
         s""
+    // scalastyle:off println
     println(s"Iter ${iter} QUERY = ${qry}")
+    // scalastyle:on println
     val df1 = snc.sql(qry)
     val res1 = df1.collect()
+    // scalastyle:off println
```
sumwale: thanks for taking care of scalastyle issues in this file

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -801,10 +802,10 @@ private[sql] final case class ColumnTableScan(
         val nullCheckAddon =
           if (notNullVar != null) s"if ($notNullVar < 0) $nullVar = $col == null;\n"
           else ""
-        assignCode =
+        stringAssignCode =
             s"($dictionary != null ? $dictionary[$dictIndex] " +
-                s": $decoder.readUTF8String($buffer, $cursorVar));\n" +
-                s"$nullCheckAddon"
+                s": $decoder.readUTF8String($buffer, $cursorVar));\n"
+        assignCode = stringAssignCode + s"$nullCheckAddon"
```
sumwale: Don't need s"" if nullCheckAddon is only value in string and just a '+' of two will do

PR #619, opened 2017-05-27
[SNAP-1486] Remove session field of CachedPlanHelperExec

sumwale:

Remove session field of CachedPlanHelperExec and instead get from the sqlContext field.

Issue is "lazy val cleanArgs" in QueryPlan as the serialization stack in the bug also suggests.
It will hold all the arguments of a case class whether or not those are transient.
cleanArgs should have been transient lazy val field in Spark which is in the Spark PR.

Also includes some scalastyle corrections in CachedPlanHelperExec.

## Patch testing

precheckin, bug 1486 test

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/51

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -619,9 +619,7 @@ task product(type: Zip) {
     delete "${snappyProductDir}/python/lib/pyspark.zip"
   }
   doLast {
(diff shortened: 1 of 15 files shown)


kneeraj:
LGTM.

PR #622, opened 2017-05-30
Jdbc cdc streaming

soubhik-c:

* Adding deleteFrom() much like putInto api for cdc ingestion.
* Much of it due to rename of TableInsertExec to TableExec and RowInsertExec to RowDMLExec
* The plan will now show RowInsert/RowUpsert/RowDelete types
* The metrics are adjusted accordingly.
* If snappy-connector/jdbc-stream-connector subproject exists, it will be compiled and published automatically along with the targets.
* distProduct will package the above connector jar in the snappy/jars folder along with other jars.
A subproject is done in order to avoid copying of rootProject build.gradle tasks/settings and many other adjustments applicable to all subprojects.

## Patch testing

It is tested with jdbc-stream-connector spark app in smart connector mode.

## ReleaseNotes.txt changes

None

## Other PRs 

None


core/src/main/scala/org/apache/spark/sql/execution/row/RowDMLExec.scala
```diff
@@ -118,22 +121,39 @@ case class RowInsertExec(_child: SparkPlan, upsert: Boolean,
...
+       |    $result += $numOperations;
+       |    ${if (numOpRowsMetric eq null) ""
+            else s"$numOpRowsMetric.${metricAdd(numOperations)};"}
        |  $rowCount = 0;
        |}
     """.stripMargin
   }
+
+  override def opType: String = if (upsert) {
+    "Upserted"
```
sumwale: Perhaps the name can be "Put" because it is different from a real upsert (arg can be renamed too).

core/src/main/scala/org/apache/spark/sql/execution/row/RowDMLExec.scala
```diff
@@ -118,22 +121,39 @@ case class RowInsertExec(_child: SparkPlan, upsert: Boolean,
...
+  } else if (delete) {
+    "Deleted"
+  } else {
+    "Inserted"
+  }
+
+  override def nodeName: String = if (upsert) {
+    "RowUpsert"
```
sumwale: RowPut?

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -635,6 +635,38 @@ private[sql] final class PreprocessTableInsertOrPut(conf: SQLConf)
         case _ => p
       }
 
+    // Check for PUT
+    // Need to eliminate subqueries here. Unlike InsertIntoTable whose
+    // subqueries have already been eliminated by special check in
+    // ResolveRelations, no such special rule has been added for PUT
```
sumwale: correct the comment for DELETE instead of PUT

core/src/main/scala/org/apache/spark/sql/sources/jdbcExtensions.scala
```diff
@@ -304,6 +304,24 @@ object JdbcExtendedUtils extends Logging {
...
 
+  /**
+   * Returns the SQL for prepare to insert or put rows into a table.
+   */
+  def getDeleteString(table: String, rddSchema: StructType): String = {
+    val sql = new StringBuilder()
+    sql.append(s"DELETE FROM $table WHERE ")
+    var fieldsLeft = rddSchema.fields.length
+    rddSchema.fields.foreach { field =>
+      sql.append(field.name).append('=').append('?')
```
sumwale: quote the names for case-sensitivity/special chars changes in https://github.com/SnappyDataInc/snappydata/pull/648

core/src/main/scala/org/apache/spark/sql/execution/row/RowDMLExec.scala
soubhik-c: done.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
soubhik-c: done.

core/src/main/scala/org/apache/spark/sql/execution/row/RowDMLExec.scala
soubhik-c: done.

core/src/main/scala/org/apache/spark/sql/sources/jdbcExtensions.scala
soubhik-c: I have added the quote just like in #648 . escapeQuote boolean is added in getDeleteString just like getInsertOrPutString method keeping possibility of future usage.

PR #625, opened 2017-05-31
SNAP-1258: [Feedback from Sumedh]

snappy-sachin:

"In Memory Size" could be just "Memory Size". Also "Total Size" is not enough for users to understand that this is memory and overflow size. Perhaps should put in brackets of title like "Total Size (Memory+Disk overflow)", or just "Memory and Disk overflow size"

## Patch testing

Tested manually.

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/cluster/src/main/scala/org/apache/spark/ui/SnappyDashboardPage.scala
+++ b/cluster/src/main/scala/org/apache/spark/ui/SnappyDashboardPage.scala
@@ -744,10 +744,10 @@ object SnappyDashboardPage{
   tableStatsColumn += ("distributionTypeTooltip" -> "Distribution Type is either PARTITIONED or REPLICATED table ")
   tableStatsColumn += ("rowCount" -> "Row Count")
   tableStatsColumn += ("rowCountTooltip" -> "Total Rows in Table")
-  tableStatsColumn += ("sizeInMemory" -> "In Memory Size")
+  tableStatsColumn += ("sizeInMemory" -> "Memory Size")
   tableStatsColumn += ("sizeInMemoryTooltip" -> "Tables Size in Memory")
   tableStatsColumn += ("totalSize" -> "Total Size")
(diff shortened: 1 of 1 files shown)


snappy-sachin:
As there is just a change in display text only, no approval is required.
Hence merging..

cluster/src/main/scala/org/apache/spark/ui/SnappyDashboardPage.scala
```diff
@@ -744,10 +744,10 @@ object SnappyDashboardPage{
   tableStatsColumn += ("distributionTypeTooltip" -> "Distribution Type is either PARTITIONED or REPLICATED table ")
   tableStatsColumn += ("rowCount" -> "Row Count")
   tableStatsColumn += ("rowCountTooltip" -> "Total Rows in Table")
-  tableStatsColumn += ("sizeInMemory" -> "In Memory Size")
+  tableStatsColumn += ("sizeInMemory" -> "Memory Size")
   tableStatsColumn += ("sizeInMemoryTooltip" -> "Tables Size in Memory")
   tableStatsColumn += ("totalSize" -> "Total Size")
-  tableStatsColumn += ("totalSizeTooltip" -> "Total Size of Tables")
+  tableStatsColumn += ("totalSizeTooltip" -> "Tables Total Size (In Memory size + Disk Overflow Size)")
```
sumwale: Don't think "Tables" is required since the whole box is for tables. Can be shortened to "Total Size (Memory + Disk Overflow Size)"

snappy-sachin: Hi Sumedh,

This is a tool tip text and needs to be specific. Similar scheme is used in members tabular list also on UI.

sumwale: ok

PR #626, opened 2017-05-31
Fix for SNAP-1505

dshirish:
Fixing a failure in SplitClusterDUnitTestBase#testTableFormChanges. 

Calling the SnappySession.clearAllCache() in  SmartConnectorHelper.createTable and dropTable functions (this already happens when tables are created/dropped in embedded mode)

The test creates a row table in connector mode, drops it in embedded mode and then creates column table with same name on the connector side.Since the cache wasn't getting cleared when table was dropped / created older cached objects were getting used.
 The test earlier was present in ThinConnectorSnappyDUnitTest (now gets called in SplitSnappyClusterDUnitTest)

The test modifications in the PR are mainly for readability.
  
## Patch testing
precheckin 

## ReleaseNotes.txt changes
NA
## Other PRs 
NA


dshirish:
Merging the fix. Had discussed the change with Neeraj.

PR #628, opened 2017-05-31
Multigrid master

ahshahid:
Requirement is to pass properties starting with  gemfire-grid.  to be passed to GemFireStore.
If the prefix is not done, it gets flitered in a function  ( because it is not belonging otherwise to store or snappy-spark).

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
@@ -27,7 +27,7 @@ import scala.collection.JavaConverters._
 import akka.actor.ActorSystem
 import com.gemstone.gemfire.distributed.internal.DistributionConfig
 import com.gemstone.gemfire.distributed.internal.locks.{DLockService, DistributedMemberLock}
-import com.gemstone.gemfire.internal.cache.GemFireCacheImpl
+import com.gemstone.gemfire.internal.cache.{GemFireCacheImpl, GemFireSparkConnectorCacheImpl}
 import com.pivotal.gemfirexd.FabricService.State
 import com.pivotal.gemfirexd.internal.engine.store.ServerGroupUtils
 import com.pivotal.gemfirexd.{FabricService, NetworkInterface}
@@ -126,6 +126,8 @@ class LeadImpl extends ServerImpl with Lead
           } else {
             Constant.STORE_PROPERTY_PREFIX + k
           }
+        } else if (k.startsWith(GemFireSparkConnectorCacheImpl.gfeGridPropPrefix)) {
+          Constant.STORE_PROPERTY_PREFIX + k
         }
         else {
           k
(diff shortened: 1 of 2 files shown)


sumwale:
@ahshahid Why not use the standard Spark convention of "spark." prefixed properties i.e. "spark.gemfire-grid.xx" or better something like "spark.gemfire.connector.grid.xx". Connectors should not add new unique prefixes to be handled in the product.

ahshahid:
Jags & me thought to keep it small & very specific to gemfire property.
Bringing Jags in loop. I am ok with either..

ahshahid:
Just to clarify, that property is actually without prefix .

sumwale:
"snappydata." prefix is one and the only special case that is allowed in SnappyData product apart from "spark." (for obvious reasons). Even for GemFireXD we changed it to "snappydata.store." and we should certainly not change that for an external connector. Please change the prefix to "spark.gemfire..." something and then this PR is not required and should be closed.

ahshahid:
i see. in which case would the conf file need full specification like spark.gemfire.-grid.xx & would this property be available in the map in GemFireStore?

ahshahid:
Sumedh, would a spark prefixed property be available in the GemFireStore.boot?

ahshahid:
@sumwale  Could you please clarify this thing , will a spark prefixed property be available in GemStore.boot? If not , what prefix is needed? It would be better if it is not  preceded by store.

sumwale:
Yes "spark." prefixed properties are available in SparkConf like what this patch is trying to do.

ahshahid:
Sumedh , but will sparkConf it self be accessible in GemFireStore.boot ?
Pls note that this spark prefixed property is needed in gemFireStore.boot

sumwale:
Yes all properties passed at boot time are available in GemFireStore.boot. All "spark." and "snappydata." ones are also set into the SparkConf of LeadImpl (like in the code you changed in this PR) so you can get it both via GemFireStore.boot or SparkEnv.get.

ahshahid:
ok. thanks

ahshahid:
and that woudl be the case on servers too?  because these props would also
be needed GemFireStore.boot of servers. Where I suppose SparkEnv does not
work ( or does it?)

sumwale:
It is available in the properties of GemFireStore.boot if set in conf/servers. It is also available in SparkEnv.get.conf fetching from driver on lead after ExecutorInitiator is started (i.e. after the GemFireStore.startExecutor call in boot). In first case this property has to be set on all servers explicitly. The second way is the cleaner and more Spark-like way where properties will be set only on driver (leads) and executors will get it during their execute.

ahshahid:
OK. Start executor would be late , becoz cache would be created by then. I
m testing explicitly passing on lead and server confs

PR #630, opened 2017-06-01
SNAP-1663

dshirish:
SNAP-1663: JDBCSourceAsColumnarStore#doInsert gets connectionType as unknown. This was observed by SachinJ  while working TX issues

In this case, JDBCSourceAsColumnarStore is created thru following code path on server (executor). GemFireXDDialect is not initialized because of which incorrect dialect gets set in ConnectionProperties in ExternalStoreUtils#validateAndGetAllProps called from ExternalStoreUtils#getExternalStoreOnExecutor

```
at java.lang.Thread.dumpStack(Thread.java:1333)
at org.apache.spark.sql.execution.columnar.impl.JDBCSourceAsColumnarStore.<init>(JDBCSourceAsColumnarStore.scala:74)
at org.apache.spark.sql.execution.columnar.ExternalStoreUtils$.getExternalStoreOnExecutor(ExternalStoreUtils.scala:562)
at org.apache.spark.sql.execution.columnar.ExternalStoreUtils.getExternalStoreOnExecutor(ExternalStoreUtils.scala)
at io.snappydata.impl.SnappyHiveCatalog$HMSQuery.call(SnappyHiveCatalog.java:348)
at java.util.concurrent.FutureTask.run(FutureTask.java:266)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
at java.lang.Thread.run(Thread.java:745)
```

As a fix Initializing GemFireXDDialect in SnappyHiveCatalog#SnappyHiveCatalog

@sumwale Let mw know if this fix can be committed as all tests seem to be passing without it but probably there is perf impact as JDBCSourceAsColumnarStore#doInsert always calls doGFXDInsert() instead of doSnappyInsert() in this case.

## Patch testing
precheckin to be run

## ReleaseNotes.txt changes
NA
## Other PRs 
NA

### DIFF
--- a/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
+++ b/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
@@ -48,7 +48,6 @@ import org.apache.spark.sql.execution.columnar.ExternalStoreUtils;
 import org.apache.spark.sql.execution.datasources.jdbc.DriverRegistry;
 import org.apache.spark.sql.hive.ExternalTableType;
 import org.apache.spark.sql.hive.SnappyStoreHiveCatalog;
-import org.apache.spark.sql.row.GemFireXDDialect;
 import org.apache.spark.sql.sources.JdbcExtendedUtils;
 import org.apache.spark.sql.store.StoreUtils;
 import org.apache.spark.sql.types.StructType;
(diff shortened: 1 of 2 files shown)


dshirish:
Merging this PR.  Sachin has confirmed that this change works with his test.

PR #632, opened 2017-06-02
SNAP-1601

hbhanawat:

1. Clean up of dependencies of a shuffledRDD is now kicked off (asynchronously) immediately after query execution. Added code to wait for cleanup of dependencies when a cachedDataFrame is being re-used. This safeguards against the case if the cleanup of a previous execution is not yet complete
2. Removed a previous checkin which removed CachedPlanHelper from the plan. This change removed the details of the nodes below the CachedPlanHelper node too. There was no information in them. Needs to be fixed properly later.

## Patch testing

Manual testing by setting thread.sleep inside BlockManagerSlaveEndpoint.scala (case RemoveShuffle).

Running precheckin.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
+++ b/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
@@ -82,8 +82,6 @@ class CachedDataFrame(df: Dataset[Row], var queryString: String,
           case None => _v
         }
         Literal(v, _dt)
-    } transform {
-      case CachedPlanHelperExec(childPlan) => childPlan
     })
 
   private lazy val lastShuffleCleanups = new Array[Future[Unit]](
@@ -170,6 +168,9 @@ class CachedDataFrame(df: Dataset[Row], var queryString: String,
       case e: Exception =>
         sparkSession.listenerManager.onFailure(name, queryExecution, e)
         throw e
+    } finally {
+      // clear the shuffle dependencies asynchronously after the execution.
+      clearCachedShuffleDeps(sparkSession.sparkContext)
     }
   }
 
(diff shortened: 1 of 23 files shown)


hbhanawat:
Run precheckin

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -304,6 +303,8 @@ class CachedDataFrame(df: Dataset[Row], var queryString: String,
       if (!hasLocalCallSite) {
         sc.clearCallSite()
       }
+      // clear the shuffle dependencies asynchronously after the execution.
+      clearCachedShuffleDeps(sc)
     }
```
sumwale: There are two other methods in Dataset that should be overridden for any required changes to collectWithHandler (had mentioned this in some other review but was ignored). See the calls to private method Dataset.collect(needCallback) which is also invoked by count() and head(). These two methods in Dataset should also be overridden in CachedDataFrame to use callbackWithHandler.

Can you take care of it in this PR because otherwise the shuffle cleanup will never happen for those two methods (e.g. if user does df.sql("...").count())?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1949,7 +1949,7 @@ object SnappySession extends Logging {
         cachedDF = evaluation._1
         queryHints = evaluation._2
       } else {
-        cachedDF.clearCachedShuffleDeps(session.sparkContext)
+        cachedDF.waitForLastShuffleCleanup
```
sumwale: convention is that for actions the empty parens should be provided (and skipped for getters)

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
hbhanawat: Done. Please review

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
hbhanawat: Done.

PR #634, opened 2017-06-02
[SNAP-1664] LowMemoryException during cluster restart

rishitesh:
When recovery was in progress it occupied almost all the off-heap memory available.
Also during recovery we use a temp UMM which has its limit set from defaults.
Also it can not do any eviction.But a side effect of exhausting all off-heap memory
in recovery is other messaging are not able to get any memory and restart hangs

## Patch testing

Manual

## ReleaseNotes.txt changes
NA

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/207

### DIFF
--- a/cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
+++ b/cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
@@ -456,9 +456,10 @@ class SnappyUnifiedMemoryManager private[memory](
 
       // Evict only limited amount for owners marked as non-evicting.
       // TODO: this can be removed once these calls are moved to execution
+      // TODO use something like "(spark.driver.maxResultSize / numPartitions) * 2"
       val doEvict = if (shouldEvict &&
           ManagedDirectBufferAllocator.nonEvictingOwners.contains(objectName)) {
-        numBytes < storagePool.memoryFree * 2
+        numBytes < math.max(0.05 * storagePool.poolSize, storagePool.memoryFree * 2)
       } else shouldEvict
 
       if (numBytes > maxMemory) {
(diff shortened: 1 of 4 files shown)


cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -495,7 +495,7 @@ class SnappyUnifiedMemoryManager private[memory](
       if (!enoughMemory) {
 
         // return immediately for OFF_HEAP with shouldEvict=false
-        if (offHeapNoEvict || tempManager) return false
+        if (offHeapNoEvict) return false
```
sumwale: Shouldn't the other call of "if (tempManager) { storagePool.acquireMemory(blockId, numBytes, numBytesToFree = 0)" also be removed to be same for both cases?

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -606,6 +606,13 @@ class SnappyUnifiedMemoryManager private[memory](
     }
     memoryForObject.clear()
   }
+
+  // Recovery is a special case. If any of the storage pool has reached 90% of
+  // max storage pool size stop recovery.
+  override def shouldStopRecovery(): Boolean = synchronized {
+    (offHeapStorageMemoryPool.memoryUsed >= (maxOffHeapStorageSize * 0.90) ) ||
+        (onHeapStorageMemoryPool.memoryUsed >= (maxHeapStorageSize * 0.90))
+  }
```
sumwale: use ">" and not ">=" so that for case where there is no off-heap configured, the check will not fail (else for only on-heap case it will fail on very first invocation)

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
rishitesh: No, for temp UMM there is no need to evict from Spark's cache. In fact, it will result in null pointer exception as Spark's artifacts are not ready.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
rishitesh: Yeah, yes..Will change it.

PR #635, opened 2017-06-02
Snap 1642

kneeraj:
Not doing plan caching at all for queries with scalar subqueries  
Added a flag to disable Tokenization. By default it is still ON.

## Patch testing

Added unit test for both.

## ReleaseNotes.txt changes

None

## Other PRs 

No.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -1885,9 +1885,13 @@ object SnappySession extends Logging {
     def apply(session: SnappySession, lp: LogicalPlan, sqlText: String, pls:
     ArrayBuffer[ParamLiteral]): CachedKey = {
 
+      val invalidate = Array(false)
+
       def normalizeExprIds: PartialFunction[Expression, Expression] = {
         case s: ScalarSubquery =>
-          s.copy(exprId = ExprId(0))
+          invalidate(0) = true
+          // s.copy(exprId = ExprId(0))
+          s
         case e: Exists =>
           e.copy(exprId = ExprId(0))
         case p: PredicateSubquery =>
@@ -1912,7 +1916,12 @@ object SnappySession extends Logging {
 
       // normalize lp so that two queries can be determined to be equal
       val tlp = lp.transform(transformExprID)
-      new CachedKey(session, tlp, sqlText, session.queryHints.hashCode(), pls.sortBy(_.pos).toArray)
+      val key = new CachedKey(session, tlp,
+        sqlText, session.queryHints.hashCode(), pls.sortBy(_.pos).toArray)
+      if (invalidate(0)) {
+        key.invalidatePlan()
+      }
+      key
     }
   }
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1889,9 +1889,13 @@ object SnappySession extends Logging {
     def apply(session: SnappySession, lp: LogicalPlan, sqlText: String, pls:
     ArrayBuffer[ParamLiteral]): CachedKey = {
 
+      val invalidate = Array(false)
```
vivekwiz: Any specific reason to avoid var?

core/src/test/scala/org/apache/spark/sql/store/TokenizationTest.scala
```diff
@@ -488,6 +489,28 @@ class TokenizationTest
...
   }
 
+  test("Test BUG SNAP-1642") {
+    SnappyTableStatsProviderService.suspendCacheInvalidation = true
+    val maxquery = s"select * from $table where a = (select max(a) from $table)"
+    val numRows = 10
+    createSimpleTableAndPoupulateData(numRows, s"$table", true)
+
+    val rs1 = snc.sql(maxquery)
+    val rows1 = rs1.collect()
```
vivekwiz: Did you want to verify this value too?

core/src/test/scala/org/apache/spark/sql/store/DisableTokenizationTest.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  override def afterAll(): Unit = {
+    System.setProperty("DISABLE_TOKENIZATION", "false")
+    super.afterAll()
+  }
+
+  after {
+    SnappyTableStatsProviderService.suspendCacheInvalidation = false
+    SnappySession.clearAllCache()
+    snc.dropTable(s"$table", ifExists = true)
```
vivekwiz: Want to do this for all tables?

vivekwiz:
Added few comments. Looks good to merge otherwise.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: No var was intentionally avoided. Using val is the recommended way. It is like final in java is preferred wherever applicable.

core/src/test/scala/org/apache/spark/sql/store/DisableTokenizationTest.scala
kneeraj: No other tables are created. Will clean the test.

core/src/test/scala/org/apache/spark/sql/store/TokenizationTest.scala
kneeraj: Main idea is to check whether cached and uncached are same or not. So validating that below.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
sumwale: Should be private. Also I think what Vivek meant was why not "private var invalidate = false" instead of using an array? A var is certainly preferred in this case instead of an array to change value.

kneeraj: No Sumedh. Vivek meant val vs var only.
BTW I used val because I was toggling the flag inside the normalizeExprIds function inside the apply method. I thought that will be better.

kneeraj: On popular demand, changing it to simple var boolean.

PR #639, opened 2017-06-05
(SNAP-1684) Add client driver classname as a driver property when creating a connection factory

ashetkar:
Changes from @hbhanawat  and @sumwale.

Add client driver classname as a driver property when creating a connection factory. It avoids 'No suitable driver' issues when driver is not auto-loaded.

## Patch testing
Manual testing with Apache Spark. 
./bin/spark-shell  --master local[*] --conf spark.snappydata.connection=localhost:1527 --jars /latest/master/snappydata-core.jar

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


dshirish:
LGTM. I have filed SNAP-1683 to add a unit test for this.

PR #640, opened 2017-06-06
Issue commit even if the tx is already committed

suranjan:
This is to ensure that at the server, we cleanup the txContext

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
@@ -535,15 +535,14 @@ final class SmartConnectorColumnRDD(
       context.addTaskCompletionListener { _ =>
         logDebug(s"The txid going to be committed is $txId " + tableName)
 
-        if ((txId ne null) && !txId.equals("null")
-        /* && !(tx.asInstanceOf[TXStateProxy]).isClosed() */ ) {
+        //if ((txId ne null) && !txId.equals("null")) {
           val ps = conn.prepareStatement(s"call sys.COMMIT_SNAPSHOT_TXID(?)")
-          ps.setString(1, txId)
+          ps.setString(1, if (txId == null) "null" else txId)
           ps.executeUpdate()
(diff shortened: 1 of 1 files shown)


kneeraj:
LGTM. @sumwale please review.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -535,15 +535,14 @@ final class SmartConnectorColumnRDD(
       context.addTaskCompletionListener { _ =>
         logDebug(s"The txid going to be committed is $txId " + tableName)
 
-        if ((txId ne null) && !txId.equals("null")
-        /* && !(tx.asInstanceOf[TXStateProxy]).isClosed() */ ) {
+        //if ((txId ne null) && !txId.equals("null")) {
           val ps = conn.prepareStatement(s"call sys.COMMIT_SNAPSHOT_TXID(?)")
-          ps.setString(1, txId)
+          ps.setString(1, if (txId == null) "null" else txId)
```
sumwale: Perhaps its cleaner to set a proper null instead of token:
```scala
if (txId ne null) {
  ps.setString(1, txId)
} else {
  ps.setNull(1, Types.VARCHAR)
}
```

sumwale:
@suranjan One minor comment

sumwale:
@suranjan Thinking about this again, didn't get what this is trying to do. The connection obtained will be pool connection which can be a completely different one having no relation to the task? What is the condition under which TXId can be null?

suranjan:
Each rdd#compute will have its own pooled connection. For a task which require more than one rdd#compute, all the connection will use same txState but then all of the connection needs to be cleaned up at server for txState when task ends.
To that end, each rdd#compute adds one task listener for cleaning up the txState at server for its connection. The first listener will commit and set the snapshotTxState to NULL. Rest all will only cleanup its txState at server with the hint that txId is "null".

suranjan:
I guess all this hackish way will not be required if all rdd#compute of a task use the same pooled connection. Will try to do that post 0.9.

PR #642, opened 2017-06-06
SNAP-1694

dshirish:
The issue is that ColumnFormatRelation#getTableName replaces all "__"  (2 consecutive _ chars) with '.' to derive row buffer table name from column table name. If the user tablename itself has 2 consecutive _ chars, then table name  returned is incorrect. This caused the stats generation code to fail with following stack trace at startup as table container was not found

```
17/06/05 08:22:41.370 PDT serverConnector<tid=0x1a> WARN snappystore: got throwable: None.get calling shut down
java.util.NoSuchElementException: None.get
at scala.None$.get(Option.scala:347)
at scala.None$.get(Option.scala:345)
at org.apache.spark.sql.execution.columnar.impl.ColumnFormatKey.getNumColumnsInTable(ColumnFormatEntry.scala:114)
at io.snappydata.SnappyEmbeddedTableStatsProviderService$$anonfun$publishColumnTableRowCountStats$2.apply(SnappyTableStatsProviderService.scala:207)
at io.snappydata.SnappyEmbeddedTableStatsProviderService$$anonfun$publishColumnTableRowCountStats$2.apply(SnappyTableStatsProviderService.scala:187)
at scala.collection.TraversableLike$WithFilter$$anonfun$foreach$1.apply(TraversableLike.scala:733)
at scala.collection.Iterator$class.foreach(Iterator.scala:893)
at scala.collection.AbstractIterator.foreach(Iterator.scala:1336)
at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
at scala.collection.TraversableLike$WithFilter.foreach(TraversableLike.scala:732)
at io.snappydata.SnappyEmbeddedTableStatsProviderService$.publishColumnTableRowCountStats(SnappyTableStatsProviderService.scala:187)
at io.snappydata.gemxd.ClusterCallbacksImpl$.publishColumnTableStats(ClusterCallbacksImpl.scala:119)
at com.pivotal.gemfirexd.internal.engine.db.FabricDatabase.postCreate(FabricDatabase.java:555)
```

As a fix using 4 consecutive '_' chars in internal column table and also using replaceFirst instead of replace in it. This reduces the possibility due to such a failure. Let me know if there exists a better way to handle this. 

## Patch testing
Running precheckin. 
Checked manually by creating a table with '__' chars.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
@@ -675,16 +675,17 @@ object ColumnFormatRelation extends Logging with StoreCallback {
 
   final def columnBatchTableName(table: String): String = {
     val tableName = if (table.indexOf('.') > 0) {
-      table.replace(".", "__")
+      table.replace(".", "____")
     } else {
-      Constant.DEFAULT_SCHEMA + "__" + table
+      Constant.DEFAULT_SCHEMA + "____" + table
     }
     Constant.SHADOW_SCHEMA_NAME + "." + tableName + Constant.SHADOW_TABLE_SUFFIX
   }
 
   final def getTableName(columnBatchTableName: String): String = {
     columnBatchTableName.substring(Constant.SHADOW_SCHEMA_NAME.length + 1,
-      columnBatchTableName.indexOf(Constant.SHADOW_TABLE_SUFFIX)).replace("__", ".")
+      columnBatchTableName.indexOf(Constant.SHADOW_TABLE_SUFFIX)).
+        replaceFirst("____", ".")
   }
 
   final def isColumnTable(tableName: String): Boolean = {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -675,16 +675,17 @@ object ColumnFormatRelation extends Logging with StoreCallback {
 
   final def columnBatchTableName(table: String): String = {
     val tableName = if (table.indexOf('.') > 0) {
-      table.replace(".", "__")
+      table.replace(".", "____")
     } else {
-      Constant.DEFAULT_SCHEMA + "__" + table
+      Constant.DEFAULT_SCHEMA + "____" + table
```
kneeraj: Shall we have a constant for this so that the next time we require to add more underscores we can just change at one place :-)

kneeraj:
Looks ok to me.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
dshirish: Done

sumwale:
Looks good. We can do a more generic solution using something like StringEscapeUtils later.

PR #644, opened 2017-06-06
    Patch from Sumedh:

vivekwiz:
    Fix for SNAP-1671: 'show tables in app' query doesnot list sample table.
    Fix for SNAP-1655: support for boolean in row tables

## Patch testing
Precheckin done.
@soubhik-c  Please confirm testcase for 1655 is working fine.
@spillai14  Please confirm testcase for 1671 is working fine.

## ReleaseNotes.txt changes
No

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/214

### DIFF
--- a/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
+++ b/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
@@ -114,7 +114,7 @@ public class SnappyHiveCatalog implements ExternalCatalog {
     return (Boolean)handleFutureResult(f);
   }
 
-  public List<ExternalTableMetaData> getNonStoreTables(boolean skipLocks) {
+  public List<ExternalTableMetaData> getHiveTables(boolean skipLocks) {
     // skip if this is already the catalog lookup thread (Hive dropTable
     //   invokes getTables again)
     if (Boolean.TRUE.equals(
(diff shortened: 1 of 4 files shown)


sumwale:
@kneeraj take a look asap. IMO this was sent quite a while back and corrects a couple of important issues and should have made into 0.9 right then.

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
```diff
@@ -270,8 +270,7 @@ public Object call() throws Exception {
               Table table = hmc.getTable(schema, tableName);
               String tableType = table.getParameters().get(
                   JdbcExtendedUtils.TABLETYPE_PROPERTY());
-              if (!ExternalTableType.Row().name().equalsIgnoreCase(tableType) &&
-                  !ExternalTableType.Index().name().equalsIgnoreCase(tableType)) {
+              if (!ExternalTableType.Row().name().equalsIgnoreCase(tableType)) {
```
kneeraj: why is index type condition removed.

kneeraj:
lgtm. a minor clarification.

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
sumwale: It will now show up in SYS.HIVETABLES but have not added it to getTables. It should be added to SYS.INDEXES or similar APIs at some point (the GUI can use SYS.HIVETABLES to show them in any case).

kneeraj: Ok.

PR #647, opened 2017-06-07
[SNAP-1703] ArrayIndexOutOfBound when query has wide schema and a fil…

rishitesh:
…ter clause

Cursor variable was not getting advanced in the case of wide schema if a filter is applied and filter value is null for a row. Decoder cursor should get moved irrespective of whether the row is processed or not.  

## Patch testing

Unit Test, precheckin

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
@@ -243,45 +243,35 @@ private[sql] final case class ColumnTableScan(
   }
 
   def splitToMethods(ctx: CodegenContext, blocks: ArrayBuffer[String]): String = {
-    if (blocks.length == 1) {
-      // inline execution if only one block
-      blocks.head
-    } else {
-      val apply = ctx.freshName("apply")
-      val functions = blocks.zipWithIndex.map { case (body, i) =>
-        val name = s"${apply}_$i"
-        val code =
-          s"""
-             |private void $name() {
-             |  $body
-             |}
+    val apply = ctx.freshName("apply")
+    val functions = blocks.zipWithIndex.map { case (body, i) =>
+      val name = s"${apply}_$i"
+      val code =
+        s"""
+           |private void $name() {
+           |  $body
+           |}
          """.stripMargin
-        ctx.addNewFunction(name, code)
-        name
-      }
-      functions.map(name => s"$name();").mkString("\n")
+      ctx.addNewFunction(name, code)
+      name
     }
+    functions.map(name => s"$name();").mkString("\n")
   }
 
   def splitMoveNextMethods(ctx: CodegenContext, blocks: ArrayBuffer[String],
       arg: String): String = {
-    if (blocks.length == 1) {
-      // inline execution if only one block
-      blocks.head
-    } else {
-      val functions = blocks.zipWithIndex.map { case (body, i) =>
-        val name = ctx.freshName("moveNext")
-        val code =
-          s"""
-             |private void $name(int $arg) {
-             |  $body
-             |}
+    val functions = blocks.zipWithIndex.map { case (body, i) =>
+      val name = ctx.freshName("moveNext")
+      val code =
+        s"""
+           |private void $name(int $arg) {
+           |  $body
+           |}
          """.stripMargin
-        ctx.addNewFunction(name, code)
-        name
-      }
-      functions.map(name => s"$name($arg);").mkString("\n")
+      ctx.addNewFunction(name, code)
+      name
     }
+    functions.map(name => s"$name($arg);").mkString("\n")
   }
 
 
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
@@ -263,9 +263,32 @@ private[sql] final case class ColumnTableScan(
     }
   }
 
+  def splitMoveNextMethods(ctx: CodegenContext, blocks: ArrayBuffer[String],
+      arg: String): String = {
+    if (blocks.length == 1) {
+      // inline execution if only one block
+      blocks.head
+    } else {
+      val apply = ctx.freshName("moveNext")
+      val functions = blocks.zipWithIndex.map { case (body, i) =>
+        val name = s"${apply}_$i"
+        val code =
+          s"""
+             |private void $name(int $arg) {
+             |  $body
+             |}
+         """.stripMargin
+        ctx.addNewFunction(name, code)
+        name
+      }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -523,18 +547,26 @@ private[sql] final case class ColumnTableScan(
...
       if (!isWideSchema) {
         moveNextCode.append(genCodeColumnNext(ctx, decoderVar, bufferVar,
-          cursorVar, batchOrdinal, attr.dataType, notNullVar)).append('\n')
+          cursorVar, batchOrdinal, attr.dataType, notNullVar, false)).append('\n')
         val (ev, bufferInit) = genCodeColumnBuffer(ctx, decoderVar, bufferVar,
           cursorVar, attr, notNullVar, weightVarName, false)
         bufferInitCode.append(bufferInit)
         ev
       } else {
+        if (isWideSchema) {
```
hbhanawat: Too many wide schema checks everywhere is making it quite complex. Need to refactor this post 0.9.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -263,9 +263,32 @@ private[sql] final case class ColumnTableScan(
...
 
+  def splitMoveNextMethods(ctx: CodegenContext, blocks: ArrayBuffer[String],
+      arg: String): String = {
+    if (blocks.length == 1) {
+      // inline execution if only one block
+      blocks.head
+    } else {
+      val apply = ctx.freshName("moveNext")
+      val functions = blocks.zipWithIndex.map { case (body, i) =>
+        val name = s"${apply}_$i"
```
hbhanawat: Can't we do a freshName in this block? Would avoid the need to append the _$i thing.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -730,10 +769,18 @@ private[sql] final case class ColumnTableScan(
...
     }
     if (notNullVar != null) {
-      val nullCode =
-        s"final int $notNullVar = $decoder.notNull($buffer, $batchOrdinal);"
-      if (moveNext.isEmpty) nullCode
-      else s"$nullCode\nif ($notNullVar == 1) $moveNext"
+      if (isWideSchema) {
+        ctx.addMutableState("int", notNullVar, "")
+        val nullCode =
+          s"$notNullVar = $decoder.notNull($buffer, $batchOrdinal);"
```
hbhanawat: I didn't understand this code. Can you elaborate why $notNullVar is a class variable?

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -263,9 +263,32 @@ private[sql] final case class ColumnTableScan(
...
+    if (blocks.length == 1) {
+      // inline execution if only one block
+      blocks.head
+    } else {
+      val apply = ctx.freshName("moveNext")
+      val functions = blocks.zipWithIndex.map { case (body, i) =>
+        val name = s"${apply}_$i"
+        val code =
+          s"""
+             |private void $name(int $arg) {
```
hbhanawat: why can't the functions be added directly to the codeGenContext using addNewFunction? This would reduce some code.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
rishitesh: Yes, we can do that as well.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
rishitesh: This variable will be used when actual read happens from the decoder. That itself will be in another method. Clubbing them was causing the problem as sometimes in the case of filters actual reads from the decoder are skipped if the filter does not match. This is why we were getting this problem of ArrayIndexOutOfBounds. So separated moving cursor part and reading part to different methods. Common variables used across these two methods need to be made class variable. Another thing that can be done is the return value from the moveNext() methods. That will require a lot of code change, so I have avoided that route. But as we say wide schema needs a re-visit after 0.9.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
rishitesh: These are very customized functions. Not sure what value it will add to CodeGenContext unless it can be very generic in nature.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
sumwale: Agree, a separate class extending this will be better.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -263,9 +263,31 @@ private[sql] final case class ColumnTableScan(
     }
   }
 
+  def splitMoveNextMethods(ctx: CodegenContext, blocks: ArrayBuffer[String],
+      arg: String): String = {
+    if (blocks.length == 1) {
+      // inline execution if only one block
+      blocks.head
+    } else {
```
sumwale: Inlining for small/single block is not worth it. Just use the common code. JIT inlines small functions in any case if possible at runtime.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -523,18 +546,26 @@ private[sql] final case class ColumnTableScan(
...
           cursorVar, attr, notNullVar, weightVarName, false)
         bufferInitCode.append(bufferInit)
         ev
       } else {
+        if (isWideSchema) {
+          if (moveNextMultCode.length > 1024) {
+            moveNextCodeBlocks.append(moveNextMultCode.toString())
+            moveNextMultCode.clear()
+          }
+        }
```
sumwale: See above comment about inlining. Just use common code and keep this clean.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
rishitesh: This change is to force inlining irrespective of JIT decision. But in our case, this case won't arise as we will be calling splitMethod only in case of the wide schema. Hence that piece code is redundant.  Will remove.

PR #648, opened 2017-06-08
Changes to support slash and special characters in column name [pacth…

SachinJanani:
… from sumedh]

* Changes to support special characters in column name
* Added  unit test

## Patch testing
precheckin

## ReleaseNotes.txt changes
NA
## Other PRs 
NA

### DIFF
--- a/cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
@@ -254,8 +254,8 @@ class SparkSQLExecuteImpl(val sql: String,
       case StringType =>
         TypeUtilities.getMetadata[String](Constant.CHAR_TYPE_BASE_PROP,
           f.metadata) match {
-          case Some(base) =>
-            val size = TypeUtilities.getMetadata[Long](
+          case Some(base) if base != "CLOB" =>
+            lazy val size = TypeUtilities.getMetadata[Long](
               Constant.CHAR_TYPE_SIZE_PROP, f.metadata)
             lazy val varcharSize = size.getOrElse(
               Constant.MAX_VARCHAR_SIZE.toLong).toInt
--- a/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
+++ b/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
@@ -328,7 +328,7 @@ public class SnappyHiveCatalog implements ExternalCatalog {
           Object value = parameters.get(StoreUtils.GEM_INDEXED_TABLE());
           String baseTable = value != null ? value.toString() : "";
           String dmls = JdbcExtendedUtils.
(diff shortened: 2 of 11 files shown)


core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
```diff
@@ -328,7 +328,7 @@ public Object call() throws Exception {
           Object value = parameters.get(StoreUtils.GEM_INDEXED_TABLE());
           String baseTable = value != null ? value.toString() : "";
           String dmls = JdbcExtendedUtils.
-              getInsertOrPutString(fullyQualifiedName, schema, false);
+              getInsertOrPutString(fullyQualifiedName, schema, false, false);
```
dshirish: getInsertOrPutString does not need to escape quotes as the insert/put formed here isn't used?

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
```diff
@@ -254,8 +254,8 @@ class SparkSQLExecuteImpl(val sql: String,
       case StringType =>
         TypeUtilities.getMetadata[String](Constant.CHAR_TYPE_BASE_PROP,
           f.metadata) match {
-          case Some(base) =>
-            val size = TypeUtilities.getMetadata[Long](
+          case Some(base) if base != "CLOB" =>
+            lazy val size = TypeUtilities.getMetadata[Long](
```
sumwale: This is related to "normalizeSchema" calls which transform these properties into something else (which was not clear to me why it was being done in the first place): https://github.com/SnappyDataInc/snappydata/blob/master/core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala#L312

We are now avoiding normalization at the lowest level to allow quoted case-sensitive names to be passed through from SQL layer as is without changing case. The earlier code there would take out this property but no longer happens at that level. We could explicitly change the schema for incoming query (for saveAsTable) in SnappySession.createTableAsSelect which only normalizes names currently (see SnappySession.scala:1133) but it is not worth it since CLOB is no longer an issue at all with the thrift driver.

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
sumwale: Normally string created will have quotes which works just fine for prepareStatement etc. The case where this has to be escaped is when the generated string is embedded into generated code i.e. a string like 'prepareStatement("SELECT ..."COL1"...")' which is compiled. Here the quotes have to be escaped.

dshirish: Ok

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
dshirish: ok

PR #654, opened 2017-06-11
[SNAP-1710] separate plan to fallback to Spark

sumwale:

- CodegenSparkFallback plan added at the very top to catch code generation failures and
  fallback to Spark plans. This plan is completely transparent and delegates everything
  to child (including "children" itself) so will never be visible in SparkPlanInfo or anywhere else.
- Switch the "disableStoreOptimizations" flag in its execute* overrides to fallback to Spark plans.
- Removed the old handling in SnappySession.load that only works for sql() API while
  above works for all cases.
- Added testcase in ColumnCacheBenchmark
- Removed OFFHEAP option from StoreUtils which refers to Gem/GemXD option and is not supported.
- Some scalastyle corrections.

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

Some AQP changes

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
@@ -19,7 +19,6 @@ package org.apache.spark.sql
 import io.snappydata.Property
 
 import org.apache.spark.sql.JoinStrategy._
-import org.apache.spark.sql.backwardcomp.ExecutedCommand
 import org.apache.spark.sql.catalyst.expressions.aggregate.{AggregateExpression, AggregateFunction, Complete, Final, ImperativeAggregate, Partial, PartialMerge}
 import org.apache.spark.sql.catalyst.expressions.{Alias, Expression, NamedExpression, RowOrdering}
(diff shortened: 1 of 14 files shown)


core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
```diff
@@ -670,10 +670,16 @@ case class CollapseCollocatedPlans(session: SparkSession) extends Rule[SparkPlan
   * Rule to collapse the partial and final aggregates if the grouping keys
   * match or are superset of the child distribution.
   */
-case class InsertCachedPlanHelper(session: SparkSession) extends Rule[SparkPlan] {
-  override def apply(plan: SparkPlan): SparkPlan = plan.transformUp {
+case class InsertCachedPlanHelper(session: SnappySession, topLevel: Boolean)
+    extends Rule[SparkPlan] {
+  private def addFallback(plan: SparkPlan): SparkPlan = {
+    if (!topLevel || session.sessionState.disableStoreOptimizations) plan
```
rishitesh: Could not get this check. Can you please put some comments here.

sumwale: done

PR #657, opened 2017-06-15
[SNAP-1714] correcting case-sensitivity handling for API calls

sumwale:

- pass through "caseSensitive" flag to ColumnTableScan and RowTableScan to
  do resolution as per case-sensitivity settings (in base PartitionedPhysicalPlan.fieldIndex);
  the method uses a temporary LocalRelation to invoke resolveQuoted
- always use case-insensitive resolution for partitioned column resolution since
  it may be case-insensitive during CREATE TABLE (and we are not storing that
      flag in the DDL itself); this means that users can't have two columns
  that just differ in case-sensitivity esp for partitioning which is a reasonable
  restriction
- SnappyBaseParser now has caseSensitive flag as a var which is updated on each
  top-level parse() call since the flag in SQLConf may have changed
- added a test in QueryTest as per the failure scenario mentioned in SNAP-1714
  with different case checks including for caseSensitiveAnalysis=false or true
  apart from quoted names
- some scalaStyle corrections

## Patch testing

precheckin

## ReleaseNotes.txt changes

Cannot have column names in the table that differ only in case-sensitivity especially
the partitioning columns.

## Other PRs 

NA

### DIFF
--- a/cluster/src/test/scala/io/snappydata/QueryTest.scala
+++ b/cluster/src/test/scala/io/snappydata/QueryTest.scala
@@ -21,7 +21,7 @@ import scala.collection.JavaConverters._
 
 import org.apache.spark.sql.execution.benchmark.ColumnCacheBenchmark
 import org.apache.spark.sql.internal.SQLConf
-import org.apache.spark.sql.{Row, SnappyContext, SparkSession}
+import org.apache.spark.sql.{AnalysisException, Row, SnappyContext, SparkSession}
 
 class QueryTest extends SnappyFunSuite {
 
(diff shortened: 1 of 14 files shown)


rishitesh:
LGTM. One clarification sought.

kneeraj:
@rishitesh I think, if column names are created with quotes then we need to maintain the case sensitivity and the query itself should contain columns inside quotes. @sumwale can clarify.

sumwale:
> Why would one want to change case sensitivity withing a session?

Maybe not, but it should work nevertheless.

PR #661, opened 2017-06-17
Changes for reverting the port randomization for spark master and few fixes 

swatisawant:

- Changes for reverting the port randomization for spark master.
- Reverting the changes for SPARK_MASTER_HOST as start-master.sh - Starts a master instance on the machine the script is executed on and not on the host provided in this parameter
 - Added task for recording all snappy and spark processID's with host details and then later using them while cleaning up the processes in multi host environment

## Patch testing

Verified the changes by running northWind bt in multi host environment

## ReleaseNotes.txt changes

NA

## Other PRs 

No


dtests/src/test/java/io/snappydata/hydra/cluster/SnappyTest.java
```diff
@@ -2291,6 +2282,20 @@ public static synchronized void HydraTask_startSnappyCluster() {
     }
   }
 
+  public static void HydraTask_recordProcessIDWithHost() {
+    snappyTest.recordSnappyProcessIDinNukeRun("LeaderLauncher");
```
sonalsagarwal: We can also shift it to the else block.

PR #669, opened 2017-06-22
SNAP-1737: 

vivekwiz:
While passing value to XD, it should converted from catalyst type to scala type. This was missed for partition pruning changes and that has been corrected.

## Patch testing
precheckin.

## ReleaseNotes.txt changes
No

## Other PRs 
No


vivekwiz:
@sumwale Doing same and also writing a test case. Will update here.

vivekwiz:
@sumwale Debugging index issue was taking time, so created another bug SNAP-1745

PR #675, opened 2017-06-30
[SNAP-1752] Use ManagedByteBuffer in CachedDataFrame

rishitesh:
Changed the ByteBufferAllocator to use ManagedByteBuffer 
for heap as well as off-heap cases.
## Patch testing

pre-checkin

## ReleaseNotes.txt changes

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
+++ b/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
@@ -412,7 +412,6 @@ object CachedDataFrame
     val buffer = new Array[Byte](4 << 10) // 4K
     // final output is written to this buffer
     val output = new ByteBufferDataOutput(4 << 10,
-      DirectBufferAllocator.instance(),
       null,
       ManagedDirectBufferAllocator.DIRECT_STORE_DATA_FRAME_OUTPUT)
     // holds intermediate bytes which are compressed and flushed to output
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -412,7 +412,6 @@ object CachedDataFrame
     val buffer = new Array[Byte](4 << 10) // 4K
     // final output is written to this buffer
     val output = new ByteBufferDataOutput(4 << 10,
-      DirectBufferAllocator.instance(),
```
sumwale: Now this will be outside of all accounting. Won't this result in a genuine OOME for large results? Please test with the original "select * from largeTable" case. I think better option will be to catch the DirectByteBuffer OOME here and convert to a LowMemoryException.

rishitesh: If a user has not set -memory-size we will be using more memory than -Xmx. JVM heap plus some off-heap memory. Would that be alright?  
A cleaner solution would be to use accounting in HeapBufferAllocator. 
The problem is we don't know which pool to access there. Storage or execution.

sumwale: JVM sets the default limit on DirectBuffer allocations as per free RAM at startup. I guess the issue might be that it could be too small in some cases? But users can increase it explicitly in any case for pure heap case using JVM standard system property. I think that's better than hitting a genuine OOME due to lack of accounting. We can keep the HeapAllocator accounting for a bit later given that we expect enterprise deployments to all use off-heap where this is not an issue.

rishitesh: Sure .Will change accordingly.

rishitesh: We should document this fact and to set JVM property -XX:MaxDirectMemorySize=<1024MB> as best practice.  1024 from default spark.driver.maxResultSize

rishitesh: Will cover this change in https://github.com/SnappyDataInc/snappydata/pull/677 as both deal code in similar area.

PR #676, opened 2017-06-30
Changes for SNAP-1678

dshirish:
Added log statements so that it will be easier to find the cluster mode by looking at the logs.

## Patch testing
precheckin

## ReleaseNotes.txt changes
Not applicable
## Other PRs 
Not applicable

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
@@ -997,7 +997,7 @@ object SnappyContext extends Logging {
   }
 
   private def resolveClusterMode(sc: SparkContext): ClusterMode = {
-    if (sc.master.startsWith(Constant.JDBC_URL_PREFIX)) {
+    val mode = if (sc.master.startsWith(Constant.JDBC_URL_PREFIX)) {
       if (ToolsCallbackInit.toolsCallback == null) {
         throw new SparkException("Missing 'io.snappydata.ToolsCallbackImpl$'" +
             " from SnappyData tools package")
@@ -1035,6 +1035,13 @@ object SnappyContext extends Logging {
         else ExternalClusterMode(sc, sc.master)
       }
     }
+    mode match {
+      case LocalMode(_, _) =>
+        logWarning(s"Initializing SnappyData in cluster mode: $mode")
+      case _ =>
+        logInfo(s"Initializing SnappyData in cluster mode: $mode")
+    }
+    mode
   }
 
   private[sql] def initGlobalSnappyContext(sc: SparkContext,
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1162,6 +1169,11 @@ object SnappyContext extends Logging {
 abstract class ClusterMode {
   val sc: SparkContext
   val url: String
+  val description: String = "Cluster mode"
+
+  override def toString: String = {
+    description + s": sc = $sc, url = $url"
```
sumwale: Use s"" all the way: s"$description: sc = $sc, url = $url"

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1035,6 +1035,13 @@ object SnappyContext extends Logging {
         else ExternalClusterMode(sc, sc.master)
       }
     }
+    mode match {
+      case LocalMode(_, _) =>
+        logWarning(s"Initializing SnappyData in cluster mode: $mode")
```
sumwale: why warning for local mode?

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1207,15 +1219,19 @@ final class BlockAndExecutorId(private[spark] var _blockId: BlockManagerId,
...
 /**
  * This is for the two cluster mode: one is the normal snappy cluster, and
  * this one is a separate local/Spark/Yarn/Mesos cluster fetching data from
  * the snappy cluster on demand that just remains like an external datastore.
  */
 case class SplitClusterMode(override val sc: SparkContext,
-    override val url: String) extends ClusterMode
+    override val url: String) extends ClusterMode {
+  override val description: String = "Split cluster mode"
+}
```
sumwale: Is this used anymore? Should be removed.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
dshirish: Changed now to info level. Earlier had kept it in warning level as per suggestion in SNAP-1678 so that users don't use local mode by mistake. But info will work being the default log level.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
dshirish: Done

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
dshirish: Had disabled this in 0.9 for users. In the code all occurrences are still to be removed. Using SNAP-1610 to track it. Changes for SNAP-1610 are ready in another branch and will merge those after opening a PR.

PR #677, opened 2017-07-03
[SNAP-1688] Use execution memory for serialized result.

rishitesh:
We have safeguarded the result serialization memory consumption from each partition. These safeguards are put in CachedDataFrame. Before copying the result iterator to heap we ask UMM for memory availability. If sufficient memory is not available the partition of result task fails. 

When accounting in UMM we take memory from the storage pool. This leads to code complexity and unused execution pool space. Also, it unnecessarily evicts some table data which otherwise could have fitted in memory

Use execution memory instead of storage memory for CachedDataFrame serialized memory.
For this, we create a memory consumer which acts on behalf of ResultTask and asks for execution memory. 

Also added a guard in execution pool to stop growing the pool beyond a storage fraction. 

Couple of test changes.

## Patch testing

pre-checkin, manual tests

## ReleaseNotes.txt changes

NA

## Other PRs 
NA

### DIFF
--- a/cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
+++ b/cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
@@ -86,8 +86,9 @@ class SnappyUnifiedMemoryManager private[memory](
 
   private val evictionFraction = SnappyUnifiedMemoryManager.getStorageEvictionFraction(conf)
 
-  private val maxHeapStorageSize =
-    (maxHeapMemory * evictionFraction).toLong
+  private val maxHeapStorageSize = (maxHeapMemory * evictionFraction).toLong
+
+  private val maxHeapExecutionSize = (maxHeapMemory * evictionFraction).toLong
 
   private val minHeapEviction = math.min(math.max(10L * 1024L * 1024L,
     (maxHeapStorageSize * 0.002).toLong), 1024L * 1024L * 1024L)
@@ -343,6 +344,15 @@ class SnappyUnifiedMemoryManager private[memory](
           getMinOffHeapEviction(numBytes))
     }
 
+    // Stop execution pool to grow beyond eviction percentage of heap.
+    if (memoryMode eq MemoryMode.ON_HEAP) {
+      if (executionPool.poolSize + numBytes > maxHeapExecutionSize) {
+        logWarning(s"MemoryManager can't allocate $numBytes bytes as it " +
+            s"would exceed maxHeapExecutionSize = $maxHeapExecutionSize")
+        return 0L
+      }
+    }
+
     /**
       * Grow the execution pool by evicting cached blocks, thereby shrinking the storage pool.
       *
@@ -466,15 +476,6 @@ class SnappyUnifiedMemoryManager private[memory](
             getMinOffHeapEviction(numBytes))
       }
 
-      // Evict only limited amount for owners marked as non-evicting.
-      // TODO: this can be removed once these calls are moved to execution
-      // TODO use something like "(spark.driver.maxResultSize / numPartitions) * 2"
-      val doEvict = if (shouldEvict &&
-          objectName.endsWith(BufferAllocator.STORE_DATA_FRAME_OUTPUT)) {
-        // don't use more than 30% of pool size for one partition result
-        numBytes < math.min(0.3 * storagePool.poolSize,
-          math.max(maxPartResultSize, storagePool.memoryFree))
-      } else shouldEvict
 
       if (numBytes > maxMemory) {
         // Fail fast if the block simply won't fit
@@ -486,7 +487,7 @@ class SnappyUnifiedMemoryManager private[memory](
       // don't borrow from execution for off-heap if shouldEvict=false since it
       // will try clearing references before calling with shouldEvict=true again
       val offHeap = memoryMode eq MemoryMode.OFF_HEAP
-      val offHeapNoEvict = !doEvict && offHeap
+      val offHeapNoEvict = !shouldEvict && offHeap
       if (numBytes > storagePool.memoryFree && !offHeapNoEvict) {
         // There is not enough free memory in the storage pool, so try to borrow free memory from
         // the execution pool.
@@ -518,7 +519,7 @@ class SnappyUnifiedMemoryManager private[memory](
           return false
         }
 
-        if (doEvict) {
+        if (shouldEvict) {
           // Sufficient memory could not be freed. Time to evict from SnappyData store.
           // val requiredBytes = numBytes - storagePool.memoryFree
           // Evict data a little more than required based on waiting tasks
--- a/cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
+++ b/cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
@@ -361,13 +361,11 @@ class SnappyUnifiedMemoryManager private[memory](
           return
         }
 
-        // Stop execution pool to grow beyond eviction percentage of heap.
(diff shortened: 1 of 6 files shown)


cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -86,8 +86,9 @@ class SnappyUnifiedMemoryManager private[memory](
 
   private val evictionFraction = SnappyUnifiedMemoryManager.getStorageEvictionFraction(conf)
 
-  private val maxHeapStorageSize =
-    (maxHeapMemory * evictionFraction).toLong
+  private val maxHeapStorageSize = (maxHeapMemory * evictionFraction).toLong
+
+  private val maxHeapExecutionSize = (maxHeapMemory * evictionFraction).toLong
```
sumwale: Do we want execution to grow into storage? I think it will be better to avoid that so maxHeapExecutionSize can be governed by "spark.memory.storageFraction" i.e. 50% by default. Also need to assert that storageFraction should be less than or equal to evictionFraction.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -409,18 +409,22 @@ object CachedDataFrame
   override def apply(context: TaskContext,
       iter: Iterator[InternalRow]): PartitionResult = {
     var count = 0
-    val buffer = new Array[Byte](4 << 10) // 4K
+    val buffer = new Array[Byte](4 << 10)
+    // 4K
```
sumwale: Idea autoformat messes this. Perhaps better to remove that dangling comment since it is obvious.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -409,18 +409,22 @@ object CachedDataFrame
...
     // final output is written to this buffer
-    val output = new ByteBufferDataOutput(4 << 10,
-      DirectBufferAllocator.instance(),
-      null,
-      ManagedDirectBufferAllocator.DIRECT_STORE_DATA_FRAME_OUTPUT)
+    var output: ByteBufferDataOutput = null
     // holds intermediate bytes which are compressed and flushed to output
-    val maxOutputBufferSize = 64 << 10 // 64K
+    val maxOutputBufferSize = 64 << 10
+    // 64K
```
sumwale: remove the comment line

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -409,18 +409,22 @@ object CachedDataFrame
...
     // can't enforce maxOutputBufferSize due to a row larger than that limit
     val bufferOutput = new Output(4 << 10, -1)
     var outputRetained = false
     try {
+      output = new ByteBufferDataOutput(4 << 10,
+        DirectBufferAllocator.instance(),
+        null,
+        ManagedDirectBufferAllocator.DIRECT_STORE_DATA_FRAME_OUTPUT)
+
```
sumwale: Shouldn't this too go into execution if off-heap is configured?

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -360,6 +361,15 @@ class SnappyUnifiedMemoryManager private[memory](
...
 
+        // Stop execution pool to grow beyond eviction percentage of heap.
+        if (memoryMode eq MemoryMode.ON_HEAP) {
+          if (executionPool.memoryUsed + extraMemoryNeeded > maxHeapExecutionSize) {
+            logWarning(s"MemoryManager can't allocate $numBytes bytes as it " +
+                s"would exceed maxHeapExecutionSize = $maxHeapExecutionSize")
+            return
+          }
+        }
+
```
sumwale: Not clear what this is accomplishing and why only for heap. What was the issue with previous way which would grow execution till storageFraction (unless there was some bug in that)?

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
rishitesh: Yes. But right now there is no clear way to do this via DirectBufferAllocator. I can add that too. Will need some more work at BufferAllocator level.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
rishitesh: Will do.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
rishitesh: Will change.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
rishitesh: This is the case when storage has not reached till storageFraction. Should we not allow execution to borrow from storage in such cases?  We can stop execution pool to grow beyond storageFraction, but some space would be wasted if the user has allocated memory, but table data is less.
The check for ON-HEAP is done as critical heap kicks in if push heap memory too much. OFF-HEAP memory can take all of storage free memory.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
sumwale: Ok needs some changes to StoreCallbacks etc so can be done later. But then you do require the 30% hack for storage above else this can push out the entire region data.

rishitesh: Yeah..  Will put it back.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
sumwale: Does this happen after eviction of execution blocks has been done or before?

rishitesh: There is no concept of execution block eviction in line with storage eviction. Individual MemoryConsumer might implement spill() method. If they don't get enough memory then they might spill their data to disk. We should discuss which all SnappyData execution structures we can spill to disk and how.

sumwale: Most common case is Sort and SortMergeJoin which have the capability to spill. What happens if storage wants its area back? Will it force spill then?

rishitesh: No, if at a time store also wants the area it would evict its own area. As execution memory is temporary after some time store can reclaim its area. But at that very moment store has to evict some of its data. It can not force spill execution structures.

sumwale: Hmm, then this needs to be re-looked at. We should assume that in a running cluster there are concurrent queries going on which can keep the execution area occupied continuously. If user wants to ingest something into storage then it will be trouble in terms of memory use. We should change this so that execution never grows into storage as per the fraction for both heap and off-heap. At least lower the max limit of execution to be much smaller, and closer to fraction rather than eviction limit.

rishitesh: Ok. Will change the limit to storage fraction for both heap & off-heap cases.

PR #678, opened 2017-07-03
[SNAP-1760] Correct null bitset expansion and reduce copying in inserts

sumwale:

- ensure that the expanded bitset size is able to accomodate the incoming ordinal
- reduce byte copying and boxing for wide-table inserts:
  - use SpecificMutableRow instead of GenericMutableRow to avoid primitive boxing
  - use own setColumn method instead of CodegenContext.setColumn that avoids cloning
    of UTF8String which is not really required
- refactored encoder close in ColumnInsertExec to use for wide-table case too
  (else those remain dangling until collected by some GC cycle)
- use a maxDeltaRows for wide-schema case too in ColumnInsertExec
- use buffer.limit() instead of capacity in ByteBufferHashMap
- corrected some scalastyle/indentation

## Patch testing

precheckin, failing test cases

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/spark/commit/5147b60ec4a627565a692dba59244bbd214cd3e4 reduces byte copying in Spark layer when reading from dictionary
Also see the store PR


sumwale:
Merging now. Will take care of any comments subsequently.

PR #682, opened 2017-07-04
[SNAP-1748] remove serialization header in the column data

sumwale:
The serialization header (i.e. typeId/classId etc written by DataSerializer) was added to the column data buffer itself as part of SNAP-1427 to reduce a buffer copy but was obsoleted by SNAP-1454 that made the column table values as full objects.

This change removes the header from the data and writes to stream when required like in normal serialization. However, note that the in-memory buffer may still contain the header because disk read will transfer the buffer as is into memory avoiding a copy but the code does not see it or deal with it explicitly.

These changes simplify code, make buffer handling more precise and also fix the issue reported in SNAP-1748.

- Remove explicit serialization header writing and maintenance in ColumnFormatValue
  and instead write it only when doing a "raw" push to channel
- Add writeSerializationHeader implementation for the overflow case which does not write to
  the channel directly rather copies to an intermediate buffer. No attempt has been made
  to reduce copying for this specific code since this should be a rare use-case.

With these changes the overflow case mentioned in SNAP-1748 works correctly. However, it frequently fails with LowMemoryException in inserts (with say 10GB size) even with off-heap because overflow writes to disk only when eviction is required and can take long so has to keep the in-memory buffer alive while new inserts start pushing and asking for memory.

I guess we can completely inhibit the overflow-only case for column tables. However, the changes to simplify handling of header are required in any case for better code maintenance (also present on SNAP-1743 branch).

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/223

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatEntry.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatEntry.scala
@@ -66,7 +66,35 @@ object ColumnFormatEntry extends Logging {
     GfxdDataSerializable.registerSqlSerializable(classOf[ColumnFormatValue])
   }
 
+  /**
+   * Size of the serialization type/class IDs of [[ColumnFormatValue]].
+   */
+  private[columnar] val VALUE_HEADER_CLASSID_SIZE = 3
+
+  /**
+   * Size of the serialization header of [[ColumnFormatValue]] including the
+   * serialization type/class IDs (3 bytes), padding (1 byte), and size (4).
+   * Padding is added for 8 byte alignment.
+   */
+  private[columnar] val VALUE_HEADER_SIZE = VALUE_HEADER_CLASSID_SIZE + 1 + 4
+
+  private[columnar] val VALUE_EMPTY_BUFFER = {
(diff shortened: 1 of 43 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatEntry.scala
```diff
@@ -399,7 +398,7 @@ final class ColumnFormatValue
   override def toData(out: DataOutput): Unit = {
     val buffer = getBufferRetain
     try {
-      val numBytes = buffer.remaining()
+      val numBytes = buffer.limit()
```
rishitesh: Some basic question. What happens when a thread tries to read the buffer when it is getting serialized, especially during overflow.

rishitesh: Ok. We are creating a duplicate buffer during scans. So it should be ok i guess.

sumwale: Right internal buffer is never touched once set so its position/limit is fixed. The getBufferRetain either returns a duplicate or slice to ensure that returned buffer always has initial position as 0.

PR #688, opened 2017-07-05
SNAP-1708

kneeraj:
Improvements in collect-debug-artifacts script

- Removed the necessity of passwordless ssh both ways.
- Added an option to extract the collected artifact.

## Patch testing

Manual

## ReleaseNotes.txt changes

None.

## Other PRs 

No.

### DIFF
--- a/cluster/sbin/collect-debug-artifacts.sh
+++ b/cluster/sbin/collect-debug-artifacts.sh
@@ -456,7 +456,6 @@ check_configs
 # host pid cwd
 
 if [ -n "${TAR_FILE}" ]; then
-  echo "Calling extract function"
   extract "${TAR_FILE}"
   exit 0
 fi
@@ -491,19 +490,19 @@ while  read -r line || [[ -n "$line" ]]; do
     echo "host: $host pid: $pid and cwd: $cwd"
   fi
 
-  ( collect_data $host $cwd )
-#  all_pids+=($!) 
+  collect_data $host $cwd &
+  all_pids+=($!) 
 done < $tmp_members_file
 
 # wait for all the collection to end on respective hosts
 # Then rsync 1 by 1
-#for p in "${all_pids[@]}"
-#do
-#  if [ "${VERBOSE}" = "1" ]; then
-#    echo "Waiting for pid ${p}"
-#  fi
-#  wait $p 2> /dev/null
-#done
+for p in "${all_pids[@]}"
+do
+  if [ "${VERBOSE}" = "1" ]; then
+    echo "Waiting for pid ${p}"
+  fi
+  wait $p 2> /dev/null
+done
 
 while  read -r line || [[ -n "$line" ]]; do
   read host cwd <<< $line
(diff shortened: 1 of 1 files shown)


cluster/sbin/collect-debug-artifacts.sh
```diff
@@ -396,31 +424,42 @@ function collect_on_remote {
...
+    echo "FILES=${files[@]} zipped and copied to ${collector_host}:$tmp_dir}"
   fi
 }
 
 check_configs
 # Assuming each line in the members info file has the following format
 # host pid cwd
 
+if [ -n "${TAR_FILE}" ]; then
+  extract "${TAR_FILE}"
```
sumwale: run this in a sub-shell "( extract ... )" since the extract function is doing a "cd" so we don't want user's shell working directory itself to change.

cluster/sbin/collect-debug-artifacts.sh
```diff
@@ -169,40 +184,45 @@ function check_configs {
...
 
+function extract {
+    debugtarzip="$1"
+    xtractdir=`dirname ${debugtarzip}`
+    tarname=`basename ${debugtarzip}`
+    cd $xtractdir
+    tar -xvf $tarname
+    dir=`echo "$tarname" | cut -d'.' -f1-5`
+    cd $dir
+    gunzip `find . -name "*.gz"`
```
sumwale: Can make this more useful by also un-tarring like:

```sh
cd $dir
for tarF in `find . -name '*.tar.gz'`; do
  ( cd "`dirname "$tarF"`" && tar xzf "`basename "$tarF"`" )
done
```

cluster/sbin/collect-debug-artifacts.sh
```diff
@@ -212,17 +232,25 @@ function collect_on_remote {
...
+  verbose="$6"
+  start_epoch="$7"
+  end_epoch="$8"
+
+  # Create a .tmpcda dir if not exists else empty it
+  tmp_dir="$data_dir/.tmpcda"
+  if [ -d ${tmp_dir} ]; then
+    rm -rf ${tmp_dir}/*
+  else
+    mkdir -p $tmp_dir
```
sumwale: Shouldn't this be done on the remote server itself? Else you are assuming the working directory is available over NFS but for persistent files we recommend having them locally.

kneeraj: @sumwale the function 'collect_on_remote' is going to run on individual remote host only. Are you meaning something else here. At no place it is assumed that NFS share will be there. Please check again.

cluster/sbin/collect-debug-artifacts.sh
kneeraj: ok.

cluster/sbin/collect-debug-artifacts.sh
kneeraj: ok.

kneeraj:
Closing this as the changes were taken care in SNAP-1708 dup PR #723

PR #691, opened 2017-07-07
Snap 1656 vivek

vivekwiz:
First set of changes to enable,
1. Secured startup (with an admin credential) of Snappydata
2. Secured JDBC connection (with any user credential) for DDL and DML statements.

These are only subset of deliverable planned and will be merged into SNAP-1656 (and not master).

## Patch testing
Changes have been tested for basic queries on row and column table. 

## ReleaseNotes.txt changes
No

## Other PRs
https://github.com/SnappyDataInc/snappy-store/pull/240


cluster/src/dunit/scala/io/snappydata/cluster/SecureQueryRoutingDUnitTest.scala
```diff
@@ -0,0 +1,229 @@
...
+import io.snappydata.test.dunit.AvailablePortHelper
+
+import org.apache.spark.Logging
+import org.apache.spark.sql.collection.Utils
+
+class SecureQueryRoutingDUnitTest(val s: String)
+    extends ClusterManagerTestBase(s) with Logging {
+
+  override def setUp(): Unit = {
+    val ldapServer = LdapTestServer.getInstance(getClass.getResource("/auth.ldif").getPath)
```
sumwale: Instead of starting and stopping in every test you could move this code to "beforeClass()" and "afterClass()" which will do once per suite.

cluster/src/dunit/scala/io/snappydata/cluster/SecureQueryRoutingDUnitTest.scala
```diff
@@ -0,0 +1,229 @@
...
+    if (!ldapServer.isServerStarted) {
+      ldapServer.startServer()
+    }
+    setSecurityProps(ldapServer)
```
sumwale: Any reason to use this separately and not use "SecurityTestUtils.startLdapServerAndGetBootProperties()"?

cluster/src/dunit/scala/io/snappydata/cluster/SecureQueryRoutingDUnitTest.scala
```diff
@@ -0,0 +1,229 @@
...
+    super.setUp()
+  }
+
+  override def tearDown2(): Unit = {
+    super.tearDown2()
+    val ldapServer = LdapTestServer.getInstance()
+    if (ldapServer.isServerStarted) {
+      ldapServer.stopService()
+    }
```
sumwale: Can be moved to "afterClass()" as mentioned.

cluster/src/dunit/scala/io/snappydata/cluster/SecureQueryRoutingDUnitTest.scala
```diff
@@ -0,0 +1,229 @@
...
+
+  override def setUp(): Unit = {
+    val ldapServer = LdapTestServer.getInstance(getClass.getResource("/auth.ldif").getPath)
+    if (!ldapServer.isServerStarted) {
+      ldapServer.startServer()
+    }
+    setSecurityProps(ldapServer)
+    super.setUp()
+  }
+
```
sumwale: Unless you are changing network server properties, you can "override val stopNetServersInTearDown = false" and explicitly stop all network servers in beforeClass/afterClass. This will also help check somewhat longer running VMs with different tests (and their cleanup) as more are added to this class.

cluster/src/dunit/scala/io/snappydata/cluster/SecureQueryRoutingDUnitTest.scala
```diff
@@ -0,0 +1,229 @@
...
+        com.pivotal.gemfirexd.Constants.AUTHENTICATION_PROVIDER_LDAP)
+    props += (AUTH_LDAP_SERVER -> s"ldap://localhost:${ldapServer.getServerPort}/")
+    props += (AUTH_LDAP_SEARCH_BASE -> "ou=ldapTesting,dc=pune,dc=gemstone,dc=com")
+    props += (AUTH_LDAP_SEARCH_DN ->
+        s"uid=$ldapAdminUser,ou=ldapTesting,dc=pune,dc=gemstone,dc=com")
+    props += (AUTH_LDAP_SEARCH_PW -> ldapAdminUser)
+    for ((k, v) <- props) System.setProperty(k, v)
+
+    props += (Attribute.USERNAME_ATTR -> snappyAdminUser)
+    props += (Attribute.PASSWORD_ATTR -> snappyAdminUser)
```
sumwale: Can directly use SecurityTestUtils.startLdapServerAndGetBootProperties as mentioned (locator port is in ClusterManagerTestBase.locPort).

cluster/src/dunit/scala/io/snappydata/cluster/SecureQueryRoutingDUnitTest.scala
```diff
@@ -0,0 +1,229 @@
...
+    // scalastyle:on println
+
+    val tableName = "order_line_col"
+    createTable1(serverHostPort, tableName, jdbcUser, jdbcUser)
+    insertRows1(20000, serverHostPort, tableName, jdbcUser, jdbcUser)
+
+    // (1 to 5).foreach(d => query())
+    query1(serverHostPort, tableName, jdbcUser, jdbcUser)
+    dropTable(serverHostPort, tableName, jdbcUser, jdbcUser)
+  }
```
sumwale: Wouldn't the main checks required here be negative checks? That is the operations should fail if there is no authorization to do those e.g. writing to other's schema.

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
```diff
@@ -375,11 +375,26 @@ private void initHMC() {
...
+      if (bootProperties.containsKey("user") && bootProperties.containsKey("password")) {
+        urlSecure = "jdbc:snappydata:" +
+            ";default-schema=" + SnappyStoreHiveCatalog.HIVE_METASTORE() +
+            ";user=" + bootProperties.get("user") +
+            ";password=" + bootProperties.get("password") +
+            ";disable-streaming=true;default-persistent=true";
+        metadataConf.setVar(HiveConf.ConfVars.METASTORE_CONNECTION_USER_NAME,
+            bootProperties.get("user").toString());
+        metadataConf.setVar(HiveConf.ConfVars.METASTOREPWD,
+            bootProperties.get("password").toString());
```
sumwale: Do you need the properties both in URL and in metadataConf, or only latter should work?

core/src/main/scala/org/apache/spark/jdbc/ConnectionUtil.scala
```diff
@@ -62,7 +62,9 @@ object ConnectionUtil {
       case SparkContext.DRIVER_IDENTIFIER => connectionProps.connProps
       case _ => connectionProps.executorConnProps
     }
-    val jdbcOptions = new JDBCOptions(connectionProps.url, "", connProps.asScala.toMap)
+    
+    val jdbcOptions = new JDBCOptions(connectionProps.url +
+      connectionProps.urlSecureSuffix, "", connProps.asScala.toMap)
```
sumwale: Any reason to add a separate "urlSecureSuffix" field and not add to the "url" itself? Alternatively it might be cleaner to put those properties in connProps and executorConnProps.

core/src/main/scala/org/apache/spark/sql/execution/ConnectionPool.scala
```diff
@@ -132,8 +134,20 @@ object ConnectionPool {
...
-    val ds = getPoolDataSource(id, poolProps, connProps, hikariCP)
+      hikariCP: Boolean, urlSuffix: String): Connection = {
+    val url = poolProps.get("url")
+    val poolPropsSecure = if (url.isDefined) {
+      val urlSecure = if (id.startsWith(Constant.SHADOW_SCHEMA_NAME)) {
+        val bootProperties = Misc.getMemStore.getBootProperties
+        if (bootProperties.containsKey("user") && bootProperties.containsKey("password")) {
+          url.get + ";user=" + bootProperties.get("user") +
+              ";password=" + bootProperties.get("password") +
+              ";default-schema=" + Misc.SNAPPY_HIVE_METASTORE + ";"
```
sumwale: Why setting DBA properties here?? This won't work for smart connector. The user/password should be coming as part of connProps.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -274,9 +275,16 @@ object ExternalStoreUtils extends Logging {
...
+    val allPoolProps = getAllPoolProperties(url, driver, poolProps, hikariCP, isEmbedded)
+    val EMPTY_URL = ""
+    val urlSecureSuffix = if (session.isDefined) {
+      val user = session.get.conf.getOption(Attribute.USERNAME_ATTR)
+      val password = session.get.conf.getOption(Attribute.PASSWORD_ATTR)
+      if (user.isDefined && password.isDefined) {
+        s";user=${user.get};password=${password.get};"
+      } else EMPTY_URL
+    } else EMPTY_URL
+    ConnectionProperties(url, urlSecureSuffix, driver, dialect, allPoolProps,
```
sumwale: Again I think filling into connProps and executorConnProps will be better and avoid the changes related to addition of "urlSecureSuffix".

core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCAppendableRelation.scala
```diff
@@ -66,9 +66,23 @@ abstract case class JDBCAppendableRelation(
...
+    val EMPTY_URL = ""
+    val user = sqlContext.conf.getConfString("snappydata.store.user", EMPTY_URL)
+    val password = sqlContext.conf.getConfString("snappydata.store.password", EMPTY_URL)
+    if (!user.equals(EMPTY_URL) && !password.equals(EMPTY_URL)) {
+      JdbcUtils.createConnectionFactory(new JDBCOptions(connProperties.url +
+        ";user=" + user +
+        ";password=" + password +
+        ";default-schema=" + SnappyStoreHiveCatalog.HIVE_METASTORE + ";",
+        table, connProperties.connProps.asScala.toMap))
+    } else connFactory
```
sumwale: This is being done in multiple places. Refactor this portion out into a separate method in Utils or somewhere that will read the properties from SQLConf and fill into connProperties (preferably into connProps and executorConnProps as mentioned before).

core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCAppendableRelation.scala
```diff
@@ -229,7 +243,7 @@ abstract case class JDBCAppendableRelation(
           JdbcExtendedUtils.executeUpdate(tableStr, conn)
           dialect match {
             case d: JdbcExtendedDialect => d.initializeTable(tableName,
-              sqlContext.conf.caseSensitiveAnalysis, conn)
+              sqlContext.conf.caseSensitiveAnalysis, conn, sysConnFactory)
```
sumwale: Why is sysConnFactory required separately? Shouldn't the "conn" itself have the relevant authorization?

core/src/main/scala/org/apache/spark/sql/hive/HiveClientUtil.scala
```diff
@@ -148,12 +148,21 @@ class HiveClientUtil(val sparkContext: SparkContext) extends Logging {
...
+        dbURL +";user=" + user.get + ";password=" + password.get + ";default-schema=" +
+            Misc.SNAPPY_HIVE_METASTORE + ";"
+      } else {
+        metadataConf.setVar(HiveConf.ConfVars.METASTORE_CONNECTION_USER_NAME,
+          Misc.SNAPPY_HIVE_METASTORE)
+        dbURL
+      }
+      logInfo(s"Using SnappyStore as metastore database, dbURL = $secureDbURL")
+      metadataConf.setVar(HiveConf.ConfVars.METASTORECONNECTURLKEY, secureDbURL)
+      metadataConf.setVar(HiveConf.ConfVars.METASTORE_CONNECTION_DRIVER, dbDriver)
```
sumwale: Duplication of most of the code in SnappyStoreHiveCatalog. Please move out into a separate method (e.g. static in SnappyStoreHiveCatalog object).

core/src/main/scala/org/apache/spark/sql/row/GemFireXDDialect.scala
```diff
@@ -171,7 +172,7 @@ abstract class GemFireXDBaseDialect extends JdbcExtendedDialect {
         "PERSISTENT_PARTITION".equalsIgnoreCase(result)) {
 
       JdbcExtendedUtils.executeUpdate(
-        s"call sys.CREATE_ALL_BUCKETS('$tableName')", conn)
+        s"call sys.CREATE_ALL_BUCKETS('$tableName')", sysConn())
```
sumwale: Why is conn not correct here?

core/src/main/scala/org/apache/spark/sql/row/JDBCMutableRelation.scala
```diff
@@ -90,9 +91,23 @@ case class JDBCMutableRelation(
...
+    val user = sqlContext.conf.getConfString("snappydata.store.user", EMPTY_URL)
+    val password = sqlContext.conf.getConfString("snappydata.store.password", EMPTY_URL)
+    if (!user.equals(EMPTY_URL) && !password.equals(EMPTY_URL)) {
+      JdbcUtils.createConnectionFactory(new JDBCOptions(connProperties.url +
+          ";user=" + user +
+          ";password=" + password +
+          ";default-schema=" + SnappyStoreHiveCatalog.HIVE_METASTORE + ";",
+        table, connProperties.connProps.asScala.toMap))
+    } else connFactory
+  }
```
sumwale: Don't get this sysConnFactory changes in multiple places. Why is the "conn" itself not the correct authenticated one? In either case this code should be refactored into a separate utility method as mentioned.

cluster/src/dunit/scala/io/snappydata/cluster/SecureQueryRoutingDUnitTest.scala
vivekwiz: Could not do beforeClass changes due to NPE.

cluster/src/dunit/scala/io/snappydata/cluster/SecureQueryRoutingDUnitTest.scala
vivekwiz: Will write more test covering this scenario.

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
vivekwiz: Done

core/src/main/scala/org/apache/spark/sql/hive/HiveClientUtil.scala
vivekwiz: They take properties from different places so a common method is not implemented.

vivekwiz:
@sumwale please review the changes again. Also see the PR in store https://github.com/SnappyDataInc/snappy-store/pull/240

vivekwiz:
@hbhanawat can you please review new target added for secured Dunit  test

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -251,10 +257,43 @@ class JDBCSourceAsColumnarStore(override val connProperties: ConnectionPropertie
...
+
+        // Hikari only take 'username'. So does Tomcat
+        def securePoolProps(props: Map[String, String]): Map[String, String] = {
+          if (props.get(Attribute.USERNAME_ALT_ATTR.toLowerCase).isEmpty &&
+              props.get(Attribute.PASSWORD_ATTR).isEmpty) {
+            props + (Attribute.USERNAME_ALT_ATTR.toLowerCase ->
+                p.get(Attribute.USERNAME_ATTR).toString) +
+                (Attribute.PASSWORD_ATTR -> p.get(Attribute.PASSWORD_ATTR).toString)
+          } else props
+        }
```
sumwale: @vivekwiz 
Why is setting system credentials needed for embedded mode? This won't work for smart connector and if it is not required for smart connector, then any reason why it is needed for embedded mode? It will be a security hole for queries routed from JDBC/ODBC if we are not very careful in code not to do it for routed queries (like for column tables which will not be checked in the GemXD parser before routing).

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -275,10 +276,34 @@ object ExternalStoreUtils extends Logging {
...
-    ConnectionProperties(url, driver, dialect, allPoolProps,
+    getConnectionProperties(session, url, driver, dialect, allPoolProps,
       connProps, executorConnProps, hikariCP)
   }
 
+  def getConnectionProperties(session: Option[SparkSession], url: String, driver: String,
+      dialect: JdbcDialect, poolProps: Map[String, String], connProps: Properties,
+      executorConnProps: Properties, hikariCP: Boolean): ConnectionProperties = {
+    if (session.isDefined && session.get.conf.getOption(Attribute.USERNAME_ATTR).isDefined &&
+        session.get.conf.getOption(Attribute.PASSWORD_ATTR).isDefined) {
```
sumwale: @vivekwiz 
Can save the result of conf.getOption and avoid lookup again. Doesn't look like much but still for low latency queries these things can add up. A cleaner scala-like way without compromising perf could be:

```scala
sessionOpt.foreach { session =>
  session.conf.getOption(Attribute.USERNAME_ATTR) match {
    case Some(user) => session.conf.getOption(Attribute.PASSWORD_ATTR) match {
      case Some(password) =>
        ...
      case None =>
    }
    case None =>
  }
}
```

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
vivekwiz: Hi @sumwale , this was needed for allowing inserts on column table. Inserts convert in put/putAll and and they do not carry credentials to remote node. I evaluated the efforts after discussions with @kneeraj and this looked safe (since connection is internal) and quick. Alternative was to carry credentials in messages like Put/PutAll and Batchupdate than can be done after 1.0 if you feel its needed. Kindly suggest.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
vivekwiz: @sumwale This code has been changed since code review. On SNAP-1656, corresponding code is:

`  def getConnectionProperties(session: Option[SparkSession], url: String, driver: String,
      dialect: JdbcDialect, poolProps: Map[String, String], connProps: Properties,
      executorConnProps: Properties, hikariCP: Boolean): ConnectionProperties = {
    if (session.isDefined) {
      val (user, password) = getCredentials(session)

      if (!user.isEmpty && !password.isEmpty) {
        def secureProps(props: Properties): Properties = {
          props.setProperty(Attribute.USERNAME_ATTR, user)
          props.setProperty(Attribute.PASSWORD_ATTR, password)
          props
        }

        // Hikari only take 'username'. So does Tomcat
        def securePoolProps(props: Map[String, String]): Map[String, String] = {
          props + (Attribute.USERNAME_ALT_ATTR.toLowerCase -> user) + (Attribute.PASSWORD_ATTR ->
              password)
        }

        ConnectionProperties(url, driver, dialect, securePoolProps(poolProps),
          secureProps(connProps), secureProps(executorConnProps), hikariCP)
      } else {
        ConnectionProperties(url, driver, dialect, poolProps, connProps, executorConnProps,
          hikariCP)
      }
    } else ConnectionProperties(url, driver, dialect, poolProps, connProps, executorConnProps,
      hikariCP)
  }

  def getCredentials(session: Option[SparkSession], prefix: String = ""): (String, String) = {
    val prefix = SnappyContext.getClusterMode(session.get.sparkContext) match {
      case ThinClientConnectorMode(_, _) => Constant.SPARK_STORE_PREFIX
      case _ => ""
    }
    (session.get.conf.get(prefix + Attribute.USERNAME_ATTR, ""),
        session.get.conf.get(prefix + Attribute.PASSWORD_ATTR, ""))
  }`

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
sumwale: @vivekwiz GemFire API like put/putAll do not do any GemXD authorization and that will work properly without any credentials. This is not safe and looks like a big security hole for routed queries as mentioned.

PR #692, opened 2017-07-07
[WIP] Support for setting scheduler pools using the set command

hbhanawat:

Added two more scheduler pools in addition to the default pool - transactional and analytical. transactional pool has FAIR scheduling while the analytical pool has FIFO scheduling. The default pool will be used if no pool is set. For now, all the three pools have equal weight i.e. same number of cores will be allocated to different pools. If there are no queries of a pool, the cores of that pool are consumed by the jobs of the other pools. A pool can be set using the set command: 

set spark.scheduler.pool=analytical;

or using 

sparkContext.setLocalProperty("spark.scheduler.pool", "analytical")

We may not document the second way because if the pool does not exist, the API may create a new pool. 

For the set command, we are explicitly throwing an exception if the pool does not exist. 

## Patch testing

Manual. Precheckin. 

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/57


jramnara:
I am not sure we should be calling these pools - Transactional, Analytical.  The term "transactional" is likely associated with transactional writes and we aren't referring to that.  Merely, a pool that allows low latency queries to continue working and making sure computational intensive ones don't hog the system. So, why not call these - "low latency" and "high throughput" ? maybe other suggestions.

jramnara:
I also think that both the built-in pools should be FAIR scheduled. We are touting a system for interactive analytics and multiple users. If no one else is competing for resources my "stage" can use all the resources. But, if someone wants to sneak in, there are likely to get the resources quickly enough.

hbhanawat:
As per our discussion Jags, I will create two pools one would be "default" and another would be "lowlatency". Both the pools would be FAIR scheduled. BTW, due to Spark 2.1 merge, I am finding it tough to merge this branch, so I have created a new pull request and closing this one. Please comment on the following PR: 

https://github.com/SnappyDataInc/snappydata/pull/700

PR #696, opened 2017-07-11
Snap 1610

dshirish:
Pending task of removing all references to old Split Cluster mode from the code. This was disabled for users in the 0.9 release (error was thrown),  but code references were still to be removed.

## Patch testing
ran precheckin excluding store (will run complete precheckin including store prior to precheckin)

## ReleaseNotes.txt changes
Not applicable

## Other PRs 
store side PR
https://github.com/SnappyDataInc/snappy-store/pull/232


core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1009,8 +1009,7 @@ object SnappyContext extends Logging {
       }.orElse(Property.McastPort.getOption(conf).collectFirst {
         case s if s.toInt > 0 =>
           val url = "mcast-port=" + s
-          if (embedded) ExternalEmbeddedMode(sc, url)
-          else SplitClusterMode(sc, url)
+          /* if (embedded) */ ExternalEmbeddedMode(sc, url)
       }).orElse(Property.SnappyConnection.getOption(conf).collectFirst {
```
sumwale: Shouldn't this throw an exception like being done for locators option above saying that snappydata.connection property should be set? Also there is a typo in that exception message "paramater"

dshirish: Done. Changed the code throw an exception.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1005,12 +1005,15 @@ object SnappyContext extends Logging {
...
-              s"Use paramater ${Property.SnappyConnection} for smart connector mode")
+              s"Use parameter ${Property.SnappyConnection} for smart connector mode")
       }.orElse(Property.McastPort.getOption(conf).collectFirst {
         case s if s.toInt > 0 =>
           val url = "mcast-port=" + s
           if (embedded) ExternalEmbeddedMode(sc, url)
-          else SplitClusterMode(sc, url)
+//          else SplitClusterMode(sc, url)
+          else throw new SparkException(
+            s"Invalid configuration. " +
```
sumwale: will be better to have parameter name in message: "Invalid configuration parameter mcast-port"

PR #697, opened 2017-07-11
Snap 1792 : Display snappy members logs on Snappy Pulse UI 

snappy-sachin:

- Display details /stats like CPU usage, Heap & Off-Heap memory usages details and members general details like type, process id, etc.
- Display members log content in incremental way. 

## Patch testing

- Manually Tested.

## ReleaseNotes.txt changes

Yes. Release Notes updated.

## Other PRs 

Store: https://github.com/SnappyDataInc/snappy-store/pull/233
Spark: https://github.com/SnappyDataInc/spark/pull/58


cluster/src/main/scala/org/apache/spark/ui/SnappyMemberDetailsPage.scala
```diff
@@ -0,0 +1,463 @@
...
+    // Verify that the normalized path of the log directory is in the working directory
+    val normalizedUri = new File(logDirectory).toURI.normalize()
+    val normalizedLogDir = new File(normalizedUri.getPath)
+    if (!Utils.isInDirectory(workDir, normalizedLogDir)) {
+      return ("Error: invalid log directory " + logDirectory, 0, 0, 0)
+    }
+
+    try {
+      val files = RollingFileAppender.getSortedRolledOverFiles(logDirectory, logFile)
+      logDebug(s"Sorted log files of type $logFile in $logDirectory:\n${files.mkString("\n")}")
```
sumwale: Does this require the log files to be on a shared NFS or similar? We recommend local filesystems for working directory so that persistent files are local to the machine and not over NFS. Most of the times users will have the log-files in the same location and we should not expect them to be available from lead directly. I think this whole code should go into a function that is executed on the selected server. What do you think @rishitesh ?

snappy-sachin: Thanks Sumedh, as discussed I will make changes to load logs from local file system on member machine.

rishitesh:
One minor comment for code refactoring. Else looks ok.

snappy-sachin:
Yes Rishi. 
I am going do refactoring of UI code collectively next time.

PR #699, opened 2017-07-12
SNAP-1794

ymahajan:

Modified code generation of DynamicFoldableExpression such that even the initMutableState splits into multiple init() functions, code will be generated properly.

## Patch testing

clean precheckin

## ReleaseNotes.txt changes

None

## Other PRs 

spark - https://github.com/SnappyDataInc/spark/pull/59


core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
```diff
@@ -254,9 +254,9 @@ case class DynamicFoldableExpression(expr: Expression) extends Expression
     val newVarIsNull = ctx.freshName("paramLiteralExprIsNull")
     val comment = ctx.registerComment(expr.toString)
     ctx.addMutableState(ctx.javaType(expr.dataType), newVar,
```
sumwale: Can you add a comment here saying something like "initialization for both variable and isNull is being done together due to dependence of latter on the variable and the two get separated due to Spark's splitExpressions -- SNAP-1794"

PR #700, opened 2017-07-12
Support for setting scheduler pools using the set command

hbhanawat:

Added a new scheduler pool in addition to the default pool - lowlatency with FAIR scheduling. The default pool will be used if no pool is set. For now, both the pools have equal weight i.e. same number of cores will be allocated to these pools. If there are no queries of a pool, the cores of that pool are consumed by the jobs of the other pools. A pool can be set using the set command: 

set spark.scheduler.pool=lowlatency;

or using 

sparkContext.setLocalProperty("spark.scheduler.pool", "lowlatency")

We may not document the second way because if the pool does not exist, the API may create a new pool. 

For the set command, we are explicitly throwing an exception if the pool does not exist. 

Right now, we are not providing any property to create new pools during run time. One can do that by updating the conf/fairscheduler.xml before starting the lead node which is the spark way. 

Secondly, This PR has increased plan cache size from 300 to 3000 and has made it configurable. Since SnappySession is part of CachedKey, cached query plans are not getting shared across sessions. The concurrency is badly hit because size of cache is only 300 and multiple concurrent sessions evict each other's plans. For now, increasing the cache size. See SNAP-1605 for more details. 

## Patch testing

Manual. Precheckin.


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1647,6 +1645,24 @@ class SnappySession(@transient private val sc: SparkContext,
...
+      case "spark.scheduler.pool" =>
+        // set the scheduler pool if the pool exists else throw an exception. 
+        if (sparkContext.getAllPools.exists(_.name == value)) {
+          sparkContext.setLocalProperty(key, value)
+        } else {
+          throw new IllegalArgumentException(s"Invalid Pool $value")
+        }
+      case _ =>
+    }
+  }
```
sumwale: Two comments:

- This handling should be moved to SnappyConf as mentioned in the other PR.

- Secondly the Spark property is a inheratable thread-local but we need the property to be per SnappySession. In particular this will not work for routed queries because the setter thread may be different that the threads used in subsequent queries. So this means that we cannot be setting the SparkContext property here and be done. Instead what we should do is:
  - keep the check for pool existence but then set a variable in SnappyConf and public getter to read it, like: sessionPool which defaults to the "default" pool (or None can indicate the same)
  - in key set action set to the new given pool and in unset case revert to "default" (or None)
  - read the current pool from SnappyConf in SnappySession.executeSQL, evaluatePlan and CachedDataFrame.collectWithHandler and set/clear the pool local property there on SparkContext. Alternatively it will be better to set it in just CodegenSparkFallback.executeWithFallback in before "f(plan)" calls assuming that threads won't be switched between plan execution and collect. This will allow it to work for sql as well as API.

hbhanawat: I was not aware about SnappyConf.keyUpdateActions, I will make the required changes. But, I already found the second issue in my testing. Working on it. BTW, this PR is WIP. :-)

sumwale: Another advantage of SnappyConf is that user's can also use the same in smart connector mode after copying the pool xml to the split cluster.

sumwale: > BTW, this PR is WIP. :-)

The review is also WIP :-)

sumwale: Actually just doing in CodegenSparkFallback will not be enough since we cache the result of the result RDD in CachedDataFrame too. So a subsequent query on a different thread can pick up a cached RDD result and then fire collect on it. So it will need to be added to CodegenSparkFallback and also to CachedDataFrame.collectWithHandler or withNewExecutionId.

hbhanawat: Yeah. But given that it is spark app, it would be better to leave it the spark way. Setting scheduler pools for embedded mode and smart connector in the same way (which is different than spark) can confuse users. In fact, I wanted to change the property name from spark.scheduler.pool to a snappy specific name, say snappydata.execution.pool, for embedded mode. This would avoid any confusion related to spark app. I discussed this with @jramnara and he was ok with spark.scheduler.pool so I left it that way. What do you say?

sumwale: I think the generic ability to set the pool using SQL can be a useful one for split cluster too. When in future we add our hive-thriftserver that uses SnappySession for BI tools, then users can run Tableau or something in split cluster and then this can be used to switch pools.

hbhanawat: Hmm. I found it confusing from app user's perspective. But, since we are not taking away the older way to set pool, I rest my case.

sumwale: Didn't get the relevance of this. How will users invoke SparkContext API over hive-thriftserver? Does Spark already provide some way?

hbhanawat: Like they are doing today. Users are using hive-thriftserver even today and may be using the pools. They will continue using it that way. We are providing a new API, to set the spark scheduler pool of Spark App, which we are not sure they need. And, the risk is that they may get confused between the embedded and non-embedded pools. But, I am okay because the old way still remains valid.

sumwale: If there is a way to set the pool property in hive thriftserver then we should honour the same instead of adding a separate one. Looks like it was added here: https://github.com/apache/spark/pull/1937/files

Note that the hive-thriftserver can run both in embedded mode (was required for 1.0 but deferred for 1.1) or in split mode so two different SQL properties will confuse users. Since our JDBC/ODBC server is also a thrift server, so that name should be fine.

hbhanawat: The name may not be appropriate as thrift server is not used for job server and the doc also says that this property is only for JDBC clients. I am not sure about the usage of this property in Spark because new config help doesn't mention it and older helps say that it is for users migrating from shark. 

I would rather go with either snappydata.scheduler.pool or the existing property. What do you say?

sumwale: > the doc also says that this property is only for JDBC clients. I am not sure about the usage of this property in Spark

Didn't get this. Weren't you just arguing above that Spark users will use the standard SparkContext.setLocalProperty way? Going by that I thought this new property handling was mostly relevant for JDBC/ODBC connections.

> I am not sure about the usage of this property in Spark because new config help doesn't mention it and older helps say that it is for users migrating from shark.

Which property you are looking at? The "spark.sql.thriftserver.scheduler.pool" is clearly the only supported way in Spark to set pool over hive-thriftserver: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L401

> I would rather go with either snappydata.scheduler.pool

Then users will have two different properties in SQLConf: "snappydata.scheduler.pool" when using SnappyContext, and "spark.sql.thriftserver.scheduler.pool" when using hive-thriftserver in smart connector mode (or when we add it in embedded mode). Whats the rationale for not using the "spark.sql.thriftserver.scheduler.pool" name that will make it consistent with existing Spark's hive-thriftserver?

hbhanawat: >> Didn't get this. Weren't you just arguing above that Spark users will use the standard SparkContext.setLocalProperty way? Going by that I thought this new property handling was mostly relevant for JDBC/ODBC connections.

What about the jobs? For smart connector,  they will use what they were using. Or they can use this new property as well. 

>>Which property you are looking at? The "spark.sql.thriftserver.scheduler.pool" is clearly the only supported way in Spark to set pool over hive-thriftserver: https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L401

I am confused because I couldn't find any documentation for this property other than the following which is for Spark 1.2. Why is it that Spark does not document it anymore? I may have to try it to understand it. 

https://spark.apache.org/docs/1.2.0/sql-programming-guide.html#migration-guide-for-shark-user

>> Whats the rationale for not using the "spark.sql.thriftserver.scheduler.pool" name that will make it consistent with existing Spark's hive-thriftserver?

As explained above, two rationale:
1. The property is documented as property for JDBC and ODBC clients. 
2. I am not sure if the property is being used for hive servers anymore.

sumwale: > What about the jobs? For smart connector, they will use what they were using.

Weren't you arguing that users will continue to use SparkContext.setLocalProperty in jobs and smart connector??

> 1. The property is documented as property for JDBC and ODBC clients.

Exactly. And as per your own stance earlier, for other cases users will continue to use SnappyContext.setLocalProperty, no? Or did I misunderstand what you were saying earlier?

> 2. I am not sure if the property is being used for hive servers anymore.

There are two cases here:

a) Either the property "spark.sql.thriftserver.scheduler.pool" is available in Spark JDBC/ODBC and users are using that. In that case we must use the same property. As per the code it is clearly available and not marked deprecated in code but for some reason is not in recent docs.

b) Or that property is not available. In that case users cannot set the pool over JDBC/ODBC in Spark currently. In that case our addition can be quite useful for Spark JDBC/ODBC too and we can use either of the property names. But we must use the SnappyConf way so that then the same property can be used for both Spark Hive JDBC/ODBC (when we add SnappySession support in that) as well as SnappyData JDBC/ODBC.

hbhanawat: >>Weren't you arguing that users will continue to use SparkContext.setLocalProperty in jobs and smart connector??

No. I was always talking about the embedded and smart connector. Not about jdbc versus scala code. 

>> Exactly. And as per your own stance earlier, for other cases users will continue to use SnappyContext.setLocalProperty, no? Or did I misunderstand what you were saying earlier

Yes, I think you misunderstood. Check the PR description. SparkContext.setLocalProperty API creates a new pool if it does not already exist. We cannot allow that. 

>> But we must use the SnappyConf way so that then the same property can be used for both Spark Hive JDBC/ODBC (when we add SnappySession support in that) as well as SnappyData JDBC/ODBC.

You mean same property name should be added to SnappyConf? I can do that as far as SnappyConf allows it.

sumwale: > No. I was always talking about the embedded and smart connector. Not about jdbc versus scala code.

So you are saying that in smart connector mode when writing SQL code (e.g. via SparkSession.sql), users will continue using SparkContext API, but when writing the same code in jobs they will use the new property? Why would that be the case or why would we recommend something like that?

The only valid case looks to be that users will use the SparkContext way when writing code, while they use the new property for SQL.

> SparkContext.setLocalProperty API creates a new pool if it does not already exist. We cannot allow that.

That doesn't make sense. How can we prevent users from creating new pools using SparkContext API? We will, of course, have some pools in the product default configuration but users can always create more or change those.

hbhanawat: I am confused. What are you recommending?

sumwale: Don't quite understand what you were recommending either. I thought you were saying:

a) Users writing Spark code will continue to use SparkContext way (apparently because that is the way they are used to). This will apply for jobs, smart connector.

b) Other cases (namely Snappy JDBC/ODBC) will use the new property.

(this doesn't have the Spark JDBC/ODBC case). But you are saying something else. Whats that and reasoning for the same?

My own preference is:

a) When using SnappySession always recommend using the property. Whether using smart connector or jobs (or Spark JDBC/ODBC in future when it will be enhanced to use SnappySession). And automatically applies to Snappy JDBC/ODBC.

b) When not using SnappySession, they will continue using the existing Spark way in any case.

hbhanawat: >> When using SnappySession always recommend using the property. 

Which property? Thrift one? 

I will make the necessary changes.

sumwale: > Which property? Thrift one?

Not sure. If that property is no longer being documented for some reason and no one is using it, then the one added in this PR is certainly better.

sumwale: Ok looked around a bit and seems like no one is using that property. Should we then stick to "snappydata.scheduler.pool"? Can also use that for low-latency scheduler implementation if we add one in future.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -886,6 +886,9 @@ object SnappyContext extends Logging {
   }
 
   private def newSnappyContext(sc: SparkContext) = {
+    if (initConf == null){
```
sumwale: space before brace

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -133,6 +131,11 @@ class CachedDataFrame(df: Dataset[Row], var queryString: String,
    */
   private def withCallback[U](name: String)(action: DataFrame => U) = {
     try {
+      val snSession = SparkSession.getActiveSession.get.asInstanceOf[SnappySession]
+      val pool = snSession.asInstanceOf[SnappySession].
+        sessionState.conf.activeSchedulerPool
+
+      snSession.sparkContext.setLocalProperty("spark.scheduler.pool", pool)
```
sumwale: Can we set lowlatency by default for pruned tasks? That will be very useful so users will not have to keep setting it explicitly. I think it can be something like:

```scala
if ((cachedRDD ne null) && cachedRDD.getNumPartitions <= 2 /* some small number */ &&
    shuffleDependencies.length == 0) && pool == "default") {
  pool = "lowlatency" // perhaps also check that this pool exists and user has not renamed/removed it
}
```

We don't expect lowlatency pool to be used for queries that can launch subqueries, or run jobs in plan execute itself, so overriding default for just the collect should be enough.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1038,6 +1041,7 @@ object SnappyContext extends Logging {
     if (!_globalSNContextInitialized) {
       contextLock.synchronized {
         if (!_globalSNContextInitialized) {
+          initConf = sc.conf
```
sumwale: Also clear initConf in SnappyContext.clearSnappyContext

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1855,7 +1855,11 @@ object SnappySession extends Logging {
         evaluatePlan(df, session, key.sqlText, key)
       }
     }
-    CacheBuilder.newBuilder().maximumSize(300).build(loader)
+    val cacheSize = Property.PlanCacheSize.getOption(SnappyContext.initConf) match {
```
sumwale: Can't you use conf from SnappyContext.globalSparkContext and get rid if this separate initConf? I think if you have planCache as lazy, then it should work fine. If you want to test this in unit test, then will probably need a separate unit test suite since this will remain once set even after stopping SparkContext. Or you can have planCache as volatile which is cleared on SnappyContext.stopSnappyContext and re-created on next access. Alternatively you can just do manual testing for this property.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -76,6 +74,7 @@ class SnappySession(@transient private val sc: SparkContext,
 
   self =>
 
+  SnappyContext.initConf = sc.conf
```
sumwale: Is this required after the initialization in global block? Probably the issue is that clear is not being done.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1855,7 +1855,11 @@ object SnappySession extends Logging {
         evaluatePlan(df, session, key.sqlText, key)
       }
     }
```
sumwale: I think you should also make planCache as lazy for this to work as expected.

hbhanawat: lazy val has performance penalty and I was trying to avoid that especially for the concurrency scenarios.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
hbhanawat: >>Can't you use conf from SnappyContext.globalSparkContext and get rid if this separate initConf? 

We need initConf during SnappySession object's initialization. This initialization happens during the lead start and SparkContext.getOrCreate returns a null. 

>> I think if you have planCache as lazy, then it should work fine. 

The performance penalty of lazy is what I was trying to avoid. This conf business is clumsy.  

>> Alternatively you can just do manual testing for this property.

I just did that.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
hbhanawat: Required for the cases when SnappyStreamingContext is created without creating SnappySession or SnappyContext. See AQPStreamingSuite.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
hbhanawat: Will do that.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
hbhanawat: This is good idea. I will add this.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
hbhanawat: Oops. Looks like I was wrong all the way. the performance penalty is only for the initialization and not for reading. Embarrassing. Will add it.

sumwale: There could have been a small penalty of volatile read, I think, but since this will then do a ConcurrentMap lookup so then that will have volatile reads in any case which will flush cache lines.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
sumwale: SnappyStreamingContext has a "val snappySession" field. Lower down in this code there is a "SnappyContext.initGlobalSnappyContext" call in constructor so that should have taken care of initialization of conf no?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
hbhanawat: Yeah. Lead processing does not benefit much from the cache lines, I suppose. So, that is ok.

hbhanawat:
@sumwale I also added the following commit after a discussion with you 64a3db3

PR #703, opened 2017-07-13
Fix for SNAP-1805 caused because function to withNewExecutionId was e…

SachinJanani:
…xecuted before it was passed as argument

* Proper passing of function as argument to withNewExecutionId method
* As UI data is stored based on the execution id and because of this issue job was executed before creation of new execution id data was not displayed on UI

## Patch testing
* Manually tested
* precheckin in progress

## ReleaseNotes.txt changes
NA
## Other PRs 
NA


SachinJanani:
Done with precheckin no new failures

PR #704, opened 2017-07-13
[SNAP-1777] increasing default member-timeout for SnappyData

sumwale:

- increasing the default member-timeout to 30s
- refactored methods a bit to do all SnappyData specific default boot setup in one place
- removed explicit pool minEvictableIdleTimeMillis added previously and let it be default

## Patch testing

precheckin

## ReleaseNotes.txt changes

Mention in best practices guide that default for member-timeout is 30s and needs to be increased further only if trouble is still seen.

## Other PRs 

NA


kneeraj:
LGTM

PR #706, opened 2017-07-13
Fix for SNAP-1723. 

kneeraj:
Functions like round etc which takes constants as argument were also being tokenized leading to the error.

When functions are there in the queries and literals are there in their argument then stop tokenization.

## Patch testing

Added junit test to reproduce the issue.

## ReleaseNotes.txt changes

None.

## Other PRs 

No.


core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -117,6 +117,8 @@ object Constant {
   val CHANGEABLE_JAR_NAME = "SNAPPY_CHANGEABLE_JAR_NAME"
 
   val RESERVOIR_AS_REGION = "spark.sql.aqp.reservoirAsRegion"
+
+  val FOLDABLE_FUNCTIONS = Set( "ROUND", "BROUND" )
```
sumwale: Other functions that depend on foldable:

- PERCENTILE (Percentile.scala)
- STACK (generators.scala)
- NTILE (windowExpressions.scala)
- STR_TO_MAP -- only 2nd & 3rd args, NAMED_STRUCT -- only odd position args (complexTypeCreator.scala)
- REFLECT and JAVA_METHOD -- only 1st and 2nd args (CallMethodViaReflection)
- XPATH, XPATH_BOOLEAN, ... all XPATH prefixed ones and only 2nd arg
- PERCENTILE_APPROX, APPROX_PERCENTILE -- only 2nd and 3rd args (ApproximatePercentile)

LIMIT argument tokenization should be turned off explicitly in Parser because it can come in a sub-query without a WHERE clause before it that would have turned tokenization off.
Same for SnappyParser.windowFrame

Tokenization for LIKE/RLIKE/REGEXP will work but will be inefficient since generated code will Pattern.compile the string for every row (as it assumes that expression can change). Since these are handled separately in SnappyParser so tokenization for only the right "termExpression" needs to be turned off.

The following functions will be more efficient on disabling and we don't expect constants to change:

- TRANSLATE (StringTranslate in stringExpressions.scala) -- ideally only 2nd and 3rd arguments
- UNIX_TIMESTAMP, TO_UNIX_TIMESTAMP, FROM_UNIXTIME, TO_UTC_TIMESTAMP, TRUNC, FROM_UTC_TIMESTAMP, NEXT_DAY (dateTimeExpressions.scala) -- ideally only 2nd argument
- LIKE, RLIKE -- only 2nd argument; seems like they can appear as just functions too as per FunctionRegistry

Selective disabling of tokenization for only required arguments will be better. The set of functions will become a map that will also list the arguments that have to disable tokenization and that can be used by parser to selectively replace instead of doing for "allExprs".

Also check if functions implementing CodegenFallback cause trouble (e.g. json_tuple) due to caching since those will ship Expression object and it could send back older one from the cached plan. If so change for all Expressions implementing CodegenFallback (e.g. using an Analyzer rule in SnappySessionState.analyzer).

Lastly two more points to track separately:

a) Right now ENABLE/DISABLE tokenization is a binary that turns on/off in the WHERE clause. If there is a sub-query in WHERE, for example, and then a further WHERE in sub-query, then tokenization will end at the end of the inner WHERE which could have gone on a bit further. A better fix for this is to turn it on globally for entire query and selectively turn it off only for clauses where it can cause trouble like being done for functions (only very few will remain like CAST dataType, ResolveInlineTables and few optimizations like RemoveLiteralFromGroupExpressions, EliminateSorts after rest of changes mentioned here are done).

b) ReorderAssociativeOperator is a rule to to fold all constants and we can track an issue to add similar for ParamLiterals for better performance. Likewise ConstantFolding will create new folded constant and equivalent rule for ParamLiteral can generate an initMutableState() with evaluated value instead of doing for each row. Similar for FoldablePropagation. Perhaps JVM will already optimize some of these?

kneeraj:
Thanks for the exhaustive feedback @sumwale . Going through it and will modify accordingly and let you know.

kneeraj:
@sumwale incorporated the comments and added the functions. Added tests for few functions, not all. LIKE/RLIKE etc. and many cases of queries where whether due to subqueries or special operators getting differently treated, the number of ParamLiterals obtained while parsing is more than what is finally seen in the executed plan. For such cases we are disabling caching and so it is like a safety net for a lot of queries whether Tokenization can go wrong.

Please review.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -676,7 +683,22 @@ class SnappyParser(session: SnappySession)
...
+            var allExprs = e.asInstanceOf[Seq[Expression]]
+            var exprs = allExprs
+            Constant.FOLDABLE_FUNCTIONS.get(n1) match {
+              case Some(args) =>
+                exprs = allExprs.zipWithIndex.collect {
+                  case (pl: ParamLiteral, index) if args.contains(index) ||
+                      // all args          // all odd args
+                      (args.head == -10) || (args.head == -1 && (index & 0x1) == 1) ||
+                      // all even args
+                      (args.head == -2 && (index & 0x2) == 1) =>
```
sumwale: even check should be: (index & 0x1) == 0

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -121,6 +121,25 @@ object Constant {
...
+    "REFLECT" -> Seq(1, 2), "JAVA_METHOD" -> Seq(1, 2), "XPATH" -> Seq(2),
+    "XPATH_BOOLEAN" -> Seq(2), "XPATH_DOUBLE" -> Seq(2),
+    "XPATH_NUMBER" -> Seq(2), "XPATH_FLOAT" -> Seq(2),
+    "XPATH_INT" -> Seq(2), "XPATH_LONG" -> Seq(2),
+    "XPATH_SHORT" -> Seq(2), "XPATH_STRING" -> Seq(2),
+    "PERCENTILE_APPROX" -> Seq(0, 1, 2), "APPROX_PERCENTILE" -> Seq(0, 1, 2),
+    "TRANSLATE" -> Seq(1, 2), "UNIX_TIMESTAMP" -> Seq(0),
+    "TO_UNIX_TIMESTAMP" -> Seq(0), "FROM_UNIX_TIMESTAMP" -> Seq(0),
+    "TO_UTC_TIMESTAMP" -> Seq(0), "FROM_UTC_TIMESTAMP" -> Seq(0),
+    "TRUNC" -> Seq(0, 1), "NEXT_DAY" -> Seq(0, 1))
```
sumwale: Missed args for couple and some args look incorrect (0 based vs 1 based I think):

- PERCENTILE: only second arg i.e. Seq(1)
- STACK: only first arg i.e. Seq(0)
- XPATH ones are second arg i.e. Seq(1)
- REFLECT, JAVA_METHOD: Seq(0, 1)
- PERCENTILE_APPROX, APPROX_PERCENTILE: Seq(1, 2)
- all *TIMESTAMP: second arg Seq(1)
- TRUNC: second arg Seq(1)
- NEXT_DAY: second arg Seq(1)

Both LIKE and RLIKE can also appear as functions in Spark so better to add those too:
- LIKE, RLIKE: Seq(1)

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -703,15 +725,29 @@ class SnappyParser(session: SnappySession)
...
+        val allExprs = e.asInstanceOf[Seq[Expression]].toList
+        var exprs = allExprs
+        Constant.FOLDABLE_FUNCTIONS.get(fn.funcName) match {
+          case Some(args) =>
+            exprs = allExprs.zipWithIndex.collect {
+              case (pl: ParamLiteral, index) if args.contains(index) ||
+                  // all args          // all odd args
+                  (args.head == -10) || (args.head == -1 && (index & 0x1) == 1) ||
+                  // all even args
+                  (args.head == -2 && (index & 0x2) == 1) =>
```
sumwale: correct even check as above

core/src/main/scala/io/snappydata/Literals.scala
kneeraj: @sumwale I considered 0 based because thats how it will be in the code. So whereever it is 0 it means the first argument of the function and so on. Therefore I used 0 for all timestamp functions and not 1 and tested the unix_timestamp function for replacement of ParamLiteral with Literal.

sumwale: @kneeraj Yes, considering that I said 2nd argument because its the format which is a constant and not the value itself e.g. `SELECT UNIX_TIMESTAMP('2016-04-08', 'yyyy-MM-dd')`
See the class comment of UnixTimestamp. Looks like Spark allows for same function to have different number of arguments. This function can take 0 arguments, 1 argument or 2 arguments, but need to disable tokenization only for the 2nd argument when specified.

kneeraj: Correct. Got it. Thanks.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
kneeraj: :-) ... my bad, silly. Corrected.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
kneeraj: done.

core/src/main/scala/io/snappydata/Literals.scala
kneeraj: Done. Modified test for UNIX_TIMESTAMP accordingly.

sumwale:
@kneeraj After looking around in the code, LIKE does cause mismatch of ParamLiterals vs reference because of LikeSimplification optimization that ends up converting the ParamLiteral to a Literal. However, that is not the case for RLIKE/REGEXP so added explicit conversion of pattern to Literal for all three in https://github.com/SnappyDataInc/snappydata/commit/97368f0510aa882c0e877fe2b0b78c1a15f3c7b9 . This also optimizes things a bit with tokenization/plan caching now happening for LIKE too.

kneeraj:
@swale This is a good change. Before this the plans consisting of LIKE/RLIKE/ etc were never cached, as in the plans were invalidated because of the catchAll type of check where we checked whether the number of ParamLiterls during the parse phase is equal to the number of PAramLiteral in the final execution plan. Thjat was a bigger hammer. With this, as long as the queries have same regex we will have plan caching. Thanks.

sumwale:
@kneeraj Yes pushed it directly to master and had asked Sonal to run on that revision.

PR #708, opened 2017-07-14
SNAP-953 - Add RPM/DEB installer packaging targets using the Netflix Nebula ospackage gradle plugin

vjr:

- Add nebula plugins `ospackage` and `ospackage-base`
- Add gradle targets `distRpm`, `distDeb`, `distInstallers` and include them in the `buildAll` and `distProduct` parent targets

## Patch testing

- Ran the new targets individually to observe creation of the RPM and DEB packages
- Ran the parent targets to observe the packages are created along with the earlier Zip/Tar targets
- Installed the DEB package with `sudo apt install ./snappydata_0.9_all.deb` to observe it is unpacking the product into `/opt/snappydata` with apparently the correct user/group permissions for all files and subfolders
- Currently unable to test the RPM package due to not having a redhat/fedora setup for the moment
- Note that when running the start-all script, logs are created in `/opt/snappydata/work/` subfolder so when uninstalling via `sudo apt remove --purge snappydata` it warns `dpkg: warning: while removing snappydata, directory '/opt/snappydata' not empty so not removed`
- Note that currently only dependency on `libc6` or `glibc` is specified because unsure from where a user might install their Java 8 packages and what the package name might be.

## ReleaseNotes.txt changes

Does this change require an entry in ReleaseNotes.txt? YES
If yes, has it been added to it? NO

## Other PRs 

N/A


vjr:
Edited the original PR comment/note to mention about being unsure what java package dependency name to use because a user could (possibly?) install Java 8 from different sources with different package names.

build.gradle
```diff
@@ -772,6 +775,37 @@ distributions {
...
+  }
+}
+
+buildRpm {
+  requires('glibc')
+}
+
+buildDeb {
+  requires('libc6')
+}
```
sumwale: Ideally we can add dependency on "java-1.8.0" for RPM and "java8-sdk" for DEB. These are virtual names and all standard java RPM/DEB (checked both openjdk and Oracle distribution/ubuntu PPAs) use these. However it is quite possible that some users just get the Oracle tar.gz and have it installed in some common place. So for DEB we can add a recommends('java8-sdk') while for RPM there is no equivalent (recent Fedora have something called "weak dependencies" but that is not available in RHEL/Centos AFAIK).

build.gradle
```diff
@@ -772,6 +775,37 @@ distributions {
     }
   }
 }
+
+ospackage {
+  packageName = distributions.main.baseName
```
sumwale: Add "-without-hadoop" suffix for "hadoop-provided" setting like being done in distTar/distZip

build.gradle
```diff
@@ -792,6 +826,16 @@ distZip {
...
+task distRpm {
+  dependsOn product
+  dependsOn buildRpm
+}
+
+task distDeb {
+  dependsOn product
+  dependsOn buildDeb
+}
+
```
sumwale: add ":packageVSD" dependency like being done for distTar/distZip so that VSD gets packaged even if these tasks are run independently

sumwale:
As suggested, if this much works well then we can add a separate "snappydata" user having the full name "SnappyData Adminstrator" that will own the directory. The preInstall script will contain the code to do "adduser" which will be as a "system" user with an invalid password, so only root can sudo to it by default. The script will be slightly different for RPM vs DEB though quite trivial so you may need to lookup the docs/examples for install scripts in RPM/DEB.

vjr:
@sumwale i have updated the PR as per your comments.

note the "without-hadoop" option RPM side does not support hyphens so i made that underscore, thus the resulting filename is like `snappydata-0.9-without_hadoop_bin.noarch.rpm` when running gradle with the `-Phadoop-provided=false` option.

i would suggest not creating any user/group like `snappydata` and leave it as it is currently (user/group set to root/root) similar to what the skype and google chrome installers are doing, otherwise if every thirdparty package creates their own user/group then it will pepper a user's system with new users/groups :-)

i feel this PR is now OK to merge but i will leave it open until i can test the RPM part later this week.

build.gradle
vjr: this seems to be handled automagically by adding the classifier/without-hadoop option checks in `distRpm` and `distDeb` targets :-)

build.gradle
vjr: done, but i had to add it to the base `buildRpm` and `buildDeb` targets instead, else build was failing with error like `classifier` method not found.

vjr:
Thanks @sumwale for the explanation about the need for separate user - I will work on that part next. In the mean time, my latest commit has removed the `bin` suffix from the generated installer package filenames and testing with clean builds look good to me.

build.gradle
```diff
@@ -862,6 +922,7 @@ task buildAll {
   dependsOn getTasksByName('product', true).collect { it.path }
   dependsOn getTasksByName('testClasses', true).collect { it.path }
   dependsOn distTar, distZip
+  dependsOn distInstallers
```
sumwale: I think distInstallers dep can be removed from buildAll because that is commonly used by devs to build all the test jars (hydra etc) and this will increase the build time (and requires rpm/deb packages to be installed?). Even the distTar, distZip tasks can be removed and replaced by dependency on only "product".

sumwale: Addendum: product dependency is not explicitly required since it is already done before "getTaskByName('product', true)..."

vjr:
I removed the packaging targets but the tar and zip are still getting created which I can't figure out why. The rpm and deb are being excluded fine from the `buildAll` target though. Saves only a few seconds from the approx. 12 minute clean buildall duration but I guess still worth it.

vjr:
Apologies for the lengthy delay, but feel free to let me know if there is anything I can do (as a regular github public user) for this ticket.

It looks like @ashetkar has merged ENT-1 as per https://github.com/SnappyDataInc/snappydata/pull/720 so not sure if I can do anything for the moment.

If I recall correctly (and based on reading this ticket info) the main pending item is to add pre/post (un)install scripts for the 'snappydata' user/group to be created/removed during rpm/deb (un)install procedure and then test on Ubuntu and RHEL/CentOS based OS.

build.gradle
```diff
@@ -829,6 +832,46 @@ distributions {
...
+buildRpm {
+  requires('glibc')
+  dependsOn ':packageVSD'
+  if (rootProject.hasProperty('hadoop-provided')) {
+    classifier 'without_hadoop'
+  }
+}
+
+buildDeb {
+  requires('libc6')
```
sumwale: Also add curl as requirement for both rpm and deb (required by snappy-job.sh and AWS checks).

vjr: Thanks @sumwale for your response(s) - I presume you will work on this internally to be able to better test this PR.

sumwale:
@vjr Made some further changes and merged for inclusion in 1.0 itself.

vjr:
yes @sumwale i see it, good stuff. i would be curious (maybe try it myself) how the uninstall procedure for the RPM/DEB goes.

PR #711, opened 2017-07-17
SNAP-1326 ALTER TABLE Support

ymahajan:

(Fill in the changes here)

- SnappyParser changes to support ALTER TABLE ADD/DROP COLUMN DDLs
- Added alterTable APIs in SnappyContext
- Added trait AlterableRelation
- Smart connector - alter table execution handling
- Added DDL routing dunits for Alter table
- scalaTests for alter table column and row tables

## Patch testing

clean precheckin, manual testing

## ReleaseNotes.txt changes

Yes

### DIFF
--- a/cluster/src/dunit/scala/io/snappydata/cluster/DDLRoutingDUnitTest.scala
+++ b/cluster/src/dunit/scala/io/snappydata/cluster/DDLRoutingDUnitTest.scala
@@ -20,7 +20,7 @@ import java.sql.{Connection, DriverManager, SQLException}
 
 import com.pivotal.gemfirexd.internal.engine.Misc
 import io.snappydata.test.dunit.{AvailablePortHelper, SerializableRunnable}
-
+import org.apache.spark.sql.SnappyContext
 import org.apache.spark.sql.collection.Utils
 
 class DDLRoutingDUnitTest(val s: String) extends ClusterManagerTestBase(s) {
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -699,6 +708,41 @@ private[sql] case class TruncateTable(ifExists: Any,
...
+}
+
+private[sql] case class AlterTableDropColumn(
+  tableIdent: TableIdentifier, column: String) extends ExecuteCommand {
+
+  override def run(session: SparkSession): Seq[Row] = {
+    val snc = session.asInstanceOf[SnappySession]
+    val catalog = snc.sessionState.catalog
+    val plan = try {
+      snc.sessionCatalog.lookupRelation(tableIdent)
```
kneeraj: Why drop column is doing a lookup of relation but add column is not? Any reason for this difference? Actually you are doing this in the actual alterTable in the snappysession class below for both cases. You can just remove from here. Isn't it?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1270,6 +1267,50 @@ class SnappySession(@transient private val sc: SparkContext,
...
+  private[sql]  def alterTable(tableIdent: QualifiedTableName, isAddColumn: Boolean,
+                              column: StructField): Unit = {
+    val plan = try {
+      sessionCatalog.lookupRelation(tableIdent)
+    } catch {
+      case tnfe: TableNotFoundException =>
+        throw tnfe
+    }
+    plan match {
+      case LogicalRelation(c: ColumnFormatRelation, _, _) =>
```
kneeraj: User should not even tru for ExternalTables. Should throw analysisexception in that case also. Also temp tables handled below should throw exception.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1270,6 +1267,50 @@ class SnappySession(@transient private val sc: SparkContext,
...
+          sessionCatalog.invalidateTable(tableIdent)
+          sessionCatalog.asInstanceOf[ConnectorCatalog].connectorHelper
+            .alterTable(tableIdent, isAddColumn, column)
+          return
+        }
+      case _ =>
+    }
+
+    plan match {
+      case LogicalRelation(ar: AlterableRelation, _, _) =>
```
kneeraj: In fcat if AlterableRelation is there and which I guess only relations of rowtables must be implementing then we should just check for that and then call the same on session which depending on the mode should either fire the procedure or call the alter table on the relation.

core/src/main/scala/org/apache/spark/sql/row/JDBCMutableRelation.scala
```diff
@@ -367,6 +368,38 @@ case class JDBCMutableRelation(
       ifExists: Boolean): Unit = {
     throw new UnsupportedOperationException()
   }
+
+  override def alterTable(tableIdent: QualifiedTableName,
```
kneeraj: Might have conflict with update/delete work on column table. Please check Sumedh's branch for his update/delete work so both of you don't have to resolve conflicts. Just a suggestion.

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
ymahajan: drop column just specifies the name of the column and not it's type, hence the lookup in this case. Also the alterTablePIs are consistent they expect a complete StructField is passed in both add and drop cases.

cluster/src/dunit/scala/io/snappydata/cluster/DDLRoutingDUnitTest.scala
```diff
@@ -158,6 +158,61 @@ class DDLRoutingDUnitTest(val s: String) extends ClusterManagerTestBase(s) {
...
+    s.execute(s"ALTER TABLE $tableName SET GATEWAYSENDER ()")
+    s.execute(s"ALTER TABLE $tableName add constraint emp_uk unique (Col1)")
+
+    verifyResultAndSchema(snc, tableName, 3)
+
+    dropTableXD(conn, tableName)
+  }
+
+  def verifyResultAndSchema(snc: SnappyContext, tableName: String, expectedColumns: Int): Unit = {
+    val dataDF = snc.sql("Select * from " + tableName)
```
ymahajan: we do verify the schema as well to see if the added column is present and dropped column is not present. but i will add a query with individual columns in projection.

cluster/src/dunit/scala/io/snappydata/cluster/DDLRoutingDUnitTest.scala
```diff
@@ -158,6 +158,61 @@ class DDLRoutingDUnitTest(val s: String) extends ClusterManagerTestBase(s) {
...
+    verifyResultAndSchema(snc, tableName, 3)
+
+    s.execute(s"ALTER TABLE $tableName ADD COLUMN Col4 INT")
+    verifyResultAndSchema(snc, tableName, 4)
+
+    s.execute(s"ALTER TABLE $tableName DROP COLUMN Col3")
+    verifyResultAndSchema(snc, tableName, 3)
+
+    // execute at store level
+    s.execute(s"ALTER TABLE $tableName SET GATEWAYSENDER ()")
```
ymahajan: ok, will check with AEL to be on the safe side.

cluster/src/dunit/scala/io/snappydata/cluster/DDLRoutingDUnitTest.scala
ymahajan: Added

core/src/main/scala/org/apache/spark/sql/row/JDBCMutableRelation.scala
ymahajan: ok, will check with him

PR #714, opened 2017-07-17
Fix for SNAP-1815

dshirish:
Throw error if user tries to configure Hive metastore using "metastore.dbUrl". 

Have kept the case of dbURL=null as it is HiveClientUtil#newClient  that seems to be used in ExternalClusterMode

## Patch testing
Will run precheckin prior to merge

## ReleaseNotes.txt changes
Not applicable
## Other PRs 
Not applicable


core/src/main/scala/org/apache/spark/sql/hive/HiveClientUtil.scala
```diff
@@ -272,8 +260,10 @@ class HiveClientUtil(val sparkContext: SparkContext) extends Logging {
     val sc = sparkContext
     Property.MetaStoreDBURL.getOption(sparkConf) match {
       case Some(url) =>
-        val driver = Property.MetaStoreDriver.getOption(sparkConf).orNull
-        (false, url, driver)
+//        val driver = Property.MetaStoreDriver.getOption(sparkConf).orNull
+//        (false, url, driver)
+        throw new IllegalArgumentException(s"Invalid property ${Property.MetaStoreDBURL}. " +
+            s"SnappyData does not allow use of property ${Property.MetaStoreDBURL}")
```
hbhanawat: Can we remove these properties from Literals.scala. In fact there is no need to read these properties and resolveMetaStoreDBProps can return (String, String) instead of (Boolean, String, String)

core/src/main/scala/org/apache/spark/sql/hive/HiveClientUtil.scala
```diff
@@ -154,18 +154,6 @@ class HiveClientUtil(val sparkContext: SparkContext) extends Logging {
...
-        metadataConf.setVar(HiveConf.ConfVars.METASTORE_CONNECTION_DRIVER,
-          dbDriver)
-      } else {
-        metadataConf.unset(
-          HiveConf.ConfVars.METASTORE_CONNECTION_DRIVER.varname)
-      }
-      metadataConf.unset(
-        HiveConf.ConfVars.METASTORE_CONNECTION_USER_NAME.varname)
     } else {
       logInfo("Using Hive metastore database, dbURL = " +
```
hbhanawat: This else should throw an exception.

dshirish: Did not remove this else since it was specifically for ExternalClusterMode. But looks like ExternalClusterMode is not much of use as discussed yesterday so can remove the else condition and probably file another issue to remove cluster modes that are not used any more (ExternalClusterMode and ExternalEmbeddedMode?)
@sumwale could you please confirm the above?

sumwale: Yes can remove those modes. As I had mentioned elsewhere the "embedded" property should be removed so then ExternalEmbeddedMode should be gone. The ExternalClusterMode is also unusable.

core/src/main/scala/org/apache/spark/sql/hive/HiveClientUtil.scala
```diff
@@ -272,8 +260,10 @@ class HiveClientUtil(val sparkContext: SparkContext) extends Logging {
     val sc = sparkContext
     Property.MetaStoreDBURL.getOption(sparkConf) match {
       case Some(url) =>
-        val driver = Property.MetaStoreDriver.getOption(sparkConf).orNull
-        (false, url, driver)
+//        val driver = Property.MetaStoreDriver.getOption(sparkConf).orNull
+//        (false, url, driver)
```
dshirish: Done

core/src/main/scala/org/apache/spark/sql/hive/HiveClientUtil.scala
dshirish: Done. Removed properties MetaStoreDBURL and MetaStoreDriver from Literals.scala. Also changed the method as per the comment. Filing a separate bug to remove obsolete cluster modes (External and ExternalEmbedded)

PR #716, opened 2017-07-18
[SNAP-671] Changes to support DSID from spark

SachinJanani:
* Added DSID as a function 
* DSID will work only in embedded mode

## Patch testing
* Added unit test 
* Tested manually
* Precheckin in progress

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


cluster/src/test/scala/org/apache/spark/sql/store/SnappyUDFTest.scala
```diff
@@ -461,4 +461,17 @@ class SnappyUDFTest extends SnappyFunSuite with BeforeAndAfterAll {
...
+
+   snc.sql("create table test123( a integer,b integer, c int) using column options()")
+    snc.sql("insert into test123 values(1,2,3)")
+    snc.sql("insert into test123 values(31,42,53)")
+    snc.sql("insert into test123 values(87,76,63)")
+    snc.sql("insert into test123 values(12,24,53)")
+
+    snc.sql("select DSID() from test123").show(5,truncate = false)
+    snc.sql("drop table test123")
+  }
```
sumwale: Check that the returned value is as expected i.e. matches String.valueOf(Misc.getMyId()

Also add a test using JDBC client and query routing (e.g. add the check to some test in QueryRoutingDUnitTest).

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1273,3 +1282,32 @@ case class ExternalClusterMode(override val sc: SparkContext,
...
 class TableNotFoundException(message: String, cause: Option[Throwable] = None)
     extends AnalysisException(message) with Serializable
+
+/**
+ * Expression that returns the dsid of the server containing the row.
+ */
+@ExpressionDescription(
+  usage = "_FUNC_() - Returns the dsid of the server containing the row.")
+case class DSID() extends LeafExpression {
+
```
sumwale: Move this out into a separate file say "io/snappydata/functions.scala" so that we can add all our functions in one place.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -129,6 +129,7 @@ class SnappySession(@transient private val sc: SparkContext,
   private[spark] val snappyContextFunctions = sessionState.contextFunctions
 
   SnappyContext.initGlobalSnappyContext(sparkContext, this)
+  sessionState.functionRegistry.registerFunction("DSID", e => DSID())
```
sumwale: Add this to a method "object functions" in the separate file as suggested above.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1273,3 +1282,32 @@ case class ExternalClusterMode(override val sc: SparkContext,
...
+  override def nullable: Boolean = false
+
+  override def dataType: DataType = StringType
+
+  override val prettyName = "DSID"
+
+  override def eval(input: InternalRow): UTF8String = {
+    UTF8String.fromString(com.gemstone.gemfire.internal.cache.GemFireCacheImpl.getInstance()
+      .getDistributionManager().getId().getId());
+  }
```
sumwale: This can throw NPE. Use String.valueOf(Misc.getMyId())

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1273,3 +1282,32 @@ case class ExternalClusterMode(override val sc: SparkContext,
...
+
+  override def doGenCode(ctx: CodegenContext, ev: ExprCode): ExprCode = {
+
+    val idTerm = ctx.freshName("dsid")
+    ctx.addMutableState("UTF8String", idTerm, "")
+    ctx.addPartitionInitializationStatement(s"$idTerm = UTF8String.fromString(com.gemstone" +
+      s".gemfire.internal.cache.GemFireCacheImpl.getInstance().getDistributionManager().getId()" +
+      s".getId());")
```
sumwale: Use String.valueOf(Misc.getMyId()) as suggested above

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1273,3 +1282,32 @@ case class ExternalClusterMode(override val sc: SparkContext,
...
+  override def eval(input: InternalRow): UTF8String = {
+    UTF8String.fromString(com.gemstone.gemfire.internal.cache.GemFireCacheImpl.getInstance()
+      .getDistributionManager().getId().getId());
+  }
+
+  override def doGenCode(ctx: CodegenContext, ev: ExprCode): ExprCode = {
+
+    val idTerm = ctx.freshName("dsid")
+    ctx.addMutableState("UTF8String", idTerm, "")
+    ctx.addPartitionInitializationStatement(s"$idTerm = UTF8String.fromString(com.gemstone" +
```
sumwale: Any reason to use "addPartitionInitializationStatement" and not put the code in third argument of addMutableState itself? This doesn't need partitionIndex.

SachinJanani: No not any specific reason i thought that initialization should go in init method

SachinJanani:
Done with precheckin and also incorporated review comments

PR #717, opened 2017-07-18
Support for default snapshot isolation in write op

suranjan:
For column table writes, start an snapshot isolation
    If any exception then rollback
    Added tests for the same

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
@@ -18,6 +18,7 @@ package org.apache.spark.sql.execution.columnar
 
 import scala.collection.mutable.ArrayBuffer
 
+import com.pivotal.gemfirexd.internal.engine.Misc
 import io.snappydata.Property
 
 import org.apache.spark.TaskContext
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnInsertExec.scala
```diff
@@ -329,6 +347,16 @@ case class ColumnInsertExec(_child: SparkPlan, partitionColumns: Seq[String],
        |${if (numInsertedRowsMetric eq null) ""
           else s"$numInsertedRowsMetric.${metricAdd(numInsertions)};"}
        |${consume(ctx, Seq(ExprCode("", "false", numInsertions)))}
+       |success = true;
+       |}
+       |finally {
+       |if (txStarted) {
+       |  if(success)
+       |    com.pivotal.gemfirexd.internal.engine.Misc.getGemFireCache().getCacheTransactionManager().commit();
```
rishitesh: The same way we started a transaction based on embedded or connector mode, we should rollback commit also in same fashion. The same code will be used for both mode and accessing a cache here will fail.

PR #718, opened 2017-07-18
Support for snapshot transactional insert in column table

suranjan:
Begin the snapshot tx before insert and commit or rollback depending on
  the success of operation

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


rishitesh:
What is the scope of this PR ?
a) Only Spark side df.insert ()  ?
b) a &  & snSession.insert(rows ) ?
c ) a, b & JDBC inserts ?

suranjan:
C. For JDBC inserts, the changes are in store side.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -64,14 +65,127 @@ class JDBCSourceAsColumnarStore(override val connProperties: ConnectionPropertie
...
+
   override def storeColumnBatch(tableName: String, batch: ColumnBatch,
       partitionId: Int, batchId: Option[String], maxDeltaRows: Int): Unit = {
     // noinspection RedundantDefaultArgument
-    tryExecute(tableName, doInsert(tableName, batch, batchId,
-      getPartitionID(tableName, partitionId), maxDeltaRows),
-      closeOnSuccess = true, onExecutor = true)
+    //val connectedInstance = getConnectedExternalStore(tableName,onExecutor = true)
+    //connectedInstance.
+    tryExecute(tableName, closeOnSuccess = false, onExecutor = true)(doInsert(tableName, batch, batchId,
```
rishitesh: Would it not be simpler if we return a connection from beginTx and use that connection to call storeColumnBatch ?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -64,14 +65,127 @@ class JDBCSourceAsColumnarStore(override val connProperties: ConnectionPropertie
...
+            ps.close()
+            SparkShellRDDHelper.snapshotTxIdForWrite.set(null)
+            logDebug(s"Rolledback ${txId} the transaction on server ")
+        }
+      }
+    }(implicitly, Some(conn))
+  }
+
+
+  def closeConnection(): Unit = {
```
rishitesh: Unused method ?

rishitesh:
Overall looks ok to merge. Some minor comments & clarifications

suranjan:
@sumwale @rishitesh 
This along with https://github.com/SnappyDataInc/snappy-store/pull/249, contains only changes required for df insert as transactional.

A separate pull request will be raised for JDBC support.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -65,13 +66,121 @@ class JDBCSourceAsColumnarStore(override val connProperties: ConnectionPropertie
...
+  def commitTx(txId: String)(implicit conn: Option[Connection]): Unit = {
+    tryExecute(tableName, closeOnSuccessOrFailure = true, onExecutor = true) {
+      (conn: Connection) => {
+        connectionType match {
+          case ConnectionType.Embedded =>
+            Misc.getGemFireCache().getCacheTransactionManager().commit()
+          case _ =>
+            logDebug(s"Going to commit ${txId} the transaction on server ")
+            val ps = conn.prepareStatement(s"call sys.COMMIT_SNAPSHOT_TXID(?)")
+            ps.setString(1, if (txId == null) "null" else txId)
```
kneeraj: If txId is null, do you still need to call the procedure?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -65,13 +66,121 @@ class JDBCSourceAsColumnarStore(override val connProperties: ConnectionPropertie
...
+
+  def rollbackTx(txId: String)(implicit conn: Option[Connection]): Unit = {
+    tryExecute(tableName, closeOnSuccessOrFailure = true, onExecutor = true) {
+      (conn: Connection) => {
+        connectionType match {
+          case ConnectionType.Embedded =>
+            Misc.getGemFireCache().getCacheTransactionManager().rollback()
+          case _ =>
+            logDebug(s"Going to rollback ${txId} the transaction on server ")
+            val ps = conn.prepareStatement(s"call sys.ROLLBACK_SNAPSHOT_TXID(?)")
```
kneeraj: Is there a way we can avoid this preparation everytime for rollback, commits, fetch etc. Ideally for one connection we just need this to happen once. Isn't it? Please think about this. If easily doable do it now but make a note of it. May be file a JIRA task if you thnk this should be done.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -65,13 +66,121 @@ class JDBCSourceAsColumnarStore(override val connProperties: ConnectionPropertie
...
+        case _ =>
+          // it should always be not null.
+          // get clears the state from connection
+          // the tx would have been committed earlier
+          // or it will be committed later
+          val txId = SparkShellRDDHelper.snapshotTxIdForWrite.get
+          if (!txId.equals("null")) {
+            val statement = conn.createStatement()
+            statement.execute(
+              s"call sys.GET_SNAPSHOT_TXID('$txId')")
```
kneeraj: So this is for setting the correct txid in the transactionobject on server?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -179,6 +290,10 @@ class JDBCSourceAsColumnarStore(override val connProperties: ConnectionPropertie
           stmt.close()
         } catch {
           // TODO: test this code
+          //TODO:Suranjan This code is no needed now.
+          // Need to rollback.
```
kneeraj: When you will add rollback. I think another PR is there. Are we doing rollback only in case of NonFatal exception or otherwise too?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -370,42 +486,41 @@ class JDBCSourceAsColumnarStore(override val connProperties: ConnectionPropertie
...
-                val numPrimaries = primaryBucketIds.size()
-                // if no local primary bucket, then select some other
-                if (numPrimaries > 0) {
-                  primaryBucketIds.getQuick(rand.nextInt(numPrimaries))
-                } else {
-                  rand.nextInt(pr.getTotalNumberOfBuckets)
-                }
+    val connection = conn match {
+      case Some(conn) => conn
+      case None => getConnection(tableName, onExecutor = true)
```
kneeraj: Why the case None was not handled before and you had to add that.

suranjan: Here we are passing the already used connection. If not passed i.e None, then gets a new pooled connection.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
suranjan: Rollback will automatically happen through is success is "false" in ColumninsertExec.
This was old code, to try to clean already inserted rows/entries from table in case of some failure. Don't think that would have worked anyway. 
Will be removing this code of deleting entries in case of failure completely. Wanted @sumwale opinion so left that.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
suranjan: This would not happen mostly in case of insert. But this used to happen in read in case of join. Will try to test it.
However, it will be no-op on server if txId is null.

core/src/main/scala/io/snappydata/impl/SparkShellRDDHelper.scala
```diff
@@ -103,7 +103,7 @@ final class SparkShellRDDHelper {
 //      }
     }
 
-    val txId = SparkShellRDDHelper.snapshotTxId.get()
+    val txId = SparkShellRDDHelper.snapshotTxIdForRead.get()
```
suranjan: Will do.

cluster/src/dunit/scala/io/snappydata/cluster/ValidateMVCCDUnitTest.scala
```diff
@@ -115,6 +115,96 @@ class ValidateMVCCDUnitTest(val s: String) extends ClusterManagerTestBase(s) wit
...
+  def testSnapshotInsertionForColumnTableDFInsertMultiThreaded(): Unit = {
+    errorInThread = null
+    val netPort1 = AvailablePortHelper.getRandomAvailableTCPPort
+    vm0.invoke(classOf[ClusterManagerTestBase], "startNetServer", netPort1)
+
+    val snc = SnappyContext(sc)
+    val tableName: String = "TESTTABLE"
+
+    snc.sql(s"create table $tableName(col1 integer, col2 String, col3 integer) using column " +
+        s"OPTIONS (PARTITION_BY 'col1', buckets '10',MAXPARTSIZE '200'," +
```
suranjan: Writing some more tests to block the insert with the use of test hook.

PR #720, opened 2017-07-18
ENT-1 - Code refactoring to enable OSS versus non-OSS modes

vjr:
@sumwale this is just the initial commit but can I get an initial review from you before I move on to the pending items viz. gradle changes and reflection based classloading after checking the new GFXD `Enterprise-Mode` property.

There is another PR on the store repo which I will link here once created.


vjr:
Also see https://github.com/SnappyDataInc/snappy-store/pull/243

vjr:
@sumwale whenever you get the time please do an initial review/comments.

build.gradle
```diff
@@ -609,6 +609,9 @@ task product(type: Zip) {
     dependsOn ":snappy-aqp_${scalaBinaryVersion}:jar"
     targetProject = project(":snappy-aqp_${scalaBinaryVersion}")
   }
+  if (rootProject.hasProperty('enterprise')) {
+    dependsOn ":snappy-enterprise_${scalaBinaryVersion}:jar"
+  }
```
sumwale: Can change this property to "snappydata.enterprise" so that its cleaner to set in user's ~/.gradle/gradle.properties if required

cluster/build.gradle
```diff
@@ -109,6 +109,7 @@ dependencies {
   }
   testCompile "org.scalatest:scalatest_${scalaBinaryVersion}:${scalatestVersion}"
 
+  // testRuntime project(path: ':snappy-enterprise_' + scalaBinaryVersion)
```
sumwale: will need to uncomment this once the project is setup

dtests/build.gradle
```diff
@@ -103,6 +103,7 @@ dependencies {
     testCompile "org.scalatest:scalatest_${scalaBinaryVersion}:${scalatestVersion}"
     testCompile project(':snappy-cluster_' + scalaBinaryVersion)
     testRuntime project(':snappy-cluster_' + scalaBinaryVersion)
+    // testRuntime project(':snappy-enterprise_' + scalaBinaryVersion)
```
sumwale: will need to uncomment this once the project is setup

settings.gradle
```diff
@@ -160,6 +160,8 @@ if (new File(rootDir, 'spark-jobserver/build.gradle').exists()) {
 if (new File(rootDir, 'aqp/build.gradle').exists()) {
   include ':snappy-aqp_' + scalaBinaryVersion
   project(':snappy-aqp_' + scalaBinaryVersion).projectDir = "$rootDir/aqp" as File
+  // include ':snappy-enterprise_' + scalaBinaryVersion
+  // project(':snappy-enterprise_' + scalaBinaryVersion).projectDir = "$rootDir/aqp/enterprise" as File
```
sumwale: will need to uncomment this once the project is setup

vjr:
Thanks @sumwale for the review I will get this part done tomorrow.

vjr:
@sumwale does the rest of the diff look OK or anything needs changing/fixing there?

sumwale:
Yes rest looks fine.

ashetkar:
Related PR https://github.com/SnappyDataInc/snappy-aqp/pull/154

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -239,6 +242,25 @@ class LeadImpl extends ServerImpl with Lead
...
 
+  private def checkAuthProvider(props: Properties): Unit = {
+    var authP = props.getProperty(Attribute.AUTH_PROVIDER)
+    if (authP == null) {
+      authP = props.getProperty(Attribute.SERVER_AUTH_PROVIDER)
+    }
+    if (authP != null && !"LDAP".equalsIgnoreCase(authP)) {
+      throw new IllegalArgumentException("LDAP is the only supported auth-provider currently.")
+    } else if (authP != null && !isEnterpriseEdition()) {
+      throw new UnsupportedOperationException("Security feature is available in SnappyData " +
```
hbhanawat: Why UnsupportedOp at one place and IllegalState at another place?

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -239,6 +243,25 @@ class LeadImpl extends ServerImpl with Lead
...
```
hbhanawat: Why UnsupportedOp at one place and IllegalState at another place?

ashetkar: Hmm, thinking was that anything other than LDAP is not legal currently, while auth-provider itself is unsupported for OSS. No strict requirement, both could be throwing same type.

build.gradle
```diff
@@ -777,15 +777,6 @@ task product(type: Zip) {
...
     }
-    // copy AQP jar from maven repo if project not present
-    if (!hasAQPProject) {
-      copy {
-        from examplesProject.configurations.testRuntime.filter {
-          it.getName().contains('snappy-aqp')
-        }
-        into "${snappyProductDir}/jars"
-      }
-    }
```
sumwale: Removing this means that one cannot run the test suite without having aqp checkout (i.e. no one outside the org)? The idea was to fetch the non-oss jar from maven and let the tests run in any case.

ashetkar: As per my understanding, the non-oss jars will not be available even on maven with 1.0. @kneeraj ?
So, yes, the tests requiring non-oss components currently fail (e,g, security tests, aqp examples and tests specifying memory-size), but others should be running fine. We'll need to exclude such tests from respective test suite depending upon the enterprise flag.

PR #722, opened 2017-07-18
[SNAP-1389] Update and delete support for column tables

sumwale:

Update and delete support for column tables including

- ColumnDeltaEncoder for delta encoding and merging of deltas
- absolute* methods to ColumnDecoder for random access that are used when merging/sorting
  deltas (will also be useful for indexes later)
- object BitSet added with methods that respect platform endian-ness so now used uniformly
  instead of Spark's BitSetMethods
- cleaning up of plans: RowDMLExec added for deleteFrom, RowMutableExec added
  in SNAP-1388 etc
- BooleanBitSetEncoder for boolean column encoding

(will add more details later)

## Patch testing

in progress

## ReleaseNotes.txt changes

Column update/delete support

## Other PRs 

NA


sumwale:
reopening against master

PR #727, opened 2017-07-20
Snap-1422

dshirish:
Clearing the connector catalog cache after every query execution in CodegenSparkFallback to avoid issues as reported in SNAP-1422.  Adding this fix as this is a usability issue. However provided a system property just in case we do not want to clear the catalog after every query.

## Patch testing
Test case as described in the bug
precheckin

## ReleaseNotes.txt changes
NA
## Other PRs 
Fixed one more issue discovered in testing for this
https://github.com/SnappyDataInc/snappy-store/pull/245

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
@@ -91,7 +91,7 @@ case class CodegenSparkFallback(var child: SparkPlan) extends UnaryExecNode {
     } finally {
       // SNAP-1422
       val cacheMetaDataForConnector =
-        java.lang.Boolean.getBoolean("SMART_CONNECTOR_CACHE_METEADATA")
+        java.lang.Boolean.getBoolean("SMART_CONNECTOR_CACHE_METADATA")
       lazy val session = sqlContext.sparkSession.asInstanceOf[SnappySession]
       if (!cacheMetaDataForConnector &&
         ExternalStoreUtils.isSmartConnectorMode(session.sparkContext)) {
--- a/core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
@@ -89,6 +89,7 @@ case class CodegenSparkFallback(var child: SparkPlan) extends UnaryExecNode {
           case None => throw t
         }
     } finally {
+      // SNAP-1422
       val cacheMetaDataForConnector =
         java.lang.Boolean.getBoolean("SMART_CONNECTOR_CACHE_METEADATA")
       lazy val session = sqlContext.sparkSession.asInstanceOf[SnappySession]
--- a/core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
@@ -23,6 +23,7 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.SnappySession
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.Attribute
+import org.apache.spark.sql.execution.columnar.ExternalStoreUtils
 import org.apache.spark.sql.internal.CodeGenerationException
 
 /**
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -198,6 +198,13 @@ object ExternalStoreUtils extends Logging {
     }
   }
 
+  def isSmartConnectorMode(sparkContext: SparkContext): Boolean = {
```
hbhanawat: Can we create a lazy val for this? The decision that app is smart connector or not can be made once.

core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
```diff
@@ -87,6 +88,15 @@ case class CodegenSparkFallback(var child: SparkPlan) extends UnaryExecNode {
             }
           case None => throw t
         }
+    } finally {
+      // SNAP-1422
+      val cacheMetaDataForConnector =
```
hbhanawat: Reading property in every query run does not make sense. This can be made a object var.

core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
```diff
@@ -87,6 +88,15 @@ case class CodegenSparkFallback(var child: SparkPlan) extends UnaryExecNode {
...
+        java.lang.Boolean.getBoolean("SMART_CONNECTOR_CACHE_METADATA")
+      lazy val session = sqlContext.sparkSession.asInstanceOf[SnappySession]
+      if (!cacheMetaDataForConnector &&
+        ExternalStoreUtils.isSmartConnectorMode(session.sparkContext)) {
+        session.sessionCatalog.invalidateAll()
```
hbhanawat: Read through SNAP-1422. This is what Sumedh is saying "There handling for specific exception types can be added which can re-fetch the meta-data in connector mode and then re-execute the entire plan". 

We can not invalidate the catalog after every query execution. We should do it only when we get an exception that too a specific one. And we, retry the query after refreshing the catalog when the exception is received. See the catch part of the CodeGenSparkFallback.

dshirish: Initial idea as was that but it does not work as the error is thrown even before plan is created. Refer to the 2nd last comment in SNAP-1422(https://jira.snappydata.io/browse/SNAP-1422?focusedCommentId=16779&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16779)

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
dshirish: Ok.

core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
dshirish: Ok, user will have to restart the system to set/unset the property.

hbhanawat: I didn't get this?

core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
hbhanawat: Then we should do it in getPlan too. You have a df and the corresponding table is deleted and now you do df.collect. This case will be handled in CodeGenSparkFallback.

sumwale: Perhaps a better option will be to refresh if SnappyConnector*Catalog.tableExists returns false. This is in addition to the refresh in exception handling mentioned before.

This should be a problem with Spark too. Is there a bug on Spark for this or any thoughts on how they have handled this?

sumwale: Ok, ignore last comment. Looked at the bug again and this case is when tableExists is incorrectly returning true. I think we should just document that particular case and provide a public API to clear the meta-data explicitly (or user can create a new session). Not worth it to take the hit of clearing the whole cache every time just for that case IMO.

PR #728, opened 2017-07-20
[Snap 1833] Fix the synchronization problem with sc.addJar()

rishitesh:

SparkContext.addJar is not job specific. For two concurrent jobs, the first spark job which gets scheduled adds jar files for two Snappy jobs. At SnappyExecutor side we keep a map of App-Name to jar path. 
But in this case, 2nd jar entry overwrites the first and 1st jar is lost forever.

To fix this issue
- Each job will set all dependent jars in job local variable. Those jars will not be added in SparkContext jar lists(sc.addJar). Instead, we will manually fetch those jars from the urls in SnappyExecutor.
- - The above urls will make a URLClassloader for the duration of jobs. This particular class loader will be set in thread context class loader.
- Change the SnappyExecutor to keep a cache of classloader. This cache is required so that multiple tasks of a single job won't be creating multiple classloaders.

## Patch testing

pre-checkin

## ReleaseNotes.txt changes

NA

## Other PRs 
https://github.com/SnappyDataInc/spark-jobserver/pull/13
https://github.com/SnappyDataInc/spark/pull/67


kneeraj:
General question before I review - What happens if the app name is same. Are we going to just document this or throw exception.

cluster/src/main/scala/org/apache/spark/sql/SnappySessionFactory.scala
```diff
@@ -56,6 +59,10 @@ object SnappySessionFactory {
       override def makeClassLoader(parent: ContextURLClassLoader): ContextURLClassLoader = {
         SnappyUtils.getSnappyContextURLClassLoader(parent)
       }
+
+      override def addJobJar(jarPath : String): Unit = {
```
hbhanawat: remove this code.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -804,21 +804,40 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
...
+      val urls = if (callbacks != null) {
+        funcDefinition.resources.map { r =>
+          ContextJarUtils.fetchFile(funcDefinition.identifier.toString(), r.uri)
+        }
+      } else {
+        funcDefinition.resources.map { r =>
+          toUrl(r)
+        }
+      }
+      val newClassLoader = new MutableURLClassLoader(urls.toArray, parentLoader)
```
hbhanawat: Will be good to use SnappyURLClassLoader

cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -42,102 +46,79 @@ class SnappyExecutor(
...
-      val diffJars = addedJarFiles.diff(newJarFiles)
-      if (diffJars.nonEmpty) {
-        diffJars.foreach(classloader.removeURL)
-        logInfo("As some of the Jars have been deleted, setting up " +
-            "a new ClassLoader for subsequent Threads")
-        diffJars.foreach(d => logInfo(s"removed jar $d"))
-
-        this.urlClassLoader = new SnappyMutableURLClassLoader(classloader.getURLs(),
-          classloader.getParent, classloader.jobJars)
-        this.replClassLoader = addReplClassLoaderIfNeeded(urlClassLoader)
```
rishitesh: Yes. Will do.

cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -42,102 +46,79 @@ class SnappyExecutor(
...
-    userClassPath.foreach { url =>
-      currentJars(url.getPath.split("/").last) = now
+  val classLoaderCache = {
+    val loader = new CacheLoader[ClassLoaderKey, SnappyMutableURLClassLoader]() {
+      override def load(key: ClassLoaderKey): SnappyMutableURLClassLoader = {
+        val appName = key.appName
+        val appNameAndJars = key.appNameAndJars
+        lazy val hadoopConf = SparkHadoopUtil.get.newConfiguration(conf)
+        val appDependencies = appNameAndJars.drop(2).toSeq
+        Misc.getCacheLogWriter.info(s"Creating ClassLoader for $appName" +
```
rishitesh: Yes, will check.

cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -42,102 +46,79 @@ class SnappyExecutor(
...
-      } else {
-        super.updateDependencies(newFiles, newJars)
+      val taskDeserializationProps = Executor.taskDeserializationProps.get()
+      if (null != taskDeserializationProps) {
+        val appDetails = taskDeserializationProps.getProperty(io.snappydata.Constant
+            .CHANGEABLE_JAR_NAME, "")
+        Misc.getCacheLogWriter.info(s"Submitted Application Details $appDetails")
+        if (!appDetails.isEmpty) {
+          val appNameAndJars = appDetails.split(",")
+          val appName = appNameAndJars(0)
```
rishitesh: ok. will remove them

cluster/src/main/scala/org/apache/spark/sql/SnappySessionFactory.scala
rishitesh: Yes, will remove.

PR #730, opened 2017-07-21
Fixes for issues found during concurrency testing

hbhanawat:

While doing jfr analysis of the concurrency test, it was realized that PooledKryoSerializer.serialize is taking lot of time. On further analysis, it was found that WholeStageCodeGenRDD is serializing the code string which is taking lot of time. It was happening because the Output.writeString was calling Output.writeString_slow. This was happening because if the buffer size is not big enough, Output calls writeString_slow. To fix this, now we determine the size of code string in the PooledKryoSerializer and create a buffer of that size. The problem could also have been fixed by adding optimized writeString/readString specifically for the code body. But decided against it to avoid array copy which would have happenend during resize as the body of the code is around 16K for a simple select query. 

For low latency pool made its minShare as 2 and increased its weight to 2. This was done so that OLTP queries get some space to execute even if a OLAP query is already executing and also it gets priority next time. 

## Patch testing

Precheckin. Will run and report the hydra test results. 

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/63
https://github.com/SnappyDataInc/snappy-store/pull/247


core/src/main/scala/org/apache/spark/serializer/PooledKryoSerializer.scala
```diff
@@ -256,8 +256,34 @@ private[spark] final class PooledKryoSerializerInstance(
...
+    val output = t match {
+      // Special handling for wholeStageCodeGenRDD
+      case p: Tuple2[_ , _] =>
+        // If it is a wholestageRDD, we know the serialization buffer needs to be
+        // bigger than the code string size. If it is not bigger, the writestring call inside
+        // WholeStageCodeGenRDD.write calls writeString_slow. Refer Output.writeString.
+        // So create a buffer of size greater than the size of code.
+        if (p._1.isInstanceOf[Product]) {
+          val rdd = p._1.asInstanceOf[Product]
+          // Hackish way to determine if it is a WholeStageRDD.
```
sumwale: Use `case (rdd: Product, _) => ...` instead of this case and instanceOf

PR #733, opened 2017-07-22
Hydra Test coverage for alterTable

swatisawant:

- Added hydra test coverage for alterTable using snappycontext, snappy shell and jdbc client
- Added support to collect thread dump of all lead members in case of failure while stopping the member
- Modified the startUp tests to run selective set of queries in order to run with 2GB of data volume by default.

## Patch testing

Verified the changes by running startUp bt and few tests from northWind bt.

## ReleaseNotes.txt changes

NA

## Other PRs 

No


swatisawant:
Merging the changes now as need to start the regression with these changes. Will incorporate the review comments in next PR.

PR #735, opened 2017-07-24
Snap 1832

vivekwiz:
SNAP-1832 found that revoke on select do not work on existing connections. Now with every revoke existing plan cache is cleared.

## Patch testing
Manual Test.
Test by QA.
All security Tests.
Precheckin pending.

## ReleaseNotes.txt changes
No

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/250


sumwale:
@vivekwiz Can you explain what has Snappy plan cache got to do with security? The authorization checks should be done by the iterators/queries on executors in any case. Latter will happen for smart connector anyhow, so as per our discussions hadn't we decided to just let GemXD do it for now?

vivekwiz:
Hi @sumwale, for optimisation reasons we are relying to have only once authorisation check per connection. This is achieved by first call of lazy val org.apache.spark.sql.row.JDBCMutableRelation#schema for RowStore. 

For column store a new dummy connection is made for each select call (see first few lines of org.apache.spark.sql.execution.columnar.JDBCAppendableRelation#scanTable). @kneeraj has also suggested to make it as lazy val (on my TODO list) for performance reason.

Test case is, 
1) start cluster with credentials of user2.
2) Open connection1, connection2, connection3 with credentials of user1, user2 and user3.
3) connection1: create row-table and column-table. Insert data in them.
4) connection3: not able to do select on  row-table and column-table.
5) connection2: grant for select to  user2 and user3 on both row-table and column-table.
6) connection3: Now able to do select on row-table and column-table.
7) connection2: revoke select given to  user2 and user3 on both row-table and column-table.
8) Name a new connection 4 with credential of user3.
9) connection4: not able to do select on  row-table and column-table.
10) connection3: Still able to do select on row-table and column-table.

This was happening due to plan caching and that is being purged after each revoke.

sumwale:
@vivekwiz Before going there, how will this work in smart connector mode? How is it possible for the pool connections to bypass security? Or is this all only for embedded mode?

As I had mentioned to Neeraj, even if we think that this is some optimization worth doing in embedded mode, then better option is to route the GRANT/REVOKE themselves like other DDL and add parsing of those on SnappyParser side. Then that can also be used from Spark side.

sumwale:
@vivekwiz To clarify: for routing not saying that anything extra is done to hive metastore or anything. The execution itself would grab a connection and execute this like done by ExternalTableDMLCmd.

vivekwiz:
This scenario was already working on SNAP-1656.

PR #731, opened 2017-07-21
SNAP-1656 Security support in snappydata

ashetkar:
## Changes proposed in this pull request
Security support in snappydata

## Patch testing
precheckin in-progress

## ReleaseNotes.txt changes
NA

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/248

PR #737, opened 2017-07-24
SNAP-1841Authenticate and authz snappy jobs in secure snappy cluster.

ashetkar:
* Users can specify credentials via '--user' option in snappy-submit.sh
  These will be passed to snappysession provided to the jobs.
* In the job code, if users provide different credentials, these override
  the ones provided at script level.
* With this change, JobServer UI will not be accessible in secure cluster.
  TODO Either make users provide credentials for UI or make it accessible to all

## Patch testing
Manual testing.

## ReleaseNotes.txt changes
NA

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/252
https://github.com/SnappyDataInc/spark-jobserver/pull/14


cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -368,6 +373,43 @@ class LeadImpl extends ServerImpl with Lead
...
+
+        def checkCredentials(userPass: Option[UserPass]): Option[AuthInfo] = {
+          userPass match {
+            case Some(u) => {
+              try {
+                val props = new Properties()
+                props.setProperty(Attribute.USERNAME_ATTR, u.user)
+                props.setProperty(Attribute.PASSWORD_ATTR, u.pass)
+                val result = FabricDatabase.getAuthenticationServiceBase.authenticate(Misc.getMemStoreBooting
+                    .getDatabaseName, props)
```
dshirish: Line length exceeds 100 chars

dshirish:
Can you add a few unit tests as well?

ashetkar:
Will add the unit test in a separate PR.

cluster/bin/snappy-job.sh
```diff
@@ -119,6 +120,15 @@ while (( "$#" )); do
...
     ;;
+    --user)
+      shift
+      username="${1:-$TOK_EMPTY}"
+      while [ -z ${password} ]; do
+        read -s -p "Enter password for ${username}: " password
+        echo ""
+      done
+      securePart="-u ${username}:${password}"
+    ;;
```
sumwale: @ashetkar Users can object to the fact that password will be sent as an argument to the command below and hence will be visible in the OS process list (even those belonging to other users). Need to discuss how this can be improved. I guess need to use a client like: https://github.com/bluebreezecf/SparkJobServerClient and allow it to read credentials from a properties file or something.

Please track this generic requirement as a task and list this under it. The task being "Secure password handling and shipping throughout the system": a) properties files for all components, b) key pair to encrypt/decrypt

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -368,6 +373,43 @@ class LeadImpl extends ServerImpl with Lead
...
+          userPass match {
+            case Some(u) => {
+              try {
+                val props = new Properties()
+                props.setProperty(Attribute.USERNAME_ATTR, u.user)
+                props.setProperty(Attribute.PASSWORD_ATTR, u.pass)
+                val result = FabricDatabase.getAuthenticationServiceBase.authenticate(Misc
+                    .getMemStoreBooting.getDatabaseName, props)
+                if (result != null) {
+                  logInfo(s"ACCESS DENIED, user [${u.user}]. $result")
```
sumwale: Add the tag used on store side throughout: GfxdConstants.TRACE_AUTHENTICATION. This will ensure a clean way to filter out auth related logging and create an audit log. Later we can also add support in our logger to send these tags to a separate file if required.

cluster/src/main/scala/org/apache/spark/sql/SnappySessionFactory.scala
```diff
@@ -66,19 +66,53 @@ trait SnappySQLJob extends SparkJobBase {
...
 
+  private def updateCredentials(snc: SnappySession, jobConfig: Config): Config = {
+    var authP = ""
+    try {
+      authP = jobConfig.getString(Constant.STORE_PROPERTY_PREFIX + com.pivotal.gemfirexd
+          .Attribute.AUTH_PROVIDER)
+    } catch {
+      case m: ConfigException.Missing => // Security not enabled.
+    }
+    if ("LDAP".equalsIgnoreCase(authP)) {
```
sumwale: should use isSecurityEnabled?

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -283,39 +283,47 @@ object ExternalStoreUtils extends Logging {
...
+        props
       }
-    } else ConnectionProperties(url, driver, dialect, poolProps, connProps, executorConnProps,
-      hikariCP)
+
+      // Hikari only take 'username'. So does Tomcat
+      def securePoolProps(props: Map[String, String]): Map[String, String] = {
+        props + (Attribute.USERNAME_ALT_ATTR.toLowerCase -> user) + (Attribute.PASSWORD_ATTR ->
+            password)
+      }
```
sumwale: Are pool properties settings required now that we key by userName and each user already has a separate pool? I think this can be removed.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -283,39 +283,47 @@ object ExternalStoreUtils extends Logging {
...
-    val prefix = SnappyContext.getClusterMode(session.get.sparkContext) match {
+  def getCredentials(session: SparkSession, prefix: String = ""): (String, String) = {
+    val prefix = SnappyContext.getClusterMode(session.sparkContext) match {
       case ThinClientConnectorMode(_, _) => Constant.SPARK_STORE_PREFIX
       case _ => ""
     }
-    (session.get.conf.get(prefix + Attribute.USERNAME_ATTR, ""),
-        session.get.conf.get(prefix + Attribute.PASSWORD_ATTR, ""))
+    (session.conf.get(prefix + Attribute.USERNAME_ATTR, ""),
+        session.conf.get(prefix + Attribute.PASSWORD_ATTR, ""))
```
sumwale: Much better than using session.get everywhere.

cluster/bin/snappy-job.sh
ashetkar: curl, on some platforms, claims to hide the credentials sent via -u option.
Nevertheless, will change to another option from curl, --netrc, which takes the credentials in a netrc file with a simple format. So --user will be removed from above and replaced with --passfile.
Will file JIRAs for the generic requirement, anyway.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
ashetkar: Done

cluster/src/main/scala/org/apache/spark/sql/SnappySessionFactory.scala
ashetkar: Done

cluster/bin/snappy-job.sh
sumwale: @ashetkar Even if curl somehow does hide the arguments (not sure if it can alter an OS "feature" but perhaps there is some kernel API to do that), this shell script itself will not and will be seen in the script arguments. Using the "--netrc" option is the proper way.

ashetkar: Yes, switched to --netrc before merging the change into master (https://github.com/SnappyDataInc/snappydata/pull/731/commits/09c66ff9060f4d5ea85eab1f0013d11d73ab0278)
But later saw some issues with it so changed to using -K or --config option (https://github.com/SnappyDataInc/snappydata/commit/b5f0fde11008b65c28415b9590c923077940c522). This one is yet to be merged into master.

ashetkar:
> Can you add a few unit tests as well?

Added in https://github.com/SnappyDataInc/snappydata/pull/763

PR #740, opened 2017-07-25
Now each pooled datasources is keyed against both table-name and user-name

vivekwiz:
Now each pooled datasources is keyed against both table-name and user-name.

Also removed redundant connection to XD from JDBCAppendableRelation.scala

## Patch testing
Manual testing of scenario. 
Dunit tests for security.
Precheckin pending.

## ReleaseNotes.txt changes
No

## Other PRs 
No

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/ConnectionPool.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/ConnectionPool.scala
@@ -81,12 +81,13 @@ object ConnectionPool {
   def getPoolDataSource(id: String, props: Map[String, String],
       connectionProps: Properties, hikariCP: Boolean): DataSource = {
     // fast lock-free path first (the usual case)
-    val dsKey = idToPoolMap.get((id, props.getOrElse(Attribute.USERNAME_ALT_ATTR.toLowerCase, "")))
+    val username = props.getOrElse(Attribute.USERNAME_ALT_ATTR.toLowerCase, "")
+    val dsKey = idToPoolMap.get((id, username))
     if (dsKey != null) {
       dsKey._1
     } else pools.synchronized {
       // double check after the global lock
-      val dsKey = idToPoolMap.get((id, props.getOrElse(Attribute.USERNAME_ALT_ATTR.toLowerCase, "")))
+      val dsKey = idToPoolMap.get((id, username))
       if (dsKey != null) {
         dsKey._1
       } else {
@@ -97,8 +98,7 @@ object ConnectionPool {
         pools.get(poolKey) match {
           case Some((newDS, ids)) =>
             ids += id
-            val err = idToPoolMap.putIfAbsent((id,
-                props.getOrElse(Attribute.USERNAME_ALT_ATTR.toLowerCase, "")), (newDS, poolKey))
+            val err = idToPoolMap.putIfAbsent((id, username), (newDS, poolKey))
             require(err == null, s"unexpected existing pool for $id: $err")
             newDS
           case None =>
@@ -118,8 +118,7 @@ object ConnectionPool {
               new TDataSource(tconf)
             }
             pools(poolKey) = (newDS, mutable.Set(id))
-            val err = idToPoolMap.putIfAbsent((id,
-                props.getOrElse(Attribute.USERNAME_ALT_ATTR.toLowerCase, "")), (newDS, poolKey))
+            val err = idToPoolMap.putIfAbsent((id, username), (newDS, poolKey))
             require(err == null, s"unexpected existing pool for $id: $err")
             newDS
         }
--- a/core/src/main/scala/org/apache/spark/sql/execution/ConnectionPool.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/ConnectionPool.scala
@@ -23,9 +23,10 @@ import javax.sql.DataSource
 import scala.collection.JavaConverters._
 import scala.collection.mutable
 
+import com.pivotal.gemfirexd.Attribute
 import com.zaxxer.hikari.util.PropertyElf
 import com.zaxxer.hikari.{HikariConfig, HikariDataSource => HDataSource}
-import org.apache.tomcat.jdbc.pool.{DataSource => TDataSource, PoolProperties}
+import org.apache.tomcat.jdbc.pool.{PoolProperties, DataSource => TDataSource}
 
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/apache/spark/sql/execution/ConnectionPool.scala
```diff
@@ -80,12 +81,12 @@ object ConnectionPool {
...
       connectionProps: Properties, hikariCP: Boolean): DataSource = {
     // fast lock-free path first (the usual case)
-    val dsKey = idToPoolMap.get(id)
+    val dsKey = idToPoolMap.get((id, props.getOrElse(Attribute.USERNAME_ALT_ATTR.toLowerCase, "")))
     if (dsKey != null) {
       dsKey._1
     } else pools.synchronized {
       // double check after the global lock
-      val dsKey = idToPoolMap.get(id)
+      val dsKey = idToPoolMap.get((id, props.getOrElse(Attribute.USERNAME_ALT_ATTR.toLowerCase, "")))
```
ashetkar: Can put the username attribute value in a val and use it at all the places.

vivekwiz:
Incorporated review comments.

core/src/main/scala/org/apache/spark/sql/execution/ConnectionPool.scala
```diff
@@ -80,12 +81,13 @@ object ConnectionPool {
   def getPoolDataSource(id: String, props: Map[String, String],
       connectionProps: Properties, hikariCP: Boolean): DataSource = {
     // fast lock-free path first (the usual case)
-    val dsKey = idToPoolMap.get(id)
+    val username = props.getOrElse(Attribute.USERNAME_ALT_ATTR.toLowerCase, "")
+    val dsKey = idToPoolMap.get((id, username))
```
sumwale: Can create the lookup key tuple only once: `val lookupKey = id -> username` and use it in all the places.

core/src/main/scala/org/apache/spark/sql/execution/ConnectionPool.scala
```diff
@@ -169,8 +171,8 @@ object ConnectionPool {
    * @return true if this was the last reference and entire pool was removed
    *         and false otherwise
    */
-  def removePoolReference(id: String): Boolean = {
-    val dsKey = idToPoolMap.remove(id)
+  def removePoolReference(id: String, userName: String): Boolean = {
+    val dsKey = idToPoolMap.remove((id, userName))
```
sumwale: This is not user specific rather needs to be done for all the users since the underlying table itself is being dropped. So iterate over the idToPoolMap and remove all entries with matching ids.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -499,21 +499,21 @@ object ExternalStoreUtils extends Logging {
...
       SnappyStoreHiveCatalog.registerRelationDestroy()
     }
   }
 
-  def removeCachedObjects(table: String): () => Iterator[Unit] = () => {
-    ConnectionPool.removePoolReference(table)
+  def removeCachedObjects(table: String, userName: String): () => Iterator[Unit] = () => {
+    ConnectionPool.removePoolReference(table, userName)
     CodeGeneration.removeCache(table)
     Iterator.empty
```
sumwale: These calls are not user specific because table itself is being dropped so needs to be cleared for all the users as mentioned above.

core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCAppendableRelation.scala
```diff
@@ -254,7 +247,8 @@ abstract case class JDBCAppendableRelation(
     val conn = connFactory()
     try {
       // clean up the connection pool and caches
-      ExternalStoreUtils.removeCachedObjects(sqlContext, table)
+      ExternalStoreUtils.removeCachedObjects(sqlContext, table,
+        connProperties.poolProps.getOrElse(Attribute.USERNAME_ALT_ATTR.toLowerCase, ""))
```
sumwale: remove the username arg as mentioned

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -294,7 +297,8 @@ abstract class BaseColumnFormatRelation(
     val conn = connFactory()
     try {
       // clean up the connection pool and caches
-      StoreUtils.removeCachedObjects(sqlContext, table)
+      StoreUtils.removeCachedObjects(sqlContext, table,
+        connProperties.poolProps.getOrElse(Attribute.USERNAME_ALT_ATTR.toLowerCase, ""))
```
sumwale: same as above; changes to this file can be reverted

core/src/main/scala/org/apache/spark/sql/row/JDBCMutableRelation.scala
```diff
@@ -300,7 +301,8 @@ case class JDBCMutableRelation(
     val conn = connFactory()
     try {
       // clean up the connection pool and caches
-      ExternalStoreUtils.removeCachedObjects(sqlContext, table)
+      ExternalStoreUtils.removeCachedObjects(sqlContext, table,
+        connProperties.poolProps.getOrElse(Attribute.USERNAME_ALT_ATTR.toLowerCase, ""))
```
sumwale: same as above

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -265,9 +265,9 @@ object StoreUtils {
     partitions
   }
 
-  def removeCachedObjects(sqlContext: SQLContext, table: String,
+  def removeCachedObjects(sqlContext: SQLContext, table: String, userName: String,
       registerDestroy: Boolean = false): Unit = {
-    ExternalStoreUtils.removeCachedObjects(sqlContext, table, registerDestroy)
+    ExternalStoreUtils.removeCachedObjects(sqlContext, table, userName, registerDestroy)
```
sumwale: as above

vivekwiz:
@sumwale all done in SNAP-1656 branch

PR #746, opened 2017-07-30
Snap 1860

suranjan:

Close the connection taken for insert to column batch. Earlier if the earlier tx state was used then commit/rollback was not called (which used to close the connection as well)
Now explicitly closing the connection.

## Patch testing

Ran precheckin
Ran a test with multiple threads doing inserts on row buffer so that flush operation happens regularly in already iniitiated tx.

## ReleaseNotes.txt changes

## Other PRs 


vivekwiz:
@suranjan may be you can checkin the test too (can modify as per your requirements).

vivekwiz:
Sorry, clicked wrong tab

kneeraj:
LGTM.

PR #747, opened 2017-07-30
[SNAP-1389] Update and delete support for column tables

sumwale:

Update and delete support for column tables including

- New Update and Delete LogicalPlans which are introduced by the parser for
  corresponding "update ... where <query>" and "delete ... where <query>" plans
- New MutableRelation interface that returns:
  - the additional key columns to be projected for firing the underlying update/delete
    which will use just the provided key columns that should be unique for each qualified row
  - SparkPlans for the update/delete operations
- implementation of the above for row tables in base JDBCMutableRelation
- some refactoring in TableInsertExec to split out CodegenSupportOnExecutor

- ColumnDeltaEncoder for delta encoding and merging of deltas
- MutatedColumnDecoder to handle all deltas as well as deletes. It just returns handle to
  ColumnDeltaDecoder while generated code makes the actual calls on it. This manual
  inlining improves performance significantly in micro-benchmarks.
- ColumnDelta: a store-side Delta to merge two ColumnDeltas
- absolute* methods to ColumnDecoder for random access that are used when merging/sorting
  deltas (will also be useful for indexes later)
- ColumnDeleteEncoder/ColumnDeleteDelta: simple encoder and delta merge which stores
  the positions deleted (can be improved to use bitset+inverse positions for large deletions)
- object BitSet added with methods that respect platform endian-ness so now used uniformly
  instead of Spark's BitSetMethods
- cleaning up of plans: RowDeleteExec added for deleteFrom, RowUpdateExec added
  in SNAP-1388 etc
- BooleanBitSetEncoder for boolean column encoding

This PR subsumes SNAP-1388 so closing that one.

## Patch testing

precheckin, extensive micro-benchmarking, some unit tests in this PR (more will be added
  subsequently)

## ReleaseNotes.txt changes

Column update/delete support

## Other PRs 

NA


sumwale:
Posting some useful discussion from SNAP-1388

@hbhanawat When will the RowDeleteExec and RowUpdateExec be called on executor? Do you plan to wire bulk updates and deletes in gemxd side to these?
  
@sumwale Oh, missed the question. Added the "onExecutor" flag so that it can be used to generate local plans in future. For example smart connector mode or JDBC single-hop can push down filters which can be applied using such plans especially for column tables.

@hbhanawat I mean we are delegating deletes and updates to gemxd. I don't see any scan happening in snappy side of the code. What is the role of generated code? Same for RowDeleteExec.
  
@sumwale Scan is the child of this plan. SnappyParser with the remaining planning phases will transform an "update ... query" into a RowUpdateExec plan on top of query plan. Same for RowDeleteExec. For column tables, there will be new implementations of these two plan in terms of what outlined in the spec.

@rishitesh Can you please elaborate the usage of keyColumns in the scheme of things. Is it that we would support bulk update & delete only for tables having a primary key? Or I am missing something?
  
@sumwale For row tables it will be either the primary key, or the internal generated long (a ROWID) -- essentially the corresponding region key, or a "RowLocation" if you may. We cannot be having a case where every row update/delete leads to another full table scan. The idea being that during the corresponding query the rows have already been located and there is a unique handle to the row which delete/update can use to quickly get to the row. The current "deleteFrom" implementation suffers from this weakness and will be unusable when table has no primary key.
Likewise for column tables, the "key" will be the batchUUID+position to uniquely locate the row to be updated/deleted. In case there is no exchange/broadcast in the query portion, then this can be further optimized to directly pick up the handle of the DeltaColumn being currently scanned.

sumwale:
Merging now. Will take care of any comments later.

PR #748, opened 2017-07-31
SNAP-1840

ymahajan:
Fixed SNAP-1840, TPCH Q22 query execution failure
Added a simple test to reproduce the issue

## Patch testing

precheckin 

## ReleaseNotes.txt changes

None

## Other PRs 

None

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/SnappyQuerySuite.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/SnappyQuerySuite.scala
@@ -1,0 +1,148 @@
+/*
+ * Copyright (c) 2016 SnappyData, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


hbhanawat:
The actual problem is why the code generation is being done on the executor side. This should not ever happen. This PR fixes the side effect but not the actual problem.

ymahajan:
See this fix in upstream Spark - https://github.com/apache/spark/pull/13569
ScalarSubqueryExec reference the child plan at execution time. Also this is not related to multi node setup(driver/executor) etc, this was reproduced in a simple scalatest added in this PR.

hbhanawat:
Tried your test, it is generating some other error and it looks like your fix fix the problem at hand. But if you see the exception on the bug, it says has the following line: basicPhysicalOperators.scala:216 which is FilterExec.doExecute.Why FilterExec.doExecute is getting called and not the code generation functions?

sumwale:
Agree with Hemant. We need to find why the difference in plans in embedded mode vs connector mode. The code generation and execution paths should be identical for the two cases (except that latter will fetch column batches using thrift connection vs former using region API).

PR #752, opened 2017-08-01
FIx for SNAP-1793 which was caused due to improper cleanup of test data

SachinJanani:
* Added method to proper clean up of the test data
* This issue is caused because the table were dropped from the catalog but the entries were not deleted from the actual store causing the issue in the next test using those tables.

## Patch testing
* Precheckin is done.All tests failing due to this issue are fixed while some other tests are failing but not related to this change

## ReleaseNotes.txt changes
NA
## Other PRs 
NA


core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitTestBase.scala
```diff
@@ -393,6 +395,18 @@ trait SplitClusterDUnitTestObject extends Logging {
     }
     value
   }
+  def validateNoActiveSnapshotTX(): Unit = {
```
kneeraj: @suranjan and @SachinJanani shall we add this to tear down for all dunit test and junit test, such that it gets invoked in all the vms.

kneeraj:
LGTM. Just a single comment/suggestion.

cluster/src/dunit/scala/io/snappydata/cluster/ClusterManagerTestBase.scala
```diff
@@ -269,6 +270,17 @@ object ClusterManagerTestBase extends Logging {
...
+   * This method is used to test HA scenarios without doing cleanup of testdata
+   */
+  def stopSparkWithoutCleanup(): Unit = {
+
+    val service = ServiceManager.currentFabricServiceInstance
+    if (service != null) {
+      service.stop(null)
+    }
+  }
+
```
sumwale: How is this different from stopAny?

SachinJanani: @sumwale They both are same I have not looked it.Will remove this method.

kneeraj:
>> Declining these changes since this seems to be a product bug that is being hidden by this change.

@sumwale why do you think this is a product bug. Reset is a test method internal. Users should not use that. If we want to make this a public api and want reset to be used by people then it should be a clear cleanup. We discussed two options yesterday. Either make reset proper, ( which is what you seem to be suggesting ) or else do the cleanup properly in super.tearDown for functions also. May be we can remove reset completely, that could be another option.

sumwale:
> Reset is a test method internal.

I assume you are talking about "cleanupTestData". It does exactly what a user would do. It gets all the tables, then drops them using dropTable. All of that are public APIs. What is it that "cleanupTestData" does which is internal? The only thing "internal" about it is that it determines the collocation chains and stream dependencies so as to drop tables in proper order.

sumwale:
Oh I see you are talking about catalog.reset which is invoked in this test explicitly. Why is that required in this test? That method should be overridden to only drop in-memory data and nothing more.

kneeraj:
Yes. This was being used in a UDF test. The test was calling catalog reset. This was removing only catalog stuff. Underlying tables were still present. Teardown was incapable of cleaning because there was nothing in the catalog.

sumwale:
Ok, so it looks like dropAllFunctions addition should suffice to fix this issue and then the override of beforeAll itself can be taken out removing the reset call. @rishitesh any other reason to call explicit catalog.reset() other than removing the user-defined functions (which will be taken care of by the dropAllFunctions added in this PR).

sumwale:
> override of beforeAll

sorry tearDown2 and not beforeAll

rishitesh:
@sumwale Yes drop function would suffice. Reset was added to clean up catalog only for tests, without creating any physical entities.

sumwale:
@SachinJanani Ok, only remaining question is that from the looks of it, removing the tearDown2 override should work fine after the dropAllFunctions addition.

SachinJanani:
@sumwale Yes removing override tearDown2 should work fine. Will remove it and run this test once

SachinJanani:
Done with cluster dunit test after incorporating review comments.No new failures hence merging this PR

PR #754, opened 2017-08-02
SNAP-1787 - Handling Array[Decimal] in both embedded and split mode

ymahajan:
- Spark 2.1 Encoders only handle Array of primitives through ScalaReflection and not generic array like Array[DecimalType]
- Overridden createDataFrame in SnappySession to avoid the Encoders route, which fixes the problem for both embedded and split mode.
- Enabled complexTypeTest

## Patch testing

precheckin 

## ReleaseNotes.txt changes

None

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/66

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/TAQTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/TAQTest.scala
@@ -80,7 +80,7 @@ class TAQTest extends SnappyFunSuite {
       tradeSize, numDays, queryNumber = 3, numIters, doInit = false)
   }
 
-  test("SNAP-1787"){
+  test("SNAP-1787") {
     val tableName = "complexTypeTable"
     val arr1 = Array(Decimal("4.92"), Decimal("51.98"))
     val arr2 = Array(4.92f, 51.98f)
--- a/cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/TAQTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/TAQTest.scala
@@ -80,23 +80,6 @@ class TAQTest extends SnappyFunSuite {
       tradeSize, numDays, queryNumber = 3, numIters, doInit = false)
   }
 
-  test("SNAP-1787") {
-    val tableName = "complexTypeTable"
-    val arr1 = Array(Decimal("4.92"), Decimal("51.98"))
-    val arr2 = Array(4.92f, 51.98f)
-    val arr3 = Array(4921212121L, 5198121212L)
-    val arr4 = Array(true, false)
-    val arr5 = Array(4.92, 51.98)
-    val data = scala.collection.mutable.ArrayBuffer[ComplexTypeData]()
-    data +=  new ComplexTypeData(arr1, arr2, arr3, arr4, arr5)
-    val rdd = snc.sparkContext.parallelize(data, 8)
-    val dataDF = snc.createDataFrame(rdd)
-    snc.createTable(s"$tableName", "column", dataDF.schema, Map.empty[String, String])
-    dataDF.write.insertInto(s"$tableName")
-    assert(snc.sql(s"select * from $tableName").count() == 1)
-    snc.dropTable(s"$tableName", true)
-  }
-
   ignore("basic query performance with JDBC") {
     val numRuns = 1000
     val numIters = 1000
@@ -553,6 +536,3 @@ object TAQTest extends Logging with Assertions {
     benchmark.run()
   }
 }
-
-case class ComplexTypeData (arr1: Array[Decimal], arr2: Array[Float],
-  arr3: Array[Long], arr4: Array[Boolean], arr5: Array[Double])
--- a/cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/TAQTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/TAQTest.scala
@@ -21,17 +21,16 @@ import java.time.{ZoneId, ZonedDateTime}
 
 import com.typesafe.config.Config
 import io.snappydata.SnappyFunSuite
-import org.scalatest.Assertions
-
 import org.apache.spark.memory.SnappyUnifiedMemoryManager
 import org.apache.spark.sql._
 import org.apache.spark.sql.execution.benchmark.TAQTest.CreateOp
 import org.apache.spark.sql.internal.SQLConf
-import org.apache.spark.sql.types.{Decimal, DecimalType, StringType, StructField, StructType}
+import org.apache.spark.sql.types._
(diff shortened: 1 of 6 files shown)


core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitTestBase.scala
```diff
@@ -150,7 +150,7 @@ trait SplitClusterDUnitTestBase extends Logging {
     doTestRowTableCreation()
   }
 
-  def _testComplexTypesForColumnTables_SNAP643(): Unit = {
+  def testComplexTypesForColumnTables_SNAP643(): Unit = {
```
ashetkar: Please override it as no-op in SplitClusterDUnitSecurityTest, else test suite may fail. See SplitClusterDUnitSecurityTest.testTableFormChanges()

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitTest.scala
```diff
@@ -437,12 +440,12 @@ object SplitClusterDUnitTest extends SplitClusterDUnitTestObject {
...
-    serializers(0).serialize(dec1)
+    serializers(0).serialize(dbl1)
     try {
-      serializers(0).serialize(dec4)
-      Assert.fail("Expected an IllegalArgumentException")
+      serializers(0).serialize(dec4) // String cannot be casted to Double
+      Assert.fail("Expected an ClassCastException")
     } catch {
-      case _: IllegalArgumentException => // expected
+      case _: ClassCastException => // expected
```
sumwale: Can you check why this is throwing ClassCastException? Which is the line that is trying to cast? Normally this should throw an IllegalArgumentException from GenericValidatingConverter.apply

cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/TAQTest.scala
```diff
@@ -81,6 +80,23 @@ class TAQTest extends SnappyFunSuite {
       tradeSize, numDays, queryNumber = 3, numIters, doInit = false)
   }
 
+  test("SNAP-1787"){
```
sumwale: space before brace

cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/TAQTest.scala
```diff
@@ -81,6 +80,23 @@ class TAQTest extends SnappyFunSuite {
...
+    val arr4 = Array(true, false)
+    val arr5 = Array(4.92, 51.98)
+    val data = scala.collection.mutable.ArrayBuffer[ComplexTypeData]()
+    data +=  new ComplexTypeData(arr1, arr2, arr3, arr4, arr5)
+    val rdd = snc.sparkContext.parallelize(data, 8)
+    val dataDF = snc.createDataFrame(rdd)
+    snc.createTable(s"$tableName", "column", dataDF.schema, Map.empty[String, String])
+    dataDF.write.insertInto(s"$tableName")
+    assert(snc.sql(s"select * from $tableName").count() == 1)
+    snc.dropTable(s"$tableName", true)
```
sumwale: This does not check for ComplexTypeSerializer? Check for its serialize/deserialize compatibility for Array[Decimal].

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitTestBase.scala
sumwale: @ashetkar Why do you anticipate that this test will fail in security test? Can you add some minimal method to base class to override and set credentials in SplitClusterDUnitSecurityTest so that all the tests get run there too with security (including the three which have been overridden as no-op)? I think it will be useful to run all the tests with security to confirm that those scenarios work (unless you think those are covered elsewhere already). In that need to handle just this complex type handling in security dunit because AFAIK there is no separate coverage for complex types in security tests.

ashetkar: Yes, the first two no-op are covered in the security dunit test.
Eventually, this test will be modified so that even the parent testcase should run fine with it but the suggestion was to make it no-op as it was disabled until now and it could result in an unnecessary failure purely because of lack credentials supplied.

ymahajan: @ashetkar i will add a no-op testComplexTypesForColumnTables_SNAP643 in SplitClusterDUnitSecurityTest. Your comment got deleted by mistake.

cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/TAQTest.scala
ymahajan: ok

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitTest.scala
ymahajan: I looked at it yesterday, but couldn't figure out the reason.not sure why string to decimal works but not string to double. will look into it today.

ymahajan:
Array[Decimal] doesn't work with latest stock spark as well.

cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/TAQTest.scala
ymahajan: How do we do this ? similar to SplitClusterDUnitTest#checkSerializationMeataData ? may be i need to understand the ComplexTypeSerializer first.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -164,6 +163,14 @@ class SnappySession(@transient private val sc: SparkContext,
     new SnappySession(sparkContext, Some(sharedState))
   }
 
+  override def createDataFrame[A <: Product : TypeTag](rdd: RDD[A]): DataFrame = {
+    SparkSession.setActiveSession(this)
+    val schema = ScalaReflection.schemaFor[A].dataType.asInstanceOf[StructType]
+    val attributeSeq = schema.toAttributes
+    val rowRDD = RDDConversions.productToRowRdd(rdd, schema.map(_.dataType))
+    Dataset.ofRows(self, LogicalRDD(attributeSeq, rowRDD)(self))
+  }
```
sumwale: Is there an upstream bug for this?

This will not work for other overloads like createDataFrame over java beans, or Seq[A] etc. Also this way is going to be very inefficient compared to the code-generated Encoders. I think better option is to add a separate method (say createDataFrameUsingRDD) instead of override, document the new method for Array[Decimal] and use the same in tests where required for now. Once fix comes upstream then we can remove.

PR #757, opened 2017-08-02
[Snap 1871] Remove custom built-in jdbc provider and instead use spark's JDBC provider

SachinJanani:
* Removed mapping of custom built-in jdbc provider and instead use spark's JDBC provider
* Modified tests to use spark jdbc provider

## Patch testing
* Precheckin

## ReleaseNotes.txt changes
Needs some changes in documentation if we have documented it will do it with help of shyja
## Other PRs 
NA

### DIFF
--- a/core/src/test/scala/org/apache/spark/jdbc/ConnectionConfTest.scala
+++ b/core/src/test/scala/org/apache/spark/jdbc/ConnectionConfTest.scala
@@ -172,23 +172,26 @@ class ConnectionConfTest extends SnappyFunSuite with Logging with BeforeAndAfter
     )
 
     // Start the database
-    DriverManager.getConnection(s"jdbc:derby:$path;create=true")
+    val conn = DriverManager.getConnection(s"jdbc:derby:$path;create=true")
+    conn.createStatement().execute("create table TEST_JDBC_TABLE_1(COL1 INTEGER,COL2 INTEGER,COL3 INTEGER)")
 
     snc.sql("DROP TABLE IF EXISTS TEST_JDBC_TABLE_1")
-    snc.sql("CREATE TABLE TEST_JDBC_TABLE_1(COL1 INTEGER,COL2 INTEGER,COL3 INTEGER)")
+    snc.sql(s"CREATE external TABLE TEST_JDBC_TABLE_1  " +
+      s"USING jdbc " +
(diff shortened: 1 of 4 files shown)


core/src/test/scala/org/apache/spark/jdbc/ConnectionConfTest.scala
```diff
@@ -157,23 +157,25 @@ class ConnectionConfTest extends SnappyFunSuite with Logging with BeforeAndAfter
...
     try {
-      rdd.foreachPartition(d => {
+      // Update will not work
+      /*rdd.foreachPartition(d => {
         val conn = ConnectionUtil.getPooledConnection("testDerby", conf)
         TaskContext.get().addTaskCompletionListener(_ => conn.close())
         val stmt = conn.prepareStatement("update TEST_JDBC_TABLE_1 set col1 = 9")
         stmt.executeUpdate()
-      })
+      })*/
```
sumwale: Can you tell what exception you see here? This is using a JDBC connection and firing an update so should have nothing to do with the provider set on Spark side.

SachinJanani: @sumwale  The ConnectionConfTest was failing with SQLException stating Table does not exist.As update was getting routed to store assuming this as row table as causing this exception.Following is the exception thrown:
```
Job aborted due to stage failure: Task 2 in stage 1.0 failed 1 times, most recent failure: Lost task 2.0 in stage 1.0 (TID 7, localhost, executor driver): java.sql.SQLSyntaxErrorException: Table/View 'TEST_JDBC_TABLE_1' does not exist.
	at org.apache.derby.impl.jdbc.SQLExceptionFactory.getSQLException(Unknown Source) 
```

sumwale: @SachinJanani I am still unable to figure what has removal of "jdbc" from inbuilt list got anything to do with the failure here. @rishitesh can you tell what is going on?

SachinJanani: @sumwale After discussing with @rishitesh fixed this issue.The issue was that I was not creating table in derby before creating external table.Merging this PR.

PR #761, opened 2017-08-03
Snap 1734

hbhanawat:

Snappy's execution happens in two phases. First phase the plan is executed to create a rdd which is then used to create a CachedDataFrame. In second phase, the CachedDataFrame is then used for further actions. For accumulating the metrics for first phase, SparkListenerSQLPlanExecutionStart is fired. This keeps the current executionID in _executionIdToData but does not add it to the active executions. This ensures that query is not shown in the UI but the new jobs that are run while the plan is being executed are tracked against this executionID. In the second phase, when the query is actually executed, SparkListenerSQLPlanExecutionStart adds the execution data to the active executions. SparkListenerSQLPlanExecutionEnd is then sent with the accumulated time of both the phases.

For consuming SparkListenerSQLPlanExecutionStart, Snappy's SQLListener has been added. 

Overridden withNewExecutionId in CachedDataFrame so that the above explained process also happens when the dataset APIs are used. 

## Patch testing

Precheckin- Manual for embedded and smart connector mode.

### DIFF
--- a/core/src/main/java/org/apache/spark/sql/hive/SnappySharedState.java
+++ b/core/src/main/java/org/apache/spark/sql/hive/SnappySharedState.java
@@ -54,6 +54,9 @@ public final class SnappySharedState extends SharedState {
 
   private static final String CATALOG_IMPLEMENTATION = "spark.sql.catalogImplementation";
 
+  /**
+   *  Create Snappy's SQL Listener instead of SQLListener
+   */
   private static SQLListener createListenerAndUI(SparkContext sc) {
     if (ExternalStoreUtils.getSQLListener().get() == null) {
       SnappySQLListener listener = new SnappySQLListener(sc.conf());
--- a/core/src/main/java/org/apache/spark/sql/hive/SnappySharedState.java
+++ b/core/src/main/java/org/apache/spark/sql/hive/SnappySharedState.java
@@ -58,7 +58,8 @@ public final class SnappySharedState extends SharedState {
    *  Create Snappy's SQL Listener instead of SQLListener
    */
   private static SQLListener createListenerAndUI(SparkContext sc) {
-    if (ExternalStoreUtils.getSQLListener().get() == null) {
+    SQLListener initListener = ExternalStoreUtils.getSQLListener().get();
+    if (initListener == null) {
       SnappySQLListener listener = new SnappySQLListener(sc.conf());
       if (ExternalStoreUtils.getSQLListener().compareAndSet(null, listener)) {
         sc.addSparkListener(listener);
@@ -67,8 +68,11 @@ public final class SnappySharedState extends SharedState {
           new SQLTab(listener, ui);
         }
       }
+      return ExternalStoreUtils.getSQLListener().get();
     }
-    return ExternalStoreUtils.getSQLListener().get();
+    else {
+      return initListener;
+    }
   }
 
   private SnappySharedState(SparkContext sparkContext) {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -199,6 +200,22 @@ class CachedDataFrame(df: Dataset[Row], var queryString: String,
     collectInternal().map(boundEnc.fromRow).toArray
   }
 
+  override def withNewExecutionId[T](body: => T): T = queryExecution.executedPlan match {
+    // don't create a new executionId for ExecutePlan since it has already done so
+    case _: ExecutePlan => body
+    case _ =>
```
sumwale: ExecutePlan does the actual execution in its plan itself so that one does not need to do a collect() for insert/delete/update. However, if user does do a collect (e.g. to get the number of rows updated/deleted/..), then it will show up separately. Is there a way to connect the two if second one is present or perhaps that is not required? Test with sql("insert into ...") with and without collect() and see how the two look. Maybe the main query is shown in first while second appears as dummy which is fine.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -563,22 +572,29 @@ object CachedDataFrame
    * Custom method to allow passing in cached SparkPlanInfo and queryExecution string.
    */
   def withNewExecutionId[T](sparkSession: SparkSession,
-      queryShortForm: String, queryLongForm: String, queryExecutionStr: String,
-      queryPlanInfo: SparkPlanInfo)(body: => T): T = {
+    queryShortForm: String, queryLongForm: String, queryExecutionStr: String,
+    queryPlanInfo: SparkPlanInfo, currentExecutionId: Option[Long] = None,
+    planProcessingTime: Long = 0)(body: => T): T = {
```
sumwale: Check indentation. Perhaps not using Project level with codeStyleSettings.xml?

core/src/main/java/org/apache/spark/sql/hive/SnappySharedState.java
```diff
@@ -49,6 +54,23 @@
...
+   */
+  private static SQLListener createListenerAndUI(SparkContext sc) {
+    if (ExternalStoreUtils.getSQLListener().get() == null) {
+      SnappySQLListener listener = new SnappySQLListener(sc.conf());
+      if (ExternalStoreUtils.getSQLListener().compareAndSet(null, listener)) {
+        sc.addSparkListener(listener);
+        SparkUI ui = sc.ui().getOrElse(null);
+        if (ui != null) {
+          new SQLTab(listener, ui);
+        }
```
hbhanawat: java file. :-)

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
hbhanawat: for snc.sql("insert into..." ).show. There are two plans. one is dummy as you said. As discussed, keeping it the way it is because merging these plans is not straightforward.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
hbhanawat: done

core/src/main/java/org/apache/spark/sql/hive/SnappySharedState.java
```diff
@@ -49,6 +54,23 @@
 
   private static final String CATALOG_IMPLEMENTATION = "spark.sql.catalogImplementation";
 
+  /**
+   *  Create Snappy's SQL Listener instead of SQLListener
+   */
+  private static SQLListener createListenerAndUI(SparkContext sc) {
+    if (ExternalStoreUtils.getSQLListener().get() == null) {
```
hbhanawat: done

PR #763, opened 2017-08-04
Snap 1656 2 Miscellaneous security related changes

ashetkar:
# Changes proposed in this pull request
- Using curl option --config instead of --netrc to provide credentials to submit snappy jobs.
- Updated the test with dataframe API for inserts into column table to hit the code path which puts rows as cached batches directly into the cached-batch table.
- Clean up the Config instance made available to SnappyStreamingJob and JavaSnapyStreamingJob APIs.
- Added a check to ensure only LDAP can be provided as a value for auth-provider property to enable  security in snappydata cluster. The check for product type is partially done, will be handled in a separate PR.

## Patch testing
precheckin in-progress

## ReleaseNotes.txt changes
NA

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/267


cluster/bin/snappy-job.sh
```diff
@@ -127,7 +127,7 @@ while (( "$#" )); do
         echo "The netrc file $passwordfile not found."
         exit 1
       fi
-      securePart=" --netrc-file ${passwordfile}"
+      securePart=" --config ${passwordfile}"
```
vivekwiz: Don't understand this change. Isn't config is too generic to identify password file?

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -239,6 +240,22 @@ class LeadImpl extends ServerImpl with Lead
...
+    if (authP != null && !"LDAP".equalsIgnoreCase(authP)) {
+      throw new IllegalArgumentException("LDAP is the only supported auth-provider currently.")
+    } else if (authP != null && !isEnterpriseEdition()) {
+      throw new UnsupportedOperationException("Security feature is available in SnappyData " +
+          "Enterprise Edition.")
+    }
+  }
+
+  // TODO Read the property from SnappyDataVersion.properties
+  private def isEnterpriseEdition(): Boolean = true
```
vivekwiz: Can you mention ticket number tracking this issue?

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -239,6 +240,22 @@ class LeadImpl extends ServerImpl with Lead
...
   }
 
+  private def checkAuthProvider(props: Properties): Unit = {
+    var authP = props.getProperty(Attribute.AUTH_PROVIDER)
+    if (authP == null) {
+      authP = props.getProperty(Attribute.SERVER_AUTH_PROVIDER)
+    }
+    if (authP != null && !"LDAP".equalsIgnoreCase(authP)) {
+      throw new IllegalArgumentException("LDAP is the only supported auth-provider currently.")
+    } else if (authP != null && !isEnterpriseEdition()) {
```
vivekwiz: This if clause should be independent, not as "else if"

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -239,6 +240,22 @@ class LeadImpl extends ServerImpl with Lead
     }
   }
 
+  private def checkAuthProvider(props: Properties): Unit = {
+    var authP = props.getProperty(Attribute.AUTH_PROVIDER)
+    if (authP == null) {
+      authP = props.getProperty(Attribute.SERVER_AUTH_PROVIDER)
+    }
```
vivekwiz: what if Attribute.AUTH_PROVIDER and Attribute.SERVER_AUTH_PROVIDER are both non null ; Attribute.AUTH_PROVIDER is ldap and  Attribute.SERVER_AUTH_PROVIDER is not ldap.

cluster/bin/snappy-job.sh
ashetkar: Yes, config file can take any config and not just the credentials.
But the --netrc-file does not seem to be working with snappy-job.sh. Couldn't figure out what the problem was, as it was not working even with explicit curl command. So switched to this option.

Is there any concern with --config?

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
ashetkar: Ticket: ENT-1

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
ashetkar: Doesn't really change anything, right? If it goes through the first if-block, exception is thrown anyway.

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitSecurityTest.scala
```diff
@@ -591,6 +596,44 @@ class SplitClusterDUnitSecurityTest(s: String)
...
+  def testSnappyJob(): Unit = {
+    // Create config file with credentials
+    val jobConfigFile = s"$snappyProductDir/conf/job.config"
+    writeToFile(s"-u $jdbcUser1:$jdbcUser1", jobConfigFile)
+
+    // Run snappy job with credentials in a config file
+    val job = s"$snappyProductDir/bin/snappy-job.sh submit --app-name CreatePartitionedRowTable " +
+        s"--class org.apache.spark.examples.snappydata.CreatePartitionedRowTable " +
+        s"--app-jar $snappyProductDir/examples/jars/quickstart.jar  " +
+        s"--passfile $jobConfigFile"
```
vivekwiz: Slightly confused with names of passFile and config (what discussed before). Does passFile means password file? How that is different from config file?

Also are you testing for --config option?

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
ashetkar: I guess gemxd doesn't support different auth schemes for server and peer authentication. AFAIK, SERVER_AUTH_PROVIDER was added to allow user to only have server authentication and skip peer authentication.

cluster/src/main/scala/org/apache/spark/sql/SnappySessionFactory.scala
```diff
@@ -98,14 +76,37 @@ trait SnappySQLJob extends SparkJobBase {
...
+trait SnappySQLJob extends SparkJobBase {
+  type C = Any
+
+  final override def validate(sc: C, config: Config): SparkJobValidation = {
+    val parentLoader = org.apache.spark.util.Utils.getContextOrSparkClassLoader
+    val currentLoader = SnappyUtils.getSnappyStoreContextLoader(parentLoader)
+    Thread.currentThread().setContextClassLoader(currentLoader)
+    SnappyJobValidate.validate(isValidJob(sc.asInstanceOf[SnappySession], SnappySessionFactory
+        .cleanJobConfig(config)))
+  }
```
vivekwiz: For a single line change why it is showing such a big diff. I dont see any change in indentation too. Can you try to restore file and later do you 2-3 lines of changes. You can restore file as:
git checkout master cluster/src/main/scala/org/apache/spark/sql/SnappySessionFactory.scala

cluster/bin/snappy-job.sh
vivekwiz: Will config take only password file?
If so, a more intuitive name will be better. But if you are not able able to work can be take care in documentation.
If not (means it will take other configuration parameters using --config option) then I am confused how that would work.

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitSecurityTest.scala
ashetkar: --passfile is the option provided by snappy-job.sh to specify file path containing the user credentials. Internally, we pass this path as --config to the curl command.
--config is the option provided by curl to specify file path containing configs.

So --config is not visible to snappy-job.sh but --passfile is. We could rename --passfile to --config but I wanted it to be different as snappy-job users should just specify credentials in it and not any other config.
So the documentation will clearly mention what the content of this file should be, if security is enabled.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
vivekwiz: yes, but better to be more intuitive than to halt and think.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
vivekwiz: Ok, but rather than relying on above information, better to handle here that may just require an extra else/if block.

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitSecurityTest.scala
vivekwiz: Ok got it.

vivekwiz:
LGTM. Have some optional comments that can be incorporated.

cluster/src/main/scala/org/apache/spark/sql/SnappySessionFactory.scala
ashetkar: Here I have moved the updateCredentials and cleanJobConfig from class to object so that it could be used outside.
I think the diff is so because git's diff algorithm first sees the trait being moved down rather than defs being moved up.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
ashetkar: I'm not sure I understood.
If AUTH_PROVIDER is specified, it should override SERVER_AUTH_PROVIDER.
Do you mean to suggest that if both are non-null and different then throw error?

Actually, for snappydata we should altogether skip checking for SERVER_AUTH_PROVIDER as we are not mentioning it in snappydata docs.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
ashetkar: Ok, will update it.

cluster/bin/snappy-job.sh
ashetkar: The other comment should explain it https://github.com/SnappyDataInc/snappydata/pull/763#discussion_r132102194

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
ashetkar: I now remember that SERVER_AUTH_PROVIDER check was added because if user specifies auth-provider as java system property (-J-D-auth-provider), it sets the SERVER_AUTH_PROVIDER property.

vivekwiz: So again, cannot be both value be set?

vivekwiz:
LGTM

PR #764, opened 2017-08-04
Fix for SNAP-1892 caused due to empty UserSpecifiedSchema instead of …

SachinJanani:
…None

* While creating external table after spark-2.1 merge UserSpecifiedSchema was passed as empty instead of None when user has not specified schema which was causing this issue.

## Patch testing
* Modified and tested exsiting tests .Also added assertion for individual rows
* precheckin in progress

## ReleaseNotes.txt changes
NA
## Other PRs 
NA


core/src/main/scala/org/apache/spark/sql/sources/StoreStrategy.scala
```diff
@@ -32,14 +32,16 @@ object StoreStrategy extends Strategy {
   def apply(plan: LogicalPlan): Seq[SparkPlan] = plan match {
 
     case CreateTable(tableDesc, mode, None) =>
-      val userSpecifiedSchema = SparkSession.getActiveSession.get
-        .asInstanceOf[SnappySession].normalizeSchema(tableDesc.schema)
+      val userSpecifiedSchema: Option[StructType] = if (tableDesc.schema.isEmpty) None
+      else
+        Some(SparkSession.getActiveSession.get.asInstanceOf[SnappySession].normalizeSchema
+        (tableDesc.schema))
```
sumwale: add braces in else it would be a scalastyle error

SachinJanani: ok will do that

PR #765, opened 2017-08-05
Snap 1883

kneeraj:

Added support of range operator in the parser.

## Patch testing

Added a unit test to generate rows using the operator and verify the result.

## ReleaseNotes.txt changes

None.

## Other PRs 

No.


core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -499,6 +499,12 @@ class SnappyParser(session: SnappySession)
   }
 
   protected final def relationFactor: Rule1[LogicalPlan] = rule {
+    identifier ~ ws ~ '(' ~ ws ~ (expression * commaSep) ~ ws ~ ')' ~>
+        ((fnname: String, e: Any) =>
+          e match {
+            case ve: Vector[Expression] => UnresolvedTableValuedFunction(fnname.toLowerCase(), ve)
+            case _ => UnresolvedTableValuedFunction(fnname, Seq.empty)
+          }) |
```
sumwale: Don't think we expect it to be the most common case, so better to move this case below the others rather than being the first rule. Note that parser will backtrack when it fails to match the rule, then go to other rules so the most common rules should be at the start.

In fact best option will be to factor out identifier which is common in this rule and tableIdentifier as has been done in many other rules like comparisonExpression. Can leave that marked for TODO for later but moving this rule at the end seems to be the better option for now.

kneeraj: I had it below first. But the parser takes range as unresolved relation and then comes out of the relationFactor completely and then expects group by etc. or where clause and then fails complaining -- Invalid input "(0". To remove this ambiguity in parsing I moved it to top.

kneeraj: Working on a simpler fix.

sumwale: Also the "ws" placements are incorrect. Convention is ws following a token and all non-terminals (so all non-terminals already have a following ws): `identifier ~ '(' ~ ws ~ (expression * commaSep) ~ ')' ~ ws ~>`

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -498,16 +498,31 @@ class SnappyParser(session: SnappySession)
         })
   }
 
+
+  protected final def tableValuedFunctionExpressions: Rule1[
+      Seq[Expression]] = rule {
+    '(' ~ ws ~ (expression + commaSep).? ~ ws ~ ')' ~ ws ~>
```
sumwale: second ws is not required and can be removed

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -498,16 +498,31 @@ class SnappyParser(session: SnappySession)
...
-            val optAlias = alias.asInstanceOf[Option[String]]
-            updatePerTableQueryHint(tableIdent, optAlias)
-            UnresolvedRelation(tableIdent, optAlias)
+            tvfe.asInstanceOf[Option[Seq[Expression]]] match {
+              case None =>
+                val optAlias = alias.asInstanceOf[Option[String]]
+                updatePerTableQueryHint(tableIdent, optAlias)
+                UnresolvedRelation(tableIdent, optAlias)
+              case Some(exprs) =>
+                UnresolvedTableValuedFunction(tableIdent.identifier.toLowerCase(), exprs)
```
sumwale: Why is explicit toLowerCase() required? The FunctionRegistry lookup is already case-insensitive and this should not be required. If really needed use org.apache.spark.sql.collection.Utils.toLowerCase(..) which will use the ENGLISH locale for lower-case conversion.

kneeraj: This was required because it does a lookup in a builtInFunctions somewhere which failed if the range was coming as RANGE. Ok will use the above suggested.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
kneeraj: ok.

PR #766, opened 2017-08-07
Snapshot cleanup

suranjan:

Wrong connection property which was used earlier but not cleaned up.
Keeping it for hive client as there are lots of constraint violation exception otherwise.
## Patch testing

precheckin

## ReleaseNotes.txt changes

## Other PRs 


core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -173,20 +173,19 @@ object ExternalStoreUtils {
...
+            Constant.DEFAULT_EMBEDDED_URL + ";host-data=false;mcast-port=0"
           case ThinClientConnectorMode(_, url) =>
-            url + ";route-query=false;skip-constraint-checks=true"
+            url + ";route-query=false"
           case ExternalEmbeddedMode(_, url) =>
-            Constant.DEFAULT_EMBEDDED_URL + ";host-data=false;skip-constraint-checks=true;" + url
+            Constant.DEFAULT_EMBEDDED_URL + ";host-data=false;" + url
           case LocalMode(_, url) =>
-            Constant.DEFAULT_EMBEDDED_URL + ";skip-constraint-checks=true;" + url
+            Constant.DEFAULT_EMBEDDED_URL + ";" + url
```
sumwale: Why has this flag been removed from here (and added below I guess)?

suranjan: It was added earlier to differentiate between snappy and jdbc connection. Now not needed as snapshot isolation on row table is not supported or will use different as it would have caused no constraint checks.
It is added back in hiveclient because, hive client without this throws lots of constraint violation exception. Discussed with @dshirish if adding in hiveclient is an issue.

PR #768, opened 2017-08-07
[SNAP-1885] Semijoin incorrect result

rishitesh:
The code to evaluate join condition is incorrect. Semijoin should traverse through all the values against a key which the code was doing.
However, the value field was not re-initialized afetr taking the next map entry against a key.

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 
NA


ymahajan:
LGTM

PR #775, opened 2017-08-09
Use same conn for rowbuffer and columnstore

suranjan:
For update/delete on column table use the same
  conn for operation on rowbuffer as well as columnstore
  It makes commit/rollback easier to do.

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


sumwale:
Why this code duplication everywhere in generated code? This is confusing and easy to get wrong (and if there is some error in one place then will need to correct all of insert/update/delete plans). Please move out these explicit commit/rollback to a separate task listener used everywhere uniformly in insert/update/delete plans.

suranjan:
@sumwale Ok. will refactor. TaskListener can only be used for commit/rollback.

Call to begin will have to be as it is? like in initCode.

sumwale:
Can't the begin be done in its constructor?

sumwale:
Obsoleted by https://github.com/SnappyDataInc/snappydata/pull/781

PR #776, opened 2017-08-10
Disabling DDLRoutingDUnitTest.testAlterRowTableRoutingFromXD which wa…

SachinJanani:
…s causing invalid configuration causing other tests to fail

* Disabling DDLRoutingDUnitTest.testAlterRowTableRoutingFromXD
* While investigating on SNAP-1868 https://jira.snappydata.io/browse/SNAP-1868 found that due to DDLRoutingDunitTest.testAlterRowTableRoutingFromXD vm2 is created with different dsid casuing it to have same vmId as that of vm0 which in turn affecting the generation same UUID for keys causing different rows to be overwritten.Causing ColumnTableCreation in SplitSnappyClusterDUnitTest to fail with assertion failure

* Needs more investigation on this as to why such invalid configuration not throwing the exception or rather should not allow the VM to start.Created separate jira for this SNAP-1912

## Patch testing
* Cluster Dunit

## ReleaseNotes.txt changes
NA
## Other PRs 

### DIFF
--- a/cluster/src/dunit/scala/io/snappydata/cluster/DDLRoutingDUnitTest.scala
+++ b/cluster/src/dunit/scala/io/snappydata/cluster/DDLRoutingDUnitTest.scala
@@ -160,7 +160,7 @@ class DDLRoutingDUnitTest(val s: String) extends ClusterManagerTestBase(s) {
     s.execute("DROP DISKSTORE d1")
   }
 
-  def testAlterRowTableRoutingFromXD(): Unit = {
+  def _testAlterRowTableRoutingFromXD(): Unit = {
     val tableName: String = "rowTableDDLRouting"
 
     vm2.invoke(classOf[ClusterManagerTestBase], "stopAny")


suranjan:
Looks good to me. Please raise an issue about not having startup failure with different distributed-system-id in same cluster.
@ymahajan Please see If the test can be re-written without using distributed-system-id or can it have different target so that the vm doesn't interfere with other tests.

SachinJanani:
Thanks @suranjan Raised a task for this SNAP-1912.Closing this PR test by executing cluster dunit

ymahajan:
@SachinJanani the ticket you added is for 1.1. The test is very important to test the alter table functionality. @suranjan it cannot be re-written without dsid as it has gatewaysender ddls. To me the main issue SNAP-1912 should be investigated further and enabled this test asap(preferably in 1.0 itself). Or other way is to have distributed-system-id configured in the ClusterManagerTestBase itself, so that all the vms will have same ds-id and we can enable this test.

sumwale:
@SachinJanani Yes, please re-enable this test ASAP. We CANNOT disable tests without approval of the author, especially when the test represents an important coverage required for a release like ALTER TABLE in this case.

SachinJanani:
@sumwale @ymahajan Will enable this test and modify it to restart vm2 with default props after test finishes so that it will not affect other tests

kneeraj:
@ymahajan and @sumwale I think we should have a seperate test target for all gateway and wan related tests, as we had previously. We should have a different tearDown for these tests where we close the snappysession and delete all persistent files.

sumwale:
@kneeraj that can be done but doesn't this failure (SNAP-1912) represent a bug or is it expected with explicit dsid setting?

kneeraj:
>> but doesn't this failure (SNAP-1912) represent a bug or is it expected with explicit dsid setting?
At least I got this understanding after looking into it briefly, ( though I did not look into it very thoroughly ) that setting a dsid on an already up DS is kind of artificial and messing up the entire generation of vmId in such a way that duplicate vmId is getting assigned. The unique id generation for region keys failed (as in generated duplicates ids across vms ) and hence the problem.

ymahajan:
@SachinJanani right, that's the right approach for now. @kneeraj this is just a simple GW DDL added in the test and doesn't make sense to add additional build target just for that. Sachin's approach is good enough and we can enable the test after that.

kneeraj:
@ymahajan can you please comment on the problem point I see in my above comment. In simpler words I feel the way we are keeping up the ds, not cleaning persistent files etc. we are injecting the trouble here by assigning id artificially which would otherwise not have happened. May be what I am thinking is wrong but if it is then it makes a lot of sense to have a seperate target for these.

PR #778, opened 2017-08-10
SNAP-1884

hbhanawat:

A replicated table is aggregated/exchanged and the resulting plan is saved as a temporary table. This temporary table when joined with SnappyHashJoin, it does not return any rows. It is because this temporary table is seen as a replicated table and the aggregation/exchange is not materialized. Changed the HashJoinExec to materialize the exchange by adding the correct shuffle dependencies.

## Patch testing

Added a test provided by @ymahajan. Precheckin.


core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoinExec.scala
```diff
@@ -183,17 +181,45 @@ case class HashJoinExec(leftKeys: Seq[Expression],
...
+    val buildShuffleDeps: Seq[Dependency[_]] = buildRDDs.flatMap(_.dependencies
+      .filter(_.isInstanceOf[ShuffleDependency[_, _, _]]))
+    val oneToOneRdds = buildRDDs.flatMap(_.dependencies).
+      filter(_.isInstanceOf[OneToOneDependency[_]]).map(_.rdd)
+    if (oneToOneRdds.isEmpty) {
+      buildShuffleDeps
+    } else {
+      buildShuffleDeps ++ findShuffleDeps(oneToOneRdds)
+    }
+  }
```
sumwale: Why is separate treatment of OneToOneDependency required here? How is the output different from a simple flatMap searching for shuffle dependencies like in the code earlier?

core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoinExec.scala
```diff
@@ -183,17 +181,45 @@ case class HashJoinExec(leftKeys: Seq[Expression],
...
+
   private lazy val (streamSideRDDs, buildSideRDDs) = {
     val streamRDDs = streamedPlan.asInstanceOf[CodegenSupport].inputRDDs()
     val buildRDDs = buildPlan.asInstanceOf[CodegenSupport].inputRDDs()
 
-    if (replicatedTableJoin) (streamRDDs, buildRDDs)
+    if (replicatedTableJoin) {
+      val streamRDD = streamRDDs.head
```
sumwale: Isn't the real problem here that "replicatedTableJoin" flag is incorrect? Are these changes required after the fix made in https://github.com/SnappyDataInc/snappydata/pull/784 that corrects the "isPartitioned" determination? Also this looks like common code in the two branches so perhaps can use it instead of repetition.

core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoinExec.scala
hbhanawat: If you check the added test, there are two problems that it exposes: 

1. If the bottom most node in the build side is a replicated table, the join is considered a replicated table join. So if the build side is a replicated table with a shuffle, the shuffled rdd needs to be materialized. We were not doing that previously. 

2. If the buildPlan has Filter as the top most node, the buildRDD becomes a MapPartitionsRDD instead of ShuffledRDD and hence the following check returns no dependencies.

buildRDDs.flatMap(_.dependencies.filter(_.isInstanceOf[ShuffleDependency[_, _, _]]))

So, during execution, the build side reutrns 0 results because the inner ShuffledRDD is not materialized. However, in the test, the ShuffledRDD is wrapped inside the MapPartitionsRDD. We need to go inside the MapPartitionsRDD to find out the dependencies to materialize the internal ShuffledRDD. 

I have merged the latest master on this branch, in case you want to try it out.

core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoinExec.scala
hbhanawat: Please see my previous comment. Even if the replicatedTableJoin is now fixed, it does not overcome this problem.

core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoinExec.scala
sumwale: Confused about 1. Why should there be a shuffle on a replicated table? Shouldn't the fix be that the partitioning of that should be such that it is compatible with other kinds of partitioning? The only exception could be RangePartitioning. Is RangePartitioning the problem here? If its a regular ClusterDistribution then the fix should be to ensure that replicated table is not shuffled.

About 2, if the problem is wrapping inside another RDD, the code should look like in SnappySession.findShuffleDependencies that doesn't do any specific checks like here for OneToOneDependency which will likely fail in cases where a PruneDependency or RangeDependency gets created for example.

hbhanawat: As per our discussion on phone, point number 1 is required because the TakeOrderedAndProjectExec is introducing a ShuffledRowRDD over the replicated table. 

Refactored the code as per your comments for point 2.

core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoinExec.scala
```diff
@@ -183,17 +181,40 @@ case class HashJoinExec(leftKeys: Seq[Expression],
   // return empty here as code of required variables is explicitly instantiated
   override def usedInputs: AttributeSet = AttributeSet.empty
 
+  private def findShuffleDependencies(rdd: RDD[_]): Seq[Dependency[_]] = {
+    rdd.dependencies.flatMap {
+      case s: ShuffleDependency[_, _, _] => s :: Nil
```
sumwale: Shouldn't this be like in SnappySession method:
```scala
      case s: ShuffleDependency[_, _, _] => if (s.rdd ne rdd) {
        s +: findShuffleDependencies(s.rdd)
      } else s :: Nil
```

PR #781, opened 2017-08-10
Snap 1904 2

suranjan:

Use the same connection for update/delete on column table

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnDeleteExec.scala
```diff
@@ -56,6 +61,42 @@ case class ColumnDeleteExec(child: SparkPlan, columnTable: String,
...
+         |   $context.addTaskCompletionListener($taskListener);
+         |}
+         | """.stripMargin
+
+    val endCode =
+      s"""
+         |""".stripMargin
+    val commitCode =
+      s"""
+       """.stripMargin
```
sumwale: empty strings don't need s""" and stripMargin rather ""

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnUpdateExec.scala
```diff
@@ -77,8 +81,44 @@ case class ColumnUpdateExec(child: SparkPlan, columnTable: String,
...
+         |   $context.addTaskCompletionListener($taskListener);
+         |}
+         | """.stripMargin
+
+    val endCode =
+      s"""
+         |""".stripMargin
+    val commitCode =
+      s"""
+       """.stripMargin
```
sumwale: same as above: empty strings don't need s""" and stripMargin rather ""

PR #782, opened 2017-08-11
Issues seen during concurrency testing

hbhanawat:

CachedDataFrame.queryExecutionString was a function which is getting evaluated every time the CDF is being executed. Changed it to lazy val so that it is evaluated once. Same for the number of partitions of the cached RDD.  

## Patch testing

precheckin

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
+++ b/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
@@ -88,7 +88,9 @@ class CachedDataFrame(session: SparkSession, queryExecution: QueryExecution,
   private lazy val boundEnc = exprEnc.resolveAndBind(logicalPlan.output,
     sparkSession.sessionState.analyzer)
 
-  def queryExecutionString: String = queryExecution.toString()
+  private lazy val queryExecutionString: String = queryExecution.toString()
+
+  private lazy val numPartitions: Int = cachedRDD.getNumPartitions
 
   private lazy val lastShuffleCleanups = new Array[Future[Unit]](
     shuffleDependencies.length)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -88,7 +88,9 @@ class CachedDataFrame(session: SparkSession, queryExecution: QueryExecution,
   private lazy val boundEnc = exprEnc.resolveAndBind(logicalPlan.output,
     sparkSession.sessionState.analyzer)
 
-  def queryExecutionString: String = queryExecution.toString()
+  private lazy val queryExecutionString: String = queryExecution.toString()
+
+  private lazy val numPartitions: Int = cachedRDD.getNumPartitions
```
sumwale: Won't this cause trouble if the pruning is different for different tokens? From what I know Soubhik had handled that by explicitly clearing the cached partitions inside cachedRDD and re-calculating in every execution.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -88,7 +88,9 @@ class CachedDataFrame(session: SparkSession, queryExecution: QueryExecution,
   private lazy val boundEnc = exprEnc.resolveAndBind(logicalPlan.output,
     sparkSession.sessionState.analyzer)
 
-  def queryExecutionString: String = queryExecution.toString()
+  private lazy val queryExecutionString: String = queryExecution.toString()
```
sumwale: Need to be able to generate appropriate plans for different tokens as discussed. @kneeraj can you track that issue?

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -146,7 +149,7 @@ class CachedDataFrame(session: SparkSession, queryExecution: QueryExecution,
       sessionState.conf.activeSchedulerPool
 
     // Check if it is pruned query, execute it automatically on the low latency pool
-    if ((cachedRDD ne null) && cachedRDD.getNumPartitions <= 2 /* some small number */ &&
+    if ((cachedRDD ne null) && isLowLatencyQuery &&
```
sumwale: Shouldn't the (cachedRDD ne null) check also go into isLowLatencyQuery flag above? I think there are some cases where cachedRDD can be null like insert/update/delete in which case isLowLatencyQuery should be false like here.

hbhanawat:
Create SNAP-1931, SNAP-1932

PR #783, opened 2017-08-11
overridedn two methods from Executor.scala.

rishitesh:

Overridden two methods from Executor.scala.
These methods have been added in Spark executor to check store related errors. 

## Patch testing

pre-checkin pending

## ReleaseNotes.txt changes

NA

## Other PRs 
https://github.com/SnappyDataInc/spark/pull/71


cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -108,6 +108,24 @@ class SnappyExecutor(
...
+    } catch {
+      case ex: CacheClosedException => true
+      case _ => false
+    }
+    return false;
+  }
+
+  override def isStoreException(t: Throwable): Boolean = {
+    if (GemFireXDUtils.retryToBeDone(t)) {
+      return true;
```
sumwale: just return "GemFireXDUtils.retryToBeDone(t)" and doesn't need another if-then check

cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -108,6 +108,24 @@ class SnappyExecutor(
...
   }
+
+  override def isStoreCloseException(t : Throwable) : Boolean = {
+    try {
+      Misc.checkIfCacheClosing(t)
+    } catch {
+      case ex: CacheClosedException => true
+      case _ => false
+    }
+    return false;
```
sumwale: Instead of "return false" the false can be moved below Misc.checkIfCacheClosing:
```scala
try {
  Misc.checkIfCacheClosing(t)
  false
} ...
```

cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -108,6 +108,24 @@ class SnappyExecutor(
       }
     }
   }
+
+  override def isStoreCloseException(t : Throwable) : Boolean = {
+    try {
+      Misc.checkIfCacheClosing(t)
+    } catch {
+      case ex: CacheClosedException => true
```
sumwale: This should catch the base "CancelException" and not the specific CacheClosedException.

cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -108,6 +108,28 @@ class SnappyExecutor(
...
+    GemFireXDUtils.retryToBeDone(t)
+  }
+
+  override def isFatalError(t: Throwable): Boolean = {
+    if (t.isInstanceOf[Error]) {
+      SystemFailure.isJVMFailureError(t.asInstanceOf[Error])
+    } else {
+      false
+    }
+  }
```
sumwale: pattern matching will be cleaner:

```scala
t match {
  case err: Error => SystemFailure.isJVMFailureError(err)
  case _ => false
}
```

cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
kneeraj: @sumwale great. This seems to be the reason for SNAP-1377. Because the ExecutorLostFailure exception got missed due to the check of only CacheClosedException.

cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
rishitesh: Ok. Will change it.

PR #786, opened 2017-08-14
[SNAP-1915] Register task listener as soon as conn is taken

suranjan:

Register task listener as soon as conn is taken
## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnBatch.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnBatch.scala
@@ -112,13 +112,6 @@ abstract class ResultSetIterator[A](conn: Connection,
     } catch {
       case NonFatal(e) => logWarning("Exception closing statement", e)
     }
-    try {
-      conn.commit()
-      conn.close()
-      logDebug("closed connection for task " + context.partitionId())
-    } catch {
-      case NonFatal(e) => logWarning("Exception closing connection", e)
-    }
     hasNextValue = false
   }
 }
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
@@ -792,10 +792,23 @@ class SmartConnectorRowRDD(_session: SnappySession,
   }
 
   override def computeResultSet(
-      thePart: Partition): (Connection, Statement, ResultSet) = {
+      thePart: Partition, context: TaskContext): (Connection, Statement, ResultSet) = {
     val helper = new SparkShellRDDHelper
     val conn: Connection = helper.getConnection(
       connProperties, thePart)
+    if (context ne null) {
+      val partitionId = context.partitionId()
+      context.addTaskCompletionListener { _ =>
+        logDebug(s"closed connection for task from listener $partitionId")
+        try {
+          conn.commit()
+          conn.close()
+          logDebug("closed connection for task " + context.partitionId())
+        } catch {
+          case NonFatal(e) => logWarning("Exception closing connection", e)
+        }
+      }
+    }
     if (isPartitioned) {
       val ps = conn.prepareStatement(
         s"call sys.SET_BUCKETS_FOR_LOCAL_EXECUTION(?, ?, ${_relDestroyVersion})")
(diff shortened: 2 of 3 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnBatch.scala
```diff
@@ -112,13 +112,6 @@ abstract class ResultSetIterator[A](conn: Connection,
...
       case NonFatal(e) => logWarning("Exception closing statement", e)
     }
-    try {
-      conn.commit()
-      conn.close()
-      logDebug("closed connection for task " + context.partitionId())
-    } catch {
-      case NonFatal(e) => logWarning("Exception closing connection", e)
-    }
     hasNextValue = false
```
vivekwiz: If you remove this how connection will be closed after usage of ColumnBatchIteratorOnRS?

suranjan: Fixed that.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -792,10 +801,23 @@ class SmartConnectorRowRDD(_session: SnappySession,
...
+        logDebug(s"closed connection for task from listener $partitionId")
+        try {
+          conn.commit()
+          conn.close()
+          logDebug("closed connection for task " + context.partitionId())
+        } catch {
+          case NonFatal(e) => logWarning("Exception closing connection", e)
+        }
+      }
+    }
```
vivekwiz: What if context is null and conn is not  null? How conn will be closed in this scenario?

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatScanRDD.scala
```diff
@@ -169,9 +170,24 @@ class RowFormatScanRDD(@transient val session: SnappySession,
...
+        try {
+          conn.commit()
+          conn.close()
+          logDebug("closed connection for task " + context.partitionId())
+        } catch {
+          case NonFatal(e) => logWarning("Exception closing connection", e)
+        }
+      }
+    }
+
```
vivekwiz: Same as above

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
suranjan: Context will not be null. In fact null check can be removed.

sumwale: Actually a good point and need to think a bit more about this. Context is null for cases when plans are prepared on executors themselves e.g. for rollover. In that case a local plan is created and executed like in ColumnBatchCreator.createAndStoreBatch where it does not create a temporary TaskContext. Perhaps a temporary taskContext can be created for those cases or need to find some other way to cleanup.

suranjan: Does this plan get created on executor. I am asking because before this change too, the close was through taskContext only in ResultSetIterator. 
This change is moving it ahead in queue.

sumwale: @suranjan Oh, didn't notice this is the smart connector RDD. That will not be created on executor. However, in general the comment about relying only on context.addTaskCompletionListener for snapshot cleanup/commit remains. The ColumnInsertExec plan, for example, will be invoked on executor and cleanup if being done only via addTaskCompletionListener is not enough since context will be null for that case. It has to be done explicitly in ColumnBatchCreator in the two places it creates the ColumnInsertExec plan explicitly.

sumwale: @suranjan just asking to review the snapshot cleanup code of TaskCompletionListener to be handled for context==null cases too (whenever ExternalStoreUtils.codeGenOnExecutor is being invoked).

suranjan: @sumwale ColumnInsertExec doesn't depend on TaskContext. We get a PoolConnection/EmbedConnection and close it ourselves. Similarly for tx commit/rollback.

To iterate over rowbuffer, we use embed connection and don't do  any commit/rollback. It just inherits the txState of thread which was doing put into rowbuffer.

This change looks to be safe.

PR #787, opened 2017-08-14
[SNAP-1877] GC issues with large dictionaries in decoding and other optimizations

sumwale:

- Now creating a long[] containing the offsets of string objects in the dictionary instead
  of populating with UTF8String upfront. This reduces number of initial objects down to one.
- Create the UTF8String on-the-fly from the offsets for large dictionaries.
- Use the UTF8String[] as before for small dictionaries with <= 1000 elements to restore
  previous performance for other cases.
- Fixed an issue of missing nextArray/Map/Struct in ResultSetDecoder. UnsafeRowDecoder is
  also missing those but it is obsolete and will be removed.
- removing unnecessary member statistics logInfo that keeps logging continuously

Data skew fixes:

- Use a power of 2 number of partitions rather than the GemFire convention of prime number
  because that is what Spark hashing is optimized for
- Put an upper limit of 500K rows in a column batch. This is to ensure that the skew has an
  upper limit of 500K. This does not effect most real-world cases since the default 24M limit
  will be hit much before, but helps in some artificial cases having one or two columns.

Different sessions end up creating different code due to indeterminate statsPredicate
ordering. Now using a consistent sort order so that generated code is identical across sessions
for the same query.

Reducing amount of generated code:

- removing the tri-state isNull which was added due to ResultSetDecoder; instead
  relying on pre-read isNull call using the new "ResultSetWithNull" interface which
  is implemented by both EmbedResultSet and thrift ClientResultSet
- removing the code that would always get the dictionary, dictionary index and then read
  the string with null checks throughout; now using the normal readUTF8String where
  required while the dictionary optimization code-path now specifically reads
  the dictionary and indexes when required and only for the required columns (i.e.
      the single key join/aggregate column for now)

TAQTest Q1 is same best performance but slightly more consistent and gives ~5% better
average (due to data skew fixes)

## Patch testing

precheckin; customer use-cases that showed the GC issues with large dictionaries

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/275


sumwale:
Merging now since this blocks SNAP-1936. Will take care of any review comments subsequently.

PR #788, opened 2017-08-14
Set user credentials into snappysession in case of prepared SQL

ashetkar:
Set user credentials into snappysession in case of prepared SQL

## Patch testing
Manual testing and security unit tests

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


ashetkar:
>> LGTM. Do you have testcase that can be added as test? Good to have.
Will add a prepared sql in the existing unit test. But I could not reproduce the issue in the unit test though this fixes the issue seen with ODBC clients.

ashetkar:
Added the test case.

PR #793, opened 2017-08-16
[SNAP-1936] use indexed cursors instead of sequential for column decoders

sumwale:

- Removed ColumnDecoder.next* methods used by sequential cursor updates.
  Instead store the baseCursor in the decoder and use it to determine the
  current cursor position for the ordinal.
- Accounting of null values is done using a "numNulls" value that is maintained
  in generated code. A new method "numNulls" in the decoders accepts the previous
  "numNulls" value and then checks if current ordinal has skipped too far
  (in which case it will recalculate the nulls for current ordinal).
- Variable length values and some decoders (like BooleanBitSetDecoder)
  that cannot jump using the ordinal use custom ways to handle it.
  Variable length ones maintain a separate last cursor position and skip
  positions reading each value sequentially if required. BooleanBitSet
  will recalculate the current word if there is a skip in position.
- Removed ColumnDecoder.absolute* methods which were for delta encoder not required now
  and just the required position is calculated using "numNullsUntilPosition" method.
- Cleaned up wide-schema handling a bit to remove unused local variables.
  Removed "moveNext" split and handling now that next* methods are gone.
- Initialize the decoder in constructor itself now and initialize the
  baseCursor to be a final val to get a small performance benefit over var.
  Also reduces generated code by removing the explicit initialize calls.
- Removed unused IntDelta and LongDelta encoders to avoid maintenance
  overhead of having to update those with any changes to ColumnDecoder.
- Moved "numNullsUntilPosition" code to BitSet.cardinality and added a test
  for the same to BitSetTest.

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/Uncompressed.scala
```diff
@@ -34,256 +33,165 @@ trait Uncompressed extends ColumnEncoding {
...
+
+  private def setCursorForVariableWidth(columnBytes: AnyRef, nonNullPosition: Int,
+      sizeWidth: Int = 4): Unit = {
+    // check sequential calls else skip as much required
+    var lastPosition = lastNonNullPosition
+    if (nonNullPosition != lastPosition + 1) {
+      var cursor = currentCursor
+      lastPosition += 1
+      do {
+        cursor += sizeWidth + ColumnEncoding.readInt(columnBytes, cursor)
```
hbhanawat: For most of the types, the length can be 2 bytes i.e. 32768. Atleast for string and non-binary types we can put this limit to compress the data further.

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/DictionaryEncoding.scala
```diff
@@ -106,96 +118,46 @@ abstract class DictionaryDecoderBase
...
-    // in the dictionary array lazily on access
-    baseCursor + ((position - numNullsUntilPosition(columnBytes, position)) << 1)
-  }
-
-  override def readUTF8String(columnBytes: AnyRef, cursor: Long): UTF8String =
-    stringDictionary.getString(columnBytes, ColumnEncoding.readShort(columnBytes, cursor))
+  // TODO: PERF: optimize for local index access case by filling
+  // in the dictionary array lazily on access
+  override def readUTF8String(columnBytes: AnyRef, nonNullPosition: Int): UTF8String =
+  stringDictionary.getString(columnBytes,
```
hbhanawat: For a join query on string column with a big dictionary (say when String is kind of primary key), what is the overhead of creation of objects in every call? Since we are trying to solve the problem of promotion of dictionary objects, another approach could be to divide the dictionary into multiple sub dictionaries. Say for e.g. after every 1000 entries, rollover to a new dictionary. Each dictionary has a start and end ordinals of columns which can be used by the decoder. Alternatively, we can put the dictionary id in addition to dictionary index inside every entry. This approach could be generic so that it can be extended when we move to a mechanism of having a common dictionary across cached batches. Theoretically, the queries should have little impact.

rishitesh: @hbhanawat How it does matter, if we create string at scan time or on a upper operator like join. The cost should be same as I understood. However in this approach if we can push some filters( filter on string columns can be looked up on a dictionary and discard the batch altogether)  to dictionary then probably we don’t have to create a lot of objects.

sumwale: > what is the overhead of creation of objects in every call?

If the column is mostly unique then object will be created at most once or not at all if there are pushed down filters. Multiple creation cases will be the in-between ones say 10K unique in total of 100K strings.

> Say for e.g. after every 1000 entries, rollover to a new dictionary.

How will this help in reducing the lifetime of objects? We will have to keep all the dictionaries alive for the whole duration of batch processing, no? Perhaps some reduction in first creation if we populate dictionaries lazily as I have mentioned in a TODO in the code. Is that what you have in mind and will be enough for the case mentioned above (10K dictionary items for total of 100K)?

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/Uncompressed.scala
sumwale: Good idea. Can even use varInt encoding like in protobuf/kryo which will use 1-5 bytes.

sumwale: Filed https://jira.snappydata.io/browse/SNAP-1968

PR #795, opened 2017-08-17
[SNAP-1935] Added a test for SnapshotGII.

rishitesh:

Added a test for SnapshotGII.

## Patch testing

precheckin pending. 

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/280


cluster/src/dunit/scala/org/apache/spark/transaction/SnapshotGIIDUnitTest.scala
```diff
@@ -0,0 +1,66 @@
...
+    vm1.invoke(classOf[ClusterManagerTestBase], "stopAny")
+
+    val data = for (i <- 1 to 50) yield (Seq(i, (i + 1), (i + 2)))
+    val rdd = snc.sparkContext.parallelize(data, 2).map(s =>
+      Data(s(0), s(1), s(2)))
+    val dataDF = snc.createDataFrame(rdd)
+    dataDF.write.insertInto(tableName)
+    vm1.invoke(restartServer(props))
+    dataDF.write.insertInto(tableName)
+    val numRows = snc.sql(s"select * from $tableName").collect().length
```
suranjan: To actually assert that GII took place in vm1, we should bring vm2 and vm3 down. Otherwise looks ok.

rishitesh: This test is only to ensure all the locks taken during GII are behaving properly. We faced issued from snappy layer when locks were not getting released. Hence added a simple test. 
A more detailed test of SnapshotGII is on SnapshotTxGIIDUnit.

PR #796, opened 2017-08-18
[WIP] Fix for SNAP-1851 to properly closing connection in case when conn co…

SachinJanani:
…mmit fails

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStore.scala
```diff
@@ -62,16 +62,19 @@ trait ExternalStore extends Serializable {
       onExecutor: Boolean = false)(f: Connection => T)
       (implicit c: Option[Connection] = None): T = {
     var success = false
-    val conn = c.getOrElse(getConnection(tableName, onExecutor))
+    var conn = c.getOrElse(getConnection(tableName, onExecutor))
```
sumwale: why var?

PR #797, opened 2017-08-18
[SNAP-1925][SNAP-1926] smart connector update/delete and other fixes

sumwale:
Apart from fixes for update/delete in smart connector mode, this includes fixes for other issues seen in testing including incorrect call to GET_SNAPSHOT_TXID in closeConnection, ColumnBatch properties not being honored in connector mode, "number of output rows" metric updated to reduce the deleted rows.

Also pushing an optimization to ParamLiteral which is quite safe. Now generated code does the cast+unboxing once in init instead of every time in the loop. Also refactored common code simplifying it.

- Fix for two variations of SNAP-1925:
  - Getting ROWID for row buffer in smart connector uses projections so the ROWID
    column should use output length instead of full schema length
  - Sometimes the columns in output are jumbled (only seen in smart connector)
    with virtual ROWID, BATCHID columns coming in between rather at the end as
    the plan actually injects. So the index of real columns have to be counted
    separately skipping the virtual columns rather than using output index.
- Correct the index passed to ColumnDelta.deltaColumnIndex which is 0-based
  and added comments to clarify the same.
- Fix ColumnBatchSize, ColumnMaxDeltaRows not being honoured in smart connector.
  Reason was that these two properties are stored as table property during CREATE
  for use by row buffer rollover thus overriding current session setting. Thin catalog
  fails to set session conf in CREATE since it uses a procedure call on embedded lead.
  Now setting a "TRANSIENT" property with those indicating whether the table property
  came from DDL or the session conf during CREATE, and skip using table property when its set.
- scalatest, dunits for SNAP-1925/1926 for embedded and connector modes
- Add tokenization for UPDATE set portions too to reduce code compilation
- Refactored common code in ParamLiteral and made it more efficient by evaluating
  cast+unboxing once in init() rather than everytime in the loop.
  Also optimized its kryo serialization to use Struct serializer rather than json conversion
- GET_SNAPSHOT_TXID call in closeConnection was buggy because it is an output parameter
  that needs to be set as parameter rather than setting the value. Somehow never hit before
  but gets hit in the new tests.
- Account for deleted rows in the "number of output rows" statistic.
  Initialize buffers in smart connector now in delete count lookup because that
  happens before any getColumnLob with this change.
- Correct last position cursor when merging multiple deltas (code re-reads the last position
    again intentionally to keep code simpler for the tiny cost of one int read again)
- Removed delete segments in JDBCSourceAsColumnarStore insert/put blocks supposed
  to be for rollback which has never been tested and not required with snapshot isolation
- Start two servers in parallel in ClusterManagerTestBase for faster dunits.
- Add dummy table lookup call to pre-initialize some parts of hive metastore
  (and avoid doing on first use where all servers try to initialize in parallel).
- Fixed a scalastyle issue in ExternalStore. Corrected an IDEA issue with SnappySharedState.

## Patch testing

precheckin with new tests

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -254,9 +255,9 @@ class JDBCSourceAsColumnarStore(private var _connProperties: ConnectionPropertie
             // or it will be committed later
             val txId = SparkShellRDDHelper.snapshotTxIdForWrite.get
             if (txId != null && !txId.equals("null")) {
-              val statement = conn.createStatement()
-              statement.execute(
-                s"call sys.GET_SNAPSHOT_TXID('$txId')")
+              val statement = conn.prepareCall("call sys.GET_SNAPSHOT_TXID(?)")
+              statement.registerOutParameter(1, java.sql.Types.VARCHAR)
+              statement.execute()
```
suranjan: Did you see any issue? I am asking because the returned value was not used. Unless the procedure itself is not executed on server due to some exception?, there shouldn't have any issue.

cluster/src/dunit/scala/io/snappydata/cluster/ClusterManagerTestBase.scala
```diff
@@ -121,7 +121,8 @@ abstract class ClusterManagerTestBase(s: String)
       }
     }
 
-    Array(vm0, vm1, vm2).foreach(_.invoke(startNode))
+    vm0.invoke(startNode)
+    Array(vm1, vm2).map(_.invokeAsync(startNode)).foreach(_.getResult)
```
kneeraj: Small things but great improvement to save time. Just curious, overall how much time it is saving in dunits approx. 10 to 20 minutes or less?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
sumwale: Yes it fails in execution with SQLState.LANG_DB2_PARAMETER_NEEDS_MARKER (42886) saying that parameter should be a marker '?' and not a value. The problem is that regardless of whether output value is used or not, the implementation of GfxdSystemProcedures.GET_SNAPSHOT_TXID needs an array which it will fill out that will only happen if registerOutParameter has been called.

If you would have registered GET_SNAPSHOT_TXID as a zero-arg function returning a string rather than a procedure with output parameter, then could have invoked it with "VALUES GET_SNAPSHOT_TXID()" which would have been cleaner.

cluster/src/dunit/scala/io/snappydata/cluster/ClusterManagerTestBase.scala
sumwale: Didn't note the total precheckin saving. But saw good improvement when running a single dunit.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -88,8 +88,6 @@ class SnappyUnifiedMemoryManager private[memory](
 
   private val maxHeapStorageSize = (maxHeapMemory * evictionFraction).toLong
 
-  private val maxHeapExecutionSize = (maxHeapMemory * evictionFraction).toLong
```
sumwale: Yes unused and incorrect (can't be same as maxHeapStorageSize).

cluster/src/test/scala/org/apache/spark/sql/store/ColumnUpdateDeleteTest.scala
```diff
@@ -35,11 +49,20 @@ class ColumnUpdateDeleteTest extends ColumnTablesTestBase {
...
+  test("SNAP-1926") {
+    ColumnUpdateDeleteTest.testSNAP1926(this.snc.snappySession)
+  }
+
   ignore("test update for all types") {
     val session = this.snc.snappySession
     // reduced size to ensure both column table and row buffer have data
     session.conf.set(Property.ColumnBatchSize.name, "100k")
     runAllTypesTest(session)
+    session.conf.unset(Property.ColumnBatchSize.name)
```
sumwale: This is for the case when it is passed the common "this.snc.snappySession" which subsequent tests can re-use. The tests that don't set any explicit ColumnBatchSize intend to use the default one and the unset is for those tests. For tests that need to use a non-default value will set it explicitly in any case.

cluster/src/test/scala/org/apache/spark/sql/store/ColumnUpdateDeleteTest.scala
```diff
@@ -275,4 +302,86 @@ object ColumnUpdateDeleteTest extends Assertions {
...
+    assert(result.toSeq.filter(_.getDouble(0) != 1.1) === Seq.empty)
+    result = session.sql("select UnitPrice, tid from order_details").collect()
+    assert(result.length === numElements)
+    assert(result.toSeq.filter(_.getDouble(0) != 1.1) === Seq.empty)
+
+    session.sql("drop table order_details")
+    session.conf.unset(Property.ColumnBatchSize.name)
+  }
+
+  def testSNAP1926(session: SnappySession): Unit = {
```
sumwale: The NPE in SNAP-1926 is not coming from code but from the Janino compilation. This test creates the exact same generated code that has been reported by Sonal to be failing with NPE in Janino. However it does not reproduce in this test so I am guessing that it is some race condition in the Janino compiler itself that manifests under stress conditions or an issue with older janino versions (though could not find any fix for this in its changelogs).

Since it is reported only in smart connector and upstream Spark uses an older version (3.0.0) compared to Snappy (3.0.6) so a bug in old Janino is the likely theory. I am hoping that with the changes in SNAP-1936 that significantly changes/reduces the generated code this will go away automatically, but if not we can recommend that Spark connector-mode users replace the janino 3.0.0 jar with the latest release 3.0.7 jar in their Spark product installation (and will ask Sonal to verify that the issue goes away with that).

sumwale:
@kneeraj Thanks. Did you look through the changes in ParamLiteral?

The change in ColumnDeltaEncoder is just a two-liner that moves position cursor back since it re-reads the current position. Its merging two sorted deltas and this code is for the copying remaining ones from the larger one. The cursor has been already incremented when the last position from larger one was read, but this code re-reads the last position to keep the code simpler so should decrement the cursor (else code will have to do separate assignment from already read position in first loop and skip cursor).

kneeraj:
Yes I did look a the ParamLiteral code and it looked fine to me.
Thanks for the explanation of the merge part.

cluster/src/dunit/scala/io/snappydata/cluster/ClusterManagerTestBase.scala
ymahajan: @sumwale  yes, this will be a good time saver. Just a question - why first vm0 started synchronously while reaming asynchronously ?  why not all three vms started  asynchronously?

ymahajan:
@sumwale  PramaLiteral code is much simplified and less error prone now from the maintenance point of view.

cluster/src/dunit/scala/io/snappydata/cluster/ClusterManagerTestBase.scala
sumwale: @ymahajan starting all three asynchronously causes occasional lock timeouts on indexes/tables for the hive meta-store resulting in startup failure. Still need to look at the cause for that (since a separate dlock is acquired before hive meta-store initialization). For that reason I also had to do the same for snappy-start-all.sh script which now starts all servers in parallel except for the first one in a multi-node cluster (single localhost startup is still sequential as before). Will look into it when I get a chance.

PR #798, opened 2017-08-20
SNAP-1660 Remove password from product logging

ashetkar:
SNAP-1660 Remove password from product logging
Also, add enterprise checks before loading AQP and Off-heap modules.

## Patch testing
Manually tested.

## ReleaseNotes.txt changes
NA

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/283

### DIFF
--- a/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
@@ -26,26 +26,23 @@ import scala.collection.JavaConverters._
 import akka.actor.ActorSystem
 import com.gemstone.gemfire.distributed.internal.DistributionConfig
 import com.gemstone.gemfire.distributed.internal.locks.{DLockService, DistributedMemberLock}
-import com.gemstone.gemfire.internal.GemFireVersion
 import com.gemstone.gemfire.internal.cache.GemFireCacheImpl
 import com.pivotal.gemfirexd.FabricService.State
 import com.pivotal.gemfirexd.internal.engine.{GfxdConstants, Misc}
 import com.pivotal.gemfirexd.internal.engine.db.FabricDatabase
 import com.pivotal.gemfirexd.internal.engine.distributed.utils.GemFireXDUtils
 import com.pivotal.gemfirexd.internal.engine.store.ServerGroupUtils
-import com.pivotal.gemfirexd.internal.shared.common.SharedUtils
 import com.pivotal.gemfirexd.internal.shared.common.sanity.SanityManager
 import com.pivotal.gemfirexd.{Attribute, FabricService, NetworkInterface}
 import com.typesafe.config.{Config, ConfigFactory}
 import io.snappydata._
 import io.snappydata.cluster.ExecutorInitiator
-import io.snappydata.gemxd.SnappyDataVersion
 import io.snappydata.util.ServiceUtils
 import org.apache.thrift.transport.TTransportException
 import spark.jobserver.JobServer
-import org.apache.spark.sql.SnappyContext
+import org.apache.spark.sql.{SnappyContext, SnappySession}
 import org.apache.spark.sql.collection.Utils
-import org.apache.spark.{SparkCallbacks, Logging, SparkConf, SparkContext, SparkException}
+import org.apache.spark.{Logging, SparkCallbacks, SparkConf, SparkContext, SparkException}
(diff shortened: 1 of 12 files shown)


cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -197,14 +197,20 @@ class LeadImpl extends ServerImpl with Lead
...
+      password.get)
 
     val confProps = conf.getAll
     val storeProps = ServiceUtils.getStoreProperties(confProps)
     checkAuthProvider(storeProps)
 
+    val pass = storeProps.remove(Attribute.PASSWORD_ATTR)
     logInfo("passing store properties as " + storeProps)
+    if (pass != null) storeProps.setProperty(Attribute.PASSWORD_ATTR, pass.asInstanceOf[String])
     super.start(storeProps, ignoreIfStarted = false)
```
sumwale: Was absence of this causing trouble in some existing tests? Perhaps will be good to add a test for this.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
```diff
@@ -320,6 +320,9 @@ object StoreCallbacksImpl extends StoreCallbacks with Logging with Serializable
 
   override def shouldStopRecovery(): Boolean =
     MemoryManagerCallback.memoryManager.shouldStopRecovery()
+
+  override def isEnterpriseEdition: Boolean =
+    SnappySession.isEnterpriseEdition
 }
```
sumwale: Not required -- see store PR comment.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1747,6 +1748,12 @@ object SnappySession extends Logging {
   private[this] val ID = new AtomicInteger(0)
   private[sql] val ExecutionKey = "EXECUTION"
 
+  lazy val isEnterpriseEdition: Boolean = {
+    GemFireVersion.getInstance(Class.forName("io.snappydata.gemxd.SnappyDataVersion"),
```
sumwale: This is the same as that added in SnappyDataVersion.isEnterpriseEdition. Please make that as lazy val and use it removing this.

ashetkar: Yes, but that is in snappy-cluster module which snappy-core cannot access, right?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
ashetkar: This was added to enable GemFireCacheImpl check for enterprise edition. It cannot access SnappyDataVersion.isEnterpriseEdition as it is in snappy-cluster and the ClusterCallbacks is available in gemfirexd module which gemfire module cannot access.

So added the check in SnappySession using reflection which this storecallback (it should really be renamed to CoreCallbacks similar to ClusterCallbacks) uses.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
sumwale: Ok, then remove SnappyDataVersion.isEnterpriseEdition and use SnappySession one throughout.

Also don't quite understand why the SnappyDataVersion class needs to be passed to GemFireVersion.getInstance(). Since this is the regular GemFireXDVersion.properties, the class used should be GemFireXDVersion for which getResourceAsStream should pass fine. Anyway the fact that SnappyDataVersion.properties is created but used only for few things while GemFireXDVersion.properties used for other things makes this whole thing quite messy. Needs a good cleanup later.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
sumwale: See the last store PR comment. It is not required since GemFireVersion is a singleton.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
ashetkar: Yes, we need to better handle version properties. Will track it in Jira.
Done the changes suggested in review. Please review.

PR #802, opened 2017-08-21
Fix for a put dml syntax variation

kneeraj:

Refactored the parser to accomodate for both tablevaluedfunctionExpressions coming via select and put dml variation syntax where you can specify column groups with the table name.

For eg.

put into tname1(col1, col2) select * from tname2

The above was throwing exception as the parser was thinkning it of as a table valued function. 

## Patch testing

Manual. Adding a test for this.

## ReleaseNotes.txt changes

None.

## Other PRs 

No.


core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -699,6 +702,16 @@ class SnappyParser(session: SnappySession)
     )
   }
 
+  protected final def relationPut: Rule1[LogicalPlan] = rule {
+    relationFactor ~ (
+        join. + ~> ((r1: LogicalPlan, joins: Any) => joins.asInstanceOf[
```
vivekwiz: Cannot be this part and parts from relation (line 697) be a common rule? Just in case, though heard you saying difficulty in code refactoring.

sumwale: In fact I think the join portion should be removed. A putInto will always be onto a single relation and doesn't make sense to have a join here (the source Dataset being put can have joins of course).

sumwale:
Also you would need to add tableValuedFunction rule to the join rule that uses relationFactor. Perhaps you can add a separate rule "relationWithExternal" that combines relationFactor and tableValuedFunction and use in both relation and join rules.

PR #803, opened 2017-08-22
SNAP-1947

ymahajan:

- Made EVICT_ACTION overflow_to_disk as default when EVICTION_BY clause is given.
- Added tests for various combinations

## Patch testing

precheckin 

## ReleaseNotes.txt changes

Yes

## Other PRs 

None


rishitesh:
The change will be applicable for row & column tables both ? or only column table ?

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -400,16 +400,11 @@ object StoreUtils {
...
+          s"$GEM_EVICTION_BY $v $GEM_OVERFLOW "
         }
       }).getOrElse(defaultEviction))
     }
 
-    if (hasOverflow) {
-      parameters.remove(OVERFLOW)
-      sb.append(s"$GEM_OVERFLOW ")
-    }
-
```
sumwale: What if user explicitly sets overflow option as false? This change will not honor that option. I think the better way will be to set hasOverflow getOrElse() block to be always true. Then note when the EVICTION_BY=NONE case (that appends EMPTY_STRING) with some boolean flag, then add the GEM_OVERFLOW in this block if its not the EVICTION_BY=NONE case.

core/src/test/scala/org/apache/spark/sql/store/ColumnTableTest.scala
```diff
@@ -625,30 +625,119 @@ class ColumnTableTest
...
+
+    snc.sql("CREATE TABLE COLUMN_TEST_TABLE6(OrderId INT ,ItemId INT) USING row options" +
+      " (PARTITION_BY 'OrderId', EVICTION_BY 'LRUMEMSIZE 200', OVERFLOW 'false')")
+    region = Misc.getRegionForTable("APP.COLUMN_TEST_TABLE6", true)
+      .asInstanceOf[PartitionedRegion]
+    assert(region.getEvictionAttributes.getAlgorithm ===
+      EvictionAlgorithm.LRU_MEMORY)
+    assert(region.getEvictionAttributes.getAction ===
+      EvictionAction.OVERFLOW_TO_DISK)
+    assert(region.getEvictionAttributes.getMaximum === 200)
```
sumwale: When overflow has been explicitly given as false, then the action here shouldn't be OVERFLOW_TO_DISK rather DESTROY. See comment in the code above.

ymahajan:
@rishitesh Yes, the change is applicable to both row and column tables. @sumwale The Overflow false case now has local-destroy evictaction

PR #804, opened 2017-08-22
SNAP-1929 INSERT in row table requires SELECT permission in SnappyJob

ashetkar:
* Fix for SNAP-1929 INSERT in row table requires SELECT permission in SnappyJob, by skipping the the auth for internal select query fired to get schema in JDBCMutableRelation
* Also, updated the snappysession and config instances correctly for snappy streaming and java jobs.

## Patch testing
Added a dunit test. prechekin in-progress.

## ReleaseNotes.txt changes
NA

## Other PRs 
NA

### DIFF
--- a/cluster/src/dunit/scala/io/snappydata/cluster/SnappySecureJob.scala
+++ b/cluster/src/dunit/scala/io/snappydata/cluster/SnappySecureJob.scala
@@ -1,0 +1,238 @@
+package io.snappydata.cluster
+
+import java.io.{FileOutputStream, PrintWriter}
+
+import com.pivotal.gemfirexd.Attribute
+import com.typesafe.config.{Config, ConfigException}
+import io.snappydata.Constant
+import org.apache.spark.sql.types.{DecimalType, IntegerType, StructField, StructType}
+import org.apache.spark.sql._
+
+class SnappySecureJob extends SnappySQLJob {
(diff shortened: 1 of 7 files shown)


sumwale:
Confused by this. Why is this an issue? What is the case where we expect that users have INSERT permission but not the SELECT permission on a table? If anything, we should disallow that combination in the first place.

ashetkar:
> Confused by this. Why is this an issue? 

Yes, there may not be a real life use case for this. This started primarily because, during testing we found that, RowStore (gemxd) mode has this behaviour. i.e. only update/delete on a table require select permission on the table and insert does fine without select.

Another thinking was that for update or delete, one would need select to decide which rows to update or delete but not so for insert.

Do you suggest to keep it as it is and also fix the behaviour in RowStore mode?

sumwale:
Yes update/delete will need SELECT permission since they will fire a separate query first to determine the qualified rows. Rowstore does not need because it opens an internal iterator with "update=true". I don't think there is anything to fix here or in rowstore mode. Simply document that for users to update/delete a table both INSERT and SELECT permissions are required. We could later even add SELECT permission automatically for a user when INSERT has been provided.

ashetkar:
Closing this PR. Retaining the JIRA ticket for further action.
Will file another PR for other changes in this PR not really related to the issue.

PR #805, opened 2017-08-22
[SNAP-1955] fixes for issues seen in parallel runs

sumwale:
SnappyData side of changes for the issues noted in SNAP-1955.

- Added a common method to setup meta-store properties in
  SnappyHiveCatalog.setCommonHiveMetastoreProperties used by both SnappyHiveCatalog
  and SnappyStoreHiveCatalog. Use BONECP pool with properties to avoid its background
  activity and a pool-size enough to accomodate multiple connection calls from
  a single hive meta-data calls (the reason BONECP was failing for SnappyStoreHiveCatalog
      and was changed to DBCP was due to its default pool max size being only 2)
- Removed system property setting for HiveConf properties which was causing trouble
  with integer property settings added above. It was unnecessary in the first place.
- Set internal-connection property as true for all pool connections so that those never
  try to boot up server if it is going down, for example, causing deadlocks.
- Made skipLocks=false in SnappyHiveCatalog init because it should have the
  DataDictionary lock when creating initial meta-tables else DDLs can get missed
  in a new node replay.
- Removed ServerStartSuite and LeaderLauncherSuite that leave around unwanted
  artifacts (e.g. locator artifacts in unit tests expecting loner JVMs) and cause
  trouble especially in parallel runs. These are covered much more comprehensively
  in dunit/hydra tests so are redundant and we are not supporting embedding cluster
  using API in any case.
- Some dunit test changes to move cleanup/stop/restart in finally blocks so
  that a failure does not lead to cascading failures of all other tests.
  Disabled DistributedIndexDUnitTest in a better way to skip its beforeClass().
- Added build output logging into buildOutput.log. This does not have output
  from smoke tests and example runs that are redirecting output explicitly
  to /dev/tty for some reason rather than into output.txt as other tests do.

## Patch testing

precheckin with and without parallel tasks setup

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/287


sumwale:
Have skipped the changes to actually do parallel precheckin runs in the PR since it has some complications (and is still showing a couple of issues that I have to yet to fix).

kneeraj:
I remember explicitly that we had to add skipLocks as it was causing hangs. Was that unnecessary?

cluster/src/dunit/scala/io/snappydata/cluster/ClusterMgrDUnitTest.scala
```diff
@@ -44,11 +44,15 @@ class ClusterMgrDUnitTest(s: String) extends ClusterManagerTestBase(s) {
     // connect with this new lead.
     // In this case servers are already running and a lead comes
     // and join
-    vm3.invoke(getClass, "startSnappyLead", startArgs)
-    vm3.invoke(getClass, "startSparkJob")
-    vm3.invoke(getClass, "startGemJob")
-    vm3.invoke(getClass, "stopSpark")
-    ClusterManagerTestBase.startSnappyLead(ClusterManagerTestBase.locatorPort, bootProps)
+    try {
+      vm3.invoke(getClass, "stopAny")
```
kneeraj: Please explain the change in this test as to why was this required.

sumwale:
No it is required for lookup calls. I think we removed for INIT too thinking that the DLock will protect it anyways, but that was a mistake since without DD write lock a node doing replay can miss DDLs from its queue (it takes a DD read lock to avoid new DDLs to be added while it replays existing ones in the queue).

cluster/src/dunit/scala/io/snappydata/cluster/ClusterMgrDUnitTest.scala
sumwale: This moves the stopAny and startSnappyLead calls in the finally block. Otherwise if there is a failure in this test then vm3 remains the lead while current one is removed, while subsequent tests will try to make the current one as lead again and hang on the lead lock which is acquired by vm3.

core/src/test/resources/log4j.properties
```diff
@@ -75,6 +75,6 @@ log4j.logger.org.apache.spark.scheduler.cluster.CoarseGrainedSchedulerBackend$Dr
 log4j.logger.org.apache.spark.storage.BlockManagerInfo=WARN
 log4j.logger.org.apache.spark.executor.SnappyExecutor=WARN
 
- log4j.logger.org.apache.spark.sql.execution.WholeStageCodegenExec=DEBUG
- log4j.logger.org.apache.spark.sql.catalyst.expressions.codegen.CodeGenerator=DEBUG
- log4j.logger.org.apache.spark.sql.store.CodeGeneration=DEBUG
+# log4j.logger.org.apache.spark.sql.execution.WholeStageCodegenExec=DEBUG
+# log4j.logger.org.apache.spark.sql.catalyst.expressions.codegen.CodeGenerator=DEBUG
+# log4j.logger.org.apache.spark.sql.store.CodeGeneration=DEBUG
```
kneeraj: Any idea why this was kept ON by default? Mistake?

kneeraj:
This one looks straight and simple. Few questions asked though. Looks good to me. On store side reviews now.

core/src/test/resources/log4j.properties
sumwale: Seems so. Was removed in #718 probably during testing.

PR #809, opened 2017-08-28
[SNAP-1972] switch to dbcp-builtin pool with maxWait

sumwale:

- Switched to dbcp-builtin pool that honours maxWait to give up on connection get
  after a timeout. Expectation is that dbcp-builtin will be in a better shape than upstream DBCP
  pool which has been unmaintained for 7+ years
- Set datanucleus.connectionPool.maxWait to 30s
- Set the datanucleus.connectionPool.testSQL to VALUES(1) to force check a connection
  (and re-create) when getting. This fixes the hang noted in SNAP-1972 jira with DBCP.
- Also set the hive scratch-dir to current working directory instead of default /tmp/hive
  where user may not have write permission (some other user created and owns the directory).
  It remains unused in any case.
- Close the SnappyThinConnectorTableStatsProvider connection in stop.

Addition of testSQL also sets the "testOnBorrow" property which means an additional round-trip. This overhead should not be noticeable with catalog caching. Going forward the catalog cache will be made common (per-user) to further minimize catalog query overheads.

## Patch testing

precheckin; manual test for scenario mentioned in SNAP-1972 jira

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
+++ b/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
@@ -88,6 +88,18 @@ public class SnappyHiveCatalog implements ExternalCatalog {
     }
   }
 
+  private static String setDefaultPath(HiveConf metadataConf,
+      HiveConf.ConfVars var, String path) {
+    String pathUsed = metadataConf.get(var.varname);
+    if (pathUsed == null || pathUsed.isEmpty() ||
+        pathUsed.equals(var.getDefaultExpr())) {
+      // set the path to provided
+      pathUsed = new java.io.File(path).getAbsolutePath();
+      metadataConf.setVar(var, pathUsed);
+    }
+    return pathUsed;
+  }
+
   /**
    * Set the common hive metastore properties and also invoke
    * the static initialization for Hive with system properties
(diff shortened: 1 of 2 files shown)


core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
```diff
@@ -102,17 +114,23 @@ public static synchronized String initCommonHiveMetaStoreProperties(
...
+    //   cannot be loaded by datanucleus which should be in system CLASSPATH).
+    // Using inbuilt DBCP pool which allows setting the max time to wait
+    // for a pool connection else BoneCP hangs if network servers are down, for example,
+    // and the thrift JDBC connection fails since its default timeout is infinite.
+    // The DBCP 1.x versions are thoroughly outdated and should not be used but
+    // the expectation is that the one bundled in datanucleus will be in better shape.
+    metadataConf.setVar(HiveConf.ConfVars.METASTORE_CONNECTION_POOLING_TYPE,
+        "dbcp-builtin");
+    // set the scratch dir inside current working directory (unused but created)
+    setDefaultPath(metadataConf, HiveConf.ConfVars.SCRATCHDIR, "./hive");
```
kneeraj: Should we create a '.hive' folder inside the cwd and use that and also ./warehouse inside it. Users need to be notified of this in the docs? Any implication on backup scripts?

sumwale: These are unused directories and remain empty so no docs or backup implication. The change is to avoid issues where user may not have permissions in default locations (which is required during hive initialization).

PR #811, opened 2017-08-28
SNAP-1966 Snappy Streaming or Snappy Java jobs may not work with security

ashetkar:
Set the credentials received from user into snappy session.

## Patch testing
Added new test cases. precheckin in-progress.

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


cluster/src/dunit/scala/io/snappydata/cluster/SnappySecureJob.scala
```diff
@@ -0,0 +1,243 @@
...
+    pw.println()
+    pw.println("Delete the records for PARTKEY 400")
+    snSession.delete(rowTable, "PS_PARTKEY = 400")
+
+    pw.println(s"Printing the contents of the $rowTable table after delete")
+    tableData = snSession.sql(s"SELECT * FROM $rowTable").collect()
+    tableData.foreach(pw.println)
+
+    pw.println("****Done****")
+  }
```
sumwale: test code in product class?

cluster/src/dunit/scala/io/snappydata/cluster/SnappySecureJob.scala
```diff
@@ -0,0 +1,243 @@
...
+            // t.getStackTrace.foreach(s => pw.println(s"${t.getMessage}\n  ${s.toString}"))
+          case t: Throwable => pw.println(s"UNEXPECTED ERROR FOR $s:\n[${t.getMessage}]")
+            t.getStackTrace.foreach(s => pw.println(s"  ${s.toString}"))
+            throw t
+        }
+      }
+    }
+
+    pw.println("****Done****")
+  }
```
sumwale: more test code here

cluster/src/dunit/scala/io/snappydata/cluster/SnappySecureJob.scala
```diff
@@ -0,0 +1,243 @@
...
+    }
+    try {
+      jobConfig.getString(Constant.STORE_PROPERTY_PREFIX + com.pivotal.gemfirexd
+          .Attribute.PASSWORD_ATTR)
+      assert(false, "Boot credentials set in job config")
+    } catch {
+      case _: ConfigException.Missing => // expected
+    }
+  }
+
```
sumwale: Why separate job types for this? I don't think it is a good idea to ask users to implement separate job types. If the job configuration contains credentials then set into session else not. Jobs will automatically fail on a secure cluster when running operations that need credentials.

cluster/src/dunit/scala/io/snappydata/cluster/SnappySecureJob.scala
ashetkar: Did not understand this. This job class is under src/dunit/ so it's not a product class, though this is modified copy of CreateReplicatedRowTable job in examples folder.
May be I should add explicit verification instead of logging to a file?

cluster/src/dunit/scala/io/snappydata/cluster/SnappySecureJob.scala
ashetkar: Which job types are you referring to?
The utility method verifySessionAndConfig() (I could put it somewhere else) simply verifies that 1) the session made available to the job has the credentials set into it and 2) job config does not have boot credentials inadvertently passed to it.
This utility method is called from each of the test snappy jobs (i.e. SnappySecureJob, SnappyStreamingJob, SnappyJavaSecureJob and SnappyJavaStreamingSecureJob).

cluster/src/dunit/scala/io/snappydata/cluster/SnappySecureJob.scala
sumwale: Oh missed that this is a test class. Looked like product code.

cluster/src/dunit/scala/io/snappydata/cluster/SnappySecureJob.scala
sumwale: Confused by this. Why are these new job types under tests? Do we expect users to use security this way by implementing new job types?

cluster/src/main/java/org/apache/spark/streaming/JavaSnappyStreamingJob.java
```diff
@@ -47,7 +48,8 @@ final public Object runJob(Object sc, Config jobConfig) {
       SnappyUtils.setSessionDependencies(context.snappySession().sparkContext(),
           this.getClass().getCanonicalName(),
           Thread.currentThread().getContextClassLoader());
-      return runSnappyJob(context, SnappySessionFactory.cleanJobConfig(jobConfig));
+      return runSnappyJob(context, SnappySessionFactory.updateCredentials(context.snappySession()
+          , jobConfig));
```
sumwale: Why is this required here when code factory already sets it? I guess the fact that SnappyStreamingContext uses a singleton SnappySession is a problem in itself.

ashetkar: This does two things: 1) Sets the credentials passed while submitting the job into the snappysession and 2) removes boot credentials from job config.

> Why is this required here when code factory already sets it?

You mean SnappySessionFactory? It does it only for SnappySQLJob and JavaSnappySQLJob

cluster/src/dunit/scala/io/snappydata/cluster/SnappySecureJob.scala
ashetkar: These new jobs **are not** meant users but unit tests only. In fact, how users implement snappy jobs will not change with security.
I have added these jobs only for the unit testing purpose. I could have used the jobs already present in the examples folder but I needed to do some specific checks hence wrote these new ones.
The credentials supplied by the user while submitting his/her job will be set into the snappysession instance made available to the job. This was already taken care of, for SnappySQLJob but was missing for SnappyStreamingJob and JavaSnappyStreamingJob.

sumwale: @ashetkar if this is the only way in which credentials with jobs are being tested, then is it enough? There should be some tests that mirror how users will do it.

cluster/src/main/java/org/apache/spark/streaming/JavaSnappyStreamingJob.java
sumwale: Hmm, ok. But I still don't understand how this is supposed to work with streaming context. The SnappyStreamingContext is a singleton, and thus is its snappySession field. How setting credentials from multiple jobs into the same singleton SnappySession will work correctly? What does this mean in the first place? Only one user can submit streaming jobs, or all streaming related jobs should be done in only one job?

The whole old streamingContext model that can be started only once is a broken design to me and should perhaps be deprecated or taken out from our jobs given that the new streaming model is fairly decent now. Please discuss with the whole team.

ashetkar:
Manually tested the last changes. Merging the PR.

PR #815, opened 2017-08-30
Snap 1981

vivekwiz:
For prepare phase avoid rules that do not handle NullType since that is what is used as placeholder for parameters. Currently for NullType param-literals were getting replaced with null.

## Patch testing
Precheckin

## ReleaseNotes.txt changes
No

## Other PRs 
No


core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -69,21 +70,49 @@ class SnappySessionState(snappySession: SnappySession)
 
   private[sql] var disableStoreOptimizations : Boolean = false
 
+  lazy val analyzerPrepare: Analyzer = new Analyzer(catalog, conf) {
+
+    def getStrategy(strategy: analyzer.Strategy): Strategy = strategy match {
+      case analyzer.FixedPoint(_) => fixedPoint
```
kneeraj: Aren't we losing the maxIterations here. Did not quite understand this.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -69,21 +70,49 @@ class SnappySessionState(snappySession: SnappySession)
 
   private[sql] var disableStoreOptimizations : Boolean = false
 
+  lazy val analyzerPrepare: Analyzer = new Analyzer(catalog, conf) {
```
kneeraj: Can you please put a comment on top that what all rules you are skipping and why. Like Promotion rules etc.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -69,21 +70,49 @@ class SnappySessionState(snappySession: SnappySession)
...
+      case batch => Batch(batch.name, getStrategy(batch.strategy), batch.rules: _*)
+    }
+
+    override val extendedResolutionRules: Seq[Rule[LogicalPlan]] =
+      getExtendedResolutionRules(this)
+
+    override val extendedCheckRules = getExtendedCheckRules
+  }
+
+  def getExtendedResolutionRules(analyzer: Analyzer): Seq[Rule[LogicalPlan]] =
```
kneeraj: Just to understand. Do we require all these extended rules in the prepare phase? For example does "put into" even works with prepare statement. Do we have a test for some of the scenarios where these rules apply.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -69,21 +70,49 @@ class SnappySessionState(snappySession: SnappySession)
...
+        new FindDataSourceTable(snappySession) ::
+        DataSourceAnalysis(conf) ::
+        ResolveRelationsExtended ::
+        AnalyzeMutableOperations(snappySession, analyzer) ::
+        ResolveQueryHints(snappySession) ::
+        (if (conf.runSQLonFile) new ResolveDataSource(snappySession) ::
+            Nil else Nil)
+
+  def getExtendedCheckRules: Seq[LogicalPlan => Unit] =
+    Seq(datasources.PreWriteCheck(conf, catalog), PrePutCheck)
```
kneeraj: Even these PreWriteCheck rules. If my understanding is right these rules will be anyways applied at the execution time. In prepare phase probbaly we can avoid.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
vivekwiz: As discussed rules are around for maintainability reason. Updated comment in code.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
vivekwiz: Same as above

vivekwiz:
Created SNAP-1987

PR #816, opened 2017-08-30
[SNAP-1973][SNAP-1970] avoid clearing hive meta-store system properties

sumwale:

The hive meta-store system properties are required to be set for static
initialization of Hive and should not be cleared because a concurrent
hive context initialization (from some other path) can see inconsistencies
like system property found but not available when actually read.

Also added file path location to file not found TestException message.

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
+++ b/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
@@ -140,10 +140,6 @@ public class SnappyHiveCatalog implements ExternalCatalog {
       System.setProperty(name, props.getProperty(name));
     }
     Hive.closeCurrent();
-    // clear the system properties else it causes trouble with integer values
-    for (String name : propertyNames) {
-      System.clearProperty(name);
-    }
 
     // set integer properties after the system properties have been used by
     // Hive static initialization so that these never go into system properties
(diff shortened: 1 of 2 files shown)


sumwale:
> shouldn't the hive context be initialized only once

Static initialization happens only once, but these errors are during each individual HiveClient creation that happens for each SnappySessionState that again reads system properties.

> Also the comment about clearing system property because of integer values.

As explained in previous PRs, not setting system properties causes default derby to be booted (and metastore_db already booted kind of issues). Setting integer values in system properties causes NumberFormatException where properties are read but values are not. Can list the hive code that causes this. What exactly are you looking for in the explanation?

kneeraj:
Sumedh, this was the comment and code in your previous check-in:

// clear the system properties else it causes trouble with integer values
for (String name : propertyNames) {		
      System.clearProperty(name);		
 }

Now in this one you have removed the clearing of the property. So whatever trouble you were talking about integer values ( probably the NumberFormatException because of lack of value you have written above) will still be there.

sumwale:
> So whatever trouble you were talking about integer values

Ah ok. The comment was inaccurate. All it needs is that integer values are not set as system properties. Clearing rest of them is not necessary.

kneeraj:
ok. Thanks. Please go ahead with your merge.

PR #821, opened 2017-08-31
[SNAP-1984] Code to copy UMM state.

rishitesh:
There is a problem with the UMM when a executor goes down and comes up again. It will create a new UMM and its pools sizes will start from 0. The old state is not maintained. Also store reads SnappyUnifiedMemoryManager reference from MemoryManagerCallback which is a static field.
This means store & spark will account in two seprate UMM instance. 

To avoid that problem piggy backed on the temporary memory manager which is create when store boots up and Spark env is not ready. 
When executor goes down it will now call close() method of UMM. UMM will copy the old state to the temporary memory manager. When spark env is ready it will copy from temp manager as before.

Also, fixed one potential problem when multiple tasks fails for some reason, then executor sometimes restarts multiple times. Although its not a big issue, avoiding this in this PR

Changed ClusterMgrDunit test to fail the executors properly. 

## Patch testing

Added test. Precheckin pending. 

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
+++ b/cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
@@ -33,7 +33,7 @@ import io.snappydata.gemxd.ClusterCallbacksImpl
 
 import org.apache.spark.deploy.SparkHadoopUtil
 import org.apache.spark.executor.SnappyCoarseGrainedExecutorBackend
-import org.apache.spark.memory.SnappyUnifiedMemoryManager
+import org.apache.spark.memory.{SnappyUnifiedMemoryManager, StoreUnifiedManager}
 import org.apache.spark.sql.SnappyContext
 import org.apache.spark.sql.collection.Utils
 import org.apache.spark.{Logging, SparkCallbacks, SparkConf, SparkEnv}
@@ -57,6 +57,7 @@ object ExecutorInitiator extends Logging {
     @volatile var stopTask = false
     private var retryTask: Boolean = false
     private val lock = new Object()
+    private val testLock = new Object()
 
     val membershipListener = new MembershipListener {
       override def quorumLost(failures: util.Set[InternalDistributedMember],
@@ -81,7 +82,17 @@ object ExecutorInitiator extends Logging {
     def setRetryFlag(retry: Boolean = true): Unit = lock.synchronized {
       retryTask = retry
       lock.notify()
+      if (retry){
+        // Not am ideal place to do clean up. Better place ?
+        SparkEnv.get.memoryManager.asInstanceOf[StoreUnifiedManager].close
+      }
     }
+
+    // Test hook. Not to be used in other situations
+    def waitForExecutor(retry: Boolean = true): Unit =
+      testLock.synchronized(testLock.wait(30000))
+
+
 
     def getRetryFlag: Boolean = lock.synchronized {
       retryTask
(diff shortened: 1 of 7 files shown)


cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -136,6 +136,37 @@ class SnappyUnifiedMemoryManager private[memory](
...
+    MemoryManagerCallback.resetMemoryManager()
+    synchronized {
+      logInfo(s" Closing Memory Manager ${this}")
+      val bootManager = MemoryManagerCallback.bootMemoryManager
+          .asInstanceOf[SnappyUnifiedMemoryManager]
+
+      memoryForObject.object2LongEntrySet().iterator().asScala foreach { entry =>
+        val (objectName, memoryMode) = entry.getKey
+        if (!objectName.equals(SPARK_CACHE) &&
+            !objectName.equals(BufferAllocator.STORE_DATA_FRAME_OUTPUT)) {
```
sumwale: endsWith for latter since off-heap uses DIRECT_STORE_DATA_FRAME_OUTPUT

cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
```diff
@@ -83,6 +84,11 @@ object ExecutorInitiator extends Logging {
       lock.notify()
     }
 
+    // Test hook. Not to be used in other situations
+    def waitForExecutor(retry: Boolean = true): Unit = testLock.synchronized {
```
sumwale: might be better to make this private[cluster] and have the test in the same package

cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
```diff
@@ -83,6 +84,11 @@ object ExecutorInitiator extends Logging {
...
+      testLock.wait (30000)
```
sumwale: remove space before paren

cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
```diff
@@ -191,10 +196,12 @@ object ExecutorInitiator extends Logging {
                     rpcenv.setupEndpoint("Executor", executor)
                   }
                   prevDriverURL = url
+                  testLock.synchronized(testLock.notify())
                 case None =>
                   // If driver url is none, already running executor is stopped.
                   prevDriverURL = ""
               }
+              setRetryFlag(false)
```
sumwale: Can you tell what moving this setting does?

rishitesh: When multiple threads get OOM exceptions every task tries to set the flag to true.  Hence sometimes the executor is stopped & started multiple times because of the loop condition. This was not harmful till now, and in actual cases very rare.  Moving set flag  does not avoid this. However it reduces the occurrences.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
rishitesh: Will change it.

rishitesh:
@The actual UMM instance is recreated when a new cache is created. However the temp manager is a static reference. You are correct in pointing out that it needs to be cleared out. Will add that check.

cluster/src/main/scala/io/snappydata/cluster/ExecutorInitiator.scala
rishitesh: It is a bit problematic, as the test uses package level private methods from memory package.

PR #823, opened 2017-09-01
Fix for SNAP-1985

hbhanawat:

Issue is with merging of two deltas. While merging, we create a decoder and initialize it properties based on the decoders that are being merged. But encoder.initialize is not called and stringMap field of the DictionaryEncoder remains uninitialized and hence a null pointer exception. This code calls the initialize passing a minimum size.
Also allows "withHeader=false" for dictionary encoding 

Fix from @sumwale 

## Patch testing

Enabled the test and precheckin.

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/store/ColumnUpdateDeleteTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/store/ColumnUpdateDeleteTest.scala
@@ -78,7 +78,7 @@ class ColumnUpdateDeleteTest extends ColumnTablesTestBase {
     session.conf.unset(Property.ColumnBatchSize.name)
   }
 
-  ignore("SNAP-1985: update delete on string type") {
+  test("SNAP-1985: update delete on string type") {
     val tableName1 = "order_line_1_col_str"
     val tableName2 = "order_line_2_ud_str"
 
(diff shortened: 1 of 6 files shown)


cluster/src/test/scala/org/apache/spark/sql/store/ColumnUpdateDeleteTest.scala
```diff
@@ -78,7 +78,7 @@ class ColumnUpdateDeleteTest extends ColumnTablesTestBase {
     session.conf.unset(Property.ColumnBatchSize.name)
   }
 
-  ignore("SNAP-1985: update delete on string type") {
+  test("SNAP-1985: update delete on string type") {
```
vivekwiz: Can you please extend this test to also execute same update using query routing path (good if you also do for prepared statement)

sumwale: @vivekwiz have it as part of concurrent tests in https://github.com/SnappyDataInc/snappydata/pull/822

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/BooleanBitSetEncoding.scala
```diff
@@ -82,7 +82,7 @@ trait BooleanBitSetEncoderBase
   }
 
   override def initialize(dataType: DataType, nullable: Boolean, initSize: Int,
-      withHeader: Boolean, allocator: BufferAllocator): Long = {
+      withHeader: Boolean, allocator: BufferAllocator, minBufferSize: Int = -1): Long = {
     byteCursor = super.initialize(dataType, nullable, initSize,
       withHeader, allocator)
```
sumwale: I think minBufferSize should be passed to super.initialize

PR #825, opened 2017-09-02
[SNAP-1993] fixes for data skew with no partition_by

sumwale:

- check the bucket sizes to pick the least loaded bucket instead of picking a random one
- use global Random instead of class specific for better distribution with concurrent calls
  (sync points not of much concern since its miniscule compared to overall batch creation/put)
- reduced the maximum number of rows per batch to 200k so that maximum skew will
  be around that number in most cases
- added a "snappydata.linkPartitionsToBuckets" SQLConf property to disable de-linking
  for all queries
- split out UTF8String tests into separate StringBenchmark from ColumnCacheBenchmark

With these changes, distribution in ColumnCacheBenchmark test, for example, is exactly equal
most of the time among the buckets. Other cases like those reported originally with 7M rows
have only ~50% difference between min and max (as compared to ~4X originally)

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/78

### DIFF
--- a/core/src/main/scala/io/snappydata/Literals.scala
+++ b/core/src/main/scala/io/snappydata/Literals.scala
@@ -111,7 +111,7 @@ object Constant {
    * Limit the maximum number of rows in a column batch (applied before
    * [[Property.ColumnBatchSize]] property).
    */
-  val MAX_ROWS_IN_BATCH = 500000
+  val MAX_ROWS_IN_BATCH = 250000
 
   val DEFAULT_SERIALIZER = "org.apache.spark.serializer.PooledKryoSerializer"
 
@@ -261,6 +261,13 @@ object Property extends Enumeration {
         "that can return very large number of rows in aggregation results. " +
         "Default value is 0b meaning no limit on the size so the optimized " +
         "hash aggregation is always used.", Some("0"))
+
+  val ForceLinkPartitionsToBuckets: SQLValue[Boolean] = SQLVal[Boolean](
+    s"${Constant.PROPERTY_PREFIX}linkPartitionsToBuckets",
+    "Property to always create one bucket per partition in column/row table scans. " +
+        "When unset or set to false, SnappyData will try to create only " +
+        "as many partitions as executor cores clubbing multiple buckets " +
+        "into each partition when possible.", Some(false), Constant.SPARK_PREFIX)
 
   val EnableExperimentalFeatures = SQLVal[Boolean](
     s"${Constant.PROPERTY_PREFIX}enable-experimental-features",
--- a/core/src/main/scala/io/snappydata/Literals.scala
+++ b/core/src/main/scala/io/snappydata/Literals.scala
@@ -264,7 +264,7 @@ object Property extends Enumeration {
 
   val ForceLinkPartitionsToBuckets: SQLValue[Boolean] = SQLVal[Boolean](
     s"${Constant.PROPERTY_PREFIX}linkPartitionsToBuckets",
-    "Property to always create one bucket per partition in column/row table scans. " +
+    "Property to treat each bucket as separate partition in column/row table scans. " +
         "When unset or set to false, SnappyData will try to create only " +
         "as many partitions as executor cores clubbing multiple buckets " +
         "into each partition when possible.", Some(false), Constant.SPARK_PREFIX)
--- a/core/src/main/scala/io/snappydata/Literals.scala
+++ b/core/src/main/scala/io/snappydata/Literals.scala
@@ -111,7 +111,7 @@ object Constant {
    * Limit the maximum number of rows in a column batch (applied before
    * [[Property.ColumnBatchSize]] property).
    */
-  val MAX_ROWS_IN_BATCH = 250000
+  val MAX_ROWS_IN_BATCH = 200000
 
   val DEFAULT_SERIALIZER = "org.apache.spark.serializer.PooledKryoSerializer"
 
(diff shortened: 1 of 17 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -539,17 +538,27 @@ class JDBCSourceAsColumnarStore(private var _connProperties: ConnectionPropertie
...
-              if (numPrimaries > 0) {
-                primaryBucketIds.getQuick(rand.nextInt(numPrimaries))
+            if (partitionId == -1) pr.synchronized {
+              val primaryBuckets = pr.getDataStore.getAllLocalPrimaryBucketRegions
+              // if no local primary bucket, then select a random bucket
+              if (primaryBuckets.isEmpty) {
+                Random.nextInt(pr.getTotalNumberOfBuckets)
               } else {
-                rand.nextInt(pr.getTotalNumberOfBuckets)
+                // select the bucket with smallest size at this point
```
hbhanawat: size here means number of cached batches. I mean the bucket that is selected for insertion has least number of cached batches not the least size. Mentioning a line in comments/ PR would help.

sumwale: Ok will do. Normally we expect cached batches to be similar in size (except when there is a lot of mix of JDBC single inserts with bulk inserts because rollover uses number of rows for policy). Avoided iteration but perhaps getTotalBytes might be better.

kneeraj:
Just one small comment here:

>> added a "snappydata.linkPartitionsToBuckets" SQLConf property to disable de-linking
for all queries
I think this should be enabled by default. Only when required you should switch it off. Reason being that each partitions will be big normally for big column tables. Isn't it?

rishitesh:
@sumwale Can you please specify the problem statement ? Why a randomized selection of buckets was an issue ?

sumwale:
@kneeraj 

> Reason being that each partitions will be big normally for big column tables. Isn't it?

Big partitions are actually good normally which is why de-linking was added. It reduces the number of partitions to be scheduled, allows JIT to kick in faster etc. It certainly helps for simple queries like "select count(*)" etc especially in local[*] runs. They are not so good for concurrent queries because a core may be occupied for a bit longer. In large clusters it is not expected to make any difference because number of buckets will likely be similar or smaller than total number of cores. So I do not think de-linking should be switched off by default until we have more solid evidence from CH-benchmark etc.

sumwale:
@rishitesh
It has been observed that random was not distributing batches over buckets uniformly even in simple cases. I suppose if there are a very large number of batches then random will gradually become more and more uniform but normally a single insert will have say 100 batches where it was not working well.

Secondly each server has its own random generation. So that means it may not be uniformly distributed across servers in any case. This may be one of the contributing factors for the observation above.

Thirdly even if we somehow have random generation across servers, it may not work very well because inserts may be going on into row buffer from JDBC layer (which will rollover), rebalance operations may have been invoked etc.

Given all these it had been discussed more than a year back that it should be distributed as per current bucket load -- see the TODO already in the code which this PR removes. This is still not being done for the smart connector mode though (which needs more meta-data to get the bucket load information).

kneeraj:
Big partitions are actually good normally which is why de-linking was added.

> Yes I understand why it was added. What I am saying is that in scenarios if there are overheads, inefficiencies in terms of dealing a lot of data at the same time, like keeping some transient in memory structure then this can have adverse effects as well. If the partition is really big then clubbing few more into one big one can be adverse in few situations and hence the defsult of treating one partiton for one partition should be good enough. May be running a tpch-h with 1 TB with de-linking and without de-linking should be done to decide the default.

sumwale:
> like keeping some transient in memory structure

Transient memory structures are per batch not per partition so they are identical

> running a tpch-h with 1 TB with de-linking and without de-linking

This will be identical. As I mentioned the only case where it can be slower is for concurrent query latency (not throughput) and when the number of buckets is substantially more than total number of cores in the system. So a CH like benchmark with small/medium cluster.

sumwale:
> running a tpch-h with 1 TB with de-linking and without de-linking

Also note that de-linking was largely for scenarios where users are quickly testing on single machines. So those should not see large deterioration and we should not be using some small drops in CH like tests to decide. The loss should be substantial and single node case (and other cases where it helps) should not deteriorate much to take a call on the default. Else we should recommend explicit setting to true for CH like cases which will be less common in practice.

kneeraj:
> Also note that de-linking was largely for scenarios where users are quickly testing on single machines. So those should not see large deterioration and we should not be using some small drops in CH like tests to decide. The loss should be substantial and single node case (and other cases where it helps) should not deteriorate much to take a call on the default. Else we should recommend explicit setting to true for CH like cases which will be less common in practice.

Yes. Thanks for explaining.

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -262,6 +262,13 @@ object Property extends Enumeration {
         "Default value is 0b meaning no limit on the size so the optimized " +
         "hash aggregation is always used.", Some("0"))
 
+  val ForceLinkPartitionsToBuckets: SQLValue[Boolean] = SQLVal[Boolean](
+    s"${Constant.PROPERTY_PREFIX}linkPartitionsToBuckets",
+    "Property to always create one bucket per partition in column/row table scans. " +
```
kneeraj: @sumwale The property is not enforcing creation. It is just enforcing one partiton per bucket. Perhaps you want to re-word to -- Property to treat each bucket as one partition -- or something like that.

sumwale: Right, good catch. Should have been other way round "create one partition per bucket"

rishitesh:
@sumwale thanks for the explanation.  LGTM

PR #827, opened 2017-09-05
[SNAP-1995] Added a pyhon KMeans example

rishitesh:
File from Sid. 

A python script example to compute KMeans from weather data. This uses SnappySession as entry point. 

## Patch testing

example suite

## ReleaseNotes.txt changes

NA

## Other PRs 

### DIFF
--- a/examples/src/main/python/KMeansWeather.py
+++ b/examples/src/main/python/KMeansWeather.py
@@ -1,0 +1,84 @@
+from __future__ import print_function
+from pyspark.conf import SparkConf
+from pyspark.context import SparkContext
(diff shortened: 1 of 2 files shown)


kneeraj:
I don't know pyhon as well as python :-)
Still I will review.

kneeraj:
LGTM. 
Just one comment:
@rishitesh This example works and therefore it is fine as to how the user can use the apis and build the model. Only comment I have is that shall we write in comment what this program is trying to do and the code is just an illustration of how to use the apis and synthetic data to build the model..

rishitesh:
@kneeraj Will add some comments.

PR #828, opened 2017-09-06
Snap 1994

vivekwiz:
Added some tests for expression used in prepared statement.

## Patch testing
Only test changes.

## ReleaseNotes.txt changes
No

## Other PRs 
No


cluster/src/test/scala/io/snappydata/cluster/PreparedQueryRoutingSingleNodeSuite.scala
```diff
@@ -470,6 +470,143 @@ class PreparedQueryRoutingSingleNodeSuite extends SnappyFunSuite with BeforeAndA
...
+        // scalastyle:off println
+        println(s"1-row($index) $i $j")
+        // scalastyle:on println
+        index += 1
+      }
+      // scalastyle:off println
+      println(s"1-Number of rows read " + index)
+      // scalastyle:on println
+      assert(index == 2)
+      assert(cacheMap.size() == 1)
```
kneeraj: Add result verification too.

kneeraj: Similarly at other places where applicable.

cluster/src/test/scala/io/snappydata/cluster/PreparedQueryRoutingSingleNodeSuite.scala
```diff
@@ -470,6 +470,143 @@ class PreparedQueryRoutingSingleNodeSuite extends SnappyFunSuite with BeforeAndA
...
+        s" (3, 3.33333, '3c')")
+    val cacheMap = SnappySession.getPlanCache.asMap()
+    assert( cacheMap.size() == 0)
+    val serverHostPort = TestUtil.startNetServer()
+    val conn = DriverManager.getConnection("jdbc:snappydata://" + serverHostPort)
+    var prepStatement0: java.sql.PreparedStatement = null
+    var prepStatement1: java.sql.PreparedStatement = null
+    var prepStatement2: java.sql.PreparedStatement = null
+    try {
+      prepStatement0 = conn.prepareStatement(s"select * from double_tab" +
```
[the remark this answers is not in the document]
vivekwiz: Done with cacheMap related changes.

Throwing exceptions from all those places where ? is not expected is pretty much complex task. May need to look into each sql function, its parameter, its expected action and respective tests. Please suggest.

cluster/src/test/scala/io/snappydata/cluster/PreparedQueryRoutingSingleNodeSuite.scala
vivekwiz: Done

kneeraj:
> Throwing exceptions from all those places where ? is not expected is pretty much complex task. May need to look into each sql function, its parameter, its expected action and respective tests. Please suggest.

For the function part at least we can do I guess. Like we have this flag:
session.sessionState.isPreparePhase

and we know the place in the parser where functions are being parsed:
Refer to the code near the FOLDABLE functions check. There we can traverse the exprs and find that out.

@sumwale please suggest.

vivekwiz:
@kneeraj the foldable functions variable is not the exhaustive list of sql functions. Please check: https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/functions.html

vivekwiz:
@kneeraj and as I said for every function you need to know its parameters.

kneeraj:
> @kneeraj the foldable functions variable is not the exhaustive list of sql functions. Please check: https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/functions.html

@vivekwiz I am not saying do that check inside Foldable functions code block. I am saying near that code, may be just above that, is the place where you can do this in the prepare phase.

vivekwiz:
@kneeraj There are 300+ sql functions. You want me to
1) write test for each of them
2) Handle in parser/elsewhere conditions checking whether function argument has ? and throw exception for same.

Kindly confirm. Should that be taken as part of SNAP-1994?

vivekwiz:
Had a work with @kneeraj and @sumwale and either of following can be done:
1. No '?' will be allowed as argument in any sql function and document same. This was @kneeraj original comment above which I did not understand.
2. @sumwale had suggested an alternative code to try and see if it works. Otherwise resort to point #1 above.

I also plan to check behaviour of some other database to see if we can have some other alternative.

vivekwiz:
Went with @kneeraj suggestion in above. Kindly review again.

Please also help draft message to user (improvement over mail with subjectline "1634: disclaimer for prepared statements in documenation"; @shyjaprabhu please note):
Prepared statements  do not accept '?' as parameter in sql functions.

cluster/src/test/scala/io/snappydata/cluster/PreparedQueryRoutingSingleNodeSuite.scala
```diff
@@ -470,6 +470,329 @@ class PreparedQueryRoutingSingleNodeSuite extends SnappyFunSuite with BeforeAndA
...
+      assert(cacheMap.size() == 3)
+      close(prepStatement2)
+
+      var prepStatement3: PreparedStatement = null
+      try {
+        prepStatement3 = conn.prepareStatement(s"select * from double_tab" +
+            s" where round(d, ?) < ?")
+        fail()
+      } catch {
+        case e: UnsupportedOperationException =>
```
kneeraj: Remove UnsupportedException. It is not expected here.

cluster/src/test/scala/io/snappydata/cluster/PreparedQueryRoutingSingleNodeSuite.scala
```diff
@@ -470,6 +470,329 @@ class PreparedQueryRoutingSingleNodeSuite extends SnappyFunSuite with BeforeAndA
...
+        case e: SQLException =>
```
kneeraj: Check if the message you send is there in the exception message or not. i.e. '?' mark should not be there in the function argument.

cluster/src/test/scala/io/snappydata/cluster/PreparedQueryRoutingSingleNodeSuite.scala
```diff
@@ -470,6 +470,329 @@ class PreparedQueryRoutingSingleNodeSuite extends SnappyFunSuite with BeforeAndA
...
+        println(s"8-row($index) $i")
+        // scalastyle:on println
+        index += 1
+        assert(i == 2 || i == 3)
+      }
+      // scalastyle:off println
+      println(s"8-Number of rows read " + index)
+      // scalastyle:on println
+      assert(index == 2)
+      assert(cacheMap.size() == 0)
```
kneeraj: Same as above.

vivekwiz:
@kneeraj please have a look at updated changes.

kneeraj:
LGTM.

sumwale:
@vivekwiz @kneeraj Not sure why we are going towards removing even existing tokenization. We have to increase not reduce. Just tested a patch that I had suggested and it worked fine.

kneeraj:
@swale vivek said that he was finding some trouble in your suggestion and so as discussed he went back to my suggestion which I approved. I will leave to @vivekwiz to explain what trouble he was facing. And anyways this is just about args in function and we are not losing much here in my opinion.

sumwale:
Below is the patch I tried. Just skips the two FOLDABLE checks in prepare phase. Am I missing the reasoning to avoid this?

```scala
diff --git a/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala b/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
index c8e83ad..f95afa7 100644
--- a/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
@@ -787,19 +787,21 @@ class SnappyParser(session: SnappySession)
             val udfName = f2.fold(new FunctionIdentifier(n1))(new FunctionIdentifier(_, Some(n1)))
             val allExprs = e.asInstanceOf[Seq[Expression]]
             var exprs = allExprs
-            Constant.FOLDABLE_FUNCTIONS.get(n1) match {
-              case Some(args) =>
-                exprs = allExprs.zipWithIndex.collect {
-                  case (pl: ParamLiteral, index) if args.contains(index) ||
-                      // all args          // all odd args
-                      (args.head == -10) || (args.head == -1 && (index & 0x1) == 1) ||
-                      // all even args
-                      (args.head == -2 && (index & 0x1) == 0) =>
-                    removeParamLiteralFromContext(pl)
-                    Literal.create(pl.value, pl.dataType)
-                  case (ex: Expression, _) => ex
-                }
-              case None =>
+            if (!session.sessionState.isPreparePhase) {
+              Constant.FOLDABLE_FUNCTIONS.get(n1) match {
+                case Some(args) =>
+                  exprs = allExprs.zipWithIndex.collect {
+                    case (pl: ParamLiteral, index) if args.contains(index) ||
+                        // all args          // all odd args
+                        (args.head == -10) || (args.head == -1 && (index & 0x1) == 1) ||
+                        // all even args
+                        (args.head == -2 && (index & 0x1) == 0) =>
+                      removeParamLiteralFromContext(pl)
+                      Literal.create(pl.value, pl.dataType)
+                    case (ex: Expression, _) => ex
+                  }
+                case None =>
+              }
             }
             val function = if (d.asInstanceOf[Option[Boolean]].isEmpty) {
               UnresolvedFunction(udfName, exprs, isDistinct = false)
@@ -829,19 +831,21 @@ class SnappyParser(session: SnappySession)
       (fn: FunctionIdentifier, e: Any) =>
         val allExprs = e.asInstanceOf[Seq[Expression]].toList
         var exprs = allExprs
-        Constant.FOLDABLE_FUNCTIONS.get(fn.funcName) match {
-          case Some(args) =>
-            exprs = allExprs.zipWithIndex.collect {
-              case (pl: ParamLiteral, index) if args.contains(index) ||
-                  // all args          // all odd args
-                  (args.head == -10) || (args.head == -1 && (index & 0x1) == 1) ||
-                  // all even args
-                  (args.head == -2 && (index & 0x1) == 0) =>
-                removeParamLiteralFromContext(pl)
-                Literal.create(pl.value, pl.dataType)
-              case (ex: Expression, _) => ex
-            }
-          case None =>
+        if (!session.sessionState.isPreparePhase) {
+          Constant.FOLDABLE_FUNCTIONS.get(fn.funcName) match {
+            case Some(args) =>
+              exprs = allExprs.zipWithIndex.collect {
+                case (pl: ParamLiteral, index) if args.contains(index) ||
+                    // all args          // all odd args
+                    (args.head == -10) || (args.head == -1 && (index & 0x1) == 1) ||
+                    // all even args
+                    (args.head == -2 && (index & 0x1) == 0) =>
+                  removeParamLiteralFromContext(pl)
+                  Literal.create(pl.value, pl.dataType)
+                case (ex: Expression, _) => ex
+              }
+            case None =>
+          }
         }
         fn match {
           case f if f.funcName.equalsIgnoreCase("TIMESTAMPADD") =>
```

This is modification to test "SNAP-1994 Test functions and expressions" that worked fine (check results instead of expecting failure in the last few lines):

```scala
      val prepStatement4 = conn.prepareStatement(s"select * from double_tab" +
          s" where round(d, ?) < ?")
      assert(cacheMap.size() == 0)
      prepStatement4.setInt(1, 2)
      prepStatement4.setDouble(2, 3.33)
      update = prepStatement4.executeQuery()
      index = 0
      while (update.next()) {
        val i = update.getInt(1)
        val j = update.getBigDecimal(2)
        // scalastyle:off println
        println(s"1-row($index) $i $j")
        // scalastyle:on println
        index += 1
        assert(i == 1 || i == 2)
      }
      // scalastyle:off println
      println(s"1-Number of rows read " + index)
      // scalastyle:on println
      assert(index == 2)
      assert(cacheMap.size() == 1)
      close(prepStatement4)
```

sumwale:
@kneeraj Didn't have time to test things earlier and go through what @vivekwiz was mentioning as the problem (or was the source of problem mentioned somewhere?). I think different constants in function args are pretty common and we should not unnecessarily put that restriction unless the fix to that is too complex for 1.0 (which is not as per my test).

vivekwiz:
@sumwale the expected problem is at the end prepare phase (SparkSQLPrepareImpl#allParamLiterals), where I am trying to get datatypes for each '?'. I had to handle some scenario specially for example using queries with case-when (SparkSQLPrepareImpl#remainingParamLiterals)  and sub-query. My worry is that if we are thinking of supporting '?' for any sql-function-parameter we may need to look at each sql-function and its parameter. 

However currently all such cases should work where '?' is given as sql-function-parameter in place of literal-value. Your change will expand the supported use case scenarios but we may still get unsupported-exception in some cases.

sumwale:
> My worry is that if we are thinking of supporting '?' for any sql-function-parameter we may need to look at each sql-function and its parameter.

Can't think of a case where it can be a problem.

> I had to handle some scenario specially for example using queries with case-when (SparkSQLPrepareImpl#remainingParamLiterals) and sub-query. 

This can be a more general problem and not related to function parameters as such. There are a bunch of known/unknown queries where ParamLiterals do not match and in those cases @kneeraj had added a catch-all to remove all ParamLiterals. Perhaps this is what you are talking about?

Lets add tests for cases you suspect (@kneeraj please help providing more queries where the catch-all comes into play) and if there are any issues then take it from there.

sumwale:
> Lets add tests for cases you suspect (@kneeraj please help providing more queries where the catch-all comes into play) and if there are any issues then take it from there.

This I meant post 1.0 to track for 1.01. Generally speaking we have to maximize tokenization unless there is no other choice. Will suggest that we go with a change like above and then take it from there.

vivekwiz:
So as suggested I will incorporate @sumwale changes and will modify my test accordingly. Will reopen SNAP-1634 (target 1.01) for widening coverage. 

@kneeraj only thing will remain is to think of any documentation. IMO we are fine either way.

sumwale:
@vivekwiz also if you can take the opportunity to re-factor that code segment into a separate method (the FOLDABLE check and replace when its not prepare phase) then will be good. Thanks.

PR #829, opened 2017-09-08
SNAP-2003 stream to big table join CQ incorrect result

ymahajan:
- HashJoinExec's inputRDDs are computed on each CQ execution.
- Added test 

## Patch testing

precheckin 

## ReleaseNotes.txt changes

No

## Other PRs 

None

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoinExec.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoinExec.scala
@@ -35,6 +35,7 @@ import org.apache.spark.sql.catalyst.plans.physical._
 import org.apache.spark.sql.collection.Utils
 import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.metric.SQLMetrics
+import org.apache.spark.sql.streaming.PhysicalDStreamPlan
 import org.apache.spark.sql.types.TypeUtilities
 import org.apache.spark.sql.{DelegateRDD, SnappySession}
 
@@ -182,6 +183,7 @@ case class HashJoinExec(leftKeys: Seq[Expression],
   // return empty here as code of required variables is explicitly instantiated
   override def usedInputs: AttributeSet = AttributeSet.empty
 
+
   private def findShuffleDependencies(rdd: RDD[_]): Seq[Dependency[_]] = {
     rdd.dependencies.flatMap {
       case s: ShuffleDependency[_, _, _] => if (s.rdd ne rdd) {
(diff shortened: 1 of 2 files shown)


sumwale:
Streaming plans should not be using cached plans in the first place. Many plans do RDD caching or can have other stale artifacts, so the correct fix is to not use plan caching at all when using streaming. I can see SnappyStreamingContext.sql using sqlUncached call so why is it not taking effect? That needs to be investigated. A better and surer option compared to sqlUncached might be a flag on SnappySession set by streaming code that will always call uncached plan.

This also needs to be ensured for the new streaming model that does not use SnappyStreamingContext.

hbhanawat:
@sumwale I think @ymahajan 's observation is correct. The problem is HashJoinExec is holding references to RDD. The following is a lazy val and hence never gets changed.

private lazy val (streamSideRDDs, buildSideRDDs)
 
It is a problem for streaming cases. 

However, I think the fix isn't correct. @ymahajan I think you need to refresh the buildRDDs as well because the kafka stream could have been the build side which is also not getting refreshed. 

One way would be to make it a def instead of lazy val. This will make sure it works always. But this would mean this function would be re-executed for every execution of non streaming queries as well. Given that it does not look like a heavy function, generic handling should be ok. I didn't like CQ specific handling. What do you think @sumwale?

hbhanawat:
BTW, forgot the mention, plan for DStream is generated once. Hence it is not the plan caching but the behaviour of this feature itself.

sumwale:
@hbhanawat @ymahajan So plan is generated once which is causing trouble? I see that SchemaDStream.executionPlan is explicitly clearing the cached field in the shuffle plan as a workaround for the same issue in shuffle plan. I think this is the wrong approach because many other plans may have some cached information which is stale but is not being covered in our tests yet. Likewise for all future plans that can have cached information this will need to be handled explicitly which is neither feasible nor a good way.

Instead executionPlan should create the QueryExecution again in every call. One simple way could be to pass a closure `() => QueryExecution` in the constructor instead of `QueryExecution`. The other uses like for schema, logicalPlan need not re-evaluate and can do so in construction as currently, but executionPlan should execute afresh (and then the explicit clearing of field in ShuffleExchange should also be removed).

hbhanawat:
I see. I checked the structured streaming code and it looks like they also are caching the logical plan but are re-generating the queryExecution everytime. However I am not clear about the perf impact. 

@ymahajan please add a test for the structured streaming as well that uses the hashjoinexec.

sumwale:
From the experience of previous testing, plan creation takes max few milliseconds. Saving that is good for point low-latency queries (as has been seen in concurrency tests) but hopefully not a noticeable impact for streaming jobs. Either way doesn't matter much since structured streaming is what is going to be the focus.

ymahajan:
@sumwale, thanks for you suggestion. Creating QueryExecution on every CQ execution doesn't work as there are PhysicalDStream and WindowDStream plans which work on internal artifacts like wrappedStream, parentStreams etc 

This has been tried in past as well, when we introduced the ShuffleExchange reuse fix post Spark 2.0 merge.  The whole registerCQ implementation is implemented considering one plan.

In my opinion, we should stick to the original fix of operator specific handling.  our focus would be on structured streaming anyways after the GA release. 

@hbhanawat I have pushed a commit which refreshes buildRDDs as well. In fact the lazy val has been changed to def and CQ specific handling has been removed. Structured streaming tests are something can be added as separate work item. Also there is a how tos to be added on structured streaming.

PR #830, opened 2017-09-08
* Fixed failure in testSparkShell

ashetkar:
* Fixed failure in testSparkShell() due to use of String.contains() instead of String.containsKey()
* Added tests for update/delete on column table with security enabled (currently fails with Jira 2004).

## Patch testing
Failing unit tests. precheckin NOT REQUIRED.

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


sumwale:
Perhaps also need to exclude "NoSuchObjectException" from the output check because that is normal when starting the shell.

PR #831, opened 2017-09-08
Added test for With Clause Fix

hbhanawat:

There was an issue with 'with' clause. That was fixed in the aqp branch. Adding a test here for the same. 

## Patch testing

Manual. Precheckin .


svbhokare:
LGTM.

PR #832, opened 2017-09-08
[AQP-293] changes for JNI UTF8String.contains

sumwale:

- convert UTF8Strings in ParamLiteral to off-heap when snappy off-heap is enabled;
  added DirectStringConsumer to acquire/release the required off-heap memory
- refactored the common code for TaskContext.get() into a separate method
  which caches the result to avoid repeated thread-local lookups (Utils.genTaskContextFunction
      should be used by snappydata code that needs it)
- SnappyParser changes:
  - added a "likeExpression" method that will directly convert to StartsWith/EndsWith/Contains
    expressions as per the like pattern with ParamLiteral so that off-heap conversion happens
    for the pattern string (else Spark's LikeSimplification replaces with a Literal)
  - enabled "inlineTable" rule which was commented out due to non-availabiity of
    UnresolvedInlineTable plan in 2.0.1 (present since 2.0.2); now INSERT ... VALUES
    or SELECT ... VALUES or just VALUES ... will work in SnappySession too (tests to be added later)
  - simplified comparisonExpression
- build support to copy JNI libraries from the AQP project if present
- test suite NativeUTF8StringPropertyCheckSuite adapted from Spark's
  UTF8StringPropertyCheckSuite
  but using off-heap UTF8Strings to test the new JNI implementation
- added off-heap test case to StringBenchmark
- updated parboiled2 to latest release
- removed an unused method in Utils

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/80

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -671,11 +671,12 @@ task product(type: Zip) {
 
   def clusterProject = project(":snappy-cluster_${scalaBinaryVersion}")
   def targetProject = clusterProject
-  def hasAQPProject = new File(rootDir, 'aqp/build.gradle').exists()
+  def isEnterpriseProduct = new File(rootDir, 'aqp/build.gradle').exists() &&
+      rootProject.hasProperty('snappydata.enterprise')
   def hasJdbcConnectorProject = new File(rootDir,
       'snappy-connectors/jdbc-stream-connector/build.gradle').exists()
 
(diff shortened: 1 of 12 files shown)


sumwale:
Merging now for the 1.0 release. Will take care of any post-merge review comments subsequently.

PR #833, opened 2017-09-08
[SNAP-2006] fix two issues seen in ColumnUpdateDeleteTests.testConcurrentOps

sumwale:

- set the positionEndCursor to the actual end of positions and not including the
  8-byte alignment bytes; the padding area may not be initialized and should not be read
- restricted the synchronized areas in ColumnFormatValue.getBufferRetain to exclude
  the Helper.getValueOnDisk call that acquires the RegionEntry lock to avoid lock
  inversion (getBufferRetain used to take ColumnFormatValue->RegionEntry while
      releaseBuffer takes RegionEntry->ColumnFormatValue)
- added asserts in ColumnDeltaEncoder to ensure that Unsafe writes never happen
  beyond the buffer boundary (to track down SNAP-1914 which is likely due to bad writes)
- renamed JDBCSourceAsColumnarStore.doInsertOrPutImpl to more appropriate doRowBufferPut
- updated ColumnUpdateDeleteTests to run the concurrency tests 5 times in a loop
  that reproduces the two issues mentioned above occasionally (and almost always
      within 5 runs)
- minor cleanup in BooleanBitSetEncoding to move cursor by 8 immediately after word write

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


sumwale:
Closing this and will open against master since AQP-293 has been merged.

PR #835, opened 2017-09-10
[SNAP-2006] fix two issues seen in ColumnUpdateDeleteTests.testConcurrentOps

sumwale:

- set the positionEndCursor to the actual end of positions and not including the
  8-byte alignment bytes; the padding area may not be initialized and should not be read
- restricted the synchronized areas in ColumnFormatValue.getBufferRetain to exclude
  the Helper.getValueOnDisk call that acquires the RegionEntry lock to avoid lock
  inversion (getBufferRetain used to take ColumnFormatValue->RegionEntry while
      releaseBuffer takes RegionEntry->ColumnFormatValue)
- added asserts in ColumnDeltaEncoder to ensure that Unsafe writes never happen
  beyond the buffer boundary (to track down SNAP-1914 which is likely due to bad writes)
- renamed JDBCSourceAsColumnarStore.doInsertOrPutImpl to more appropriate doRowBufferPut
- updated ColumnUpdateDeleteTests to run the concurrency tests 3 times in a loop
  that reproduces the two issues mentioned above occasionally (and almost always
      within 3 runs)
- fixing failure in PreparedQueryRoutingSingleNodeSuite due to cache map size changes
  after AQP-293 that uses ParamLiteral for contains/startsWith/endsWith/equals cases in LIKE
  pattern; skip the result size check for equals case due to SNAP-2004 for now
- minor cleanup in BooleanBitSetEncoding to move cursor by 8 immediately after word write
- avoid rollover in concurrent update/delete tests; rollover is causing extra rows to be selected in
  scan in dunit tests (possibly due to remote iteration?) which will be tracked separately
- change eol style of KMeansWeather.py to unix

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

Previous PR against AQP-293: https://github.com/SnappyDataInc/snappydata/pull/833


sumwale:
Previous PR https://github.com/SnappyDataInc/snappydata/pull/833 was approved so merging.

PR #836, opened 2017-09-10
[SNAP-788] skip explicit initialization of row buffer

sumwale:

Parent row buffer table buckets will get initialized when child column table
is initialized, so skip its explicit initialization (else it triggers
  background creation of child buckets when column table is created
  interfering with its initialization and delaying it due to locks)

## Patch testing

precheckin with store precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


PradeepSurale:

PR #837, opened 2017-09-11
[SNAP-2007] fix deadlock between UMM and RegionEntry

sumwale:

- free the buffer in a separate thread in background
- more precise data for eviction (with background it should not be too much eviction)

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/304

### DIFF
--- a/cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
+++ b/cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
@@ -17,7 +17,7 @@
 package org.apache.spark.memory
 
 import java.nio.ByteBuffer
-import java.util.concurrent.atomic.{AtomicInteger}
+import java.util.concurrent.atomic.AtomicInteger
 import java.util.function.Consumer
 
 import scala.collection.JavaConverters._
@@ -91,8 +91,9 @@ class SnappyUnifiedMemoryManager private[memory](
 
   private[memory] val maxHeapStorageSize = (maxHeapMemory * evictionFraction).toLong
 
+  private val maxHeapEviction = 1024L * 1024L * 1024L
   private val minHeapEviction = math.min(math.max(10L * 1024L * 1024L,
-    (maxHeapStorageSize * 0.002).toLong), 1024L * 1024L * 1024L)
+    (maxHeapStorageSize * 0.002).toLong), maxHeapEviction)
 
   private[memory] val wrapperStats = new MemoryManagerStatsWrapper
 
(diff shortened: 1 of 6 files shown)


cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -339,16 +340,14 @@ class SnappyUnifiedMemoryManager private[memory](
...
-      required * math.max(100L, waitingThreads + 1)))
+    math.max(required, math.min(math.max(minHeapEviction, required * waitingThreads),
+      maxHeapEviction))
   }
 
   private def getMinOffHeapEviction(required: Long): Long = {
     // off-heap calculations are precise so evict exactly as much as required
-    (required * math.max(1, math.min(4, threadsWaitingForStorage.get()))) -
-        offHeapStorageMemoryPool.memoryFree
+    math.max(0, required - offHeapStorageMemoryPool.memoryFree + (1024 * 1024))
```
kneeraj: What is the reason behind evicting 1 MB extra? Also at least 4 times the required - free was being evicted previously and not now? What am I missing?

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
```diff
@@ -339,16 +340,14 @@ class SnappyUnifiedMemoryManager private[memory](
   }
 
   private def getMinHeapEviction(required: Long): Long = {
-    // evict at least 100 entries to reduce GC cycles
     val waitingThreads = threadsWaitingForStorage.get()
-    math.max(required * waitingThreads, math.min(minHeapEviction,
-      required * math.max(100L, waitingThreads + 1)))
+    math.max(required, math.min(math.max(minHeapEviction, required * waitingThreads),
```
kneeraj: Just to understand: I think with the previous code there was a possibility of this function returning more than maxHeapEviction i.e. 1 GB. Right?
Also, I see that @rishitesh idea was to keep 100 factor to simulate at least 100 entry eviction at once ( assuming number of waiting threads is much less and required * 100 is less than 10 MB) but that is gone now. I guess he did this so that smaller eviction does not happen too often. But here again that chance is created for the same.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
sumwale: Earlier it was evicting too much especially because the putAll would pre-allocate bytes for all the values being put (and then thread would evict 4 times of even that).

1M is just arbitrary number to ensure that eviction does not fall short by just a small amount because putAll pre-allocation is an estimate.

cluster/src/main/scala/org/apache/spark/memory/SnappyUnifiedMemoryManager.scala
sumwale: This change of 100 factor was by me. I am still thinking of all pros and cons here. Will revert the heap change for now and come back to it later and with more testing.

kneeraj:
The change to avoid deadlock looks good to me to avoid the deadlock. I am still reasoning however if a lot of putter threads ask for release and the actual release does not happen because the thread from the pool does not get a chance to quickly release the memory. Other question is that will this also have effect on multiple threads trying to do eviction and which was not the case when happening in the same thread?

kneeraj:
Perhaps, this might lead to OoM situations which was fully avoided in sync case.
Can we think of an alternative where we timeout on umm lock and make it retryable.

rishitesh:
@kneeraj Actual eviction still looks like single threaded. Only release buffer will be multi threaded.
Now to the question of getting an OOM in this case which is very likely , as now we are dependent on scheduling of the new threads.

sumwale:
@kneeraj @rishitesh Yes, I was also worried about that. Please review the updated code where async is sent as a flag to release and only from Gem layers it is passed as true. Will open store PR shortly.

rishitesh:
Looks good to merge.

PR #839, opened 2017-09-11
[SNAP-2012] skip locked entries in evictor

sumwale:
Fix as suggested by @rishitesh to use Unsafe API to try acquire monitor on RegionEntry.
These are the SnappyData side changes corresponding to the actual fix on store side.

- use "skipLockedEntries" flag as try from SnappyUMM evictor call
- revert previous changes for SNAP-2007
- added synchronized in a couple of places in SnappyUMM where it was required
- minor changes to store memoryObject instead of multiple calls since it is a method
- update jetty to latest 9.2.x version in an attempt to fix occasional "bad request" errors
  seen currently on dashboard

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/307

### DIFF
--- a/ReleaseNotes.txt
+++ b/ReleaseNotes.txt
@@ -7,19 +7,196 @@
 # ADD THE JIRA TICKET ID, IF APPLICABLE.                                                           #
 ####################################################################################################
 
-Release 1.0
+Release 1.0.0 (WIP)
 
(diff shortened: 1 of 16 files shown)


kneeraj:
lgtm

PR #844, opened 2017-09-13
[SNAP-2020] track in-progress insert size to avoid data skew

sumwale:

With many concurrent inserts/partitions on a node, significant data skew in inserts
was still observed (on machines with large number of cores like 32) due to same
smallest bucket being chosen by multiple partitions.

This change now tracks the in-progress size for bucket and adds that to determine smallest bucket.

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

The atomic long field in BucketRegion to keep track of in-progress size was added in https://github.com/SnappyDataInc/snappy-store/commit/8410b4d30aa82a6a2af6dc2d2481d913dc4b60cc

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
@@ -85,8 +85,18 @@ class JDBCSourceAsColumnarStore(private var _connProperties: ConnectionPropertie
   override def storeColumnBatch(tableName: String, batch: ColumnBatch,
       partitionId: Int, batchId: Long, maxDeltaRows: Int,
       conn: Option[Connection]): Unit = {
-    doInsertOrPut(tableName, batch, batchId, getPartitionID(tableName, partitionId),
-      maxDeltaRows, conn)
+    if (partitionId >= 0) {
+      doInsertOrPut(tableName, batch, batchId, partitionId, maxDeltaRows, conn)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
```diff
@@ -553,26 +566,27 @@ class JDBCSourceAsColumnarStore(private var _connProperties: ConnectionPropertie
...
+                  val bucketSize = getInProgressBucketSize(bucket, shift)
                   if (bucketSize < minBucketSize ||
                       (bucketSize == minBucketSize && Random.nextBoolean())) {
                     smallestBucket = bucket
                     minBucketSize = bucketSize
                   }
                 }
-                smallestBucket.getId
+                // update the in-progress size of the chosen bucket
+                val batchSize = batchSizeInBytes
```
rishitesh: Not required val.

suranjan:
LGTM.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/JDBCSourceAsColumnarStore.scala
sumwale: @rishitesh Oh forgot why this was added. The "batchSizeInBytes" is a call by name closure that is supposed to be evaluated only when required. With the val it will be unnecessarily evaluated twice. Adding this back.

PR #845, opened 2017-09-13
[SNAP-2022] Remove the check which tested if any lead is already stopped, in snappy-stop-all.sh

ashetkar:
[SNAP-2022] Remove the check which tested if any lead is already stopped, in snappy-stop-all.sh
* Add a check for rowstore, so that 'sbin/snappy-stop-all.sh rowstore' doesn't see the message.
* Add a while + case to handle option in any order.

## Patch testing
Manual

## ReleaseNotes.txt changes
NA

## Other PRs 
NA

### DIFF
--- a/cluster/sbin/snappy-start-all.sh
+++ b/cluster/sbin/snappy-start-all.sh
@@ -32,16 +32,31 @@ if [ -f "${MEMBERS_FILE}" ]; then
   rm $MEMBERS_FILE
 fi
 
-# Check for foreground start
 BACKGROUND=-bg
-if [ "$1" = "-bg" -o "$1" = "--background" ]; then
+clustermode=
+
+while (( "$#" )); do
+  param="$1"
+  case $param in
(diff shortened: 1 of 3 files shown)


cluster/sbin/snappy-start-all.sh
```diff
@@ -32,23 +32,38 @@ if [ -f "${MEMBERS_FILE}" ]; then
...
+    ;;
+    -fg | --foreground)
+      # Check for foreground start
+      BACKGROUND=""
+    ;;
+    rowstore)
+      clustermode="rowstore"
+    ;;
+    *)
+      echo "Invalid option: $param"
```
kneeraj: You care for this echo? I think later somewhere down it will complain. Anyways no harm.

kneeraj: Isn't this more standard:

while test $# -gt 0
do
     param="$1"
     shift
done
}

PR #852, opened 2017-09-15
[SNAP-2029] added option to prefer primaries for queries

sumwale:

New "snappydata.preferPrimaries" option to prefer primary buckets for queries.

Avoid double memory at the cost of reduced scalability but still having a hot backup.
See discussion on Slack: https://snappydata-public.slack.com/archives/C0DCF0UGG/p1505460492000378

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -269,6 +269,12 @@ object Property extends Enumeration {
         "as many partitions as executor cores clubbing multiple buckets " +
         "into each partition when possible.", Some(false), Constant.SPARK_PREFIX)
 
+  val ForcePrimariesInQueryRouting: SQLValue[Boolean] = SQLVal[Boolean](
+    s"${Constant.PROPERTY_PREFIX}forcePrimaries",
+    "Property to prefer using primary buckets in queries. This reduces " +
```
kneeraj: Is this forcing or is this just a preference?
And too much of query routing term :-), may be we can use PreferPrimaryPartitions.

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -120,7 +121,8 @@ object StoreUtils {
   }
 
   private[sql] def getBucketPreferredLocations(region: PartitionedRegion,
-      bucketId: Int, forWrite: Boolean): Seq[String] = {
+      bucketId: Int, forWrite: Boolean,
+      preferPrimaries: Boolean = false): Seq[String] = {
     if (forWrite) {
```
kneeraj: We could have the condition like this no?
if (forWrite || preferPrimaries)

core/src/main/scala/io/snappydata/Literals.scala
sumwale: Will change name

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
sumwale: forWrite will select only the primary. This change will still have all of primary and secondaries albeit with primary at the start (i.e. in worst case if query partition can't be scheduled on a primary for long time then it will use secondary).

kneeraj: yes thought so. Was just thinking will there be a situation like that. Because getNodeForBucketWrite is a blocking call only afaik so either it will get or that call will hang. Isn't it?

sumwale: Not that. Spark will reschedule the partitions if cores are busy on primary after sometime.

kneeraj: Ok. Thanks.

PR #854, opened 2017-09-15
SNAP-2031 -- q77 of tpcds was failing

kneeraj:

Fixed a parser issue where AS can be optional in namedExpression rule. q77 of tpcds was failing.

## Patch testing

added a unit test.

## ReleaseNotes.txt changes

None.

## Other PRs 

No


kneeraj:
> Its possible that this can introduce ambiguities in the parser.

Why do you say so? I checked the SqlBase.g4 and there the namedExpression has AS as optional.
BTW the tpcds queries all passed and the precheckin is also good. So merging.

PR #857, opened 2017-09-16
[SNAP-2033] use proper sort ordering for update/delete

sumwale:

The update/delete plans expect that input is ordered on batchIds in each
partition to minimize delta creation since they will end the previous
delta when a new batchId run-length is seen.

- instead of explicit check for Exchange in the plan to add sort on batchId
  for above, use SparkPlan's outputOrdering and requiredChildOrdering for the
  same which will be handled by EnsureRequirements automatically
- ColumnTableScan sends out ordering as that on batchId for update/delete case,
  while those plans expect the child to be ordered on the same
- update the condition to introduce exchange; earlier it used to match
  number of child partitions against target table buckets but that introduces
  unnecessary exchange for bucket pruning; now carrying the table buckets in
  OrderlessHashPartitioning and comparing that before introducing exchange
- skip DelegateRDD wrapping (for preferred locations) for the case of child having
  partition pruning since exchange will not be there in such a case

With above changes the second large update in SNAP-1914 takes just ~8mins as opposed
to over an hour before on a single AWS machine (gp2 disk).

## Patch testing

precheckin, manual TPCH testing with large number of updates as mentioned in SNAP-1914

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
@@ -19,13 +19,12 @@ package io.snappydata
 import java.io.File
 
 import io.snappydata.impl.LeadImpl
-import org.apache.hadoop.conf.Configuration
 
+import org.apache.spark.SparkContext
 import org.apache.spark.sql.catalyst.expressions.{Attribute, Expression}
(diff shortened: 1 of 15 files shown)


rishitesh:
@sumwale I don't understand the partition pruning part much. Rest all look goo to go.

PR #858, opened 2017-09-16
[SNAP-2034] remove column batch from region if all rows of batch are deleted

sumwale:

- using CacheListener implementation on the server-side to determine the case
  of full batch deletion after the merge of deltas (delete deltas already have
      original batch count);
  ColumnFormatEncoder itself implements the CacheListener as expected by the
  changes on store side to avoid adding more internal clauses to grammar
- also check if first delete delta put itself causes entire batch to be deleted
  then fire a delete directly instead of putting delete deltas
- renamed "tableName" to "columnTableName" in JDBCSourceAsColumnarStore to clear
  ambiguities in table names being used in the class

Fix for two issues seen in manual testing by @kneeraj
- ColumnTableScan uses only the delete count to skip iteration if no columns are
  being projected (e.g. select count(*)); but it also was happening for the
  "delete from table" kind of query since it projects only internal columns;
  now moved the deleteCountCheck/deleteCheck below to check for output rather
  than inside the projected column code generation loop
- Decrement deleted row count from SnappyTableStatsProviderService iteration
  by accounting in ColumnFormatKey.getColumnBatchRowCount. Removed the TODO
  there about that and added a TODO that the iterator should be transactional.
- Added test for "delete from table" to ColumnUpdateDeleteTests (issue does not
    reproduce) and also to airline table in TokenizationTest where it reproduces

## Patch testing

precheckin, manual testing; unit tests will be added soon

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/317

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnDelta.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnDelta.scala
@@ -19,7 +19,7 @@ package org.apache.spark.sql.execution.columnar.impl
 
 import java.nio.ByteBuffer
 
-import com.gemstone.gemfire.cache.{EntryEvent, Region}
+import com.gemstone.gemfire.cache.{EntryEvent, EntryNotFoundException, Region}
 import com.gemstone.gemfire.internal.cache.delta.Delta
 import com.gemstone.gemfire.internal.cache.versions.{VersionSource, VersionTag}
 import com.gemstone.gemfire.internal.cache.{DiskEntry, EntryEventImpl}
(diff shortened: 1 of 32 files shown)


sumwale:
> ColumnRegion.destroy will update the stats. Right?

Yes.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnDelta.scala
```diff
@@ -165,4 +165,50 @@ object ColumnDelta {
...
+        case _: EntryNotFoundException => // ignore
+      }
+    }
+
+    val numColumns = key.getNumColumnsInTable(columnTableName)
+    // delete the stats rows first
+    destroyKey(key.withColumnIndex(ColumnFormatEntry.STATROW_COL_INDEX))
+    destroyKey(key.withColumnIndex(ColumnFormatEntry.DELTA_STATROW_COL_INDEX))
+    // column values and deltas next
+    for (columnIndex <- 1 to numColumns) {
```
rishitesh: What will happen if the cache is getting closed when delete is inside this loop ? Some entries will be left behind ?

sumwale: This entire set of operations is under the same transaction as original operation, so if operation fails due to node failure then ideally entire thing should be rolled back. We are not there yet for persistent transactions of course. Otherwise do you have any solution?

rishitesh: That's fine. I somehow got confused it won't be under transaction.

PR #873, opened 2017-10-09
Test to identify the recovery time with persistent row tables (10GB, 20GB and 50GB data size)

swatisawant:

- Added a test to identify the recovery time with row tables (10GB, 20GB and 50GB data size)
- Changes from Sumedh to avoid redirecting output twice while executing scripts in framework
- Removed  '-bg' option added as a default parameter while using snappy-start-all.sh script

## Patch testing

Verified the changes by Running clusterRestartWithPersistentRecovery.bt

## ReleaseNotes.txt changes

NA

## Other PRs 

No


dtests/src/test/java/io/snappydata/hydra/cluster/SnappyTest.java
```diff
@@ -2353,6 +2358,12 @@ public int getMyTid() {
   public static synchronized void HydraTask_startSnappyCluster() {
     if (forceStart) {
       startSnappyCluster();
+      try {
+        Thread.sleep(120000);
```
rishitesh: Why sleep ? Can't we check for status file and fail after certain time.

swatisawant: This is to avoid the stopSnappy cluster immediately after restart and verify the recovery in UI.

kneeraj:
Recovery time determination with row tables should also contains these cases.

1. One test case should have few indexes on the row tables. Try having single column and composite column indexes.
2. Have an option to switch On/Off PERSIST-INDEXES flag in the test [ by default it is ON ]. This is important because the product tries to rebuild all the indexes on startup if it finds that the index and tables are out of sync. We need to time this as well.
3. I think the tables are only persistent which are being used in the test. We need to definitely add eviction in the mix in one of the variations. Memory configured should be such that quite a good amount of data should be on disk. I mean not everything should be in memory. 
4. On recovery apart from recording the timings, we should also fire some diagnostics queries to see whether the primary buckets and secondary buckets and their corresponding indexes are all contaning data as expected.

swatisawant:
1. This test contains indexes(composite column indexes) on the row tables. I will add single column indexes test.
2. Will add an option to switch On/Off PERSIST-INDEXES flag in the test
3.  With 50 GB data test, quite a good amount of data was on disk.  I will add a test with eviction in the mix(persistent  and non- persistent tables)
4. I will add some diagnostics queries to see whether the primary buckets and secondary buckets and their corresponding indexes are all containing data as expected.

swatisawant:
Yes. I will add a test to achieve this (with eviction in the mix of persistent and non- persistent tables)

PR #876, opened 2017-10-11
Modified the parser to understand FETCH FIRST syntax also.

kneeraj:

So with this parser change along with existing syntax of LIMIT snappy will also understand FETCH FIRST syntaxx. Looked like an easier option to add for tools which explore data using this syntax rather than LIMIT.

## Patch testing

Added a unit test. Will be committing it soon. Precheckin underway.

## ReleaseNotes.txt changes

No.

## Other PRs 

No.


core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -679,6 +680,10 @@ class SnappyParser(session: SnappySession)
     }
   }
 
+  protected final def fetchExpression: Rule1[Expression] = rule {
+    FETCH ~ FIRST ~ TOKENIZE_END ~ expression ~ ((ROW|ROWS) ~ ONLY) ~> ((f: Expression) => f)
```
vivekwiz: Just curious, can we replace 'expression' with more specific i.e. numericLiteral? What error do we get now if limit/fetch-first value we get now is -ve or  double?

core/src/test/scala/org/apache/spark/sql/store/TokenizationTest.scala
```diff
@@ -305,6 +305,19 @@ class TokenizationTest
...
+      // test fetch first syntax also
+      // Cache map size doesn't change because fetch first converts into limit
+      // itself and the exact same thing is fired above
+      query = s"select * from $table where a = 0 fetch first 1 row only"
+      res1 = snc.sql(query).collect()
+      assert( cacheMap.size() == 4)
+
+      // Cache map size doesn't change because fetch first converts into limit
+      // itself and the exact same thing is fired above
+      query = s"select * from $table where a = 0 fetch first 10 rows only"
```
vivekwiz: Query can have order by and value of the rows returned should be verified.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
kneeraj: Since the change is more like providing an alias to LIMIT clause so I used the same rule of expression which LIMIT is using.

sumwale: @vivekwiz LIMIT in Spark allows for arbitrary expressions that evaluate to a constant integer.

core/src/test/scala/org/apache/spark/sql/store/TokenizationTest.scala
kneeraj: Done.

Issue #862, opened 2017-09-19
streaming registerCQ miss records to insert when changing duration seconds and slide seconds


niko2014:
# Env:
snappydata-0.9
-J-Dspark.streaming.backpressure.enabled=true
1 locator, 3 dataserver, 1 leader
stream Table created with `using kafka_stream`
Kafka version : `0.10.1.0` 

# code

I read from streamTableA `(duration 6 seconds, slide 6 seconds)` and write to columnTableA.

```
        val dStream = snsc.registerCQ("select * from streamTableA window (duration 6 seconds, slide 6 seconds)");
        import snsc.snappySession.sqlContext.implicits._
        dStream.foreachDataFrame(df => {
            df.as[SnappyImageRow].write.insertInto("columnTableA")
        })
```

# before streaming started : 

There are `1000000 * 3 = 3,000,000` records to consume (3 partition , each with 1000000 records )

kafka offset : 
```
[2017-09-19 11:02:48,916] WARN WARNING: ConsumerOffsetChecker is deprecated and will be dropped in releases following 0.9.0. Use ConsumerGroupCommand instead. (kafka.tools.ConsumerOffsetChecker$)
Group           Topic                          Pid Offset          logSize         Lag             Owner
snappyStreaming01 stored_capture_features_01     0   220505899       221505899       1000000         none
snappyStreaming01 stored_capture_features_01     1   219352291       220352291       1000000         none
snappyStreaming01 stored_capture_features_01     2   219307305       220307305       1000000         none
```

columnTableA row count: `25,034,770`

# streaming ended

kafka offset : 
```
[2017-09-19 11:24:25,679] WARN WARNING: ConsumerOffsetChecker is deprecated and will be dropped in releases following 0.9.0. Use ConsumerGroupCommand instead. (kafka.tools.ConsumerOffsetChecker$)
Group           Topic                          Pid Offset          logSize         Lag             Owner
snappyStreaming01 stored_capture_features_01     0   221505899       221505899       0               snappyStreaming01_node78-1505791154261-790ae794-0
snappyStreaming01 stored_capture_features_01     1   220352291       220352291       0               snappyStreaming01_node78-1505791154261-790ae794-0
snappyStreaming01 stored_capture_features_01     2   220307305       220307305       0               snappyStreaming01_node78-1505791154261-790ae794-0
```

columnTableA row count: `25,722,888`

# consumed records != inserted row count

columnTableA row count change :
[...]

PR #882, opened 2017-10-20
Kafka 0.10 APIs + Structured streaming test framework

ymahajan:

- Removed Kafka 0.8 APIs from the product. 
- Upgraded product/test code to use Kafka 0.10 APIs   
- Retained both old SchemaDStream/CQ and new structured streaming
- Both streaming modules are based on Kafka 0.10 APIs now
- Fixed an issue reported here https://github.com/SnappyDataInc/snappydata/issues/862
- Cleaned up old unit tests, added more unit tests

## Patch testing

precheckin 

## ReleaseNotes.txt changes

Yes

## Other PRs
 
AQP - https://github.com/SnappyDataInc/snappy-aqp/pull/165


sumwale:
@ymahajan Before reviewing, can we discuss a bit on the implications for users? Are there users using Kafka 0.8 with old streaming model with Spark 2.x? Would it not be better to retain both the models for sometime because Spark still has both. Have separate modules for the two like in Spark.
@jramnara

ymahajan:
@sumwale @jramnara I don't see much value in maintaining three different streaming implementations. Kafka 0.8 is quite old now, even spark's structured streaming doesn't use it. It has been only implemented for Kafka 0.10. Spark's old streaming has separate external modules for 0.8 and 0.10. Combining them for us in product has it's own challenges in bundling, maintaining and adding test cases for both versions for old streaming, which anyways will be removed from the product eventually.

jramnara:
Less complex the better. Do we have any stats on which version customers are using most? I suppose, 0.8 kafka is sufficient. Do some google searches.

ymahajan:
@jramnara , Based on looking at stackoverflow and other forums, most of the kafka users are using 0.10, 0.11, 1.0 etc. And this 0.10 release is about to become almost 2 years stale now. Even structured streaming doesn't support anything less than 0.10 version. Also using 0.8 with spark streaming we needed to support both kafka and direct kafka streaming receiver. 

@rishitesh Jags wants to keep both streaming engine support for now. May be when we release structured streaming we will deprecate the legacy SchemaDStream, but for now we cannot remove it altogether. Also the legacy streaming supports features like stream to stream join which are not yet implemented in structured streaming.

core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
```diff
@@ -48,31 +53,112 @@ final class DirectKafkaStreamRelation(
...
+        })
+      }
+    stream
+  }
+}
+
+/**
+  * Copied from org.apache.spark.sql.kafka010.JsonUtils
+  */
+private object JsonUtils {
```
rishitesh: JsonUtil can be a separate class. Only saw usage in test classes. If so we can mode it to test packages.

PR #892, opened 2017-11-03
[SNAP-2084] Handled dropStorageMemoryForObject in DefaultMemoryManager.

rishitesh:

We keep a map of memory used against each region. When we drop the region we release the total memory accumulated. This is not supported in DefaultMemoryManager( which gets used in local & connector mode). Hence the problem.
While acquiring memory for table it uses Spark's UMM , but does not release that memory when the table gets dropped.

## Patch testing

manual test. Pre-checkin pending

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
+++ b/core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) 2017 SnappyData, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+package org.apache.spark.memory
+
+import java.nio.ByteBuffer
+
+import com.gemstone.gemfire.internal.snappy.UMMMemoryTracker
+import com.gemstone.gemfire.internal.snappy.memory.MemoryManagerStats
+import io.snappydata.collection.ObjectLongHashMap
+
+import org.apache.spark.{Logging, SparkEnv}
+import org.apache.spark.storage.BlockId
+
+
+class DefaultMemoryManager extends StoreUnifiedManager with Logging {
+
+  private val memoryForObject: ObjectLongHashMap[(String, MemoryMode)]
+  = ObjectLongHashMap.withExpectedSize[(String, MemoryMode)](16)
+
+
+  private val managerId = "DefaultMemoryManager"
+
+  override def acquireStorageMemoryForObject(objectName: String,
+      blockId: BlockId,
+      numBytes: Long,
+      memoryMode: MemoryMode,
+      buffer: UMMMemoryTracker,
+      shouldEvict: Boolean): Boolean = {
+    logDebug(s"Acquiring DefaultManager memory for $objectName $numBytes")
+    if (SparkEnv.get ne null) {
+      val success = SparkEnv.get.memoryManager.acquireStorageMemory(blockId, numBytes, memoryMode)
+      memoryForObject.addTo(objectName -> memoryMode, numBytes)
+      success
+    } else {
+      true
+    }
+  }
+
+  // This is not be called in connector mode. This api is called from
+  // region.clear hence only applicable for local mode
+  override def dropStorageMemoryForObject(
+      objectName: String,
+      memoryMode: MemoryMode,
+      ignoreNumBytes: Long): Long = {
+    if (SparkEnv.get ne null) {
+      val key = objectName -> memoryMode
+      val bytesToBeFreed = memoryForObject.getLong(key)
+      val numBytes = Math.max(0, bytesToBeFreed - ignoreNumBytes)
+      logDebug(s"Dropping $managerId memory for $objectName =" +
+          s" $numBytes (registered=$bytesToBeFreed)")
+      if (numBytes > 0) {
+        SparkEnv.get.memoryManager.releaseStorageMemory(numBytes, memoryMode)
+        memoryForObject.removeAsLong(key)
+      }
+    }
+    0L
+  }
+
+  override def releaseStorageMemoryForObject(
+      objectName: String,
+      numBytes: Long,
+      memoryMode: MemoryMode): Unit = {
+    logDebug(s"Releasing DefaultManager memory for $objectName $numBytes")
+    if (SparkEnv.get ne null) {
+      SparkEnv.get.memoryManager.releaseStorageMemory(numBytes, memoryMode)
+      val key = objectName -> memoryMode
+      if (memoryForObject.containsKey(key)) {
+        if (memoryForObject.addTo(key, -numBytes) == numBytes) {
+          memoryForObject.removeAsLong(key)
+        }
+      }
+    }
+  }
+
+  override def getStoragePoolMemoryUsed(memoryMode: MemoryMode): Long = 0L
+
+  override def getStoragePoolSize(memoryMode: MemoryMode): Long = 0L
+
+  override def getExecutionPoolUsedMemory(memoryMode: MemoryMode): Long = 0L
+
+  override def getExecutionPoolSize(memoryMode: MemoryMode): Long = 0L
+
+  override def getOffHeapMemory(objectName: String): Long = 0L
+
+  override def hasOffHeap: Boolean = false
+
+  override def logStats(): Unit = logInfo("No stats for NoOpSnappyMemoryManager")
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
```diff
@@ -0,0 +1,116 @@
...
+  override def acquireStorageMemoryForObject(objectName: String,
+      blockId: BlockId,
+      numBytes: Long,
+      memoryMode: MemoryMode,
+      buffer: UMMMemoryTracker,
+      shouldEvict: Boolean): Boolean = {
+    logDebug(s"Acquiring DefaultManager memory for $objectName $numBytes")
+    if (SparkEnv.get ne null) {
+      val success = SparkEnv.get.memoryManager.acquireStorageMemory(blockId, numBytes, memoryMode)
+      memoryForObject.addTo(objectName -> memoryMode, numBytes)
```
sumwale: better to store SparkEnv.get volatile value in a val rather than calling twice (just for the case where executor is just being closed and this can throw NPE in second call)

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
```diff
@@ -0,0 +1,116 @@
...
+      memoryMode: MemoryMode,
+      ignoreNumBytes: Long): Long = {
+    if (SparkEnv.get ne null) {
+      val key = objectName -> memoryMode
+      val bytesToBeFreed = memoryForObject.getLong(key)
+      val numBytes = Math.max(0, bytesToBeFreed - ignoreNumBytes)
+      logDebug(s"Dropping $managerId memory for $objectName =" +
+          s" $numBytes (registered=$bytesToBeFreed)")
+      if (numBytes > 0) {
+        SparkEnv.get.memoryManager.releaseStorageMemory(numBytes, memoryMode)
```
sumwale: better to store SparkEnv.get volatile value in a val rather than calling twice; also sync for memoryForObject on SparkEnv's memoryManager

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
```diff
@@ -0,0 +1,116 @@
...
+    0L
+  }
+
+  override def releaseStorageMemoryForObject(
+      objectName: String,
+      numBytes: Long,
+      memoryMode: MemoryMode): Unit = {
+    logDebug(s"Releasing DefaultManager memory for $objectName $numBytes")
+    if (SparkEnv.get ne null) {
+      SparkEnv.get.memoryManager.releaseStorageMemory(numBytes, memoryMode)
```
sumwale: better to store SparkEnv.get volatile value in a val rather than calling twice

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
sumwale: memoryForObject doesn't need to be synchronized?

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
```diff
@@ -0,0 +1,116 @@
...
+
+  // This is not be called in connector mode. This api is called from
+  // region.clear hence only applicable for local mode
+  override def dropStorageMemoryForObject(
+      objectName: String,
+      memoryMode: MemoryMode,
+      ignoreNumBytes: Long): Long = {
+    if (SparkEnv.get ne null) {
+      val key = objectName -> memoryMode
+      val bytesToBeFreed = memoryForObject.getLong(key)
```
sumwale: memoryForObject should be synchronized on SparkEnv's memoryManager

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
```diff
@@ -0,0 +1,116 @@
...
+      numBytes: Long,
+      memoryMode: MemoryMode): Unit = {
+    logDebug(s"Releasing DefaultManager memory for $objectName $numBytes")
+    if (SparkEnv.get ne null) {
+      SparkEnv.get.memoryManager.releaseStorageMemory(numBytes, memoryMode)
+      val key = objectName -> memoryMode
+      if (memoryForObject.containsKey(key)) {
+        if (memoryForObject.addTo(key, -numBytes) == numBytes) {
+          memoryForObject.removeAsLong(key)
+        }
```
sumwale: sync

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
```diff
@@ -0,0 +1,116 @@
...
+  override def shouldStopRecovery(): Boolean = false
+
+  override def initMemoryStats(stats: MemoryManagerStats): Unit = {}
+
+  override def close(): Unit = {}
+
+  /**
+    * Clears the internal map
+    */
+  override def clear(): Unit = {}
```
sumwale: Have you verified that none of these empty methods are invoked from the store layer? Perhaps they don't need to be in interface at all or else at least some of them should return proper values reading from SparkEnv's memoryManager (like offheap related ones).

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
rishitesh: Ok. Will change.

rishitesh: Will change.

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
rishitesh: Ok. Will change.

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
rishitesh: Will change.

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
rishitesh: The empty methods means they are only called from store layer, and makes sense in Snappy UMM. As default will be used in environment where store is not there, hence empty.

sumwale: Isn't SNAP-2084 for the local[*] mode where SnappyUMM may not have been explicitly configured? For that case callbacks from store will happen. For example whether off-heap has been configured or not which will effect the BufferAllocator.

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
```diff
@@ -0,0 +1,119 @@
...
+      memoryMode: MemoryMode,
+      buffer: UMMMemoryTracker,
+      shouldEvict: Boolean): Boolean = synchronized {
+    logDebug(s"Acquiring DefaultManager memory for $objectName $numBytes")
+    val env = SparkEnv.get
+    if (env ne null) {
+      val success = env.memoryManager.acquireStorageMemory(blockId, numBytes, memoryMode)
+      memoryForObject.addTo(objectName -> memoryMode, numBytes)
+      success
+    } else {
```
sumwale: Wouldn't it be better to do a sync on env.memoryManager because its call already uses the same?

```scala
  override def acquireStorageMemoryForObject(objectName: String,
      blockId: BlockId,
      numBytes: Long,
      memoryMode: MemoryMode,
      buffer: UMMMemoryTracker,
      shouldEvict: Boolean): Boolean = {
    logDebug(s"Acquiring DefaultManager memory for $objectName $numBytes")
    val env = SparkEnv.get
    if (env ne null) {
     env.memoryManager.synchronized {
       val success = env.memoryManager.acquireStorageMemory(blockId, numBytes, memoryMode)
       memoryForObject.addTo(objectName -> memoryMode, numBytes)
       success
    }
   } else {
```

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
```diff
@@ -0,0 +1,119 @@
...
+    if (env ne null) {
+      val key = objectName -> memoryMode
+      val bytesToBeFreed = memoryForObject.getLong(key)
+      val numBytes = Math.max(0, bytesToBeFreed - ignoreNumBytes)
+      logDebug(s"Dropping $managerId memory for $objectName =" +
+          s" $numBytes (registered=$bytesToBeFreed)")
+      if (numBytes > 0) {
+        env.memoryManager.releaseStorageMemory(numBytes, memoryMode)
+        memoryForObject.removeAsLong(key)
+      }
```
sumwale: see above comment about sync

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
sumwale: see above

core/src/main/scala/org/apache/spark/memory/DefaultMemoryManager.scala
rishitesh: You mean to say if a local mode application configures offheap. Ok. In such scenarios we might need these APIs.

PR #895, opened 2017-11-08
[SNAP-2101] smart connector performance fixes and related issues

sumwale:

- update store link to SNAP-2101 having the fixes to PK and elimination of buffer copies
- adding partitioningColumns to ColumnPartitionResolver to enable Hash1Index creation
- fix SnappyUMM with smart connector:
  - region evictor should be nop instead of failing due to no GemFireCache
  - old entries cleaner should not be invoked when no GemFireCache
  - try to load ManagedBufferAllocator in smart connector mode too so that
    the temporary scan buffers can use the same (else will continue to use
        unaccounted ByteBuffer.allocateDirect with a log warning)
- fix SmartConnectorColumnRDD serialization with kryo
- make IF a reserved keyword to match the corresponding change in store parser
- added asserts in smart connector column insert/scan that the received chunk
  should not be a "reference" (i.e. coming from ColumnFormatValue) rather
  owned by itself as received using the thrift channels

With above changes (+ the store ones), the performance for smart connector mode
in ColumnCacheBenchmark has improved by more than 2X and now within expected
range: from 12-13ns per row to 5-6ns per row. It is now 3-4X faster than Spark
caching and 2-3X faster than direct Parquet scan having compression=none and
entirely in OS buffers.

## Patch testing

precheckin -Pstore

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/336


core/src/main/scala/org/apache/spark/sql/execution/ddl.scala
```diff
@@ -75,7 +75,7 @@ private[sql] case class DropTableCommand(ifExists: Boolean,
   override def run(session: SparkSession): Seq[Row] = {
     val snc = session.asInstanceOf[SnappySession]
     val catalog = snc.sessionState.catalog
-    snc.dropTable(catalog.newQualifiedTableName(tableIdent), ifExists, resolveRelation = false)
+    snc.dropTable(catalog.newQualifiedTableName(tableIdent), ifExists, resolveRelation = true)
```
rishitesh: Why we needed to change this ?

sumwale: Was missed in previous fix. The "resolveRelation=false" route is only for smart connector mode when the required classes for an external table may not be present on embedded-side so relation should not be attempted to be materialized.

sumwale:
@rishitesh Without SnappyUMM it will fallback to the "DefaultMemoryManager" you have for buffer allocations. Which was one of the reasons for my comments on that PR. Will add clarifications on that PR about "changeOffHeapOwnerToStorage".

PR #896, opened 2017-11-18
Fix the case of remote pull from smart connector

sumwale:

- SparkConnectorRDDHelper.scala: changed the UNION ALL query to a much smaller
  IN query that is now correctly converted to getAll on store-side (note the
      use of "select *" rather than projection that still cannot be dealt by store)
- SparkConnectorRDDHelper.scala: added the optimized path for thrift connection/statement
  to set properties on it rather than making procedure call round-trips;
- SparkConnectorRDDHelper.scala: also removed obsolete code used by old accessor/peer
  smart connector; copied its code for "orphanBuckets" i.e. data nodes having no network server
- removed some unnecessary commit() calls since snapshot commit is explicitly invoked
- make bucket sizes consistent in embedded and smart connector; these now always
  go into the table properties else the defaults can be interpreted differently
  in embedded vs smart connector
- fixed case-sensitivity of table properties and table name making it consistent (case-insensitive)
- renamed ExecutorMultiBucketLocalShellPartition to SmartExecutorBucketPartition and made
  it kryo serializable
- reduced number of default buckets for a localhost-only cluster
- increased executor locality wait by default in embedded mode

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/337

### DIFF
--- a/cluster/src/dunit/scala/io/snappydata/cluster/SplitSnappyClusterDUnitTest.scala
+++ b/cluster/src/dunit/scala/io/snappydata/cluster/SplitSnappyClusterDUnitTest.scala
@@ -66,6 +66,8 @@ class SplitSnappyClusterDUnitTest(s: String)
   }
 
   def testCreateTablesFromOtherTables(): Unit = {
+    // stop a network server to test remote fetch
+    vm0.invoke(classOf[ClusterManagerTestBase], "stopNetworkServers")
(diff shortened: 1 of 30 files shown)


sumwale:

PR #897, opened 2017-11-21
SNAP-2110

kneeraj:

Modified the log collection script to filter out unwanted lines from logs.
A list of patterns can be specified in the conf file and lines matching those patterns will not be collected by the script.

## Patch testing

Manual.

## ReleaseNotes.txt changes

None.

## Other PRs 

No.


cluster/sbin/collect-debug-artifacts.sh
```diff
@@ -436,6 +439,20 @@ function collect_on_remote {
     fi
   done
 
+  # remove all unwanted lines from logs
+  if [ ! -z "${invert_pattern}" ] ; then
```
sumwale: Can be "-n" instead of "! -z"

cluster/sbin/collect-debug-artifacts.sh
```diff
@@ -436,6 +439,20 @@ function collect_on_remote {
...
+    cd "${tmp_dir}"
+    for i in `ls *.log*`
+      do
+        invert_log_file_name=$i".inv"
```
sumwale: Double quote should be around whole to be effective: "$i.inv"

Though the for loop above will mess up any spaces in file names already. You will need something like:

```
for i in `ls *.log* | sed 's/^/"/;s/$/"/'`; do
```
to surround each filename in double quotes explicitly

sumwale:
@kneeraj do you want to get this in for 1.0.1?

PradeepSurale:

cluster/sbin/collect-debug-artifacts.sh
bisenpiyush: Modified as suggested

PR #911, opened 2017-12-15
[SNAP-338] initialize the hive catalog in background

sumwale:

- acquire the DLock from inside initHMC()
- catalog initialization has two phases:
  - first init is creating hive client
  - second is checking for catalog consistency (only on lead); this second phase itself
    has to wait for first phase to complete;GemFireStore.getExternalCatalog(boolean fullInit)
    refers to only first phase or both depdending on whether "fullInit" is true
- changes to move catalog initialization to background
- moved SnappyHiveCatalog.SKIP_HIVE_TABLE_CALLS to store HiveTablesVTI; fixes an NPE where
  HiveTablesVTI tries to initialize SnappyHiveCatalog from inside hive-metastore thread itself

## Patch testing

precheckin -Pstore

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/346


core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
```diff
@@ -222,11 +237,11 @@ public String catalogSchemaName() {
   }
 
   @Override
-  public void stop() {
+  public void close() {
     HMSQuery q = getHMSQuery();
     q.resetValues(HMSQuery.CLOSE_HMC, null, null, true);
     try {
-      this.hmsQueriesExecutorService.submit(q).get();
+      this.hmsQueriesExecutorService.submit(q).get(5, TimeUnit.SECONDS);
```
kneeraj: Is it ok to ignore exception here as it might timeout now?

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -86,7 +86,13 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
 
   val sparkConf: SparkConf = snappySession.sparkContext.getConf
 
-  private[sql] var client = metadataHive
+  private var _client = metadataHive
+
+  private[sql] def client = {
+    // check initialized meta-store (including initial consistency check)
+    Misc.getMemStoreBooting.getExternalCatalog
```
kneeraj: Any special reason to force initialize here? First access anyways initializes it.

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
sumwale: This is during shutdown so any exceptions are ignored. Timeout should be fine because node is stopping in any case. Primarily to avoid blocking shutdown here for long (and sometimes it is the Cancellation that triggers any blocked operation to give up).

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
sumwale: Primarily to ensure that catalog consistency check is complete before any catalog reads/writes. The catalog consistency check has been moved to background task as part of these changes and don't block the startup, so on first use the check has to ensured complete (or else throw an exception if there is still some unexpected inconsistency that could not be repaired).

PR #912, opened 2017-12-16
[SNAP-338] new quick launcher and background jobserver start by default

sumwale:
Primarily two changes:
* a faster launcher that avoids loading any other classes (other than gemfire-shared and JNA)
* jobserver startup (and thus the global SnappySession initialization) in background

- added a new QuickLauncher class having minimal dependencies (gemfire-shared)
  to reduce the startup time of the launcher itself in start/stop/status
  (the main locator/server/lead process is a separate one in any case)
- separate "snappy-launcher" project for above and support in snappy script
  with an option in snappy-env.sh to override with "SNAPPY_NO_QUICK_LAUNCH"
- refactored LeadImpl and related code to start snappy-store initialization in
  parallel with Spark initialization
- mark lead as RUNNING on lead after Spark initialization and start jobserver
  in background by default
- a property "jobserver.waitForInitialization" to force wait for jobserver
  initialization before marking the status as RUNNING
- added "withHiveExceptionHandling" for catalog calls where it was missing;
  avoid HiveTableVTI lookup from hive metastore client itself
- fixed some issues in the new QuickLauncher:
  - print the correct log file being used eventually if a conf/log4j.properties
    has been configured
  - skip printing other DistributedMembers in the cluster in the status message
  - skip printing the stacktrace if server is already running (SNAP-1645)

## Patch testing

precheckin -Pstore

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/347


sumwale:
Merging now. Will take care of any review comments subsequently.

 Revert

PR #918, opened 2017-12-27
[SNAP-2127] use separate delta disk-stores for row buffer regions

sumwale:
Follow up PR for SNAP-2121 as per review comments

- if a disk-store has been defined for a column table, then use the corresponding
  delta disk-store instead of a global one
- update store link for the store layer changes

## Patch testing

precheckin

## ReleaseNotes.txt changes

Note the new implicit delta disk stores in the docs.

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/350


kneeraj:
LGTM. Still looking at store side changes.

PR #922, opened 2018-01-05
Snap 2114

kneeraj:

No plan caching is attempted for any sort of external tables now.

## Patch testing

Added a junit test to verify the same

## ReleaseNotes.txt changes

None.

## Other PRs 

No.


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -2059,6 +2059,14 @@ object SnappySession extends Logging {
...
       throw new EntryExistsException("uncached plan", cdf)
     }
+
+    // Try plan caching for only snappy tables
+    if (executedPlan.find(t => t match {
+      case dsc: DataSourceScanExec => !dsc.relation.isInstanceOf[DependentRelation]
+      case _ => false
+    }).isDefined) {
+      throw new EntryExistsException("uncached plan", cdf)
+    }
```
sumwale: Can you combine with the search above instead of a separate searches?

```scala
executedPlan.find {
  case _: InMemoryTableScanExec => true
  case ...
}
```

kneeraj: I initially thought so but decided against it for better code clarity. Anyways if plan caching works then this be just done once. For non plan caching cases this is a very small cost. Can do it still if you feel so.

kneeraj: Ok, will do. Just saw the code again. Code clarity won't be affected.

sumwale: These small costs have added up to reduce the best performance of point queries from ~2ms at the time of 0.7-perf merge to ~3ms now, so best to avoid this. I don't see any loss of code clarity with merging the two. Add comments in individual matches statements if you think so.

sumwale:
@kneeraj This also means that if user is joining snappy table with an external table, for example, or in-memory cache table, then tokenization won't happen. We should track why tokenization is not working with external/in-memory tables.

kneeraj:
We can track that.
BTW why I thought of disabling it entirely for all external tables is that for each separate kind of external type the implementation of scan/buildScan can be different and if it caches some value which is what I thought is happening in the gemtable scan.
Other thing is that for most of the external tables the data fetch will be so expensive that this won't matter much.
But yes I was not comfortable with cached table and a parquet external table too. But went ahead for correctness as row table, column table and aqp tables are completely in our control and not other types.

sumwale:
Ok, we can enable selectively for external sources and in-memory cache later.

kneeraj:
Filed SNAP-2160 -- https://jira.snappydata.io/browse/SNAP-2160

PR #923, opened 2018-01-05
Snap 2125

kneeraj:

Added a setter to disable plan caching per session. All can be used to disable across session. Will be completed in implementation if agreed upon. Very small change will be required.

## Patch testing

Added a test for that.

## ReleaseNotes.txt changes

Plan caching can be disabled per session if required.

## Other PRs 

None.


core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -508,6 +511,14 @@ abstract class SnappyDDLParser(session: SparkSession)
     SET ~ (
         CURRENT.? ~ SCHEMA ~ '='.? ~ ws ~ identifier ~>
             ((schemaName: String) => SetSchema(schemaName)) |
+        DISABLE ~ PLAN ~ CACHING ~ (ALL ~ push(true)).? ~>
+            ((all: Any) => {
+              val allSessions = all match {
```
rishitesh: can be simplified by "all.isDefined"

rishitesh: val allSessions = all.isDefined

rishitesh:
IMO disabling across all won't be a great idea. What would be the use case where one will need that. 
One minor comment. LGTM

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -508,6 +511,14 @@ abstract class SnappyDDLParser(session: SparkSession)
...
+                case None => false
+                case Some(_) => true
+              }
+              SetDisablePlanCaching(allSessions)
+            }) |
```
sumwale: Why a new parser rule? IMO "SET DISABLE PLAN CACHING" looks odd in either case. Why not just have a session property like snappydata.planCaching so "set snappydata.planCaching=true/false" will be fine. You will need to add the key to the list in SnappyConf.keyUpdateActions for which existing plan cache is disabled.

kneeraj: @rishitesh and @swale answer to both questions from you too is same:

i.e. I wanted to provide something like DISABLE ALL.
Let's say there is a system and multiple sessions are there and similar type of queries are being run from different session. Actually most common use case will be like this where different clients are working and issuing similar queries. Now if we found out in a running session that there is something wrong with plan caching of the type of query they are using then they can just get hold of one session and disable it right across. If it is session wise then all the user has to issue this individually on their session.
If we want to maintain it just single session wise then I can remove ALL and adopt the simpler one. Please suggest.

sumwale: @kneeraj Don't quite understand why global setting is related to parser changes. I don't see SetDisablePlanCachingCommand doing anything different for "all" in this PR so it won't work in any case.

For global property you can have a separate property, then set it on SparkConf or System.setProperty in SnappyConf.keyUpdateActions (like already done for SchedulerPool), then the default can read the global setting when initializing SnappySession.disablePlanCaching:

```scala
private[sql] var disablePlanCaching: Boolean = sessionState.conf.getConfString(<sessionKey>,
    sparkContext.conf.get(<globalKey>, "false")).toBoolean
```

For the global case you will also need to traverse all open sessions and change which I guess you were asking. But I think it is reasonable that global setting only applies to all new sessions while all open sessions will have to change theirs explicitly. Otherwise there will be unnecessary overhead of maintaining/cleaning global list or else the disablePlanCaching will have to become a function which is again unnecessary overhead to cater to a very unlikely case.

kneeraj: @swale yes that change is not there because I was reluctant to add global list and was just thinking how else I can do it and for now just go ahead with single session. Yes for the same reason I was asking about how to get the list of all active sessions a few days back.

Ok will change as suggested.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -219,6 +219,10 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
   @transient
   private[sql] var currentKey: SnappySession.CachedKey = _
 
+  @transient
+  private[sql] def planCaching: Boolean = sessionState.conf.getConfString(
+    Property.PlanCaching.name, "true").toBoolean
+
```
sumwale: Why a def and not var? Overhead of conf read multiple times in every parse will further reduce parser performance. Already its performance is down by over 50% (in ColumnTableTest.compare parser performance) compared to 0.9 release which I was planning to look into soon. Yes, this means that "planCachingAll" flag will not effect existing sessions but too high an overhead just for an extreme config option that will be used very rarely, if at all. Please change to val.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1831,6 +1835,8 @@ object SnappySession extends Logging {
   private[this] val ID = new AtomicInteger(0)
   private[sql] val ExecutionKey = "EXECUTION"
 
+  def tokenizationDisabled: Boolean = java.lang.Boolean.getBoolean("DISABLE_TOKENIZATION")
```
sumwale: Same as above. Also this name can be changed to be more in-line with other names (e.g. snappydata.sql.tokenization). Isn't it better to make this a conf property instead of a system property?

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -274,6 +274,14 @@ object Property extends Enumeration {
         "scalability of queries in the interest of reduced memory usage for " +
         "secondary buckets. Default is false.", Some(false), Constant.SPARK_PREFIX)
 
+  val PlanCaching: SQLValue[Boolean] = SQLVal[Boolean](
+    s"${Constant.PROPERTY_PREFIX}planCaching",
+    "Property to set/unset plan caching", Some(true))
+
+  val PlanCachingAll: SQLValue[Boolean] = SQLVal[Boolean](
+    s"${Constant.PROPERTY_PREFIX}planCachingAll",
+    "Property to set/unset plan caching on all sessions", Some(true))
```
sumwale: I think its better to use proper qualified name like spark does for all its properties. For example: snappydata.sql.planCaching and snappydata.sql.planCachingAll

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -499,6 +499,25 @@ class SnappyConf(@transient val session: SnappySession)
...
+        case None =>
+      }
+
+    case Property.PlanCachingAll.name =>
+      value match {
+        case Some(boolval) => {
+          val disable_tokenization = !boolval.toString.toBoolean
+          System.setProperty("DISABLE_TOKENIZATION", disable_tokenization.toString)
+          if (disable_tokenization) SnappySession.getPlanCache.asMap().clear()
+        }
```
sumwale: Why mixing tokenization with planCaching flag? The two are separate things. I think better to make tokenization a separate property (e.g. snappydata.sql.tokenization=true/false) like mentioned in previous comment.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -499,6 +499,25 @@ class SnappyConf(@transient val session: SnappySession)
           pool.toString
         case Some(pool) => throw new IllegalArgumentException(s"Invalid Pool $pool")
       }
+
+    case Property.PlanCaching.name =>
+      value match {
+        case Some(boolval) => {
+          if (boolval.toString.toBoolean) session.clearPlanCache()
+        }
```
sumwale: As mentioned, the planCaching should be made a var instead of def and that should then be updated here.

core/src/main/scala/io/snappydata/Literals.scala
kneeraj: ok. I don't thnk there is any existing property like this put in sql from our side. Will add sql.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: Yes I was not liking it. I thought if I put it as val then the expression won't get evaluated again and again. Will check.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: This is an existing name which I just changed the place from parser to snappysession object. Will do the needful.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
kneeraj: ok.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
kneeraj: Ok. will add the property looks good.
BTW We do tokenize to see if we cache plan or not. So won't they go together. So did not quite understand the comment - "Why these two things are being mixed"

sumwale: Tokenization has significant value even if plan caching is disabled. Because its major benefit lies in being able to produce same generated code for two different logical plans that differ only in the embedded constants. Plan caching is primarily for low latency queries (< 50ms or so), while tokenization helps all kinds of queries (e.g. we have seen first runs of queries that require compilation+JIT to be many times slower, or the TPCH runs with different values where most of the gains are due to tokenization).

kneeraj: Actually I don't understand this part well. I mean generated code is part of the plan which is getting cached or not? So plan caching before Tokenization only helped when the same query ( with the same constants ) ran again and again. If a new query with just different constants made the engine produce new plan and hence new generated code. So yes they are different but if different queries are being run ( even differeing in constants ) then doesn't plan caching and Tokenization in some sense means the same thing.

sumwale: Plan caching is different from code generation cache. Latter is common across plans/sessions which is where the biggest advantage of tokenization comes.

Plan caching reduces the work to recreate LogicalPlan=>optimize=>...=>physical  SparkPlan generation. This is normally only a factor for concurrent short queries (in the range of few ms). The physical plan then generates code which is cached in Spark's CodeGenerator.cache that does exact generated code string lookup. That is where tokenization allows to generate same code.

> So a new query with just different constants made the engine produce new plan and hence new generated code.

Plan cache is for a session. Why would same query or query patterns in different sessions produce different generated code string? That would a critical bug. I had fixed one issue in ColumnTableScan MIN/MAX predicate code where code string would be different due to predicate ordering being unpredictable due to hash map iteration (so had changed to sorted consistent ordering). We need to ensure that generated code across sessions for same query patterns (i.e. differing only in constants) should be identical.

ymahajan: @sumwale, just curious to know if we can have cases where planCache will be a cache hit but for same query code generation cache will be a cache miss. I mean if the sparkPlan/executedPlan is available in planCache will there always be corresponding generated code in Code generation cache. Or size/capacity of these caches will also play a role here. planCache can hold 3000 plans but default value for spark.sql.codegen.cacheSize is 250

sumwale: That's possible. In snappy-spark the Spark codegen cache is 1000

PR #924, opened 2018-01-08
* Set overflow-to-disk as default evict-action unless explicitly set otherwise

ashetkar:
    * By default, column and row tables will have heap-based eviction enabled with overflow-to-disk
      as evict action.
    * By default, sample tables will continue to have eviction disabled.
    * Removed OVERFLOW as DDL option.
    * Users can configure eviction for tables using only the 'eviction_by' option.
    * overflow-to-disk will be the only evict action allowed now.
    * Updated tests using overflow option.

## Patch testing
Precheckin is running.

## ReleaseNotes.txt changes
Tables are now by default created with evictaction as overflow-to-disk

## Other PRs 
https://github.com/SnappyDataInc/snappy-aqp/pull/164


sumwale:
@ashetkar @rishitesh Case a) should be disallowed. We don't want the eviction action to be destroy ever for user tables.

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -417,7 +417,7 @@ object StoreUtils {
     // if OVERFLOW has been provided, then use HEAPPERCENT as the default
     // eviction policy (unless overridden explicitly)
     val hasOverflow = parameters.get(OVERFLOW).map(_.toBoolean)
-        .getOrElse(!isRowTable && !parameters.contains(EVICTION_BY))
+        .getOrElse(!parameters.contains(EVICTION_BY))
```
sumwale: I think this can be changed to just getOrElse(true) because we always want overflow as the only eviction action (and the default one if not explicitly specified).

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -428,6 +428,9 @@ object StoreUtils {
           if (hasOverflow) {
             overflowAdded = true
             s"$GEM_EVICTION_BY $v $GEM_OVERFLOW "
+          } else if (!parameters.contains(OVERFLOW)) {
+            // SNAP-1501 Set overflow as default evict-action unless specified otherwise.
+            s"$GEM_EVICTION_BY $v $GEM_OVERFLOW "
           } else {
             s"$GEM_EVICTION_BY $v "
           }
```
sumwale: Not needed with above mentioned change. Instead you can check that when EVICTION_BY is specified then hasOverflow has to be true i.e. if (hasOverflow) ... else throw Utils.analysisException(...)

core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -444,6 +447,9 @@ object StoreUtils {
...
```
sumwale: same as previous comment

core/src/test/scala/org/apache/spark/sql/store/ColumnTableTest.scala
```diff
@@ -738,6 +736,24 @@ class ColumnTableTest
...
+    snc.sql("DROP TABLE IF EXISTS COLUMN_TEST_TABLE6")
+
+    snc.sql("CREATE TABLE COLUMN_TEST_TABLE6(OrderId INT, ItemId INT) USING row")
+    val r = Misc.getRegionForTable("APP.COLUMN_TEST_TABLE6", true)
+        .asInstanceOf[DistributedRegion]
+    assert(region.getEvictionAttributes.getAlgorithm ===
+        EvictionAlgorithm.LRU_HEAP)
+    assert(region.getEvictionAttributes.getAction ===
+        EvictionAction.OVERFLOW_TO_DISK)
+    snc.sql("DROP TABLE IF EXISTS COLUMN_TEST_TABLE6")
```
sumwale: Can add a check that specifying EVICTION_BY and OVERFLOW=false throws an exception as per comments above.

ashetkar:
>> @ashetkar @rishitesh Case a) should be disallowed. We don't want the eviction action to be destroy ever for user tables.

@rishitesh @sumwale Is this what you are suggesting?
For column table:

OVERFLOW | EVICTION_BY | RESULT
------------ | ------------- | -------------
FALSE | DOES NOT MATTER | < THROW ERROR >
TRUE | DOES NOT MATTER | OVERFLOW_TO_DISK
NOT GIVEN | DOES NOT MATTER | OVERFLOW_TO_DISK

For row table:

OVERFLOW | EVICTION_BY | RESULT
------------ | ------------- | -------------
FALSE | DOES NOT MATTER | LOCAL_DESTROY
TRUE | DOES NOT MATTER | OVERFLOW_TO_DISK
NOT GIVEN | DOES NOT MATTER | OVERFLOW_TO_DISK

sumwale:
Yes, but I am saying the same for both i.e. row table should be same as column table. If there is EVICTION_BY, then overflow=true is the only choice and the default. And default for EVICTION_BY is LRUHEAPPERCENT like currently.

If some specialized use-case does come for local destroy then we can discuss then but from whatever we know so far, it is nothing other than trouble so removing it completely is best.

sumwale:
@ashetkar will need to keep OVERFLOW as an allowed option in DDL for backward compatibility. Check that overflow cannot be false with explicit eviction_by and throw illegal argument exception for that combo.

ashetkar:
Incorporated the review comments. Merging.

PR #925, opened 2018-01-11
[SNAP-2141] fix updates on complex types

sumwale:
Bunch of fixes for complex types and other changes seen in testing for those.

- pass the underlying real encoder to the column writer generated code instead
  of delta encoder which doesn't override to delegate to the real encode for complex
  type writes; also send "nullEncoder" separately which will be the delta encoder
  since null position is written as relative position in delta and not of the base column
- also fixed a code generation issue in DeltaWriter for the case of merge of multiple complex
  type updates
- fixed a major performance problem with LIMIT which would not trim off rows in an
  individual partition if possible; now limit the rows grabbed from individual partitions
  too; also updated the partition "scaling" logic to be as per current Spark code
- provide proper datatype and name over SQL queries by resolving the full names for
  those types recursively; refactored the code in ExternalStoreUtils.getColumnMetadata for that
- make JSON conversion for complex types over SQL routed queries as default so they
  display meaningfully for end-user by default instead of showing BLOB hex bytes
- allow for hashJoinSize and hashAggregateSize to use strings with k/m/g/t suffixes
  instead of always forcing specifying in bytes; also interpret negative hashAggregateSize
  as disabling it completely
- added release of old buffer after expansion in ColumnEncoder
- reduced the minimum batch size for inserting into row buffer DF insert to be half of
  the maximum delta rows limit rather than a tenth of it which can be too small in some cases
  and create many small batches

## Patch testing

Manual testing so far. Unit tests to be added shortly.

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -714,13 +715,16 @@ object CachedDataFrame
       // totalParts in runJob.
       var numPartsToTry = 1L
       if (partsScanned > 0) {
-        // If we didn't find any rows after the first iteration, just try all
-        // partitions next. Otherwise, interpolate the number of partitions
-        // we need to try, but overestimate it by 50%.
+        // If we didn't find any rows after the previous iteration, quadruple and retry.
+        // Otherwise, interpolate the number of partitions we need to try, but overestimate
+        // it by 50%. We also cap the estimation in the end.
+        val limitScaleUpFactor = Math.max(session.sessionState.conf.limitScaleUpFactor, 2)
```
vivekwiz: Can we guard this number to remain positive or zero

sumwale: @vivekwiz The math.max() will make it 2 if negative or zero. Do you mean throw an exception here if the value specified in the conf is <= 1?

vivekwiz: No I wanted to handle negative value from conf. But yes Max will take care.

PR #926, opened 2018-01-11
SNAP-2088 Fix for ArrayIndexOutOfBoundException

ashetkar:
* Updating the null count in case the dictionary array is not null.
* Moving the initialization of dictionaryIndex variable inside if(dictionaryArray != null) block.

## Patch testing
New unit test, manual testing, precheckin

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -797,12 +797,26 @@ private[sql] final case class ColumnTableScan(
...
+            s"""
+               |if ($decoder.isNullAt($buffer, $batchOrdinal)) {
+               |  $dictionaryIndexVar = $dictionaryVar.size();
+               |  $numNullsVar++;
+               |} else {
+               |  $dictionaryIndexVar = $updateDecoder == null
+               |    ? $decoder.readDictionaryIndex($buffer, $nonNullPosition)
+               |    : $updateDecoder.readDictionaryIndex();
+               |}
+               """.stripMargin, "false", dictionaryIndexVar)
```
sumwale: Ideally could also initialize the isNullVar to avoid doing twice but can be noted and done later (or I can take care in my set of changes).

cluster/src/dunit/scala/io/snappydata/externalstore/ColumnTableDUnitTest.scala
```diff
@@ -685,6 +685,38 @@ class ColumnTableDUnitTest(s: String) extends ClusterManagerTestBase(s) {
...
+
+    1 to 500 foreach { i =>
+      if (i % 2 == 0) {
+        snc.sql(s"insert into $t1 values (${Random.nextInt}, 'name_$i', null, 'country_$i')")
+        snc.sql(s"insert into $t2 values (${Random.nextInt}, 'name_$i', null, 'country_$i')")
+      } else {
+        snc.sql(s"insert into $t1 values (${Random.nextInt}, 'name_$i', 'city_$i', 'country_$i')")
+        snc.sql(s"insert into $t2 values (${Random.nextInt}, 'name_$i', 'city_$i', 'country_$i')")
+      }
+    }
```
sumwale: Better to do this in one insert with Spark APIs which also allows inserting much larger set of rows in a reasonable time: `snc.range(10000).selectExpr("cast ((rand() * 100000) as int)", "case when id % 2 = 0 then null else concat("city_", cast(id % 10) as string)").insertInto...`

cluster/src/dunit/scala/io/snappydata/externalstore/ColumnTableDUnitTest.scala
```diff
@@ -685,6 +685,38 @@ class ColumnTableDUnitTest(s: String) extends ClusterManagerTestBase(s) {
...
+    snc.sql(s"select distinct city from $t1").show
+    snc.sql(s"select distinct city from $t2 order by city").show
+    var df = snc.sql(s"select count(*) from $t1 where city is null")
+    var cnt = df.collect()(0).getLong(0)
+    assert(cnt == 250, s"$cnt records found in $t1 with null city, expected 250")
+
+    df = snc.sql(s"select count(*) from $t2 where city is null")
+    cnt = df.collect()(0).getLong(0)
+    assert(cnt == 250, s"$cnt records found in $t2 with null city, expected 250")
+
```
sumwale: Please add a query that compares this count against a group by result (that will have the dictionary optimization applied). For that its better to limit the number of cities (e.g. using i%10 as the suffix): `select count(*) from $t1 group by city`

sumwale:
@ashetkar Can you also run precheckin disabling fallback for code generation exception in CodegenSparkFallback (see lines 60-70 in CodegenSparkFallback.scala). This is just to make sure that this change does not trigger any new code generation failures that get silently ignored (perhaps adding a property to enable/disable it will be good). AFAIK the only known one as of now is AQP-292.

ashetkar:
Ran the precheckin with fallback set to false and found below failures related to code generation.
They do not look caused by the changes here.
 - ColumnCacheBenchmark - "insert more than 64k data" - code grows beyond 64K
 - AQPStreamingSuite.scala
   - SNAP-463 topk tables data with SQL. Case 1 with time interval > batch interval - Line 214 - Expression "scan_isNull" is not an rvalue
   - SNAP-463 topk tables data with SQL. Case 2 with time interval < batch interval - Line 214 - Expression "scan_isNull" is not an rvalue
   - SNAP-463 topk tables data with SQL. Case 3 with time interval = batch interval - Line 214 - Expression "scan_isNull" is not an rvalue
   - SNAP-463 topk tables data with SQL. - without time interval - Line 214 - Expression "scan_isNull" is not an rvalue

sumwale:
AQP failures are likely the same as AQP-292 which is being tracked. ColumnCacheBenchmark wide-table insert should not be dependent on codegen fallback. Is this in inserts or query, and can you check with @rishitesh if that is expected? Does not look related to these changes either way.

If you can update the test as suggested, then please do and merge.

ashetkar:
Filed a jira for the failures (SNAP-2159). AQP-292 seemed different, though looked at only the top-level error.
> If you can update the test as suggested, then please do and merge.

Have already added a new test (disabled), please review.

sumwale:
@ashetkar New test looks good -- please enable and merge. @rishitesh can you comment on whether the failure seen in wide-table test is expected without codegen fallback?

PR #928, opened 2018-01-12
fit to gradle 5.0

jerryldh:

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


dhavalstr:

sumwale:
@jerryldh Does this change alone allow this to work with gradle 4.x or upcoming 5? There were some other issues with gradle 4.x due to which I had skipped it when upgrading gradle to 3.5.1 (and the fact that 4.x was barely out of the door then).

I don't think this works with current 3.5.x based scripts and will break the current builds, so let me know if you are interested in porting the whole set of gradle scripts to 4.4.x/5.0.

PR #930, opened 2018-01-13
Jnj security

ahshahid:
This PR is for addressing following defects
1) [Snap -2154](https://jira.snappydata.io/browse/SNAP-2154)
2) [Snap-2153](https://jira.snappydata.io/browse/SNAP-2153)
3) [Snap-2152](https://jira.snappydata.io/browse/SNAP-2152)
4) [Snap-2155](https://jira.snappydata.io/browse/SNAP-2155)

The changes suggested are:
1)In the SnappyStoreHiveCatalog, replace '-' with '_' when generating schema for a user with user name containing '-'
2) Reuse the credentials from the snappysession passed as optional parameter.  If you have any suggestion to handle it in a cleaner way pls let me know.
3) Added a new constructor which allows creating SnappyStreamingContext using the existing SnappyContext ( on the lines of using an existing SparkContext). As discussed in previous mail thread, this change is not must for the JnJ pilot & can be reverted.
4) Changing the order of setting of the active SnappyStreamingContext which needs to happen before starting the streams.
5) Clearing snappysession & calling destroy relations only if stopSparkContex is true ( the later part is needed or not, I am not sure)
6) Added a new method to create checkpointed SnappyStreamingContext which takes existing SnappySession as param from which credentials can be retrieved.

For bug SNAP-2154 , a corresponding PR has been generated in the store repo too.
[PR 356](https://github.com/SnappyDataInc/snappy-store/pull/356)


core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -117,7 +116,9 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
       // In smart connector, property name is different.
       user = snappySession.conf.get(Constant.SPARK_STORE_PREFIX + Attribute.USERNAME_ATTR, "")
     }
-    val defaultName = if (user.isEmpty) Constant.DEFAULT_SCHEMA else formatDatabaseName(user)
+    val defaultName = IdUtil.getUserAuthorizationId(
+      if (user.isEmpty) Constant.DEFAULT_SCHEMA else formatDatabaseName(user)).
+      replace('-', '_')
```
vivekwiz: replace can be done for only else part

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
```diff
@@ -279,7 +311,9 @@ object SnappyStreamingContext extends Logging {
       createOnError: Boolean = false
       ): SnappyStreamingContext = {
     ACTIVATION_LOCK.synchronized {
-      getActive.getOrElse { getOrCreate(checkpointPath, creatingFunc, hadoopConf, createOnError) }
+     getActive.getOrElse { getOrCreate(checkpointPath, creatingFunc,
```
vivekwiz: I think formatting went wrong here

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
```diff
@@ -252,6 +283,7 @@ object SnappyStreamingContext extends Logging {
   def getActiveOrCreate(creatingFunc: () => SnappyStreamingContext): SnappyStreamingContext = {
     ACTIVATION_LOCK.synchronized {
       getActive.getOrElse { creatingFunc() }
+
```
vivekwiz: unncessary

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
```diff
@@ -307,7 +341,23 @@ object SnappyStreamingContext extends Logging {
       ): SnappyStreamingContext = {
     val checkpointOption = CheckpointReader.read(
       checkpointPath, new SparkConf(), hadoopConf, createOnError)
-    checkpointOption.map(new SnappyStreamingContext(null, _, null)).getOrElse(creatingFunc())
+    checkpointOption.map(new SnappyStreamingContext(null, _, null)).
```
vivekwiz: formatting?

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
```diff
@@ -307,7 +341,23 @@ object SnappyStreamingContext extends Logging {
...
+      getOrElse(creatingFunc())
+
+  }
+
+  def getOrCreateWithUseCredential(
```
vivekwiz: What is this function for?

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
```diff
@@ -56,7 +59,28 @@ class SnappyStreamingContext protected[spark](
...
+
+    val isSmartConnProp = if (user.isEmpty) {
+      // In smart connector, property name is different.
+      user = csn.conf.get(Constant.SPARK_STORE_PREFIX + Attribute.USERNAME_ATTR, "")
+      true
+    } else {
+      false
+    }
+    if (!user.isEmpty) {
+      val password = csn.conf.get(Attribute.PASSWORD_ATTR, "")
```
ashetkar: Why not checking isSmartConnProp for password too?

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
```diff
@@ -56,7 +59,28 @@ class SnappyStreamingContext protected[spark](
...
+
+      snappySession.sessionState.conf.setConfString(
+        if (isSmartConnProp) {
+          Constant.SPARK_STORE_PREFIX + Attribute.USERNAME_ATTR
+        }
+      else Attribute.USERNAME_ATTR, user)
+      snappySession.sessionState.conf.setConfString(Attribute.PASSWORD_ATTR, password)
```
ashetkar: Same as above.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
ahshahid: That is true, but since DEFAULT_SCHEMA though right now app, is hypothetically is changed to include - , then it can create trouble. Code is meant to be that at the end of all processing of name, if '-' remains , it is to be changed to '_'

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
ahshahid: corrected.

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
ahshahid: true

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
ahshahid: corrected

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
ahshahid: I have added a doc comment. This function is to fix Snap-2153 where a streaming context created from check point does not have user credentials passed to the job.

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
ahshahid: Thanks Amogh for pointing that out. I was under the impression only user name is having different attrib. but looks like password to has different key

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
ahshahid: fixed

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
```diff
@@ -56,7 +59,31 @@ class SnappyStreamingContext protected[spark](
...
+        val key = Constant.SPARK_STORE_PREFIX + Attribute.PASSWORD_ATTR
+        key -> csn.conf.get(key, "")
+      } else {
+        val key = Attribute.PASSWORD_ATTR
+        key -> csn.conf.get(key, "")
+      }
+      snappySession.sessionState.conf.setConfString(userKey, user)
+      snappySession.sessionState.conf.setConfString(passwordKey, password)
+    }
+  })
```
sumwale: Concise way:

```scala
val attrs = Attribute.USERNAME_ATTR -> Attribute.PASSWORD_ATTR
val smartAttrs = (Constant.SPARK_STORE_PREFIX + attrs._1) -> (Constant.SPARK_STORE_PREFIX + attrs._2)
// exists (or find) instead of foreach to break out on first match
Seq(attrs, smartAttrs).exists { case (userKey, passKey) =>
  if (csn.conf.contains(userKey)) {
    snappySession.sessionState.conf.setConfString(userKey, csn.conf.get(userKey)
    snappySession.sessionState.conf.setConfString(passKey, csn.conf.get(passKey, "")
    true
  } else false
}
```

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
```diff
@@ -56,7 +59,31 @@ class SnappyStreamingContext protected[spark](
         "both SparkContext and checkpoint as null")
   }
 
-  val snappySession = new SnappySession(sc)
+  val snappySession = reuseSnappySession.getOrElse(new SnappySession(sc))
+  currentSnappySession.foreach(csn => {
```
sumwale: If reuseSnappySession is None, then won't currentSnappySession be used? Or that case should never arise?

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
ahshahid: @sumwale I did not get the question.  are you talking about the situation where you have two users with name differing only by  - & _ . Hmm. Yes that will lead to clash... . This is a temporary measure, I suppose. I looked at changing the Hive related class to allow special character but decided against it as it would have required us to maintain another 3rd party repo branch internally. Once this JnJ is out of way, can work for a better fix as suggested.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -134,9 +134,11 @@ class LeadImpl extends ServerImpl with Lead
     propNames = bootProperties.stringPropertyNames().iterator()
     while (propNames.hasNext) {
       val propName = propNames.next()
-      if (propName.startsWith(STORE_PREFIX)) {
+      if (propName.startsWith(STORE_PREFIX) ) {
```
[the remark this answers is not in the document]
ahshahid: will fix it

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -134,9 +134,11 @@ class LeadImpl extends ServerImpl with Lead
...
         storeProperties.setProperty(propName.substring(
           STORE_PREFIX.length), bootProperties.getProperty(propName))
+      } else if (propName.startsWith(GemFireSparkConnectorCacheImpl.connectorPrefix) ) {
+        storeProperties.setProperty(propName, bootProperties.getProperty(propName))
```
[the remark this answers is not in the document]
ahshahid: After talking to Neeraj, we decided that a task will be opened which will be about incorporating the review comments which require code change.  Right now my JnJ test env. is not working & I do not have automated test in place to test the changes after incorporating feedback.

core/src/main/scala/org/apache/spark/streaming/SnappyStreamingContext.scala
ahshahid: Yes that case will never arise, because the function from which currentSnappySession is passed, the reuseSnappySession is passed as none.

PR #932, opened 2018-01-16
perf improvements over SNAP-2125

kneeraj:

Corrected some perf mistakes pointed in review comment of SNAP-2125
https://github.com/SnappyDataInc/snappydata/pull/923

## Patch testing

Same test passed.

## ReleaseNotes.txt changes

Need to add newly added properties in documentation.

## Other PRs 

https://github.com/SnappyDataInc/snappydata/pull/923

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -220,7 +220,7 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
   private[sql] var currentKey: SnappySession.CachedKey = _
 
   @transient
-  private[sql] def planCaching: Boolean = sessionState.conf.getConfString(
+  private[sql] var planCaching: Boolean = sessionState.conf.getConfString(
     Property.PlanCaching.name, "true").toBoolean
 
   /**
@@ -1835,7 +1835,7 @@ object SnappySession extends Logging {
   private[this] val ID = new AtomicInteger(0)
   private[sql] val ExecutionKey = "EXECUTION"
 
-  def tokenizationDisabled: Boolean = java.lang.Boolean.getBoolean("DISABLE_TOKENIZATION")
+  var tokenize: Boolean = true
 
   lazy val isEnterpriseEdition: Boolean = {
     GemFireCacheImpl.setGFXDSystem(true)
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -1835,7 +1835,7 @@ object SnappySession extends Logging {
   private[this] val ID = new AtomicInteger(0)
   private[sql] val ExecutionKey = "EXECUTION"
 
-  var tokenize: Boolean = true
+  private[sql] var tokenize: Boolean = true
 
   lazy val isEnterpriseEdition: Boolean = {
     GemFireCacheImpl.setGFXDSystem(true)
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -220,8 +220,7 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
   private[sql] var currentKey: SnappySession.CachedKey = _
 
   @transient
-  private[sql] var planCaching: Boolean = sessionState.conf.getConfString(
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -220,7 +220,7 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
   private[sql] var currentKey: SnappySession.CachedKey = _
 
   @transient
-  private[sql] def planCaching: Boolean = sessionState.conf.getConfString(
+  private[sql] var planCaching: Boolean = sessionState.conf.getConfString(
     Property.PlanCaching.name, "true").toBoolean
```
sumwale: cleaner type-safe way: Property.PlanCaching.get(sessionState.conf)

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1835,7 +1835,7 @@ object SnappySession extends Logging {
   private[this] val ID = new AtomicInteger(0)
   private[sql] val ExecutionKey = "EXECUTION"
 
-  def tokenizationDisabled: Boolean = java.lang.Boolean.getBoolean("DISABLE_TOKENIZATION")
+  var tokenize: Boolean = true
```
sumwale: should be initialized like for planCaching: Property.Tokenize.get(sessionState.conf)

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -503,21 +503,31 @@ class SnappyConf(@transient val session: SnappySession)
     case Property.PlanCaching.name =>
       value match {
         case Some(boolval) => {
-          if (boolval.toString.toBoolean) session.clearPlanCache()
+          if (boolval.toString.toBoolean) {
+            session.clearPlanCache()
+          }
+          session.planCaching = boolval.toString.toBoolean
```
sumwale: can store boolVal.toString.toBoolean

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -503,21 +503,31 @@ class SnappyConf(@transient val session: SnappySession)
...
       }
 
     case Property.PlanCachingAll.name =>
       value match {
         case Some(boolval) => {
-          val disable_tokenization = !boolval.toString.toBoolean
-          System.setProperty("DISABLE_TOKENIZATION", disable_tokenization.toString)
-          if (disable_tokenization) SnappySession.getPlanCache.asMap().clear()
+          val clear_cache = !boolval.toString.toBoolean
+          if (clear_cache) SnappySession.getPlanCache.asMap().clear()
```
sumwale: current session value should be set: session.planCaching = !clear_cache
also naming convention is camel case "clearCache"

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
sumwale: cal also be made private[sql] like for planCaching

kneeraj: >> should be initialized like for planCaching: Property.Tokenize.get(sessionState.conf)

This is more like a global variable. No session is attached to it at the time of initialization. So kept it as "true". May be this property should be in global sqlconf.
Is there one like that?

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
kneeraj: Thought but ignored. Hardly matters for this use case.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
kneeraj: Does not matter as I am using this session to disable across all sessions. If I do selectively in this session ( like planCaching above ) then this session in some sense is getting differently treated than others.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
sumwale: Oh, ok. Then Property.Tokenize is better be a SparkConf Val and not SQLVal. Then in SnappyContext.initGlobalSparkContext you could set it: SnappySession.tokenize = Property.Tokenize.get(sc.conf)

PR #933, opened 2018-01-17
[SNAP-2102] added memory+disk optimized column batch iterators

sumwale:

- added new base ClusteredColumnIterator that fetches and returns all the columns
  of a batch together
- disk read implementation of above "ColumnFormatIterator"
  - this maintains the current running set of column batches being iterated and when a batch
    is known to be finished, then it will collect the in-memory only batches to return
  - batch is known to be finished when the hash table array index changes (getMapTableIndex
    method added for the same in store's CustomEntryConcurrentHashMap)
  - hashCode of ColumnFormatKey changed to return same hashCode for all columns of a
    batch to enable above so that all columns of a batch are clustered in the same hash chain
  - for batches that have even one overflow entry are pushed into a DiskBlockSorter to be
    iterated at the end; if some columns are in memory even then their disk positions are
    used because it is very likely with multiple concurrent iterators that those blocks
    would be evicted by the time they are iterated
- pass the set of projection indexes from JDBCAppendableRelation down to the above iterator
- updated ColumnBatchIterator to use the ClusteredColumnIterator and expect it to provide
  the individual columns instead of doing itself; removed all region.get calls
- added skipEvictionForEntry to StoreCallbacks that skips stats rows
- changed hashJoinSize & hashAggregateSize to use snappydata.sql prefix instead of just
  snappydata
- changed koloboke maps to use "justPut" instead of put for slight perf improvement

TODO: remote iterator implementation of ClusteredColumnIterator being added

## Patch testing

precheckin, manual performance tests

Some performance numbers:

```
Off-heap of 6GB, 500M rows totaling 8GB in-memory data (6-7GB on disk)

SNAP-2102

Java HotSpot(TM) 64-Bit Server VM 1.8.0_151-b12 on Linux 4.10.0-38-generic
Intel(R) Core(TM) i7-5600U CPU @ 2.60GHz
Cache random keys:  Best/Avg Time(ms)    Rate(M/s)   Per Row(ns)   Relative
----------------------------------------------------------------------------
snappy = T              866 /  955        577.3           1.7       1.0X

Master

Cache random keys:  Best/Avg Time(ms)    Rate(M/s)   Per Row(ns)   Relative
----------------------------------------------------------------------------
snappy = T             3482 / 4985        143.6           7.0       1.0X
```

The large difference here despite only some data being on-disk is due to the fact that master
keeps reading disk blocks randomly and besides being slow on disk reads, it also cannot
take any advantage of OS buffers (in this case there are plenty of buffers left).

```
Off-heap 8g, 1Billion rows totaling 16GB in-memory data and 12-13GB on disk

SNAP-2102

Java HotSpot(TM) 64-Bit Server VM 1.8.0_151-b12 on Linux 4.10.0-38-generic
Intel(R) Core(TM) i7-5600U CPU @ 2.60GHz
Cache random keys:  Best/Avg Time(ms)    Rate(M/s)   Per Row(ns)   Relative
----------------------------------------------------------------------------
snappy = T           11686 / 12189         85.6          11.7       1.0X

Master

Cache random keys:  Best/Avg Time(ms)    Rate(M/s)   Per Row(ns)   Relative
----------------------------------------------------------------------------
snappy = T          21794 / 48013         45.9          21.8       1.0X
```

Here 50% of data is on disk and 50% in memory. The difference in best timings is about 2X
but that in average is much larger. The behaviour on master is very erratic with timings
varying from 22s to 1.4min while on SNAP-2102 are nearly constant. This is again likely a
result of random disk reads and continuous evictions/fault-ins.

For comparison with parquet, with snappy compression it is about 2-3X slower than SNAP-2102
when all data is on disk and just 1GB off-heap (1.5 billion rows) while with gzip compression it is
5-6X slower. Was expecting latter to be slightly faster given that disk size is quite a bit smaller
but seems like gzip decompression is very heavy. Also both snappy/gzip with parquet consume
nearly all of CPU while SnappyData takes only ~25% of CPU so it will scale much better too.

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/358


sumwale:
One other observation related to current master behaviour. It gives quite comparable numbers if nearly all data is on-disk and very small memory is configured (e.g. 1GB off-heap with 24GB data). I expect this to change quite a bit on non-SSD disks but on SSDs the difference is 20-30%. As the memory is made larger then it starts misbehaving with numbers all over the place instead of improving. This looks like a result of continuous fault-ins and evictions, apart from lots of unnecessary disk reads. Going by OS behaviour what is likely happening is that due to continuous memory pages being moved around, the OS also starts to aggressively swap out pages as the memory pressure increases. Since we recommend 80% or so of RAM, this will actually end up in worst possible behaviour if all of data does not fit in memory. SNAP-2102 needs to be tested more for different kinds of queries as some further tuning might be in order (post merge).

sumwale:
@kneeraj
Pushed a better remote iterator in https://github.com/SnappyDataInc/snappydata/pull/933/commits/fc165686fb956bd863e8a9116b49a0c8c57bcdc5

Still working on a test for it as it seems no test for remote iteration exists in current test suite
(precheckin passed without this while some test should have failed with a ClassCastException).
Have tested manually so far with a bypass hook to assign bucketIds to partitions randomly.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/RemoteEntriesIterator.scala
```diff
@@ -0,0 +1,97 @@
...
+    // get the full entries for all stats rows in one getAll
+    val msg = new GetAllExecutorMessage(pr, statsKeys.asInstanceOf[Iterator[AnyRef]].toArray,
+      null, null, null, null, null, null, txState, null, false, false)
+    statsKeys.zip(GemFireResultSet.callGetAllExecutorMessage(msg).asScala.toIterator)
+  }
+
+  /**
+   * Full projection including all of delta and meta-data columns (except base stats entry)
+   */
+  private val fullProjection = {
```
kneeraj: So the projection list is passed so that only the required columns are fetched. Now the delta buffer and the bit masks also needs to be brought I understand. But can you please explain this maths or add a comment in the code itself that explains what all is being desired to be brought.

(projection.length * (ColumnDelta.USED_MAX_DEPTH + 1) - DELETE_MASK_COL_INDEX + STATROW_COL_INDEX)

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/RemoteEntriesIterator.scala
```diff
@@ -0,0 +1,97 @@
...
+    currentStatsKey = p._1.asInstanceOf[ColumnFormatKey]
+    NonLocalRegionEntry.newEntry(currentStatsKey, p._2, pr, null)
+  }
+
+  override def getColumnValue(column: Int): AnyRef = {
+    if (currentValueMap.size() == 0) {
+      // fetch all the projected columns for current batch
+      val fetchKeys = fullProjection.map(c =>
+        new ColumnFormatKey(currentStatsKey.uuid, currentStatsKey.partitionId, c))
+      val msg = new GetAllExecutorMessage(pr, fetchKeys.asInstanceOf[Array[AnyRef]],
```
kneeraj: For column table scan the txstate is all local. What is the point of passing a local txid to remote get message?

core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -206,16 +206,16 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
...
-          // snapshot (also pass the same transaction to getNumColumnsInTable
-          //   for reading value and delete count)
-          val itr = pr.localEntriesIterator(null.asInstanceOf[InternalRegionFunctionContext],
-            true, false, true, null).asInstanceOf[PartitionedRegion#PRLocalScanIterator]
           var numColumnsInTable = -1
           // Resetting PR numRows in cached batch as this will be calculated every time.
           var rowsInColumnBatch = 0L
           var offHeapSize = 0L
           if (container ne null) {
+            // TODO: this should use a transactional iterator to get a consistent
```
kneeraj: Shifting inside ne null is good. But this is by inspection or is it fixing some reported problem? and shouldn't he container be always non null?

sumwale: Right null check looks redundant. However, the main change here is not the shifting but the "includeValues" parameter pr.localEntriesIterator which is now false. So stats iteration will always use map iteration only and not attempt to sort by disk entries etc (stats rows should always be in-memory).

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatIterator.scala
```diff
@@ -0,0 +1,399 @@
...
+
+  private lazy val (readerId, diskPosition, diskEntries) = {
+    if (canOverflow) {
+      val sortManager = distributedRegion.getDiskStore.getSortManager
+      currentDiskSorter = sortManager.getSorter(distributedRegion, fullScan, null)
+      (sortManager.newReaderId(), new DiskPosition, new java.util.ArrayList[DiskEntryPage]())
+    } else (0, null, null)
+  }
+
+  setRegion(baseRegion)
```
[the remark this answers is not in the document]
sumwale: This is the normal body of primary constructor in scala. Any statements in the body are for its primary constructor. How else would one invoke a method in primary constructor?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/RemoteEntriesIterator.scala
sumwale: The math comes directly from the loops below that fill it. Will add a comment nevertheless.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/RemoteEntriesIterator.scala
sumwale: My assumption is that this should also automatically take snapshot on remote node. So subsequent getAll calls should also read from the same snapshot. If it does not work correctly that way, then its a bug.

kneeraj: I don't think it will work as per the expectation you are having at this point. @suranjan can you please see this and track it if needs attention.

PR #934, opened 2018-01-18
[SNAP-1501] OVERFLOW=false disables eviction if EVICTION_BY is absent.

ashetkar:
* OVERFLOW=false disables eviction if EVICTION_BY is absent.
* Updated test cases.

## Patch testing
unit tests.
precheckin running

## ReleaseNotes.txt changes
A note about OVERFLOW option is deprecated.

## Other PRs 
NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
+++ b/core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
@@ -414,7 +414,8 @@ object StoreUtils {
     parameters.remove(PARTITIONER).foreach(v =>
       sb.append(GEM_PARTITIONER).append('\'').append(v).append("' "))
 
-    val defaultEviction = s"$GEM_HEAPPERCENT $GEM_OVERFLOW"
+    val overflow = parameters.get(OVERFLOW).map((_.toBoolean)).getOrElse(true)
+    val defaultEviction = if (overflow) s"$GEM_HEAPPERCENT $GEM_OVERFLOW" else EMPTY_STRING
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -414,15 +414,16 @@ object StoreUtils {
     parameters.remove(PARTITIONER).foreach(v =>
       sb.append(GEM_PARTITIONER).append('\'').append(v).append("' "))
 
-    val defaultEviction = s"$GEM_HEAPPERCENT $GEM_OVERFLOW"
+    val overflow = parameters.get(OVERFLOW).map((_.toBoolean)).getOrElse(true)
```
kneeraj: This can lead to bad error btw if the user makes some spelling mistake in writing "false" or "true".
Blindly doing _.toBoolean will give error.

ashetkar: It throws below message, which user can use to find out the issue. This should be okay, right?
`
snappy> create table coltab (id int) using column options (overflow 'fals');
ERROR 38000: (SQLState=38000 Severity=20000) (Server=localhost/127.0.0.1[1528] Thread=ThriftProcessor-0) The exception 'com.pivotal.gemfirexd.internal.engine.jdbc.GemFireXDRuntimeException: myID: 127.0.0.1(26557)<v1>:5443, caused by java.lang.IllegalArgumentException: For input string: "fals"' was thrown while evaluating an expression.
`

PR #936, opened 2018-01-22
Parser performance improvements

sumwale:
Some parser performance improvements to recover the regression over 0.9 release.
Also optimized and enhanced numeric/decimal literal handling.

- moved paramLiteral/literal rule below the identifier one in primary; this was changed as part
  of parser fixes to avoid treating INTERVAL literal as identifier (d015588) but cause perf trouble due to:
  - rejecting identifier rule in case of mismatch is fast most of the time since it has to only
    check against [a-z_] pattern which is done using a bitset check but rejecting literals
    requires running through many rules that are ORed and parser has to check and
    reject each one of those
  - moreover, if tokenization is enabled then the rule has to do double the work of
    first rejecting paramLiteral and then rejecting literal in non-match cases
  - identifiers are overall more common
- two changes to correct above:
  - broke out the intervalLiteral rule from inside literal and check it independently first
    before identifier but remaining ones are moved back to be below identifier
  - combined paramLiteral and literal into a single "paramOrLiteral" rule that will check
    for tokenize inside the body and return ParamLiteral or Literal accordingly

above was major factor for the perf degradation by 20-30% over 0.9 in ColumnTableTest.compare parser performance

- allow for float literals using F/f suffix; also allow lower-case for double/long suffix d/l
- avoid creating higher precision DecimalType because it will lead to casts on other side
  of comparison if column type is narrower, for example (the tag creation is no longer
    a problem since its lazy and not used in Spark any more)
- changed wholeStageEnabled to be a var in SnappySession updated like other properties
  instead of TOKENIZE_BEGIN having to read it from conf every time
- change FOLDABLE_FUNCTIONS map to be a koloboke one which is >2X faster than scala immutable map
- changed keywords maps to be OpenHashSets rather than scala mutable hashsets
- replace scala TrieMaps with java ConcurrentHashMaps in parser/session
  since former are quite a bit slower
- reset PlanCaching/Tokenization properties back to defaults in case they are unset
  (for None match) instead of no change
- added a "snappydata.sql.parser.traceError" proper to print detailed parser backtracking
  traces in case of a parse error
- mark RETURNS as neither reserved, nor non-reserved;
  fixes failure in TokenizationTest.SNAP-2031 tpcds that uses "returns" as a named alias
  without AS as per the changes to remove optional AS with identifier that fix ambiguities
  for cases like "SELECT ... FROM (SELECT ...) table where" (with table as just identifier and
  not strictidentifier the parser can easily confuse between this and when there is no alias)

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
@@ -90,14 +90,14 @@ class SparkSQLExecuteImpl(val sql: String,
   // check for query hint to serialize complex types as JSON strings
   private[this] val complexTypeAsJson = session.getPreviousQueryHints.get(
     QueryHint.ComplexTypeAsJson.toString) match {
-    case None => true
-    case Some(v) => Misc.parseBoolean(v)
+    case null => true
+    case v => Misc.parseBoolean(v)
   }
 
   private val (allAsClob, columnsAsClob) = session.getPreviousQueryHints.get(
     QueryHint.ColumnsAsClob.toString) match {
-    case None => (false, Set.empty[String])
-    case Some(v) => Utils.parseColumnsAsClob(v)
+    case null => (false, Set.empty[String])
+    case v => Utils.parseColumnsAsClob(v)
   }
 
   override def packRows(msg: LeadNodeExecutorMsg,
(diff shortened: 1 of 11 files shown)


sumwale:
Merging for 1.0.1 release builds. Will take care of any review comments subsequently.

PR #937, opened 2018-01-22
[SNAP-2088] fixes for queries with filters on columns with nulls

sumwale:
Brief overview of null handling in generated code and the new approach.

#### Pre SNAP-1936

The initial code for null handling in ColumnTableScan was simple one where there was always
sequential iteration over all the rows for all the projected columns. The generated code
maintained current cursor position of decoder that indicated the position of current non-null
value to be read while nullability check was done with a separate call to decoder.

#### Changes in SNAP-1936 and its weaknesses

This changed completely with #793 (SNAP-1936) that changed sequential cursors to random
ones to handle cases of high filtering where cursors of other columns (having no filter
conditions but being projected) need not be updated anymore for every row. It complicated
the nullability handling because now the code needs to determine "number of nulls till this
point" to enable decoder to read the non-null value and it was too inefficient to determine
this afresh in every iteration. The solution adopted there was to track "number of nulls till
now" and update that with a call to decoder which in turn will track the next null position
and only if current ordinal has jumped past it without touching it, it will recompute number
of nulls till current ordinal. Further to avoid two calls, one for number of nulls and other for
whether current value is null, it used a convention of returning negative number of nulls
to denote latter as true.

This approach had one remaining bug as in SNAP-2118 fixed in #929 (for non-nulls only) where
same ordinal could be read multiple times but nextNullPosition tracker would update
immediately on first read of an ordinal expecting that the next call would always be greater
than previous (so will give erroneous result for nullability if a call with same ordinal is made
again).

This was further muddied by changes in SNAP-1743 #905 that erroneously assumed that
nextNullPosition was only for determine nullability and missed the "number of nulls till this
point" functionality achieved by the same. The motivation for those changes was the
observation that scan performance over null columns was erratic with JVM sometimes
optimizing "nextNullPosition" code correctly and sometimes not while direct bitset read
was always optimized (and a bit faster even if both were optimized due to no writes). One
error in the thinking of SNAP-1936 in terms of performance was the design to do two things
in a single function call so added extra negation operations, while forgetting the fact that
we need the functions to be inlined in any case so two functions calls may have been just fine.

Just for comparison, Spark's ColumnVector creates arrays of values of full sizes with blanks
at null values so it never has to adjust the position as per the number of nulls. For nulls it
has a boolean array again of the full size. It is, of course, much slower because of extra writes
and is not a feasible approach in this implementation.

#### Approach in this PR

This PR avoids the pitfalls above and also brings performance back to pre-SNAP-1936 level
for all cases. The basic idea being similar to SNAP-1936 but with a few changes:

a) handling of multiple calls to same ordinal by updating the position only when ordinal goes
    past (and optimized handling when it has gone just past i.e. is +1)

b) inlined code segments in areas that are expected to be inlined to get decent performance

c) assume that an optimal nullable decoder will always use the approach of "nextNullPosition"
   tracking to enable inlining in generated code above, so the ColumnDecoder interface has
   added appropriate methods

Specifically note this code fragment in ColumnTableScan code generation:

```scala
    s"""
       |int $nextNullPosition = $decoder.getNextNullPosition();
       |if ($batchOrdinal < $nextNullPosition ||
       |    // check case when batchOrdinal has gone just past nextNullPosition
       |    ($batchOrdinal == $nextNullPosition + 1 &&
       |     $batchOrdinal < ($nextNullPosition = $decoder.findNextNullPosition(
       |       $buffer, $nextNullPosition, $numNullsVar++))) ||
       |    // check if batchOrdinal has moved ahead by more than one due to filters
       |    ($batchOrdinal != $nextNullPosition && (($numNullsVar =
       |       $decoder.numNulls($buffer, $batchOrdinal, $numNullsVar)) == 0 ||
       |       $batchOrdinal != $decoder.getNextNullPosition())))""".stripMargin
```

combines all the three conditions in a single if, preferring performance over code readability.
The first case is when ordinal has still not reached nextNullPosition (so no change in numNulls
till ordinal and current value is also not null). The second case is when ordinal has moved
just past nextNullPosition in which case update the nextNullPosition from nulls bitset
(and runLength in future) and check again. The last case is when ordinal has jumped much
past due to filters, and in that case calculate the new numNulls (the == 0 check is redudandant
but is the most performant way to do it inline given the lack of comma operator in java).

- Added new methods in ColumnDecoder to take care of filters skipping nulls and making the
  running null count invalid:
  - getNextNullPosition: returns the next position that is null
  - findNextNullPosition: would search for and update the next null position beyond the
    provided on
  - numNulls: count the nulls from given position to given new one (old count also provided)
  These are supposed to be invoked by generated code tracking the number of nulls.
- Limit update delta handling to two deltas because more than two will never be used
  else hit on scan performance is high due to increasing number of cache lines
- Moved test for part2 of SNAP-2088 from ColumnTableDUnitTest to QueryTest.
  Fixed scalastyle errors in ColumnTableDUnitTest
- Added many more queries in QueryTest for SNAP-2088 with filters and checking
  full result set against Spark cache results
- initial corrections to update handling (still some work left here to complete)

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/UpdatedColumnDecoder.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/UpdatedColumnDecoder.scala
@@ -31,10 +31,8 @@ import org.apache.spark.sql.types._
  * a nullable or non-nullable version as appropriate.
  */
 final class UpdatedColumnDecoder(decoder: ColumnDecoder, field: StructField,
-    delta1Position: Int, delta1: ColumnDeltaDecoder,
-    delta2Position: Int, delta2: ColumnDeltaDecoder)
-    extends UpdatedColumnDecoderBase(decoder, field, delta1Position, delta1,
-      delta2Position, delta2) {
+    delta1: ColumnDeltaDecoder, delta2: ColumnDeltaDecoder)
+    extends UpdatedColumnDecoderBase(decoder, field, delta1, delta2) {
 
   protected def nullable: Boolean = false
 
@@ -45,10 +43,8 @@ final class UpdatedColumnDecoder(decoder: ColumnDecoder, field: StructField,
  * Nullable version of [[UpdatedColumnDecoder]].
  */
 final class UpdatedColumnDecoderNullable(decoder: ColumnDecoder, field: StructField,
-    delta1Position: Int, delta1: ColumnDeltaDecoder,
-    delta2Position: Int, delta2: ColumnDeltaDecoder)
-    extends UpdatedColumnDecoderBase(decoder, field, delta1Position, delta1,
-      delta2Position, delta2) {
+    delta1: ColumnDeltaDecoder, delta2: ColumnDeltaDecoder)
+    extends UpdatedColumnDecoderBase(decoder, field, delta1, delta2) {
 
   protected def nullable: Boolean = true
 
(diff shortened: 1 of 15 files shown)


sumwale:
One more point to note: when it says random reads from decoders, it still assumes a forward-only cursor. The decoders can fail in various ways (some silently) if ordinal moves back and has to always either go forward or stay at the same position. This is ensured in ColumnTableScan generated code loop over a column batch and is unlikely to ever change, but still is something to keep in mind when working with ColumnTableScan.

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/UpdatedColumnDecoder.scala
```diff
@@ -139,56 +129,36 @@ abstract class UpdatedColumnDecoderBase(decoder: ColumnDecoder, field: StructFie
...
-    }
+    do {
+      // skip the position in current delta
+      skipUpdatedPosition(nextDeltaBuffer)
+      // update the cursor and keep on till ordinal is not reached
+      nextUpdated = moveToNextUpdatedPosition(nextUpdated)
+    } while (nextUpdated < ordinal)
+    nextUpdatedPosition = nextUpdated
+    if (nextUpdated > ordinal) return true
     currentDeltaBuffer = nextDeltaBuffer
```
ashetkar: Do currentDeltaBuffer and nextUpdatedPosition need to be updated if (nextUpdatedPosition == ordinal)?

sumwale: This will change and nextDeltaBuffer is being removed. There will only be currentDeltaBuffer which will be updated when ordinal moves past nextUpdatedPosition so only currentDeltaBuffer will be required.

sumwale:
@ashetkar changes to the update decoder have been added in https://github.com/SnappyDataInc/snappydata/pull/937/commits/eeb99a06dde41f5d156d8f7b15655c34ebf8be8c. You may find it easier to just read the full ColumnDeltaDecoder and UpdatedColumnDecoder files instead of the diff.

PR #941, opened 2018-01-29
[SNAP-2158] DataFrame update API. 

rishitesh:

Added update API to DataFrame.
Both row and column tables can be updated df.write.update("table")

Also added a utility method to get JDBC data frames using pooled connection settings. Otherwise, JDBC relation will create a new connection every time, which is very expensive in streaming applications.

Added a check in SnappyHashJoin to skip scanning of tables, if the built map size is zero. This is helpful if some join operation is happening in an empty dataframe.

## Patch testing
pre-checkin

## ReleaseNotes.txt changes
Na

## Other PRs 
NA


PradeepSurale:

PR #943, opened 2018-01-31
[SNAP-2156] added test for the case

sumwale:

- test to check that JDBC query execution returns a count for such routed DMLs
- enabled spark.memory.debugFill for unit tests to detect many cases of incorrect memory
   read/write

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/359

### DIFF
--- a/core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitTest.scala
+++ b/core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitTest.scala
@@ -18,7 +18,7 @@ package io.snappydata.cluster
 
 import java.io.PrintWriter
 import java.nio.file.{Files, Paths}
-import java.sql.{Blob, Clob, Connection, DriverManager, ResultSet, Statement, Timestamp}
+import java.sql.{Blob, Clob, Connection, DriverManager, ResultSet, SQLException, Statement, Timestamp}
 import java.util.Properties
 
 import scala.collection.JavaConverters._
@@ -26,6 +26,7 @@ import scala.collection.mutable.ArrayBuffer
 import scala.language.{implicitConversions, postfixOps}
 import scala.sys.process._
 import scala.util.Random
+
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.pivotal.gemfirexd.Attribute
 import com.pivotal.gemfirexd.snappy.ComplexTypeSerializer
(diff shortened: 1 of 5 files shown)


core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitTest.scala
```diff
@@ -209,6 +209,13 @@ object SplitClusterDUnitTest extends SplitClusterDUnitTestObject {
       selectFromTableUsingJDBC("splitModeTable1", 1005, stmt)
     }
 
+    // check for SNAP-2156/2164
+    val updateSql = s"update splitModeTable1 set col1 = 100 where exists " +
+        s"(select 1 from splitModeTable1 t where t.col1 = splitModeTable1.col1 and t.col1 = 1234)"
+    assert(!stmt.execute(updateSql))
+    assert(stmt.getUpdateCount == 0)
+    assert(stmt.executeUpdate(updateSql) == 0)
```
kneeraj: We would have tests also where return is positive count too?

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/ColumnEncoding.scala
```diff
@@ -634,7 +634,7 @@ trait ColumnEncoder extends ColumnEncoding {
     }
     val baseTypeOffset = offset(position).toInt
     // initialize the null bytes to zeros
-    allocator.clearBuffer(columnData, baseTypeOffset, numNullBytes)
+    allocator.fill(columnData, 0, baseTypeOffset, numNullBytes)
```
kneeraj: Memsetting to '0' for the full buffer. Did it cause a problem or what? clearBuffer does the same thing too. Right?

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitTest.scala
sumwale: Ok, though the issue was about these operations throwing exceptions. Updates being proper is tested in details elsewhere like ColumnUpdateDeleteTests.

core/src/main/scala/org/apache/spark/sql/execution/columnar/encoding/ColumnEncoding.scala
sumwale: fill is a more generic method so store code was refactored a bit to remove clearBuffer and instead use fill. The recent addition being of "spark.memory.debugFill" property now honoured by store's BufferAllocator too to fill initialized/freed memory with non-zero bytes (so that any code errors of reading from those will lead to incorrect results)

PR #543, opened 2017-03-22
Pruning partitions for predicates based on partitioning columns

soubhik-c:
- Changed ParamLiteral to inherit from Literal instead of containing it.
- Ensured predicate pushdown to datasources happen for pruning to take
effect.
- Using a partitionPruner closure that is used for RowBuffer as well as
ColumnScan

prechecking bug fixes

- handling shuffledependency during reset
- checking CodegenContext references to generate multiple literal term
for the same ParamLiteral
- marking ParamLiteral as foldable=false because ConstantFolding
shouldn't happen during optimization rather everytime new literals are
passed.
- serializing the dataType efficiently and correctly using kryo. (can
    eliminate dataType altogether, todo next)
- handling preparedStatement ParamLiteral constant replace for Smart and
Embedded connector mode due to filter pushdown now happening as
expected.
- removing redundant pruning test

- Putting converted value in SmartConnector mode while executing prepared
statement, filters in smart connector mode now puts in converted to scala
values instead of catalyst values.

- Adding index pruning support and therefore changing it to explicit
ShuffleExchange instead of going through EnsureRequirements so that
numpartitions are aligned with pruned ColScan and base table RowScan.

## Changes proposed in this pull request

(Fill in the changes here)

## Patch testing

- Added new tests specifically verifying buckets used while querying.
- Asserted bucket changes on constant change with tokenization and plan caching.

## ReleaseNotes.txt changes

(Does this change require an entry in ReleaseNotes.txt? If yes, has it been added to it?)

## Other PRs 

(Does this change require changes in other projects- store, spark, spark-jobserver, aqp? Add the links of PR of the other subprojects that are related to this change)

PR #945, opened 2018-02-01
[SNAP-2124] rows missed in update due to incorrect stats row read

sumwale:
Fix for the issue is in ColumnTableScan when determining number of fields in the table.
Other changes done while reviewing related code and generated code mentioned below.

- rows missed in scan of update/delete due to incorrect number of fields given in stats row read
  which included the virtual columns (batchID, ordinalID, ...) added for update/delete operations
- added test using the sample data provided in SNAP-2124
- don't disable tokenization if planCaching is disabled; corrected planCaching handling
  to explicitly invalidate the cached plan if set
- added missing ParamLiteral/Literal combinations in MIN/MAX predicate filtering
- add a ParamLiteral only once to mutable state in case it is referred multiple times
- correct subexpression elimination comment for child expression in
  DynamicFoldableExpression and add elimination for itself
- removed obsolete SKIP_SPS_PRECOMPILE flag

Incorporating some of pending feedback in partitioning pruning PR #543

- system property "DISABLE_PARTITION_PRUNING" changed to
   "snappydata.sql.partitionPruning" session property
- use better hashCode and equals in ParamLiteral

## Patch testing

precheckin with new test

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -161,16 +161,36 @@ private[sql] final case class ColumnTableScan(
         statsFor(a).lowerBound <= l && l <= statsFor(a).upperBound
       case EqualTo(l: DynamicReplacableConstant, a: AttributeReference) =>
         statsFor(a).lowerBound <= l && l <= statsFor(a).upperBound
+      case EqualTo(a: AttributeReference, l: Literal) =>
```
rishitesh: Sumedh, can you please clarify this for me. We have added extra filter conditions. If they were absent before, although skipping would have been bad, but all rows should have got reflected, isn't it?

sumwale: These conditions were added on code inspection and debugging. The test there was applying the filter due to the DynamicReplaceableConstant condition already present. As mentioned the real issue the numColumnsInStatBlob change below.

Where I got mislead while debugging was that turning off tokenization caused the test to pass causing me to chase the tokenization code while the real problem was that filter was getting skipped for EqualTo when tokenization is disabled.

PR #946, opened 2018-02-02
[SNAP-2191] Disable zeppelin interpreter from within lead process when security is enabled

ashetkar:
* Disable zeppelin interpreter from within lead process when security is enabled
* Also, fixes an issue where UI remained un-secured due to temporary addition of property prefix, on latest master.

## Patch testing
precheckin, manually tested.

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


ashetkar:
Test validation is added for both the fixes: 1) Ensure Zeppelin interpreter is not initialized and 2) Pulse UI uses Snappy authenticator when security is enabled.
Both these assertions are made inside a snappy job which run on a secure cluster via SplitClusterDUnitSecurityTest.scala

PR #950, opened 2018-02-05
[SNAP-2192] delay rollover in column updates to pre-commit

sumwale:

- added "delayRollover" flag to scan RDDs and SnapshotConnectionListener in updates
  to disable rollover during operation and do it explicitly, if required, just before commit
- flag is passed through two routes:
  - ColumnUpdateExec overrides delayRollover which is passed to SnapshotConnectionListener;
    this is the case where update starts its own transaction due to intermediate shuffle
  - flag passed to row/column scan RDDs during construction (via Relation instance) which
    is the case where there is no shuffle and scan started transaction is used
  though issue is only for second case above, still delayRollover has been added to both
- added testOnBorrow flag to tomcat pool (hikari has it by default) to clean failed connections
- use empty string TXId for no TX instead of "null" string
- enabled test for SNAP-2124 to dunit tests too (disable pruning check in split mode where
    pruning with smart connector is still missing)

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/362


kneeraj:
Is it fair to say that the design problem is only due to an update which affects variable length columns and hence the rollover. For fixed width columns update on them should not cause size change and hence rollover.

PR #951, opened 2018-02-05
[SNAP-2180] Explicitly initialized memoryMap on UMM start.

rishitesh:

Explicitly initialized memoryMap on UMM start.
When executor goes away we store the UMM stats in another memory manager which is not associated with SparkEnv. Once executors come back we transfer the memory consumption stats to UMM. But this is done in first memory access. Till that time Pulse shows zero memory consumed in UI.

Added a init method to initialize UMM once SparkEnv is ready.

## Patch testing

precheckin, manual testing. 

## ReleaseNotes.txt changes

NA

## Other PRs 

Na


core/src/test/scala/org/apache/spark/sql/store/TokenizationTest.scala
```diff
@@ -318,7 +318,7 @@ class TokenizationTest
           assert(r.get(0) == r.get(1) && r.get(2) == i)
         })
       }
-      assert(cacheMap.size() == 1)
+      assert(cacheMap.size() == 0)
```
sumwale: Pretty sure this test passed when the changes were made. Not sure what happened here but these changes look correct.

PR #952, opened 2018-02-05
[SNAP-2194] pass partition pruner to smart connector

sumwale:

- pass the partition pruner to SmartConnectorColumnRDD
- added bucketId to SmartExecutorBucketPartition which can be different from partition index
  for the case of pruning; use that to get the partitionId/bucketId instead of index
- added check for pruned select in existing test for SNAP-2124

Update/delete still don't prune rather result in an exchange after the select part
(which is pruned). This is due to special OrderlessHashPartitioning still lying in snappy-spark
only and used in cluster mode. Moving that to snappy-core will be tracked in separate issue.

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


sumwale:

PR #954, opened 2018-02-08
[SNAP-2204] search through aliases (e.g. for VIEWs) for colocated join keys

sumwale:

Search the plan expressions to find any matching view aliases to determine the
underlying column for a join key. A physical plan for a query on join VIEW looks like:

```
Join Inner, (gen_attr_0#347 = gen_attr_2#349)
:- Project [ID#204 AS gen_attr_0#347]
:  +- Filter isnotnull(ID#204)
:     +- Relation[ID#204,ADDR#205] ColumnFormatRelation[APP.VIEWCOLTABLE]
+- Project [ID#248 AS gen_attr_2#349, ADDR#249 AS gen_attr_1#350]
   +- Filter isnotnull(ID#248)
      +- Relation[ID#248,ADDR#249] RowFormatRelation[APP.VIEWROWTABLE]
```

The underlying column is buried inside the Project alias while the join condition
is on generated attributes (gen_attr_0 and gen_attr_2).

Also changes to avoid multiple unapplies of ExtractEquiJoinKeys, PhysicalOperation rather
do it once and then match on the extracted children.

Added unit tests for above.

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
@@ -16,6 +16,7 @@
  */
 package org.apache.spark.sql
 
+import scala.annotation.tailrec
 import scala.util.control.NonFatal
 
 import io.snappydata.Property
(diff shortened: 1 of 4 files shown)


sumwale:
@rishitesh @vivekwiz when looking at the changes in "Files changed" tab, you can avoid the indentation change by adding "?w=1" at the end of URL i.e. https://github.com/SnappyDataInc/snappydata/pull/954/files?w=1

core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
```diff
@@ -78,119 +79,142 @@ private[sql] trait SnappyStrategies {
...
+                joinType, joins.BuildRight, replicatedTableJoin = false)
+            } else if (buildLeft) {
+              makeLocalHashJoin(leftKeys, rightKeys, left, right, condition,
+                joinType, joins.BuildLeft, replicatedTableJoin = false)
+            } else if (RowOrdering.isOrderable(leftKeys)) {
+              joins.SortMergeJoinExec(leftKeys, rightKeys, joinType, condition,
+                planLater(left), planLater(right)) :: Nil
+            } else Nil
+          }
+          // broadcast joins preferred over exchange+local hash join or SMJ
```
vivekwiz: Why is this reference of SMJ here, when in both if and else cases below hash join is being done.

vivekwiz: I did not see reference for SortMergeJoinExec other than line number 104.

sumwale: This is the case of "isCollocatedJoin" when the join is a collocated one but too large for a hash join. Otherwise it will return Nil both here and below if all custom join plans fail. In that case Spark's normal join (exchange+SMJ for equi-join and NLJ for non-equi joins) strategies will take effect which is applied after these strategies. Keeping the call for SMJ here for colocated case helps in avoiding an unnecessary exchange that Spark strategies may end up applying.

vivekwiz: Thanks Sumedh. I did not see reference for  Spark's normal join (exchange+SMJ). So is that default and comes into play if Nil is returned from here?

sumwale: Yes these strategies apply first, then SparkStrategies (JoinSelection in SparkStrategies.scala). See DefaultPlanner.strategies in SnappySessionState which adds the snappy strategies and then all the spark ones.

PR #958, opened 2018-02-11
Fixed putinto inner join cache perf and related issues.

rishitesh:
Fixed putinto inner join cache perf and related issues.
## Patch testing
prechecking

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/92

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/execution/SnappyTableMutableAPISuite.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/execution/SnappyTableMutableAPISuite.scala
@@ -345,23 +345,42 @@ class SnappyTableMutableAPISuite extends SnappyFunSuite with Logging with Before
     assert(resultdf.contains(Row(88, 88, 90)))
   }
 
+  test("PutInto Cache Test") {
+    val snc = new SnappySession(sc)
+    val rdd = sc.parallelize(data1, 2).map(s => Data(s(0), s(1), s(2)))
+    val df1 = snc.createDataFrame(rdd)
+    val rdd2 = sc.parallelize(data2, 2).map(s => Data(s(0), s(1), s(2)))
+    val df2 = snc.createDataFrame(rdd2)
+    snc.sql("set spark.sql.defaultSizeInBytes=1000")
+    snc.createTable("col_table", "column", df1.schema, Map("key_columns" -> "col2"))
+
+    df1.write.insertInto("col_table") // insert initial data
+    df2.write.putInto("col_table") // update & insert subsequent data
+    df2.write.putInto("col_table")
+    val resultdf = snc.table("col_table").collect()
+    assert(resultdf.length == 7)
+    assert(resultdf.contains(Row(8, 8, 8)))
+    assert(resultdf.contains(Row(88, 88, 90)))
+  }
+
   test("PutInto with API for pure inserts") {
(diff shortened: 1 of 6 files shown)


cluster/src/test/scala/org/apache/spark/sql/execution/SnappyTableMutableAPISuite.scala
```diff
@@ -345,24 +345,43 @@ class SnappyTableMutableAPISuite extends SnappyFunSuite with Logging with Before
...
+  }
+
   test("PutInto with API for pure inserts") {
     val snc = new SnappySession(sc)
     val rdd = sc.parallelize(data1, 2).map(s => Data(s(0), s(1), s(2)))
     val df1 = snc.createDataFrame(rdd)
     val df2 = snc.range(100, 110).selectExpr("id", "id", "id")
-
-    snc.createTable("col_table", "column", df1.schema, Map("key_columns" -> "col2"))
+    snc.sql("set spark.sql.defaultSizeInBytes=1000")
```
sumwale: What does "defaultSizeInBytes" do?

core/src/main/scala/org/apache/spark/sql/internal/ColumnTableBulkOps.scala
```diff
@@ -69,42 +54,60 @@ object ColumnTableBulkOps {
...
+        updateSubQuery = analyzedUpdate.child
 
-        val doInsertJoin = if (updateSubQuery.statistics.sizeInBytes <= cacheSize) {
+        val doInsertJoin = if (subQuery.statistics.sizeInBytes <= cacheSize) {
           val joinDS = new Dataset(sparkSession,
             updateSubQuery, RowEncoder(updateSubQuery.schema))
+          if (sparkSession.sharedState.cacheManager
+              .lookupCachedData(updateSubQuery).isDefined){
+            joinDS.unpersist(blocking = true)
+          }
```
sumwale: This means every session will retain a cached result, so won't this keep growing unncessarily? What was the problem with earlier one which used to uncache at the end of executeCollect()? Perhaps what is required there is equivalent of "blocking = true" there.

cluster/src/test/scala/org/apache/spark/sql/execution/SnappyTableMutableAPISuite.scala
rishitesh: This sets the size of a RDD if there is no reasonable statistics about its statistics.

core/src/main/scala/org/apache/spark/sql/internal/ColumnTableBulkOps.scala
rishitesh: This is workaround. Before I was putting it in contextObject and was clearing after execute all.But looks like contextObject itself was getting cleared. Hence the cached data remained as it is.

sumwale: Ok, but this will keep growing especially if updateSubQuery is different each time (which will normally be the case). I think you can have the key "CACHED_PUTINTO_UPDATE_PLAN" define in SnappySession object, and do the cleanup in SnappySession.clearContext. Later we can think of how to make this generic.

PR #960, opened 2018-02-11
[SNAP-2214][SNAP-2036] OOME after restart with heap, projection pushdown

sumwale:
Changes for SNAP-2214:

Issue of OOME was seen when testing for performance problem with spark_partition_id
function group by queries (SNAP-2036) which are used to see data distribution (perf is problem
  even in embedded especially if some data has overflowed to disk). The issue is that
decompression after recovery is not updating the UMM for the heap storage case. The reason
for this is that the "regionContext" and "diskId" are never set for a heap ColumnFormatValue so
it does not update the UMM for the region.

Changes for SNAP-2036:

All columns of a table get selected in the ColumnTableScan due to pushdown being skipped
entirely if there is a non-deterministic function (like spark_partition_id) anywhere.
This causes performance problem not only in smart connector (that pulls all columns
    unnecessarily) but also in embedded mode when some data is overflowed to disk
where it ends up read it from disk and decompressing even though they get projected
out later in the plan.

SNAP-2214

- explicitly acquire memory from UMM in compression/decompression
- release this memory in finally block in ColumnFormatValue compress/decompress
- use isManagedDirect consistently for explicit accounting in acquire/release
  (though non-managed direct memory case should never arise here after cache is initialized)

SNAP-2036
- added a PhysicalScan class taking from Spark's PhysicalOperation to extract
  filters/projections for table scan operator
- this does not skip the projection/filter entirely for non-deterministic functions
  that was added in SPARK-10316 rather removes non-deterministic filters at the end
  which are what cause trouble (e.g. rand() gets injected twice in the plan)
- added unit test to SnappySQLQuerySuite adapting from the spark test using row table
- removed unnecessary local variable in DSID
- removed PhysicalOperation unapply in LinkPartitionsToBuckets for index relation
  because it is traversing down the entire plan in any case

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/367

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/SnappySQLQuerySuite.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/SnappySQLQuerySuite.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 SnappyData, Inc. All rights reserved.
+ * Copyright (c) 2018 SnappyData, Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you
  * may not use this file except in compliance with the License. You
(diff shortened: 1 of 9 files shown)


sumwale:
@rishitesh It will be easier to review with whitespace changes hidden: https://github.com/SnappyDataInc/snappydata/pull/960/files?w=1

PR #963, opened 2018-02-13
[AQP-292] evaluate grouping keys explicitly if aggregates are using them

sumwale:
For the case of dictionary optimized aggregate, the grouping key is not materialized rather
only the dictionary index is made use of (except for first insert in map for a key where
    the full key is looked up separately). This caused trouble for the case when one of the
aggregate expressions uses the grouping key e.g. select id, count(id) from table group by id
because the key code would not be materialized.

It caused the compilation of snappy aggregate plan to fail which would fallback to spark
aggregation correctly. For some reason the fallback plan would sometimes fail with
a ClassCastException.

This change does not address the ClassCastException itself rather only fixes the snappydata
hash aggregation plan so that fallback spark plan is not used. Former may need to be tracked
separately in AQP JIRA.

- search for referred grouping keys in aggregate expressions
- send the matching indexes to ObjectHashMapAccessor generation and explicitly
  materialize them for dictionary optimized aggregate case
- added a test for above that checks for presence of snappy aggregation plan

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/SnappySQLQuerySuite.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/SnappySQLQuerySuite.scala
@@ -35,9 +35,11 @@
 
 package org.apache.spark.sql
 
-import io.snappydata.SnappyFunSuite
+import io.snappydata.{Property, SnappyFunSuite}
 import org.scalatest.Matchers._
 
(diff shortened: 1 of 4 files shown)


sumwale:
Merging, will take care of any review comments subsequently.

PR #965, opened 2018-02-15
[SNAP-2217] fix thread-unsafe paths in stats service and other cleanups

sumwale:

- use the new VTI to query for stats from smart connector instead of procedure
  with serialized blob; also added sync blocks to correct thread-unsafe paths and
  recreate the statement in case of any exception due to node failures or otherwise
- also corrected thread-unsafe mutable maps that are value of member info map;
  changed to use java CHM consistently
- changed explicit thread creation in getMembersStatsOnDemand with a Future
  which is both cleaner and does not require starting thread every time
- the new VTI calls will force the stats to be updated else thin stats service can
  fall 2 times behind (due to delay in both that service and embedded service runs)
- removed obsolete and unused ExternalClusterMode (SNAP-1837)
- replaced two start methods in TableStatsProviderService with a single one
- added jobserver.waitForInitialization flag to split dunit tests that submit jobs
  and need that to be ready immediately after start to fix occasional failures

## Patch testing

precheckin and manual

## ReleaseNotes.txt changes

Add information on the new VTI in docs @shyjaprabhu 

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/369

### DIFF
--- a/core/src/main/scala/io/snappydata/SnappyThinConnectorTableStatsProvider.scala
+++ b/core/src/main/scala/io/snappydata/SnappyThinConnectorTableStatsProvider.scala
@@ -19,14 +19,13 @@
 
 package io.snappydata
 
-import java.sql.{CallableStatement, Connection}
+import java.sql.{Connection, PreparedStatement, ResultSet}
 import java.util.{Timer, TimerTask}
 
-import scala.collection.JavaConverters._
+import scala.collection.mutable.ArrayBuffer
 import scala.util.control.NonFatal
 
-import com.gemstone.gemfire.internal.ByteArrayDataInput
-import com.gemstone.gemfire.{CancelException, DataSerializer}
+import com.gemstone.gemfire.CancelException
 import com.pivotal.gemfirexd.Attribute
 import com.pivotal.gemfirexd.internal.engine.ui.{SnappyExternalTableStats, SnappyIndexStats, SnappyRegionStats}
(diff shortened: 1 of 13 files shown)


core/src/main/scala/io/snappydata/SnappyThinConnectorTableStatsProvider.scala
```diff
@@ -108,23 +111,29 @@ object SnappyThinConnectorTableStatsProvider extends TableStatsProviderService {
...
+          isColumnTable, isReplicatedTable)
+      }
+      (regionStats, Nil, Nil)
     } catch {
-      case e: Exception =>
+      case NonFatal(e) =>
         logWarning("Warning: unable to retrieve table stats " +
             "from SnappyData cluster due to " + e.toString)
         logDebug("Exception stack trace: ", e)
         closeConnection()
```
rishitesh: I am not sure closing connection here is a good idea. It might be some temporary unavailability.

rishitesh: Probably should re-initialize connection for safety.

sumwale: Yes, connection is reinitialized by "executeStatsStmt" if it has been closed.

snappy-sachin:
Looks good to me.

PR #966, opened 2018-02-15
Condition in exist was incorrect.

rishitesh:
Condition in exist was incorrect.
Corrected the logic of existence join, which was looking for null values earlier. 
## Patch testing

Unit test added.

## ReleaseNotes.txt changes

NA

## Other PRs 

Na

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/execution/SnappyTableMutableAPISuite.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/execution/SnappyTableMutableAPISuite.scala
@@ -640,4 +640,40 @@ class SnappyTableMutableAPISuite extends SnappyFunSuite with Logging with Before
       df2.write.deleteFrom("col_table")
     }
   }
+  
+  // This test should be moved to a query suite.
+  test("Double exists") {
+    val snc = new SnappySession(sc)
+    snc.sql("create table r1(col1 INT, col2 STRING, col3 String, col4 Int)" +
+        " using row ")
+    snc.sql("create table r2(col1 INT, col2 STRING, col3 String, col4 Int)" +
+        " using row")
+
+    snc.sql("create table r3(col1 INT, col2 STRING, col3 String, col4 Int)" +
+        " using row")
+
+    snc.insert("r1", Row(1, "1", "1", 100))
+    snc.insert("r1", Row(2, "2", "2", 2))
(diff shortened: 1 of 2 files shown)


cluster/src/test/scala/org/apache/spark/sql/execution/SnappyTableMutableAPISuite.scala
```diff
@@ -640,4 +640,40 @@ class SnappyTableMutableAPISuite extends SnappyFunSuite with Logging with Before
...
+        "(exists (select col1 from r2 where r2.col1=r1.col1) " +
+        "or exists(select col1 from r3 where r3.col1=r1.col1))")
+
+    df.show
+    val result = df.collect()
+    assert(result.contains(Row(1, "1", "1", 100)))
+    assert(result.contains(Row(2, "2", "2", 2)))
+    assert(result.contains(Row(4, "4", "4", 4)))
+    assert(!result.contains(Row(7, "7", "7", 7)))
+  }
```
sumwale: For full check change to: checkAnswer(df, Seq(Row(1, "1", "1", 100), Row(2, "2", "2", 2), ...))

cluster/src/test/scala/org/apache/spark/sql/execution/SnappyTableMutableAPISuite.scala
```diff
@@ -640,4 +640,38 @@ class SnappyTableMutableAPISuite extends SnappyFunSuite with Logging with Before
...
+    snc.insert("r3", Row(4, "4", "4", 4))
+
+    val df = snc.sql("select * from r1 where " +
+        "(exists (select col1 from r2 where r2.col1=r1.col1) " +
+        "or exists(select col1 from r3 where r3.col1=r1.col1))")
+
+    val result = df.collect()
+    checkAnswer(df, Seq(Row(1, "1", "1", 100),
+      Row(2, "2", "2", 2), Row(4, "4", "4", 4) ))
+    assert(!result.contains(Row(7, "7", "7", 7)))
```
sumwale: Actually this assert is now redundant since checkAnswer does exact result check i.e. there must be no other rows than provided (and also accounts for sorting if there is a Sort in the plan else unsorted comparison). Hence had suggested checkAnswer.

sumwale:
@rishitesh move this test to SnappySQLQuerySuite as your comment also says

PR #969, opened 2018-02-17
Misc fixes, BUCKETS column in dashboard tables display

sumwale:

- added bucket count to dashboard and when collecting table stats;
  decreased width of size columns from 250px to 200px since they were showing too wide on
  bit lower resolution monitors (like 1280x1024)
- use a linked hash map when putting columns of a batch so that the stats row is ensured
  to be the last one written (to handle cases of partial writes in failures); store layer putAll
  from JDBC batch insert/put already uses a linked hash map so smart connector mode is fine
- force link partitions to buckets in case of any non-deterministic expressions in table scans;
  specifically this allows queries involving spark_partition_id() to create one-to-one partition
  to bucket mapping without having to set the "linkPartitionsToBuckets" property explicitly
- added a "ParseException" extension to AnalysisException and wrap all parse error in the
  same that also converts to better SQLException (spark parser has similar wrapper already)
- moved StoreDataSourceStrategy from datasources package name to sources to allow for
  normal info-level logs (the datasources package is marked with WARN level in the default
    store-log4j.properties to suppress large number of logs from external datasource readers)

## Patch testing

precheckin and manual

## ReleaseNotes.txt changes

Update the dashboard snapshots

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/370


snappy-sachin:
UI changes looks good to me.

PR #971, opened 2018-02-23
[SNAP-2225] Different results of nearly identical queries , due to join order

rishitesh:
Removed OrderlessHashpartition class. 
a) Handled bucket linking in TableExec and table scan. Removed the tableBucket param from partitioning class. If delink is enabled just keep numPartition as numBuckets. 

b) Also removed custom changes in HashPartition. We won't store bucket information in HashPartitioning. Instead based on the flag "linkPartitionToBucket" we can determine the number of partitions to be either numBuckets or num cores assigned to the executor.
This will help in enabling delinkPartition in Smart connector mode as well. This will be tracked with a separate ticket. 

c) Fixed couple of issues with partition pruning.
d) Fixed one issue with deleteFrom from IndextTest of the test suite. 
e) Removed ClusterSnappyJoinSuite, as both in core & cluster join order optimization should be applied. 

## Patch testing

precheckin

## ReleaseNotes.txt changes
NA

## Other PRs 
https://github.com/SnappyDataInc/spark/pull/95


cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
```diff
@@ -19,42 +19,39 @@ package io.snappydata
...
 
-  override def checkOrderlessHashPartitioning(partitioning: Partitioning): Option[
-      (Seq[Expression], Seq[Seq[Attribute]], Int, Int, Int)] = partitioning match {
-    case p: OrderlessHashPartitioning => Some(p.expressions, p.aliases,
-      p.numPartitions, p.numBuckets, p.tableBuckets)
+  override def checkHashPartitioning(partitioning: Partitioning): Option[
+      (Seq[Expression], Int, Int)] = partitioning match {
+    case p: HashPartitioning => Some(p.expressions, p.numPartitions, p.numBuckets)
     case _ => None
   }
```
sumwale: Same question. What does buckets here change? Since this won't work in smart connector so what is the difference in behaviour?

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -243,6 +248,87 @@ class SnappySessionState(snappySession: SnappySession)
...
+              r.resolveQuoted(colName, analysis.caseInsensitiveResolution)
+                  .getOrElse(throw new AnalysisException(
+                    s"""Cannot resolve column "$colName" among (${r.output})""")))
+            partCols
+          case _ => Nil
+        }
+        case _ => Nil
+      }
+    }
+
```
sumwale: Above two methods look similar to those in HashJoinStrategies (apart from Nil vs Int.MaxValue difference). Can be refactored into a base trait.

Or perhaps once this is applied, then that code in HashJoinStrategies can be removed and substituted by a simpler comparison (or this can determine colocated join here itself and inject a separate "ColocatedJoin" plan which can be resolved appropriately in HashJoinStrategies).

core/src/test/scala/org/apache/spark/sql/store/SnappyJoinSuite.scala
```diff
@@ -498,8 +501,41 @@ class SnappyJoinSuite extends SnappyFunSuite with BeforeAndAfterAll {
...
+        s"$t1 P JOIN $t2 R ON P.OrderId = R.OrderId" +
+        s" AND P.OrderRef <=> R.OrderRef")
+    // TODO Why an exchange is needed for coalesce.
+    checkForShuffle(nullableEquality.logicalPlan, snc, shuffleExpected = true)
+    assert(nullableEquality.count() === 500)
+
+    val withCoalesce = snc.sql(s"select P.OrderRef, P.description from " +
+        s"$t1 P JOIN $t2 R ON P.OrderId = R.OrderId" +
+        s" AND coalesce(P.OrderRef,0) = coalesce(R.OrderRef,0)")
+    // TODO Why an exchange is needed for coalesce.
```
sumwale: Good question. It should not.

sumwale:
@rishitesh Can you also add a test to check pruning for update/delete after these changes? If they still do not prune then track in a separate ticket.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
rishitesh: Refactored into a single trait.

core/src/test/scala/org/apache/spark/sql/store/SnappyJoinSuite.scala
rishitesh: The semanticEqual() check for an attribute & its Coalesce returns false.  We need to override semanticEqual method for some of these functions.

cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
rishitesh: Yes, In connector mode the scan will happen bucket wise always. When scan starts based on "linkBucketsToPartitions" flag it will determine number of scan partitions. This functionality can also be taken up at optimizer level, so that it works for both connector and embedded mode.

rishitesh:
@sumwale There are already test for update/select pruning. I have added one for delete as well.

PR #973, opened 2018-02-28
Fix for SNAP-1931, SNAP-1932 and SNAP-1906

kneeraj:
Fixes the issue of incomplete plan info in UI due to plan caching changes.
This also addresses SNAP-1931 and SNAP-1906

Chose an inexpensive way of substituting just the current value in the ParamLiteral nodes in the plan and overriding the toString() method to use the currentValue instead of the constant value field.
Expensive transformation has been avoided for every run. Instead used once per cacheddataframe to just collect all the paralliteral nodes. The Spark metrics get properly reflected. Previously due to transformstion these fields were getting null and hence the problem. 

## Patch testing

Manual testing

## ReleaseNotes.txt changes

None.

## Other PRs 

No.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
+++ b/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
@@ -34,12 +34,13 @@ import com.gemstone.gemfire.internal.shared.unsafe.{DirectBufferAllocator, Unsaf
 import com.gemstone.gemfire.internal.{ByteArrayDataInput, ByteBufferDataOutput}
 import com.pivotal.gemfirexd.internal.shared.common.reference.SQLState
 import org.apache.spark._
+import _root_.io.snappydata.collection.IntObjectHashMap
 import org.apache.spark.io.CompressionCodec
 import org.apache.spark.memory.DefaultMemoryConsumer
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.codegen.CodeAndComment
-import org.apache.spark.sql.catalyst.expressions.{Literal, LiteralValue, ParamLiteral, UnsafeProjection, UnsafeRow}
+import org.apache.spark.sql.catalyst.expressions.{LiteralValue, ParamLiteral, UnsafeProjection, UnsafeRow}
 import org.apache.spark.sql.collection.Utils
 import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.aggregate.CollectAggregateExec
@@ -94,6 +95,10 @@ class CachedDataFrame(session: SparkSession, queryExecution: QueryExecution,
     shuffleDependencies.length)
 
   private lazy val rowConverter = UnsafeProjection.create(queryExecution.executedPlan.schema)
+
+  // Map to cache the param literal positions to show correct plan info with
+  // new set of constants
+  private val paramLiterals = IntObjectHashMap.withExpectedSize[ParamLiteral](8)
 
   private[sql] def clearCachedShuffleDeps(sc: SparkContext): Unit = {
     val numShuffleDeps = shuffleDependencies.length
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
```diff
@@ -189,6 +188,14 @@ final class ParamLiteral(override val value: Any, _dataType: DataType, val pos:
     }
     ev.copy(initCode, isNullLocal, valueLocal)
   }
+
+  var currentValue: Any = value
```
sumwale: restrict to private[sql]

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
```diff
@@ -189,6 +188,14 @@ final class ParamLiteral(override val value: Any, _dataType: DataType, val pos:
...
+
+  override def toString: String = currentValue match {
+    case null => "null"
+    case binary: Array[Byte] => s"0x" + DatatypeConverter.printHexBinary(binary)
```
sumwale: s"" is useless here

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -557,17 +555,31 @@ object CachedDataFrame
     m
   }
 
+  // Some helper fields to show correct plan Info on UI as
+  // per current set of constants even though a cached plan
+  // is executed
+  var pls: Map[Int, ParamLiteral] = null
```
sumwale: As noted in general email, never use scala immutable maps for frequent lookups/writes. Change to IntObjectHashMap.withExpectedSize(8) or something and change to val. Also the visibility of this and the one below can be private. A better name like "paramLiterals" will be better.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -557,17 +555,31 @@ object CachedDataFrame
...
+        plan.transformAllExpressions {
+          case pl@ParamLiteral(_, _, _p) =>
+            pls = pls + (_p -> pl)
+            pl
         }
-        Literal(v, _dt)
-    })
+      }
+      if (allLiterals.length == pls.size) {
+        allLiterals.map(x => {
```
sumwale: change to foreach instead of map else it will unnecessarily create a Seq[Unit]

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -557,17 +555,31 @@ object CachedDataFrame
...
+          val pl = pls.getOrElse(x.position, emptyParameter).currentValue = x.value
```
sumwale: Prefer match to getOrElse in performance sensitive code. Latter creates a closure on the fly. With IntObjectHashMap a match against null.

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
kneeraj: Ok.

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
kneeraj: Yes. Just copied from Literals toString. Will remove.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
kneeraj: Ok.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
kneeraj: Ok.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
kneeraj: Ok.

kneeraj:
Yes it was an oversight. I thought it is inside the CachedDataFrame class only and not in object CachedDataFrame. Will rectify.
For Snap-1932 as discussed yesterday we should have the regexp replacement stuff. Right now it is lazy val and is wrong only as you pointed.

kneeraj:
>> Also track a unit test in the UI suite that is being worked on.

Yes. Will file a task for this.

PR #974, opened 2018-03-01
SNAP[2220] Corrected Row count updated/inserted in column table putInto 

rishitesh:
a) Gathered updated & inserted row count separately and converted that into an RDD. This RDD is sent to upper layer for showing row count in Snappy-shell

b) Added one validation to avoid updating dataframe with no columns to update. 

c) One test is moved to more appropriate SnappySqlQuerySuite. 
## Patch testing
Manual testing/precheckin

## ReleaseNotes.txt changes
NA
## Other PRs 

NA


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnPutIntoExec.scala
```diff
@@ -18,23 +18,35 @@ package org.apache.spark.sql.execution.columnar
...
+    sqlContext.sparkContext.parallelize(resultRow, 1)
+  }
+
+  override def executeCollect(): Array[InternalRow] = {
     // First update the rows which are present in the table
-    updatePlan.execute()
+    val u = updatePlan.executeCollectPublic().map(_.apply(0).asInstanceOf[Long]).toSeq.foldLeft(0L)(_ + _)
     // Then insert the rows which are not there in the table
-    insertPlan.execute()
+    val i = insertPlan.executeCollectPublic().map(_.apply(0).asInstanceOf[Long]).toSeq.foldLeft(0L)(_ + _)
```
sumwale: Why executeCollectPublic() and not executeCollect() both here and above?

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnPutIntoExec.scala
```diff
@@ -18,23 +18,35 @@ package org.apache.spark.sql.execution.columnar
...
+    val resultRow = new UnsafeRow(1)
+    val data = new Array[Byte](32)
+    resultRow.pointTo(data, 32)
+    resultRow.setLong(0, i + u)
+    Seq(resultRow).toArray
```
sumwale: Array(resultRow)

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnPutIntoExec.scala
```diff
@@ -18,23 +18,35 @@ package org.apache.spark.sql.execution.columnar
...
-
   override protected def doExecute(): RDD[InternalRow] = {
+    val resultRow = executeCollect()
+    sqlContext.sparkContext.parallelize(resultRow, 1)
+  }
+
+  override def executeCollect(): Array[InternalRow] = {
     // First update the rows which are present in the table
-    updatePlan.execute()
+    val u = updatePlan.executeCollectPublic().map(_.apply(0).asInstanceOf[Long]).toSeq.foldLeft(0L)(_ + _)
```
sumwale: getLong(0) instead of apply(0).asInstanceOf[Long] (both here and below)

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -861,27 +861,13 @@ private[sql] final class PreprocessTableInsertOrPut(conf: SQLConf)
...
-          l match {
-            case LogicalRelation(ps: PartitionedDataSourceScan, _, _) =>
-              if (!ps.partitionColumns.forall(a => child.output.exists(e =>
-                comp(e, a.toUpperCase)))) {
-                throw new AnalysisException(s"${child.output.mkString(",")}" +
-                    s" columns in the WHERE clause of the DELETE FROM statement must " +
-                    s"have all the parititioning column(s) ${ps.partitionColumns.mkString(",")}.")
-              }
-            case _ =>
-          }
```
sumwale: I think this is trying to enforce that partitioning columns should be part of delete, at least for row tables. This is a good thing to enforce else the deleteFrom will be very inefficient -- every row delete on row table will lead to a full scan in store side which will be very expensive. I think the problem in SNAP-2157 is that no partitioning has been defined for the table (try if that works by adding a partition_by C_CUSTKEY option). If that's the case perhaps the fix is to change default partitioning to be on primary key columns, if defined, for row tables.

rishitesh: DeleteFromTable only works if all the columns of the table match. Although the column names might not match. Hence it will automatically include column tables. But yes that validation also got skipped in this change for MutableRelations. Adding them back.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnPutIntoExec.scala
rishitesh: Will change

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnPutIntoExec.scala
rishitesh: Will change.

core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnPutIntoExec.scala
rishitesh: Will change. Leftover from some experiment.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
rishitesh: I am still not able to pinpoint the exact cause of SNAP-2157, as to why it was only failing for decimal types. When a primary key is defined it takes the join key as taken as PK. If partition key is there PK + partition keys are considered.

PR #979, opened 2018-03-10
[SNAP-2243][SNAP-2188] procedure for smart connector iteration and fixes

sumwale:
Includes fixes for perf issues noted in SNAP-2243 for all iterators (disk iterator, smart connector
and remote iterator). Smart connector has been changed to use a procedure to table scan for
performance as noted in SNAP-2188 and required for SNAP-2243.

Many other fixes for issues seen in testing. Like Filters in RowFormatScanRDD not getting
updated for subsequent executions, but issue was not seen due to another issue where
index columns were incorrect prefixed with null like "nullID" so there was no filter pushdown
happening in row tables.

- move the getDeletedRowCount() call in ColumnTableScan to be after the stats row filtering
  else it leads to all projected buffers to be fetched for smart connector or from disk/remote
- Implementation of StoreCallbacks.columnTableScan that translates Filters to Expressions
  and generates the code to apply the same locally to ColumnBatchIterator stats rows
- Changed smart connector iterator to use the new COLUMN_TABLE_SCAN procedure instead
  of multiple queries.
- Added passing of Filters to the plans and recreation of those in getPartitions if
  the parameter values have changed for ParamLiterals
- Fixed RowFormatScanRDD to regenerate filter clause in getPartitions if the parameter
  values have changed for ParamLiterals (not seen earlier because index columns were
    incorrect which has been fixed in store)
- Perf fix to ColumnFormatIterator: keep track of updated delta stats separately with
  forced faultin like the full stats in DiskMultiColumnBatch so that entire batch does
  not need to be read if filter can skip using stats
- Perf fix to RemoteEntriesIterator:
  - fetch both full stats and delta stats rows when fetching keys first time
  - sort and ensure both rows of a batch are together when fetching other columns
- Updated ParamLiteral serialization to replace its value with the updated one in LiteralValue
  since parameter may have changed (but the base Literal.value is a val and cannot be changed)
- check if Spark is using IP addresses in smart connector cluster, and if so then
  extract the address portion from the returned bucket-server mapping instead of
  using host names so that NODE_LOCAL routing happens for connector jobs
- set the default shuffle partitions to be defaultParallelism (which uses the total cores in system)
  with an upper cap of the Spark setting
- clear plan cache when shuffle partition setting changes to enable it be applied on all plans
- fixed a missing release of storage memory in smart connector iterator after decompression
  when the decompressor has acquired new buffer (using ByteOrder as an indication in
      the returned buffer whether it was decompressed or not)
- Corrected RDD to be cleared in CachedDataFrame to use either the cachedRDD or the
  last used one for execution.
- Added handling of the new DECOMPRESS_IF_IN_MEMORY fetch type to return self (or null)
  if decompression cannot replace the underlying in-memory value
- Updated ColumnFormatValue to store disk RegionEntry since DiskId can change.
- Improved performance of Snappy stats iterator to avoid lookup of deleted bitmask column for
  every stats column, rather iterate both of them and add negative size for deletes if any.
- Refactored PooledKryoSerializer to add generic methods that accept closures
- Added more transient expected exception types to SnappyTestRunner.
- Updated store link.

## Patch testing

precheckin, manual testing for the issues.

Specific unit tests for the issues will be added in subsequent PRs.

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/372


sumwale:
Merging now. Will take care of any review comments subsequently.

PR #980, opened 2018-03-11
[SNAP-2244] stats for delta column batches

sumwale:

- Add delta column stats with same format/schema as full stats. Columns that have seen
  no updates will have the three fields set as null (so cost 3 bits in UnsafeRow).
- Refactored stats generation in ColumnInsertExec to move into object for use by update
  and also removed explicit handling for wide-column schema. Now the full code for stats
  is not generated in one shot rather each column variable holds own code so the
  GenerateUnsafeProjection code generation will split automatically if required.
- Added update delta stats row generation using above in ColumnUpdateExec.
- Added merge of multiple delta stats as part of ColumnDelta.apply.
- Refactored CodeGeneration a bit to allow for code generation of UnsafeProjection for above
  merge code that needs to write the final merged UnsafeRow.
- Removed checks/add of delta stats in delete bitmask column which was never required since
  delete bit mask is fetched separately in any case after all stats checks are done
  (and will lead to read of all projected columns).
- Skip pushed down filters for types that do not maintain stats (e.g. calendar, complex types)
- Added delta stats count in ColumnTableScan and stats service
  (ColumnFormatKey.getColumnBatchRowCount)

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnDelta.scala
```diff
@@ -77,37 +80,106 @@ final class ColumnDelta extends ColumnFormatValue with Delta {
...
+      // returns null if either of the sides is null). The "AddStats" extension
+      // to Add operator is to encapsulate that behaviour.
+      val lower =
+        if (newLower == null) oldLower
+        else if (oldLower == null) newLower
+        else if (ordering.lt(oldLower, newLower)) oldLower else newLower
+      val upper =
+        if (newUpper == null) oldUpper
+        else if (oldUpper == null) newUpper
+        else if (ordering.lt(oldLower, newLower)) newLower else oldLower
```
vivekwiz: hi @sumwale, I think this comparison should be of upper values. Also if we are comparing, here the upper of both should preside. But I did not get why this comparison is needed?

sumwale: Yes, should be upper comparison -- thanks for spotting. Was having failure in a test due to the same.

PR #993, opened 2018-03-22
added a method to identify if the table is internal column batch table

ahshahid:
Pls refer to bug
[ENT-22](https://snappydata.atlassian.net/browse/ENT-22)

Added a method in the Callback to determine if the table is internal column batch storing table.
Linked PR request
[SnappyStore](https://github.com/SnappyDataInc/snappy-store/pull/376)
## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
```diff
@@ -184,6 +184,10 @@ object StoreCallbacksImpl extends StoreCallbacks with Logging with Serializable
     list
   }
 
+  override def isInternalBatchTable(tableName: String): Boolean =
```
ashetkar: Can you use the existing method isColumnTable() for the same? No need to write this one.

PR #995, opened 2018-03-29
Fix for bug snap-2255. Closing the connection so as to return the poo…

ahshahid:
…led connection back to pool
Pls refer to bug SNAP-2255
[SNAP-2255](https://jira.snappydata.io/browse/SNAP-2255)
Returning the PooledConnection to the pool by closing it , leaving the underlying EmbedConnection intact. Without this the system becomes unusable after certain number of queries.
I have modified the code in the ColumnBatch , though I have not tried to reproduce the error from that code path.
Added bug test

## ReleaseNotes.txt changes

## Other PRs 
[Corresponding PR in Store](https://github.com/SnappyDataInc/snappy-store/pull/377)

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnBatch.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnBatch.scala
@@ -45,7 +45,8 @@ case class ColumnBatch(numRows: Int, buffers: Array[ByteBuffer],
     statsData: Array[Byte], deltaIndexes: Array[Int])
 
 abstract class ResultSetIterator[A](conn: Connection,
-    stmt: Statement, rs: ResultSet, context: TaskContext)
+    stmt: Statement, rs: ResultSet, context: TaskContext,
+    closeConnectionOnResultsClose: Boolean = true)
     extends Iterator[A] with Logging {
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnBatch.scala
```diff
@@ -85,30 +85,40 @@ abstract class ResultSetIterator[A](conn: Connection,
...
+          stmt.close()
+        }
+      } catch {
+        case NonFatal(e) => logWarning("Exception closing statement", e)
+      }
+      hasNextValue = false
+    } finally {
+      try {
+        if (conn != null) {
+          conn.close()
```
sumwale: Looked at the calls to ResultSetIterator. Two of the paths probably require the connection close but the one from ColumnBatchIteratorOnRS is closed separately in SmartConnectorColumnRDD AFTER the commit is called, so you cannot close the connection here. So you probably need a boolean in the constructor whether to close the connection or not and pass as false at least from ColumnBatchIteratorOnRS.

@suranjan Please check this if there was some deliberate reason why connection close was removed for all paths. From what I recall this code used to have a connection close earlier.

PR #996, opened 2018-04-02
Snap 2080

kneeraj:

creating index on columns added through 'alter table' was failing. The string passed to gemxd layer was containing 'clob' for column type even for varchar and char column added. Now the change consults the exact data type and forms the string accordingly.

## Patch testing

Added a test.

## ReleaseNotes.txt changes

No.

## Other PRs 

No

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/row/JDBCMutableRelation.scala
+++ b/core/src/main/scala/org/apache/spark/sql/row/JDBCMutableRelation.scala
@@ -19,9 +19,7 @@ package org.apache.spark.sql.row
 import java.sql.Connection
 
 import scala.collection.mutable
-
-import io.snappydata.SnappyTableStatsProviderService
-
+import io.snappydata.{Constant, SnappyTableStatsProviderService}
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/row/JDBCMutableRelation.scala
```diff
@@ -441,6 +439,26 @@ case class JDBCMutableRelation(
     throw new UnsupportedOperationException()
   }
 
+  private def getDataType(column: StructField): String = {
```
rishitesh: There is a method  GemFireXDBaseDialect.getJDBCType. Will that method work here?  The code looks very similar.

kneeraj: Let me check. We have the dialect object there. Should work.

kneeraj:
Rishi has given a go ahead. He forgot to approve. Going ahead with the merge.

PR #998, opened 2018-04-03
[SNAP-2241] Handled catalog database creation.

rishitesh:

If a function is created the very first thing after creating the schema, it fails as the schema is
not registered in Hive catalog. In table creation we handle the catalog db creation if the db is not present.
Similar handling added for functions.

## Patch testing

precheckin

## ReleaseNotes.txt changes

na

## Other PRs 

NA

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/store/SnappyUDFTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/store/SnappyUDFTest.scala
@@ -23,6 +23,7 @@ import com.pivotal.gemfirexd.internal.engine.Misc
 import io.snappydata.SnappyFunSuite
 import org.scalatest.BeforeAndAfterAll
 
+import org.apache.spark.jdbc.{ConnectionConfBuilder, ConnectionUtil}
 import org.apache.spark.sql.udf.UserDefinedFunctionsDUnitTest._
 
 case class OrderData(ref: Int, description: String, price: Long,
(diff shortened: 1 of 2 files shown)


cluster/src/test/scala/org/apache/spark/sql/store/SnappyUDFTest.scala
```diff
@@ -472,9 +473,48 @@ class SnappyUDFTest extends SnappyFunSuite with BeforeAndAfterAll {
...
+      snc.sql("CREATE TABLE trade.rr_test_table(OrderRef INT NOT NULL, description String, " +
+          "price BIGINT, serviceTax DECIMAL, surcharge Float, purchase_date DATE, time Timestamp)")
+
+      snc.sql("CREATE TABLE trade.col_test_table(OrderRef INT NOT NULL," +
+          " description String, price  " +
+          "LONG, serviceTax DECIMAL, surcharge Float, purchase_date DATE, time Timestamp) " +
+          "using column options(PARTITION_BY 'OrderRef')")
+
+
+      snc.sql("select TRADE.strudf(description) from trade.col_test_table").collect()
```
kneeraj: If I set the current schema as "TRADE" on the jdbcconn and then use the function without specifying the "schemaname" in the function, then will it work?

cluster/src/test/scala/org/apache/spark/sql/store/SnappyUDFTest.scala
```diff
@@ -472,9 +473,48 @@ class SnappyUDFTest extends SnappyFunSuite with BeforeAndAfterAll {
...
+          " @Override public String call(String s){ " +
+          "               return s + s; " +
+          "}" +
+          "}"
+      val file = createUDFClass("StringUDF", udfText)
+      val jar = createJarFile(Seq(file))
+
+      snc.sql(s"CREATE FUNCTION TRADE.STRUDF AS StringUDF " +
+          s"RETURNS STRING USING JAR " +
+          s"'$jar'")
```
kneeraj: Minor format comment. May be you can move the $jar in the second line itself.

PR #1002, opened 2018-04-12
Add test to reproduce SNAP-2230 and changes in existing dmlOps row table tests to have primary key constraints

swatisawant:

- Added a test to reproduce SNAP-2230. This test includes:
   1. Single insert ops with concurrency in dmlOps test
   2. Data Size: initial 1 GB data loading and then inserts for at least 20 million records
   3. The cluster should include 2 servers, 1 lead and 1 locator node
   4. Each server have 25 threads performing insert ops
   5. Server HA after every half an hour 
   6. Two times cluster restart at the end of the test

- Modified existing dmlOps row table tests to have primary key constraints

## Patch testing

Ran the new bt and existing dmOps bt with the changes to have primary key constraints

## ReleaseNotes.txt changes

None

## Other PRs 

No


dtests/src/test/java/io/snappydata/hydra/testDMLOps/SnappyDMLOpsUtil.java
```diff
@@ -1185,6 +1186,11 @@ public String getRowFromCSV(String tableName, int randTable) {
     //    File.separator + csvFileName);
     try (Stream<String> lines = Files.lines(Paths.get(csvFilePath + File.separator + csvFileName))) {
       row = lines.skip(insertCounter).findFirst().get();
+    } catch (NoSuchElementException nse) {
```
sonalsagarwal: Since it not required in all tests, this can be made flag based, we have a flag with the bug number, why we need to re-insert the data.

swatisawant: Done.

PR #1003, opened 2018-04-16
Snap 2128

hemanthmeka:
1: SNAP-2128: Support Store specific DDLs in SnappyDDLParser
2: adding support for CALL syntax


dhavalstr:

hemanthmeka:
Made DISK_STORE and CALL as nonReservedKeyword.

PR #1008, opened 2018-04-23
Fix for bug SNAP-2308. The fix involves allowing query with decimal n…

ahshahid:
…umbers suffixed by 'BD' to be parsed correctly & recognized as numeric literals

Pls refer to bug SNAP-2308
[SNAP-2308](https://jira.snappydata.io/browse/SNAP-2308)

Changes to recognize BD as a valid suffix for the decimal number and allowing numeric literal to be created out of it.

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


ahshahid:
Does the "BD" suffix if explicitly given, should also work for row store parser?

ahshahid:
pls review. this needs to be part of any build to be given to jnj

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -175,8 +184,8 @@ class SnappyParser(session: SnappySession)
 
   protected final def numericLiteral: Rule1[Expression] = rule {
     capture(plusOrMinus.? ~ Consts.numeric. + ~ (Consts.exponent ~
-        plusOrMinus.? ~ CharPredicate.Digit. +).? ~ Consts.numericSuffix.?) ~
-        delimiter ~> ((s: String) => toNumericLiteral(s))
+        plusOrMinus.? ~ CharPredicate.Digit. +).? ~ Consts.numericSuffix.? ~
+        Consts.numericSuffix.?) ~ delimiter ~> ((s: String) => toNumericLiteral(s))
```
sumwale: This will cause suffixes like "bl", "df" to also be accepted (with last char used to determine) which is not proper.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -103,10 +103,18 @@ class SnappyParser(session: SnappySession)
...
-          java.lang.Double.parseDouble(s.substring(0, len - 1)), DoubleType)
+        if (s.length > 2 && (s.charAt(len - 2) match {
+          case 'B' | 'b' => true
+          case _ => false
+        })) {
+          return toDecimalLiteral(s.substring(0, len - 2), checkExactNumeric = false)
+        } else {
+          return newTokenizedLiteral(
+            java.lang.Double.parseDouble(s.substring(0, len - 1)), DoubleType)
+        }
```
sumwale: Add checks for valid suffixes too:

```scala
if (s.length > 2) {
  s.charAt(len - 2) match {
    case 'B' | 'b' => toDecimalLiteral(...)
    case c if Character.isDigit(c) => newTokenalizedLiteral(...)
    case _ => throw new ParseException(...)
  }
}
```

Likewise for non 'D' | 'd' cases, the previous char should be checked that it should be a digit and never a numericSuffix -- see next comment.

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
ahshahid: yes. that is true, but how to avoid that in parsing phase, i could not figure out. I am hoping that at later stage in analysis phase exception will be thrown.

sumwale: Either add checks for valid suffixes for each of the cases like mentioned below. Or you could use: ~ (ignoreCase("BD") | Consts.numericSuffix).? instead of two Consts.numericSuffix (and remove 'B' from numericSuffix). I think the first approach will be more efficient so just add checks for valid suffixes like mentioned in the comment below.

core/src/test/scala/org/apache/spark/sql/store/ColumnTableTest.scala
```diff
@@ -138,6 +138,47 @@ class ColumnTableTest
...
+    query = s"SELECT sum(Col1) as summ FROM $tableName having summ > .001f"
+    result = snc.sql(query)
+    result.collect
+    snc.sql(s"create or replace view $viewName as ($query)")
+
+    query = s"SELECT sum(Col1) as summ FROM $tableName having summ > .001d"
+    result = snc.sql(query)
+    result.collect
+    snc.sql(s"create or replace view $viewName as ($query)")
+
```
sumwale: Can you also add a couple of checks for invalid suffixes like "bl", "fd" that they should throw ParseExceptions?

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -103,16 +103,32 @@ class SnappyParser(session: SnappySession)
...
+      case 'F' | 'f' => if (Character.isDigit(s.charAt(len - 2))) {
         return newTokenizedLiteral(
           java.lang.Float.parseFloat(s.substring(0, len - 1)), FloatType)
-      case 'L' | 'l' =>
+      } else {
+        throw new ParseException(s"Found non numeric token $s")
+      }
+      case 'L' | 'l' => if (Character.isDigit(s.charAt(len - 2))) {
```
sumwale: To be safe might be good to add length check too here and the case before: if (s.length > 2 && Character.isDigit(...))

ahshahid: Is it needed? Because I suppose the parser rule itself would ensure atleast one number followed by suffix, so I deliberately avoided the check.

PR #1020, opened 2018-04-30
Ent 27

ahshahid:
Initial changes to push filter predicates to scan level.
I am not sure if the use of check "GemFireRelation" to push predicate is the way to go.
Also there is some bug in reconstructing query in replacing literal resulting in StringIndexOutOfBounds.
I will debug tomorrow ( unless you want to take a look at it).  Without try catch some of the connector tests will fail.
Right now all the connector tests are passing ...

Associated pull requests:
[ENT-27 on spark](https://github.com/SnappyDataInc/spark/pull/100)
[ENT-27 on snappy-connectors](https://github.com/SnappyDataInc/snappy-connectors/pull/16)
## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


sumwale:
Don't understand these changes. A whole new strategy which is largely a copy has been added to what end? What is the issue that this is attempting to fix? Why was it not possible to fix the current store strategies instead? If filters are not being pushed down then why not? Does lack of pushed down filters cause incorrect results or this is a perf fix? And lastly why is this a problem only with GemFireRelation and not with all external relations? And if filters are not getting pushed down for other external relations too, then that is the problem that needs to be fixed.

This needs more discussion. A temporary build can be provided for JnJ but this definitely looks like a wrong approach to fix whatever is the underlying problem.

@ahshahid @kneeraj

sumwale:
I think you can add transformation to Literals in StoreDataSourceStrategy. A case for other LogicalRelations like "case l@LogicalRelation(b, _, _) => planLater(plan.transformAllExpressions(...))".

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -163,6 +163,7 @@ object Constant {
     // and the Spark impls create the seed in constructor rather than in generated code
     "RAND" -> Array.emptyIntArray, "RANDN" -> Array.emptyIntArray,
     "LIKE" -> Array(1), "RLIKE" -> Array(1)))
+
```
sumwale: can be reverted

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -2134,6 +2135,7 @@ object SnappySession extends Logging {
       }
       sb.toString
     } else text
+
   }
 
   private def newId(): Int = {
```
sumwale: can be reverted

core/src/main/scala/org/apache/spark/sql/execution/sources/StoreDataSourceStrategy.scala
```diff
@@ -72,6 +72,20 @@ private[sql] object StoreDataSourceStrategy extends Strategy {
           0,
           Nil,
           (a, f) => t.buildUnsafeScan(a.map(_.name).toArray, f.toArray)) :: Nil
+      case LogicalRelation(_, _, _) => {
+        var foundParamLiteral = false
+        val tp = plan.transformAllExpressions {
+          case pl: ParamLiteral => {
+            foundParamLiteral = true
+            pl.asLiteral
+          }
```
sumwale: braces in case are unnecessary

core/src/main/scala/org/apache/spark/sql/execution/sources/StoreDataSourceStrategy.scala
```diff
@@ -72,6 +72,20 @@ private[sql] object StoreDataSourceStrategy extends Strategy {
           0,
           Nil,
           (a, f) => t.buildUnsafeScan(a.map(_.name).toArray, f.toArray)) :: Nil
+      case LogicalRelation(_, _, _) => {
+        var foundParamLiteral = false
+        val tp = plan.transformAllExpressions {
```
sumwale: Can you add a comment here on why this has been added (like in https://github.com/SnappyDataInc/snappydata/pull/1035/files#diff-d32d8c68c3b9a5c7e05772e506b0ace5)

PR #1021, opened 2018-04-30
Snap 2297 -- deployment of packages as a ddl command

kneeraj:

deploy packages as ddl command and then make sure that jars get downloaded and we persist it too so that on restarts and new nodes joining all of them see the deployed jars.

## Patch testing

Manual testing with Cassandra cluster. HA testing and new node joining testing also.
The plan is to add a bt which will be run as part of hydra runner ( smoke )

## ReleaseNotes.txt changes

Deployment of maven and Spark packages can be done through the newly introduced ddl. 

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/385


cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -234,6 +235,11 @@ class LeadImpl extends ServerImpl with Lead
         markLauncherRunning(if (startupString ne null) s"Starting $startupString" else null)
       }
 
+      // Add a URL classloader to the main thread so that new URIs can be added
+      val parent = Thread.currentThread().getContextClassLoader
```
rishitesh: Its a common practice to use Utils.getContextOrSparkClassLoader. You can use it for parent classloader.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -663,6 +669,15 @@ class LeadImpl extends ServerImpl with Lead
   }
 }
 
+class ExtendibleURLClassLoader(parent: ClassLoader)
```
rishitesh: Theoretically, this classloader should replace install jar and its corresponding changes in job server and executors. We should track this and do the enhancement and do away with install jars. Else there will be too many options for jar loading.

cluster/src/main/scala/org/apache/spark/sql/streaming/SnappyStreamingContextFactory.scala
```diff
@@ -84,7 +84,15 @@ class SnappyStreamingContextFactory extends SparkContextFactory {
       // If Job class directly refers to any jars which has been provided
       // by install_jars, this can help.
       override def makeClassLoader(parent: ContextURLClassLoader): ContextURLClassLoader = {
-        SnappyUtils.getSnappyContextURLClassLoader(parent)
+        val cl = SnappyUtils.getSnappyContextURLClassLoader(parent)
```
rishitesh: lead.urlclassloader can be passed as the parent to SnappyUtils.getSnappyContextURLClassLoader(parent) ?

cluster/src/main/scala/io/snappydata/gemxd/ClusterCallbacksImpl.scala
```diff
@@ -125,4 +124,17 @@ object ClusterCallbacksImpl extends ClusterCallbacks with Logging {
     }
     GemFireVersion.getClusterType
   }
+
+  override def setLeadClassLoader(): Unit = {
```
rishitesh: Is it for query execution? If so can't we do at SparkSQLExecuteImpl ?

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -75,6 +74,8 @@ class LeadImpl extends ServerImpl with Lead
   private var remoteInterpreterServerClass: Class[_] = _
   private var remoteInterpreterServerObj: Any = _
 
+  var urlclassloader: ExtendibleURLClassLoader = _
```
rishitesh: Can it be made a val? We are just adding urls to it and not changing the reference as such.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -234,6 +235,11 @@ class LeadImpl extends ServerImpl with Lead
         markLauncherRunning(if (startupString ne null) s"Starting $startupString" else null)
       }
 
+      // Add a URL classloader to the main thread so that new URIs can be added
+      val parent = Thread.currentThread().getContextClassLoader
+      urlclassloader = new ExtendibleURLClassLoader(parent)
+      Thread.currentThread().setContextClassLoader(urlclassloader)
```
rishitesh: Do we have to set the classloader in start method?  I guess only query & job path should suffice?
You are already taking care of adding all urls at restart.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
rishitesh: Also, this extension class does not do much. Can't it be a simple URLClassloader?

cluster/src/main/scala/org/apache/spark/sql/SnappySessionFactory.scala
```diff
@@ -88,7 +87,15 @@ object SnappySessionFactory {
       // If Job class directly refers to any jars which has been provided
       // by install_jars, this can help.
       override def makeClassLoader(parent: ContextURLClassLoader): ContextURLClassLoader = {
-        SnappyUtils.getSnappyContextURLClassLoader(parent)
+        val cl = SnappyUtils.getSnappyContextURLClassLoader(parent)
```
rishitesh: urlClassloader as parent?

cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
```diff
@@ -47,4 +53,34 @@ object ToolsCallbackImpl extends ToolsCallback {
...
+      snappyexecutor.updateMainLoader(jars)
+    }
+  }
+
+  override def getAllGlobalCmnds(): Array[String] = {
+    GemFireXDUtils.waitForNodeInitialization()
+    Misc.getMemStore.getGlobalCmdRgn.values().toArray.map(_.asInstanceOf[String])
+  }
+
+  override def removePackage(alias: String): Unit = {
```
rishitesh: Can you please explain here a bit, how remove package would work? What's the scope of it? I believe it's only removing from the region and not from the actual classloader. Should we not change lead & executors classloaders here?

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
kneeraj: Yes my intention is to remove install jar. That will be the next step. However in rowstore mode it should continue to work. That is why did not remove it eagerly.

cluster/src/main/scala/org/apache/spark/sql/streaming/SnappyStreamingContextFactory.scala
kneeraj: Hmm.. May be will change and see. Because parent will already = lead.uelclassloader. and in that case additional addUrl won't be required.

cluster/src/main/scala/io/snappydata/gemxd/ClusterCallbacksImpl.scala
kneeraj: LeadNodeExecutorMsg execute calls the sparksqlexecuteimpl thingy. Just a layer above that.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
kneeraj: May be lazy val.

PR #1024, opened 2018-05-02
taking creds from globalSparkContext to shutdown the store

ahshahid:
Precheckin tests fail as once a secured system is created in a test, it is not possible to shutdown the system using sparkContext.stop as creds are not passed.
Change is to retrieve creds from globalSparkContext ( assuming this is the context created during boot) & passing it to Utils for shutdown.
This fixes the failure of BugTest when run in the suite.

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
@@ -31,8 +31,9 @@ class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
   private val sysUser = "gemfire10"
 
   override def beforeAll(): Unit = {
-   this.stopAll()
+    this.stopAll()
   }
+  
   protected override def newSparkConf(addOn: (SparkConf) => SparkConf): SparkConf = {
     val ldapProperties = SecurityTestUtils.startLdapServerAndGetBootProperties(0, 0, sysUser,
       getClass.getResource("/auth.ldif").getPath)
@@ -70,10 +71,7 @@ class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
     }
     System.clearProperty(Constant.STORE_PROPERTY_PREFIX + Attribute.USERNAME_ATTR)
     System.clearProperty(Constant.STORE_PROPERTY_PREFIX + Attribute.PASSWORD_ATTR)
-  //  System.setProperty(Attribute.AUTH_PROVIDER, "NONE")
-  //  System.setProperty(Constant.STORE_PROPERTY_PREFIX + Attribute.AUTH_PROVIDER, "NONE")
     System.setProperty("gemfirexd.authentication.required", "false")
-
   }
 
   test("Bug SNAP-2255 connection pool exhaustion") {
@@ -83,7 +81,6 @@ class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
     val snc1 = snc.newSession()
     snc1.snappySession.conf.set(Attribute.USERNAME_ATTR, user1)
     snc1.snappySession.conf.set(Attribute.PASSWORD_ATTR, user1)
-
 
     snc1.sql(s"create table test (id  integer," +
         s" name STRING) using column")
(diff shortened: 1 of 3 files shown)


ahshahid:
Though this itself appears to be a security hole, as any one can shutdown spark context.

ahshahid:
I suppose we  should have a stop method in snappy context , in which before calling sparkContext.stop, we set the user field in Object SnappyContext & in the call back listener , before invoking fabric server's stop, we check if the global spark Context's user matches with the user of the snappy context invoking the shutdown & then allowing it to go head.
But if the user directly stop's using spark context, what should we do?
Even if we don't shutdown the fabric data base, spark would be stopped.
If the user creates a new spark context, will he/she get a proper system back, reusing the existing fabric server?

ahshahid:
pls review. this is needed to pass the BugTest:testConnectionPool exhaustion

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1150,7 +1151,17 @@ object SnappyContext extends Logging {
       // clear current hive catalog connection
       Hive.closeCurrent()
       if (ExternalStoreUtils.isLocalMode(sc)) {
-        ServiceUtils.invokeStopFabricServer(sc)
+        val user = sc.getConf.get(Constant.STORE_PROPERTY_PREFIX + Attribute.USERNAME_ATTR, "")
+        val props = if (!user.isEmpty) {
+          val prps = new java.util.Properties();
+          val pass = sc.getConf.get(Constant.STORE_PROPERTY_PREFIX + Attribute.PASSWORD_ATTR, "")
+          prps.put(com.pivotal.gemfirexd.Attribute.USERNAME_ATTR, user)
```
sumwale: sc.getConf will create a copy of SparkConf in every call. Use sc.conf instead which has package-level visibility (and is visible here).

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1150,7 +1151,17 @@ object SnappyContext extends Logging {
       // clear current hive catalog connection
       Hive.closeCurrent()
       if (ExternalStoreUtils.isLocalMode(sc)) {
-        ServiceUtils.invokeStopFabricServer(sc)
+        val user = sc.getConf.get(Constant.STORE_PROPERTY_PREFIX + Attribute.USERNAME_ATTR, "")
+        val props = if (!user.isEmpty) {
```
sumwale: cleaner way is pattern matching:

sc.conf.getOption(Constant.STORE_PROPERTY_PREFIX + Attribute.USERNAME_ATTR) match {
  case Some(user) => ...
  case None => null
}

cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
```diff
@@ -20,18 +20,19 @@ import java.util.Properties
...
-
 import org.junit.Assert.{assertEquals, assertFalse, assertTrue}
 
 import org.apache.spark.SparkConf
 
 class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
   private val sysUser = "gemfire10"
 
+  override def beforeAll(): Unit = {
+   this.stopAll()
```
sumwale: 2 space indentation

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
sumwale: @ashetkar
Also I see that explicit "Constant.STORE_PROPERTY_PREFIX + Attribute.USERNAME_ATTR" has been used in all places for security user/password checks. For smart connector the prefix changes to "spark.snappydata.store" which is also checked I think.

This should instead be added as property in io.snappydata.Property (Literals.scala) like other store properties there. Then: Property.User.getOption(...) which will automatically check for both prefixes (snappydata.store and spark.snappydata.store). Need not be part of these changes but I think this cleanup should be done at some point.

PR #1027, opened 2018-05-04
[SNAP-2312] Handled int overflow case.

rishitesh:

For a large number of distinct keys ObjectHashSet might have entries near to 1 << 30.
Multiplying that to 8 causes int overflow to a negative number.
## Patch testing

precheckin 

## Other PRs 

NA


core/src/main/scala/io/snappydata/collection/ObjectHashSet.scala
```diff
@@ -290,7 +290,7 @@ final class ObjectHashSet[T <: AnyRef : ClassTag](initialCapacity: Int,
 
   private def freeMemoryOnTaskCompletion(): Unit = {
     taskContext.addTaskCompletionListener { _ =>
-      consumer.freeMemory(totalSize)
+        consumer.freeMemory(totalSize)
```
sumwale: incorrect extra indentation added?

rishitesh: Oh. ok. Will fix it.

PR #1032, opened 2018-05-10
- Modified all scripts in recovery bts to work as per snappy syntax f…

swatisawant:

- Modified all scripts in recovery bts to work as per snappy syntax for create table ddl
- Code refactoring changes

## Patch testing

Verified the changes by running recovery bts with the small volume of data.

## ReleaseNotes.txt changes

NA

## Other PRs 

None


core/src/main/scala/org/apache/spark/sql/store/StoreUtils.scala
```diff
@@ -472,7 +472,7 @@ object StoreUtils {
     } else {
       parameters.remove(DISKSTORE).foreach { v =>
         if (isPersistent) sb.append(s"'$v' ")
-        else throw Utils.analysisException(
+        else if (!isPersistent && !overflow) throw Utils.analysisException(
```
spillai14: Can you please explain this  check  and why is it required.

swatisawant: It is a fix for SNAP-2336(PERSISTENCE as none does not work with user-specified diskstore in overflow configuration). Made a change to allow this.

PR #1039, opened 2018-05-22
Travis CI test

cruizen:

Creating a simple .travis.yml file to be able to build from Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,14 @@ language: java
 
 jdk:
   - oraclejdk8
+  
+git:
+  depth: false
 
 notifications:
-  email: false
+  email:
+    recipients:
+      - process@snappydata.io
   
 # whitelist
 branches:
@@ -15,4 +20,4 @@ branches:
 sudo: false 
 
 script: 
-  ./gradlew product
+  ./gradlew precheckin
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ notifications:
 # whitelist
 branches:
   only:
-    - master
+    - trilok_travis_ci
     
 sudo: false 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: java
+
+jdk:
+  - oraclejdk8
+
+notifications:
+  email: false
+  
+# whitelist
+branches:
+  only:
+    - master
+    
+sudo: false 
+
+script: 
+  ./gradlew product
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ notifications:
 branches:
   only:
     - trilok_travis_ci
+    - master
     
 sudo: false 
 


cruizen:
I have disabled pull request builds for now in travis-ci.com 
The build status for this pull request can be seen at https://travis-ci.com/SnappyDataInc/snappydata/builds/74100932

.travis.yml
```diff
@@ -0,0 +1,18 @@
+language: java
+
+jdk:
+  - oraclejdk8
+
+notifications:
+  email: false
```
sumwale: Will be good to have email notifications. We use "process@snappydata.io" for github notifications so that same can be used here.

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+  
+# whitelist
+branches:
+  only:
+    - trilok_travis_ci
+    - master
```
sumwale: Shouldn't this be on all branches for which PRs are created? That way one can see the status of precheckin before a merge (and a successful pass will become a requirement for merge) rather than after the merge.

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+    
+sudo: false 
+
+script: 
+  ./gradlew product
```
sumwale: This should be to run all the unit tests: ./gradlew precheckin -Pstore

sumwale: To save time, this one can skip store tests: ./gradlew precheckin, while there can be a separate configuration for PR merges in snappy-store repository that can run with -Pstore

sumwale:
One other question: where does the configuration specify how to update the git checkout? We need to be able to update both snappydata as well as submodules using "git pull && git submodule sync --recursive && git submodule update --init --recursive"

.travis.yml
cruizen: Sure. Just demoing that it is possible to whitelist or exclude branches. I can comment out or remove this part. 
The idea of pull request builds is to add a pre-checkin successful build verification. This build is triggered on creating a PR, not at the time of merge.

.travis.yml
cruizen: The product target takes about 30 minutes to run on Travis's (free, with shared infrastructure for machines/ VMs) environment. Precheckin even escluding store tests will run for several hours (more than 4 hours without parallel tests), although we could try to optimise that by caching etc.  

I get that the point of PR build is to run tests though. Shall we identify a smaller smoke tests suite for that purpose? 

Branch builds can be scheduled with chron jobs to run less frequently. We can run the full test suites and build the product there.

.travis.yml
cruizen: Pull Request builds do not trigger email notifications in Travis. https://docs.travis-ci.com/user/notifications/#Configuring-email-notifications 

Emails are sent when, on the given branch: 
 - a build was just broken or still is broken.
 - a previously broken build was just fixed.
It is possible to change this frequency and the default conditions that trigger the notification.

The process email ID should work since it's already configured in GitHub. 

notifications:
  email:
    recipients: 
      - process@snappydata.io

cruizen:
>> where does the configuration specify how to update the git checkout? ...
It looks like a full git clone is performed (with a default depth of 50) since the Travis build environment is ephemeral (I think the submodule sync is redundant here unlike in git pull on an existing clone). Submodules are updated by default using ' git submodule update --init --recursive'.

Looking at the build log though, it takes about 5 minutes to clone the source. Most of the build time is spent in the assemble target - need to find out how much of that is downloading things versus compiling source, tests. This build went through some retries apparently and succeeded on 3rd or 4th attempt.

cruizen:
Build jobs time out in 50 minutes in Travis, causing a build to fail. The precheckin target fails therefore. Will need to break it down into smaller jobs.

.travis.yml
cruizen: Build is triggered on Travis site using GitHub notifications and cannot be triggered from the command line. Travis supports environment variables though. The Travis command line client can trigger a GitHub repo sync.

cruizen:
Closing this since the free Travis CI is inadequate to build our full targets (times out on the small VM). Only the product target can be built. However, that doesn't serve the purpose. We can revisit other options in future.

PR #1041, opened 2018-05-25
Added test to reproduce bug SNAP-2332. Probable fix by removing param…

ahshahid:
… literals from the aggregate functions in prepare phase for prepared statements
In case of Prepared Statements, in the prepare phase, the foldable functions param literals are not converted into token literals which causes problem.
I think that in case of predefined aggregate functions any constant parameters would be for internal usage of the function ( in most cases)  & hence two queries differing only on the basis of the constant params may be considered separate queries in terms of cached plan and hence the conversion of param literals to Token Literals be done in prepare phase itself.. 
(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
@@ -16,94 +16,24 @@
  */
 package org.apache.spark.sql.store
 
-import com.pivotal.gemfirexd.Attribute
-import com.pivotal.gemfirexd.security.{LdapTestServer, SecurityTestUtils}
-import io.snappydata.{Constant, SnappyFunSuite}
-import org.junit.Assert.assertEquals
+import java.sql.DriverManager
+
+import com.pivotal.gemfirexd.TestUtil
+import io.snappydata.SnappyFunSuite
 import org.scalatest.BeforeAndAfterAll
 
-import org.apache.spark.SparkConf
-
 class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
-  private val sysUser = "gemfire10"
 
   override def beforeAll(): Unit = {
-    this.stopAll()
-  }
-
-  protected override def newSparkConf(addOn: (SparkConf) => SparkConf): SparkConf = {
-    val ldapProperties = SecurityTestUtils.startLdapServerAndGetBootProperties(0, 0, sysUser,
-      getClass.getResource("/auth.ldif").getPath)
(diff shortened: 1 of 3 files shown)


cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
```diff
@@ -16,312 +16,321 @@
...
+        "first(C.field45) as field45," +
+        " first(C.subfield38) as subfield38, " +
+        "first(C.field40) as field40" +
+        " FROM tab1 A  LEFT JOIN tab7 B " +
+        " ON A.field20 = B.field20 AND " +
+        " B.globalfield16 = A.localfield16 LEFT JOIN field29_hier C ON " +
+        " A.field3 = field42 WHERE A.localfield16 ='0L' " +
+        " GROUP BY A.field14, A.field19, " +
+        "A.field9, A.field20, A.field11, A.field3, A.localfield16 " +
+        "having ( SUM(A.field13 ) > 0.001F or SUM(A.field13 ) < -0.001F) );"
```
sumwale: The table and view looks the same as in SNAP-2342 test above. Can you move the two into a common method instead of duplicating?

sumwale:
This is assuming that precheckin is clean after removing the "isPreparePhase" check. @vivekwiz, @kneeraj can you tell the reason why FOLDABLE_FUNCTIONS was being skipped for prepare phase?

vivekwiz:
I think that was temporarily done and to address the issue SNAP-1634 was
left open.

Regards,
Vivek

sumwale:
> Pls review the change . Disallowing parameterization of spark functions. throwing ParseException in that case.

Is this for all functions or only for those parameters of functions that don't expect parameterized literals otherwise ("foldable" functions as listed in Constant.FOLDABLE_FUNCTIONS)?

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -833,12 +832,17 @@ class SnappyParser(session: SnappySession)
               (args(0) == -10) || (args(0) == -1 && (index & 0x1) == 1) ||
               // all even args
               (args(0) == -2 && (index & 0x1) == 0) =>
+            l match {
+              case pl: ParamLiteral  if pl.tokenized => throw new ParseException(
+                s"function $fnName cannot have parameterized arguments")
```
sumwale: I think this should be "if pl.tokenized && isPreparePhase". Also the error message can include the specific parameter number: "Function $fnName cannot have parameterized argument ${index + 1}"

ahshahid: ok

PR #1044, opened 2018-06-01
[SNAP-2348] Invalid decompress call on stats row.

rishitesh:
Even if delta stats rows are not compressed ColumnFormatValue considered it as a compressed value and tries to decompress.

Stats rows are written as UnsafeRow. Unsafe row irrespective of nullability keeps nullbits.
So if 100 columns are  there stats row will contain 100 * 3 and UnsafeRow will contain
300 nullbits plus bits required for 8 bytes word allignment. Hence setting unused columns
as null does not really saves much.

These nullbits are set based on platform endianness. SD ColumnFormatValue
assumes LITTLE_ENDIAN bytes. However, Unsafe row itself does not force any endianness
and picks up from the platform. Hence a lower byte of the bit set  can be represented as
[11111110]. The 0th bit is for batch count which can never be null.
This causes SD ColumnFormatValue to understand stats row as a compressed byte
array( as first int is 1, hence -1 which after taking a negative
equals to 1 i.e LZ4 compression codec id ).
Hence setting each 3rd bit( null count stats) with not null flag. This will never cause
the word to be read as negative number.

## Patch testing

pre-checkin pending.

## ReleaseNotes.txt changes

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnUpdateExec.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnUpdateExec.scala
@@ -264,7 +264,7 @@ case class ColumnUpdateExec(child: SparkPlan, columnTable: String,
           val nullcountExpr = ExprCode(s"final int $nullCount = -1;", "false", nullCount)
           // write null for unchanged columns (by this update)
           (ColumnStatsSchema(field.name, field.dataType,
-            nullCountNullable = false).schema, allNullsExprs :+ nullcountExpr)
+            nullCountNullable = true).schema, allNullsExprs :+ nullcountExpr)
         case u => ColumnWriter.genCodeColumnStats(ctx, field,
           s"$deltaEncoders[$u].getRealEncoder()", nullCountNullable = true)
       }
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnUpdateExec.scala
```diff
@@ -241,19 +241,30 @@ case class ColumnUpdateExec(child: SparkPlan, columnTable: String,
...
+    // assumes LITTLE_ENDIAN bytes.
     val allNullsExprs = Seq(ExprCode("", "true", ""),
-      ExprCode("", "true", ""), ExprCode("", "true", ""))
+      ExprCode("", "true", ""))
     val (statsSchema, stats) = tableSchema.indices.map { i =>
       val field = tableSchema(i)
       tableToUpdateIndex.get(i) match {
         case null =>
+          val nullCount = ctx.freshName("nullCount")
+          val nullcountExpr = ExprCode(s"final int $nullCount = -1;", "false", nullCount)
```
sumwale: Do you need a new variable here? Shouldn't directly embedded value work: ExprCode("", "false", "-1")?

rishitesh: Yes, that will also work. Will change it.

sumwale:
@rishitesh Is it possible to add some workaround for previous format as on current master? Perhaps check for the size returned should be > 0 and < 2*maxBatchSize, and then decompress should not throw exception. If its easy enough then try to handle else ignore.

rishitesh:
@sumwale Any particular reason why you want the workaround? After this fix is merged do you foresee any such issue?

sumwale:
For some users who are running on previous master. Ignore if it looks a bit more work.

PR #1045, opened 2018-06-01
[SNAP-2349][SNAP-2364] scope synchronized blocks in ColumnFormatValue; skip batch if stats row is missing from snapshot

sumwale:

- deadlock reported in SNAP-2349 is fixed by reducing the scope of synchronized blocks
  in ColumnFormatValue to only when reading/writing fields:
  - move transformValueRetain outside of sync block from all calling points and instead
    set a boolean to indicate that the call has to be made
  - decompress/compress methods invoked by transformValueRetain now have a sync block
    at the start when reading/writing the values; the code that actually does
    compression/decompression is outside any sync and there is just another sync block
    if the underlying value is to be replaced in the region
  - made compressionCodecId as volatile so just reads on it can be done without any
    synchronization (like in copy method)

What this means is that two concurrent threads can possibly compress/decompress twice
unnecessarily and replace the region value. This should not happen for off-heap storage
due to the refCount <= 2 check but can happen for on-heap storage.

- skip batch if the stats row is missing; this is already handled for in-memory batches
  and the same has been added for on-disk batches in this PR
- also changed DiskMultiColumnBatch to set entry in the constructor itself instead of later

## Patch testing

precheckin, new tests to be added shortly

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatEntry.scala
```diff
@@ -430,61 +438,85 @@ class ColumnFormatValue extends SerializedDiskBuffer
...
 
   private def decompressValue(buffer: ByteBuffer, onlyIfStored: Boolean): ColumnFormatValue = {
-    if (this.decompressionState != 0) {
+    var position = 0
+    var typeId = 0
+    var context: RegionEntryContext = null
+    var isDirect = false
+    var doReplace = false
+    var updateStats = false
+    val doDecompress = synchronized(if (this.decompressionState != 0) {
```
rishitesh: The bracket after synchronized looks odd. Looks like its synchronizing on the return value of the expression. Is that the intention?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatEntry.scala
```diff
@@ -430,61 +438,85 @@ class ColumnFormatValue extends SerializedDiskBuffer
...
             if (!StoreCallbacksImpl.acquireStorageMemory(context.getFullPath,
               numBytes, buffer = null, offHeap = false, shouldEvict = true)) {
               throw LocalRegion.lowMemoryException(null, numBytes)
             }
           }
           val newBuffer = transferToStorage(decompressed, allocator)
-          // update the statistics before changing self
-          if (updateStats) {
-            context.updateMemoryStats(this, newValue)
+          val releaseStorageMemory = synchronized {
```
rishitesh: Can transfertoStorage be also called for cases where we are actually replacing the buffer.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatEntry.scala
sumwale: No, synchronized is always on the object ("this" in this case) e.g. obj.synchronized.

A bit odd usage to minimize indentation changes.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatEntry.scala
sumwale: transferToStorage is primarily for off-heap buffers that hold a Runnable inside which will release the storage in UMM and the buffer itself (either by GC or by explicit UnsafeHolder.releaseDirectBuffer). This Runnable inside the Cleaner in DirectByteBuffer will be changed to "FreeStoreBuffer".

The other change is just transforming buffer type if required, as per storage type: e.g. heap to off-heap if storage type is off-heap but incoming is heap. This case should normally not happen in our code paths.

So this call is for changing an existing buffer which will then be stuffed into ColumnFormatValue. For example, compression will use temporary memory (ideally execution memory but was not being done due to LME issues) which will then be transferred to storage accounting if compression has reduced the size enough else it is released right away.

rishitesh: My question was a little different. I was not very clear. We are using "newBuffer" value only if certain conditions are met (like no other thread has changed "this.columnBuffer").  So why call transferToStorage if we ar enot going to use newBuffer i.e in the else part in line 512.

sumwale: That was done to keep transferToStorage outside the sync block because that can also call UMM. I guess this should also eagerly release newBuffer in case it was not stored in region so will add that.

PR #1051, opened 2018-06-12
External relation limit

ahshahid:
As per discussions with Sudhir & implementing his suggestion ( which I agree is a good way ), to avoid the snappydata cluster to become inoperable,  any query which fetches all the data from the external relation ( right now specifically done for GemFireRelation only)  will have an implicit limit applied to it. This application of limit is a feature of external table & can be switched off, default is on). However queries which contain aggregate function , group by,  specific projection, where clause or is part of create table as select will not have the limit criteria applied.

The logic used is that in case a query fullfills the above criteria & is a candidate of limit, a dummy condition IsNotNull on the first column of table , is introduced. That this is a fictitious condition which is ensured that it will be pushed down to the relation level and is created using TaggedAttribute with GemFireLimitTag which marks it as a special condition. Once the analysis phase is over, the name of the attribute in the condition is changed to start with GemFireLimitTag . Thus this Filter is identified in the GemFireRelation & limit applied.

Also pls note that, TaggedAttribute is used throughout & not converted into AttributeReference because down the line in spark there is code which replaces the AttributeReference with Column's Attribute  if is AttributeReference with expcr Id same as Attribute's exprid. If that happens there is no way of knowing in GemFireRelation if it is a true condition or fictitious condition.

associated PRs
[connector changes](https://github.com/SnappyDataInc/snappy-connectors/pull/20)
[aqp changes](https://github.com/SnappyDataInc/snappy-aqp/pull/167)
## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

I will see if I can clean the case of CreateTableAsSelect further. Pls let me know if you have any suggestion for it.


ahshahid:
@sumwale @kneeraj  @rishitesh @suranjan @ashetkar .
please review this . I am running a combined precheckin of SNAP-2232 & this PR

PR #1052, opened 2018-06-13
SNAP-2341: Table shown in dashboard even after create table is killed

snappy-sachin:

- Adding check for existing tables look up in catalog before returning last updated snapshots in case tables stats results in empty. Reset table stats map if no table is present in catalog.

## Patch testing

 - Manually

## ReleaseNotes.txt changes

 - NA

## Other PRs 

 - None

### DIFF
--- a/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
+++ b/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
@@ -193,7 +193,10 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
       }
     }
     catch {
-      case NonFatal(e) => log.warn(e.getMessage, e)
+      case NonFatal(e) => {
+        log.warn("Exception occurred while collecting Table Statistics: "
+            + e.getMessage, e)
+      }
     }
 
     try {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -207,11 +210,28 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
...
+        allTablesList ++= allStoreTables.get(schema).asScala.map(tbl => {
+          schema.concat(".").concat(tbl)
+        })
+      })
+
+      if(allTablesList.size == 0 && tableSizeInfo.size > 0){
+        // No table is present in any db schema, reset tableSizeInfo
+        tableSizeInfo = Map.empty[String, SnappyRegionStats]
+      }
+
```
sumwale: Please fix the formatting. In fact in most of your checkins. There should be space between "if" and opening parenthesis. "if()". Please use "Project" code settings with codeStyleSettings.xml copied to .idea as noted in build docs. IDEA >= 16 must be showing this in red since this violates code style guidelines.

snappy-sachin: Code formatted as suggested.

I have followed snappydata documentation while Setting up IntelliJ IDEA and also have added codeStyleSettings.xml into .idea directory. But I never get any code styling error automatically. If I reformat code then it uses the specified code style codeStyleSettings.xml. Therefore I always do it manually before committing the code.

core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -207,19 +210,36 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
...
+      // Return last successfully updated tableSizeInfo
+      // Check for if any table is present in catalog
+      val allStoreTables = Misc.getMemStore.getExternalCatalog.getAllStoreTablesInCatalog(true)
+      val allTablesList = scala.collection.mutable.Buffer.empty[String]
+      allStoreTables.asScala.keys.foreach(schema => {
+        allTablesList ++= allStoreTables.get(schema).asScala.map(tbl => {
+          schema.concat(".").concat(tbl)
+        })
+      })
+
```
sumwale: Why store in a list when you only need to know whether size is 0 or not? Also shouldn't the externalTables and indexStats also be cleared.

This size==0 does not look to be proper. I think you should simply use a flag to indicate the case when result was successfully acquired:

```scala
      if (dataServers != null && dataServers.size() > 0) {
        result = FunctionService.onMembers(dataServers)
            // .withCollector(new GfxdListResultCollector())
            .execute(SnappyRegionStatsCollectorFunction.ID).getResult(5, TimeUnit.SECONDS).
            asInstanceOf[java.util.ArrayList[SnappyRegionStatsCollectorResult]]
            .asScala
        resultObtained = true
      }
```

And use that flag instead of size check. Then there is no need to go back to catalog again so this new code can be removed.

snappy-sachin: Feedback is incorporated in the code.

PR #1053, opened 2018-06-13
[SNAP-2186] Fixed offheap size for PR

rishitesh:
a) While determining the off heap size of a PR the iterator only considered primary buckets.
Changed it to iterate over secondary buckets as well. 
b) Also added a utility method in DataGenerator to generate unique values based on a range.
Some formatting changes.
## Patch testing
Manual testing. Pre checkin pending. 

## ReleaseNotes.txt changes
NA
## Other PRs 
NA

### DIFF
--- a/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
+++ b/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
@@ -142,9 +142,9 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
         val id = memMap.get("id").toString
 
         var memberStats: MemberStatistics = {
-          if(dssUUID != null && membersInfo.contains(dssUUID.toString)) {
+          if (dssUUID != null && membersInfo.contains(dssUUID.toString)) {
             membersInfo(dssUUID.toString)
-          } else if(membersInfo.contains(id)) {
+          } else if (membersInfo.contains(id)) {
             membersInfo(id)
           } else {
             null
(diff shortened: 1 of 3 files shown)


core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -257,18 +257,20 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
...
               val key = re.getRawKey.asInstanceOf[ColumnFormatKey]
-              if (numColumnsInTable < 0) {
-                numColumnsInTable = key.getNumColumnsInTable(table)
+              if (itr.getHostedBucketRegion.getBucketAdvisor.isPrimary) {
+                if (numColumnsInTable < 0) {
+                  numColumnsInTable = key.getNumColumnsInTable(table)
+                }
+                rowsInColumnBatch += key.getColumnBatchRowCount(itr, re,
+                  numColumnsInTable)
               }
```
sumwale: Bit of perf optimization possible. ColumnFormatKey.getColumnBatchRowCount also does itr.getHostedBucketRegion. In fact that's the only reason it uses itr. Can you change getColumnBatchRowCount to be passed BucketRegion instead of PREntriesIterator and then get the BucketRegion only once from iterator? Will be cleaner API for ColumnBatchKey interface too.

rishitesh: Sure. Will change it.

PR #1054, opened 2018-06-18
Snap 2358 Sorted Column Batches on partitioning keys

vibhaska:
Now user can create sorted Column Batches on partitioning keys using DDL mentioned below. This will keep a column batch in sorted manner that can be leveraged for better performance of point queries, range queries and colocated join queries on partitioning columns. For more details please refer https://jira.snappydata.io/browse/SNAP-2358

TODO:
1. Would open Jira tickets for pending items or any suggestion from code review.
2. Taking care of one known debugging issues. Also take care of similar bugs.

## Patch testing
Unit test
Precheckin

## ReleaseNotes.txt changes
A sample DDL to create table with sorted partitioning columns is,
    session.sql(s"create table $colTableName (id int, addr string, status boolean) " +
        s"using column options(buckets '$numBuckets', partition_by 'id SORTING ASC' " + s")")

If no sorting is required, above DDL would be,
    session.sql(s"create table $colTableName (id int, addr string, status boolean) " +
        s"using column options(buckets '$numBuckets', partition_by 'id' " + s")")

Valid sorting identifiers are,
SORTING ASC
SORTING DESC
SORTING Ascending
SORTING Descending

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/395


PradeepSurale:

PR #1056, opened 2018-06-19
[SNAP-2381] global lock to serialize concurrent puts

sumwale:

- global lock in putInto execution to serialize concurrent puts
- added lock to session context and clear in putInto plan's executeCollect

## Patch testing

new unit test to be added

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -27,8 +27,9 @@ import scala.concurrent.Future
 import scala.language.implicitConversions
 import scala.reflect.runtime.universe.{TypeTag, typeOf}
 import scala.util.control.NonFatal
+
 import com.gemstone.gemfire.internal.GemFireVersion
-import com.gemstone.gemfire.internal.cache.GemFireCacheImpl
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -396,11 +399,53 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
   private[sql] def getHashVar(ctx: CodegenContext,
       keyVars: Seq[String]): Option[String] = getContextObject(ctx, "H", keyVars)
 
-  private[sql] def clearContext(): Unit = synchronized {
-    getContextObject[LogicalPlan](SnappySession.CACHED_PUTINTO_UPDATE_PLAN).
-        foreach { cachedPlan =>
-          sharedState.cacheManager.uncacheQuery(this, cachedPlan, blocking = true)
+  private[sql] def cachePutInto(plan: Option[LogicalPlan], table: String): Boolean = {
```
rishitesh: I would not prefer keeping an operation specific logic in SnappySession. We should move it to some other class if possible.

PR #1058, opened 2018-06-25
[SNAP-2347] Fix for row tables getting dropped.

rishitesh:
There was a small window when the connection is lost to the JDBC source JDBCMutable relation might drop the table, thinking of a failed initialization. 

## Patch testing

pre-checkin

## ReleaseNotes.txt changes

NA

## Other PRs 
Na

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCAppendableRelation.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/JDBCAppendableRelation.scala
@@ -66,6 +66,7 @@ abstract case class JDBCAppendableRelation(
   override val needConversion: Boolean = false
 
   var tableExists: Boolean = _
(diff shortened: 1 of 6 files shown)


kneeraj:
@rishitesh essentially we are keeping a member variable as a flag. Is there a chance that when a cached relation is returned we may have problems? Is it possible to do the same thing passing around local variables.

rishitesh:
@kneeraj A cached relation will not call createRelation , hence the destroy in case of a failure will not get called.

kneeraj:
Ok. Not insisting but I think I would clear the field to false once 'createRelation' has succeeded anyways.
Rest looks fine to me.

rishitesh:
@kneeraj Ok I can add that.

rishitesh:
@kneeraj This variable is used only when a new relation is getting created, so we are safe here.

PR #1060, opened 2018-06-26
Change to let 'lead' with warnings

kneeraj:

if deployed jars/coordinates not present during restart do not stop lead from coming up. Start and spit out warnings in logs.

## Patch testing

Manual

## ReleaseNotes.txt changes

None

## Other PRs 

No

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
@@ -502,10 +502,10 @@ abstract class SnappyDDLParser(session: SparkSession)
         (REPOS ~ stringLiteral).? ~ (PATH ~ stringLiteral).? ~>
         ((alias: TableIdentifier, packages: String, repos: Any, path: Any) => DeployCommand(
           packages, alias.identifier, repos.asInstanceOf[Option[String]],
-          path.asInstanceOf[Option[String]]))) |
+          path.asInstanceOf[Option[String]], false))) |
       JAR ~ tableIdentifier ~ stringLiteral ~>
           ((alias: TableIdentifier, commaSepPaths: String) => DeployJarCommand(
-        alias.identifier, commaSepPaths))) |
+        alias.identifier, commaSepPaths, false))) |
     UNDEPLOY ~ tableIdentifier ~> ((alias: TableIdentifier) => UnDeployCommand(alias.identifier)) |
     LIST ~ (
       PACKAGES ~> (() => ListPackageJarsCommand(true)) |
@@ -761,34 +761,55 @@ case class DeployCommand(
     coordinates: String,
     alias: String,
     repos: Option[String],
-    jarCache: Option[String]) extends RunnableCommand {
+    jarCache: Option[String],
+    restart: Boolean) extends RunnableCommand {
 
   override def run(sparkSession: SparkSession): Seq[Row] = {
-    val jarsstr = SparkSubmitUtils.resolveMavenCoordinates(coordinates, repos, jarCache)
-    if (jarsstr.nonEmpty) {
-      val jars = jarsstr.split(",")
-      val sc = sparkSession.sparkContext
-      val uris = jars.map(j => sc.env.rpcEnv.fileServer.addFile(new File(j)))
-      SnappySession.addJarURIs(uris)
-      Utils.mapExecutors[Unit](sparkSession.sparkContext, () => {
-        ToolsCallbackInit.toolsCallback.addURIsToExecutorClassLoader(uris)
-        Iterator.empty
-      })
-      val deployCmd = s"$coordinates|${repos.getOrElse("")}|${jarCache.getOrElse("")}"
-      ToolsCallbackInit.toolsCallback.addURIs(alias, jars, deployCmd)
+    try {
+      val jarsstr = SparkSubmitUtils.resolveMavenCoordinates(coordinates, repos, jarCache)
+      if (jarsstr.nonEmpty) {
+        val jars = jarsstr.split(",")
+        val sc = sparkSession.sparkContext
+        val uris = jars.map(j => sc.env.rpcEnv.fileServer.addFile(new File(j)))
+        SnappySession.addJarURIs(uris)
+        Utils.mapExecutors[Unit](sparkSession.sparkContext, () => {
+          ToolsCallbackInit.toolsCallback.addURIsToExecutorClassLoader(uris)
+          Iterator.empty
+        })
+        val deployCmd = s"$coordinates|${repos.getOrElse("")}|${jarCache.getOrElse("")}"
+        ToolsCallbackInit.toolsCallback.addURIs(alias, jars, deployCmd)
+      }
+      Seq.empty[Row]
+    } catch {
+      case ex: Throwable => {
+        restart match {
+          case true => {
+            logWarning(s"Following mvn coordinate" +
+                s" could not be resolved during restart: ${coordinates}")
+            Seq.empty[Row]
+          }
+          case false => throw ex
+        }
+      }
     }
-    Seq.empty[Row]
   }
 }
 
 case class DeployJarCommand(
     alias: String,
-    paths: String) extends RunnableCommand {
+    paths: String,
+    restart: Boolean) extends RunnableCommand {
 
   override def run(sparkSession: SparkSession): Seq[Row] = {
     if (paths.nonEmpty) {
       val jars = paths.split(",")
       val sc = sparkSession.sparkContext
+      val availableUris = jars.filter(j => new File(j).exists())
+      val unavailableUris = jars.filterNot(j => new File(j).exists())
+      if (unavailableUris.nonEmpty && restart) {
+        logWarning(s"Following jars are unavailable" +
+            s" for deployment during restart: ${unavailableUris.deep.mkString(",")}")
+      }
       val uris = jars.map(j => sc.env.rpcEnv.fileServer.addFile(new File(j)))
       SnappySession.addJarURIs(uris)
       Utils.mapExecutors[Unit](sparkSession.sparkContext, () => {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -761,35 +761,56 @@ case class DeployCommand(
...
       val jars = paths.split(",")
       val sc = sparkSession.sparkContext
-      val uris = jars.map(j => sc.env.rpcEnv.fileServer.addFile(new File(j)))
+      val availableUris = jars.filter(j => new File(j).exists())
+      val unavailableUris = jars.filterNot(j => new File(j).exists())
+      if (unavailableUris.nonEmpty && restart) {
+        logWarning(s"Following jars are unavailable" +
+            s" for deployment during restart: ${unavailableUris.deep.mkString(",")}")
+      }
+      val uris = availableUris.map(j => sc.env.rpcEnv.fileServer.addFile(new File(j)))
```
ashetkar: Can there be any exception other than path-not-found while deploying jars? If yes, we should log them too and proceed rather than letting them fail lead start-up (Similar to try-catch block in DeployCommand.run()).

kneeraj: Since at this time i.e. deploy time apart from file existence nothing else of the jar will be used so even if the jar file itself is ill-formatted or whatever it would not be known as this will be just added in the list of urls of the class loader. So no other exception expected here.

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -761,35 +761,56 @@ case class DeployCommand(
...
+        Utils.mapExecutors[Unit](sparkSession.sparkContext, () => {
+          ToolsCallbackInit.toolsCallback.addURIsToExecutorClassLoader(uris)
+          Iterator.empty
+        })
+        val deployCmd = s"$coordinates|${repos.getOrElse("")}|${jarCache.getOrElse("")}"
+        ToolsCallbackInit.toolsCallback.addURIs(alias, jars, deployCmd)
+      }
+      Seq.empty[Row]
+    } catch {
+      case ex: Throwable => {
```
sumwale: A catch of all Throwables should check for SystemFailure and cancellation. Like:

```scala
  case ex: Throwable =>
    if (ex.isInstanceOf[Error] && SystemFailure.isJVMFailureError(e.asInstanceOf[Error])) {
      SystemFailure.initiateFailure(ex)
      // If this ever returns, rethrow the error. We're poisoned
      // now, so don't let this thread continue.
      throw ex
    }
    Misc.checkIfCacheClosing(ex)
    ...    
```

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -761,35 +761,56 @@ case class DeployCommand(
...
+    paths: String,
+    restart: Boolean) extends RunnableCommand {
 
   override def run(sparkSession: SparkSession): Seq[Row] = {
     if (paths.nonEmpty) {
       val jars = paths.split(",")
       val sc = sparkSession.sparkContext
-      val uris = jars.map(j => sc.env.rpcEnv.fileServer.addFile(new File(j)))
+      val availableUris = jars.filter(j => new File(j).exists())
+      val unavailableUris = jars.filterNot(j => new File(j).exists())
```
sumwale: Use partition rather than filter+filterNot. Also use java nio APIs that throw proper IOExceptions: `val (availableUris, unavailableUris) = jars.partition(f => Files.isReadable(Paths.get(f)))`

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -761,35 +761,56 @@ case class DeployCommand(
...
+    } catch {
+      case ex: Throwable => {
+        restart match {
+          case true => {
+            logWarning(s"Following mvn coordinate" +
+                s" could not be resolved during restart: ${coordinates}")
+            Seq.empty[Row]
+          }
+          case false => throw ex
+        }
```
sumwale: Matching for boolean is a bit of overkill. Just if (restart) { ... } else throw ex is better (Idea will also show a warning for the same).

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
kneeraj: Ok

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
kneeraj: ok.

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
kneeraj: ok.

kneeraj:
>> Other than that, not sure if the implications of this are desirable. This means that cluster may come up successfully but subsequent accesses will fail at run time. Is there some scenario where we expect that jars will no longer be available after restart? I think a more realistic scenario is that jar path may have changed after restart. So what can be provided is a configuration option to remap paths (as a special case remap to empty would mean ignore), but by default this should continue to fail on restart. I think needs to be discussed with team.

It was originally designed to fail at restart but perhaps became too restrictive in a cloud like environment. The request for change came from @jramnara . His pods were changed and there the mapping to persistent volumes was not there for the new one and it started failing. Or the jars location was not on a persistent volume. So for him things started to fail and he could not even bring up tyhe cluster.
According to Jags:
------------------------------------------------------
"I can see from teh start_leader.log that one of my deployed JARs is not found and the leader craps out. This is bad."
"Unless, the jars are completely under our management (cannot be some directory the user can specify in the command) the Lead should always come up. Log errors if appropriate. 
So, the lead not coming up is a bigger problem. " 
-----------------------------------------------------

Remapping was one thing which I thought but generally any additional start up parameter has its own set of problems. Like they tend to stay permanently with the scripts. Issuing a fresh deploy should be equally easy if not more. The old deploy entry will just be overwritten.
Yes we can discuss this again. Jags, Amogh and I had discussed this in a thread a couple of days back.

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
kneeraj: Also in the core classes ( parser here ) we should not use gemfire classes. Right? So you mean dhave a method in ToolsCallbackInitImpl which perhaps can be used everywhere.

sumwale: GemFire classes can be used just fine and lots of code in core does. But if code depends on GemFireCache then it cannot be used directly because it will fail in smart connector (if that piece of code is executed in smart connector too). Here there is no such problem.

sumwale: Agree about checkIfCacheClosing. Not needed.

sumwale:
@kneeraj @jramnara 

Don't quite understand this. Now this is saying that just bring up the lead even if potentially required jars are missing and no remap-like parameter is added. How is cluster supposed to be operational after this? What about genuine deployment error by user where jars have been removed and then whole cluster becomes unusable? Admin is supposed to redeploy always, but how will user even know what all to redeploy? In an actual scenario there can be tons of such jars. And what happens in the interim period when ops come in?

For specific scenarios like k8s nodes moving around and jars no longer available, a remap-like parameter should be added as mentioned. In fact the remap parameter can automatically do the redeploy at very start if that is the better way. That is the safe way to bring up the cluster.

Other than that don't understand how jar deployment is supposed to work with k8s if no shared volume is available. In many cases the jars have to be available on both the leads as well as servers (for deserialization, reference to those classes inside RDD.compute etc). I will suggest the proper fix to this is to allow for hadoop URLs for jar deployment, if not allowed already, then users can provide s3 or similar shared URLs easily. This is how any external data sources are supposed to be referred in any case (which these jars will likely refer to), so allow the same for jars. That is the safe and clean way to fix the problem with k8s or cloud deployments in general. If this does not sound good, then needs further discussion. IMO this should not be merged as it is and will suggest closing this PR for now.

sumwale:
@kneeraj @jramnara 

As an alternative to remap, we could provide a system property to skip the deploy on error like in this PR. And use the "-initial-scripts" argument to provide a SQL script to redeploy all the required jars. It should be noted with warnings that user has to take care that all the missing jars are listed in "-initial-scripts". This should be only used as a temporary hack in some extreme scenarios (maybe for user debugging) and should not be a permanent setting.

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
kneeraj: Ok. Thanks. Got confused about gemfire classes.

kneeraj:
@swale not having the jar on a shared volume was more of a mistake from @jramnara and was not intended. In my opinion the docs should clearly call out the best practices for jar management as far as snappy is concerned.
 
Also we already have ddls to list jars and coordinates. When the user does
> list jars;
it outputs the following columns:
alias          |coordinate     |isPackage

we can add a fourth column like "status" or "deployed" ( can think of a better name ) and this column can be filled as per whether the jar was a success deployment or not. This seems to be useful.

>> In fact we already have such a property in store "gemfirexd.datadictionary.allow-startup-errors" (Property.DDLREPLAY_ALLOW_RESTART_WITH_ERRORS) which can also be honored here.

Regarding providing anything at the startup time or additional scripts etc. I think we have seen the fallout of this in one of customer's deployment whether a temporary startup argument ended up being in the customer's scripts for ever. So I would refrain from doing anything special at any restart. This looks to be a hassle for the administrators. Only permanent changes should be there in the scripts in my opinion.

If some job starts failing then using the jar/coordinates related utilities to figure out valid/invalid jars might be better and then redeploying for ones which were failures.

PR #1064, opened 2018-06-27
Fixes for SNAP-2382: Column table mysteriously changes to Row table on dashboard after LME in data serever

snappy-sachin:

- Setting correct Table Type after collecting all table stats in lead node.
   This fixes problem of converting table type from Column to Row when there is column store drop due to some reasons.

## Patch testing

 - Tested with hack in code.

## ReleaseNotes.txt changes

 - None

## Other PRs 

 - None

### DIFF
--- a/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
+++ b/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
@@ -222,11 +222,13 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
     if (resultObtained) {
       // Return updated tableSizeInfo
       val regionStats = result.flatMap(_.getRegionStats.asScala).map(rs => {
-        hiveTables.map(ht => {
+        hiveTables.foreach(ht => {
           val tableName = rs.getTableName
+          val schema = tableName.substring(0, tableName.indexOf("."));
           val tname = tableName.substring(tableName.indexOf(".") + 1);
           // Set whether table is column table or not
-          if (ht.entityName.equalsIgnoreCase(tname)) {
+          if (ht.schema.toString.equalsIgnoreCase(schema)
+              && ht.entityName.equalsIgnoreCase(tname)) {
             if (ht.tableType.equalsIgnoreCase("COLUMN"))
               rs.setColumnTable(true)
             else
--- a/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
+++ b/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
@@ -23,6 +23,7 @@ import java.util.concurrent.{ConcurrentHashMap, TimeUnit}
 import java.util.function.BiFunction
 
 import scala.collection.JavaConverters._
+import scala.collection.mutable
 import scala.language.implicitConversions
 import scala.util.control.NonFatal
 
@@ -181,7 +182,6 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
   override def getStatsFromAllServers(sc: Option[SparkContext] = None): (Seq[SnappyRegionStats],
       Seq[SnappyIndexStats], Seq[SnappyExternalTableStats]) = {
     var result = new java.util.ArrayList[SnappyRegionStatsCollectorResult]().asScala
-    var externalTables = scala.collection.mutable.Buffer.empty[SnappyExternalTableStats]
     val dataServers = GfxdMessage.getAllDataStores
     var resultObtained: Boolean = false
     try {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -196,28 +196,44 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
...
+    if (result.flatMap(_.getRegionStats.asScala).size == 0) {
       // Return last updated tableSizeInfo
       (tableSizeInfo.values.toSeq,
        result.flatMap(_.getIndexStats.asScala),
        externalTables)
     } else {
       // Return updated tableSizeInfo
-      (result.flatMap(_.getRegionStats.asScala),
+      val regionStats = result.flatMap(_.getRegionStats.asScala).map(rs => {
+        hiveTables.map(ht => {
```
rishitesh: We can convert hiveTables to a set. Then it will become easier to lookup the current region name in the set. It will avoid iterating the whole list for each table

core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -196,28 +196,44 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
...
 
+    val hiveTables = Misc.getMemStore.getExternalCatalog.getHiveTables(true).asScala
+
     try {
       // External Tables
-      val hiveTables: java.util.List[ExternalTableMetaData] =
-        Misc.getMemStore.getExternalCatalog.getHiveTables(true)
-      externalTables = hiveTables.asScala.collect {
-        case table if table.tableType.equalsIgnoreCase("EXTERNAL") =>
+      externalTables = hiveTables.collect {
```
rishitesh: externalTables need not be a var. Can be a val if we are not changing it

core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -196,28 +196,44 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
       case NonFatal(e) => log.warn(e.getMessage, e)
     }
 
+    val hiveTables = Misc.getMemStore.getExternalCatalog.getHiveTables(true).asScala
```
rishitesh: Should we wait for hive catalog initialization? What if hive is not initialized in lead yet?

rishitesh: Probably not, as Hive catalog would have got initialized while creating SnappyContext.

core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
kneeraj: And since we are being too careful here :-) then please change 'hiveTables.map' to 'hiveTables.foreach'
BTW don't we support same column table under different schema? If yes, then won't this
tableName.substring(tableName.indexOf(".") + 1);
mean we are just taking care of table name and not fully qualified names.

snappy-sachin: Adding check on table schema while determining table type.
Changed hiveTables.map to hiveTables.foreach.

snappy-sachin: Also for easier look up, creating and using HashMap built from hiveTables, this will avoid iterating over hiveTables for each and every Table.

core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
snappy-sachin: Done. Also externalTables initialization moved inside block.

snappy-sachin:
All feedback comments are incorporated.
Hence merging.

core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -201,26 +201,49 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
...
       }
     }
 
     if (resultObtained) {
+      // Return updated tableSizeInfo
+      // Map to hold hive table type against table names as keys
+      val tableTypesMap: mutable.HashMap[String, String] = mutable.HashMap.empty[String, String]
+      hiveTables.foreach(ht => {
+        val key = ht.schema.toString.concat("." + ht.entityName)
+        tableTypesMap.put(key.toUpperCase, ht.tableType)
```
sumwale: @snappy-sachin This is an inconsistency in ExternalTableMetaData that is sometimes holding the StructType object in schema field with full name in entityName (for getHiveTableMetaData) while for getHiveTables it is returning the schema name of table with just the name in entityName. Please file a separate issue to correct this inconsistency.

core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -201,26 +201,49 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
...
+      })
+
+      val regionStats = result.flatMap(_.getRegionStats.asScala).map(rs => {
+        val tableName = rs.getTableName
+        if (tableTypesMap.contains(tableName.toUpperCase)
+            && tableTypesMap.get(tableName.toUpperCase).get.equalsIgnoreCase("COLUMN")) {
+          rs.setColumnTable(true)
+        } else {
+          rs.setColumnTable(false)
+        }
```
sumwale: Double lookup for contains then get is not good. Cleaner scala way:

```scala
val isColumnTable = tableTypesMap.get(tableName.toUpperCase) match {
  case Some(t) if "COLUMN".equalsIgnoreCase(t) => true
  case _ => false
}
rs.setColumnTable(isColumnTable)
```

PR #1066, opened 2018-07-03
- Fix/modify scripts for performance testing

swatisawant:

- Added configuration template files for stock spark and snappy for 1 GB and 100 GB data sizes
- Added template files to be used while running the benchmark tests on stock spark
- Fixed few issues with stock spark and snappy scripts in order to generate TPCH perf numbers with 1 GB and 100 GB data
    The fixes include:
     - Adding missing parameters for stock spark table creation and query execution scripts
     -  Change in stock spark script for starting spark master using the hostname instead of localhost/ipAddress 
     - Change in the checkout location for stock spark scripts. Original scripts were using snappy spark instead of stock spark. 
     - Environment related fixes in scripts
     - Change in snappy cluster start script to use snappy-start-all script instead of individual scripts for starting the members

## Patch testing

Generated perf numbers for stock sprak 2.1.1, 2.3.1 and SnappyData latest master using modified scripts.

## ReleaseNotes.txt changes

NA

## Other PRs 

No


tests/benchmark/snappy/dynamic/PerfRun.conf.template1
```diff
@@ -0,0 +1,109 @@
...
+#sparkProperties="-J-Xmx2g -spark.network.timeout=300s -spark.driver.maxResultSize=2g -spark.shuffle.sort.bypassMergeThreshold=28"
+sparkProperties="-heap-size=10g -memory-size=40g $commonProperties"
+
+#Spark Sql properties are specified while executing query
+#spark.sql.shuffle.partitions=${shufflePartitions},spark.sql.inMemoryColumnarStorage.compressed=${inMemoryColumnarStorageCompressed}
+#spark.sql.inMemoryColumnarStorage.compressed=false
+sparkSqlProperties="\"\""
+
+#location of checkout
+#SnappyData=/data1/users/perfuser/snappyBuilds/1.0.0/snappydata-1.0.0-bin
```
cruizen: Can we replace /data1/users/perfuser/ with an environment variable $DATA_DIR ?

```
if [ -z "$DATA_DIR" ] 
then
	echo "Please set \$DATA_DIR"
else
	DATA_DIR=~; echo "Setting up base directory at $DATA_DIR"
```

tests/benchmark/spark/3_GenerateResult.sh
```diff
@@ -10,7 +10,7 @@ latestProp=$directory/latestProp.props
 
 cd $SnappyData/snappy-spark
 echo snappy-spark = $(git rev-parse HEAD)_$(git log -1 --format=%cd) >> $latestProp
-cd ../build-artifacts/scala-2.10/snappy/benchmark/snappy/
```
cruizen: 2.11 now

swatisawant: I have commented out this command from script as its not required in case of spark. Will remove it from the file.

tests/benchmark/snappy/dynamic/PerfRun.conf.template1
swatisawant: I think, setting up the path variables in conf would be better here instead of environment variable. For each PerfRun, the conf template has to be modified into PerfRun.conf file according to the test run requirements.
I will just remove the hard coded paths and specify the description for same in template. Users will have to create their own copy of PerfRun.conf for running the test.

PR #1071, opened 2018-07-06
Snap 2398

kneeraj:

Smart Connector can use deployed jars now as long as path is visible on the driver node of the spark cluster.
Added a flag to fail if required when deployed jars are for some reason not available.
By default it just logs a warning.

## Patch testing

Manual testing

## ReleaseNotes.txt changes

Documentation of deploy jar should cover this.

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/403

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SmartConnectorHelper.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SmartConnectorHelper.scala
@@ -17,7 +17,7 @@
 package org.apache.spark.sql
 
 import java.io._
-import java.net.URI
+import java.net.{URI, URL}
 import java.nio.file.{Files, Paths}
 import java.sql.{CallableStatement, Connection, SQLException}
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/sql/SmartConnectorHelper.scala
```diff
@@ -74,6 +80,10 @@ class SmartConnectorHelper(snappySession: SnappySession) extends Logging {
     createUDFStmt = conn.prepareCall(createUDFString)
     dropUDFStmt = conn.prepareCall(dropUDFString)
     alterTableStmt = conn.prepareCall(alterTableStmtString)
+    getJarsStmt = conn.prepareCall(getJarsStmtString)
+    if (sc != null) {
+      executeGetJarsStmt(sc)
```
rishitesh: IMO it should be in a try catch. Log a warning if something goes bad in terms of deployed jars. Failing it seems a littlel too restrictive.

kneeraj: If you see presence of jar is not mandatory. For that it logs only a warning. But failure will indicate connection level problem which can affect smart connection otherwise so I don't think explicit handling is necessary.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
```diff
@@ -621,6 +621,9 @@ object StoreCallbacksImpl extends StoreCallbacks with Logging with Serializable
   override def clearConnectionPools(): Unit = {
     ConnectionPool.clear()
   }
+
+  override def getLeadClassLoader() : URLClassLoader =
```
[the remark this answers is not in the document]
kneeraj: Ok. Will refactor in the next change.

PR #1079, opened 2018-07-11
Added a timeout for hive queries wait so that we can fail fast

kneeraj:
This causes delay in shutdown etc. also even if hive query is failing due to cluster being down.

Do not wait for more than 30 seconds for any hive query. Added a timeout.

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
+++ b/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
@@ -269,7 +269,8 @@ public class SnappyHiveCatalog implements ExternalCatalog {
 
   private <T> T handleFutureResult(Future<T> f) {
     try {
-      return f.get();
+      // Time out if it takes more than 30 seconds
+      return f.get(30, TimeUnit.SECONDS);
     } catch (ExecutionException e) {
       throw new RuntimeException(e.getCause());
     } catch (Exception e) {


core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
```diff
@@ -269,7 +269,8 @@ private HMSQuery getHMSQuery() {
 
   private <T> T handleFutureResult(Future<T> f) {
     try {
-      return f.get();
+      // Time out if it takes more than 30 seconds
+      return f.get(30, TimeUnit.SECONDS);
```
rishitesh: Won't it be best to shutdown "hmsQueriesExecutorService" in case of shutdown, which calls catalog.close();

rishitesh: Ok. anyway, that is called. So why was the queries waiting?

kneeraj: The query waits for the result on a future. A different thread runs the query and gets the result. The  hive query kept on retrying ( from hive code ) and the future was indefinite wait. It looks like the hive thingy retries for a very very long time...
So thought of timing this out after 30 seconds. Please mind, this is the query on the metastore.

PradeepSurale:

PR #1084, opened 2018-07-14
Row Level Security changes

ahshahid:
Please review the initial changes for Row Level Security
The first cut of change is for
1) Creation of Policy
2) Using the Policy filter in the query 
Not yet implemented : 
1) Dropping of Policy
2) Invalidation of Plan when LDAP group changes
3) Dependency of Policy on Table
4) owner check that only table owner can create policy
5) Show Policy command
6) GemXd query rerouting to lead node for tables with policy
7) DUnit test
8) Comprehensive unit test.

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


cluster/src/test/scala/org/apache/spark/sql/store/SecurityBugTest.scala
```diff
@@ -75,7 +75,7 @@ class SecurityBugTest extends SnappyFunSuite with BeforeAndAfterAll {
     System.setProperty("gemfirexd.authentication.required", "false")
   }
 
-  test("Bug SNAP-2255 connection pool exhaustion") {
+  ignore("Bug SNAP-2255 connection pool exhaustion") {
```
sumwale: Why disabled?

core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
```diff
@@ -397,6 +397,8 @@ object SnappyParserConsts {
   final val WHERE: Keyword = reservedKeyword("where")
   final val WITH: Keyword = reservedKeyword("with")
   final val FUNCTION: Keyword = reservedKeyword("function")
+  final val POLICY: Keyword = reservedKeyword("policy")
```
sumwale: This is not reserved in any of the dbs including Postgres (e.g. see https://www.petefreitag.com/tools/sql_reserved_words_checker/?word=POLICY). Can this not be made nonReservedKeyword?

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -210,6 +224,8 @@ abstract class SnappyDDLParser(session: SparkSession)
   final def SETS: Rule0 = rule { keyword(Consts.SETS) }
   final def LATERAL: Rule0 = rule { keyword(Consts.LATERAL) }
 
+  private val CURRENT_USER = "CURRENT_USER"
+
```
sumwale: This also needs support for something similar to CURRENT_GROUPS to return the groups of a user. Can be tracked for next.

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -291,6 +307,68 @@ abstract class SnappyDDLParser(session: SparkSession)
...
+      forOpt match {
+        case Some(v) => v.asInstanceOf[String].trim
+        case None => SnappyParserConsts.SELECT.upper
+      })
+  }
+
+  protected final def policyTo: Rule1[Seq[String]] = rule {
+    (TO ~
+        (CURRENT ~ '_' ~ USER ~ push(CURRENT_USER)  |
+            (LDAPGROUP ~ ws ~ ':' ~ ws ~
```
sumwale: Might be cleaner to make this consistent with LDAPGROUP: make CURRENT_USER as a nonReservedKeyword and use that instead of using '_' this way.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1402,6 +1434,36 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
+            "for smart connector mode")
+      case _ =>
+    }
+
+    plan match {
+      case LogicalRelation(rls: RowLevelSecurityRelation, _, _) =>
+        sessionCatalog.invalidateTable(tableIdent)
+        rls.enableOrDisableRowLevelSecurity(tableIdent, enableRls)
+        // TODO : Asif : Is it needed what does this do?
+        SnappyStoreHiveCatalog.registerRelationDestroy()
```
sumwale: This will force refetch of meta-data by smart connector (bucket info etc) among other things. Perhaps can be avoided but does not harm and good to have.

core/src/main/scala/org/apache/spark/sql/aqp/SnappyContextFunctions.scala
```diff
@@ -33,7 +34,18 @@ class SnappyContextFunctions {
...
+
+    registry.registerFunction("CURRENT_USER",
+      e => {
+        if (! e.isEmpty) {
+          throw new AnalysisException("Argument(s)  passed for zero arg function " +
+              s"CURRENT_USER")
+        }
+        CurrentUser()
+      })
+  }
```
sumwale: I think this was specifically for AQP functions. There is another list in SnappyDataFunctions.registerSnappyFunctions. Also adding use the overload with "info" field then it will appear in DESCRIBE FUNCTION.

core/src/main/scala/org/apache/spark/sql/execution/SecurityUtils.scala
```diff
@@ -79,4 +83,22 @@ object SecurityUtils {
...
+  def allowPolicyOp(currentUser: String, table: QualifiedTableName,
+      session: SnappySession): Boolean = {
+    if (Misc.isSecurityEnabled) {
+      val connProps = ExternalStoreUtils.validateAndGetAllProps(Some(session), mutable.Map.empty)
+      val pooledConn = ExternalStoreUtils.getConnection("policyConn", connProps, false)
+      try {
+        val stmt = pooledConn.createStatement()
+        stmt.executeQuery(s"select 1 from sys.systables s, sys.sysschemas c where " +
+            s"s.tablename = '${table.table}' and s.tableschemaname = '${table.schemaName}' and " +
+            s"s.schemaid = c.schemaid and c.authorizationid = '$currentUser'").next()
```
sumwale: Will be much better to use something similar to GfxdSystemProcedures.authorizeTableOperation.

cluster/src/test/scala/org/apache/spark/sql/store/SecurityBugTest.scala
ahshahid: mistake. did it while testing..

ahshahid: thanks for catching it

core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
ahshahid: ok. will do it

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
ahshahid: yes. that needs to be provided.

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
ahshahid: OK. will do that

ahshahid: Actually this CURRENT_USER is ambiguous because there is added a  CURRENT_USER function too.  Dae Song got confused. I had mentioned this in the call . It would have been better if they were different.

core/src/main/scala/org/apache/spark/sql/aqp/SnappyContextFunctions.scala
ahshahid: Oh I did not know that . So I renamed the function name used for registering aqp functions to indicate common usage. I also incidentally renamed it as registerSnappyFunctions. Will move it to functions.scala after the merge. Adding the info.

core/src/main/scala/org/apache/spark/sql/execution/SecurityUtils.scala
ahshahid: will work on this after merge..

PR #1091, opened 2018-07-17
Minor automation in build script

ashetkar:
Automate inclusion of enterprise artifacts, if available, into connectors directory.

## Patch testing
./gradlew distTar

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


build.gradle
```diff
@@ -71,7 +71,7 @@ allprojects {
   apply plugin: 'nebula.ospackage-base'
 
   group = 'io.snappydata'
-  version = '1.0.2-SNAPSHOT'
+  version = '1.0.2'
```
kneeraj: We will have an RC build first and then within 10 days should have the 1.0.2

build.gradle
```diff
@@ -121,7 +121,7 @@ allprojects {
     janinoVersion = '3.0.8'
     derbyVersion = '10.12.1.1'
     pegdownVersion = '1.6.0'
-    snappyStoreVersion = '1.6.2-SNAPSHOT'
+    snappyStoreVersion = '1.6.2'
```
kneeraj: Plz look at above comment...

build.gradle
```diff
@@ -858,17 +857,45 @@ task product(type: Zip) {
...
-    if (isEnterpriseProduct && hasAqpProject) {
-      // copy enterprise shared libraries for optimized JNI calls
-      copy {
-        from aqpProject.projectDir.path + '/lib'
-        into "${snappyProductDir}/jars"
+    if (isEnterpriseProduct) {
+      if (hasAqpProject) {
+        // copy enterprise shared libraries for optimized JNI calls
+        copy {
+          from aqpProject.projectDir.path + '/lib'
```
kneeraj: If the intention is to just copy only the native libs then would it be better to have some regex for those files, instead of full dir. You might not want to add other jars which you keep here ... I don't know...may be it is ok...but please think...

build.gradle
```diff
@@ -858,17 +857,45 @@ task product(type: Zip) {
...
+          into "${snappyProductDir}/connectors"
+        }
+        copy {
+          from gfeFunctionProject.jar.destinationDir
+          into "${snappyProductDir}/connectors"
+        }
+        copy {
+          from "${gemfireConnectorProject.projectDir}/examples/quickstart/data"
+          into "${snappyProductDir}/connectors"
+          include "persons.jar"
```
kneeraj: Shouldn't we place in the quickstart folder. May be main quickstart. Copying here in connector seems odd.

build.gradle
ashetkar: Ok

build.gradle
ashetkar: As that's required by the connector project, thought it'll be better to have it in one place. @ahshahid ?

We also place JDBC client jar in here. Can restructure the directory into connectors/jdbc, connectors/gemfire, connectors/gemfire/quickstart and connectors/jdbc.

build.gradle
ashetkar: There is only a single jar in respective directories. If more get added in future, we can update the script with appropriate regex.

PR #1094, opened 2018-07-18
[ENT-34] Avoid setting snappy user name.

rishitesh:
Security code used to set snappy username in all pool connection obtained in the system.
Added a check to allow it only for snappy.

## Patch testing
precheckin
## ReleaseNotes.txt changes
NA
## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
@@ -334,7 +334,12 @@ object ExternalStoreUtils {
       hikariCP: Boolean): ConnectionProperties = {
     val (user, password) = getCredentials(session)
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/ExternalStoreUtils.scala
```diff
@@ -334,7 +334,12 @@ object ExternalStoreUtils {
       hikariCP: Boolean): ConnectionProperties = {
     val (user, password) = getCredentials(session)
 
-    if (!user.isEmpty && !password.isEmpty) {
+    val isSnappy = dialect match {
+      case GemFireXDDialect | GemFireXDClientDialect => true
+      case _ => false
+    }
+
+    if (!user.isEmpty && !password.isEmpty && isSnappy) {
```
ashetkar: Minor: May be evaluate isSnappy only if credentials are available? Or vice versa.

PR #1096, opened 2018-07-19
Fix to SNAP-2438

kneeraj:

currentschema should be considered for plan caching

## Patch testing

Added a unit test

## ReleaseNotes.txt changes

No

## Other PRs 

None


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -2289,29 +2290,32 @@ object SnappySession extends Logging {
   }
 }
 
-final class CachedKey(val session: SnappySession, private val lp: LogicalPlan,
+final class CachedKey(val session: SnappySession,
+   val currSchema: String, private val lp: LogicalPlan,
     val sqlText: String, val hintHashcode: Int) {
```
sumwale: indentation is different for the two lines

PR #1102, opened 2018-07-20
Fix for SNAP-2440. 

kneeraj:
Will revisit for a more decent fix if required after 102

Do not do anything by default. I mean no extra analysis rule or new node type.
Instead just catch the Analysis exception and retry with new UnresolvedStar node or UnresolvedAttributeNode as the case may be with unqualified projection name.

## Patch testing

Added a unit test

## ReleaseNotes.txt changes

Yes.

## Other PRs 

No.

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/MiscTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/MiscTest.scala
@@ -94,22 +94,24 @@ class MiscTest extends SnappyFunSuite with Logging {
     snc.sql(s"insert into test values (1, 2), (4, 5), (6, 7)")
     val sqlstrs = Seq(s"select app.test.* from app.test",
       s"select app.test.col1, app.test.col2 from app.test",
-      s"select app.test.col1 from test",
       s"select col1, col2 from app.test",
       s"select * from app.test",
       s"select test.* from test")
-    sqlstrs.foreach(sqlstr => snc.sql(sqlstr).collect().size === 3)
+    sqlstrs.foreach(sqlstr => {
+      val res = snc.sql(sqlstr).collect()
+      assert(res.size === 3)
+      assert(res(0).get(0) != res(0).get(1))
+    })
 
-    //    val badsqls = Seq(s"select apppp.test.* from app.test",
-    //      s"select app.test.col99, app.test.col2 from app.test",
-    //      s"select testt.* from app.test",
-    //      s"select test.*.* from test")
-    //    badsqls.foreach(sqlstr =>
-    //      try {
-    //        snc.sql(sqlstr)
-    //        fail(s"expected analysis exception for $sqlstr")
-    //      } catch {
-    //        case ae: AnalysisException => // expected ... ignore
-    //      })
+    val badsqls = Seq(s"select apppp.test.* from app.test",
+      s"select app.test.col99, app.test.col2 from app.test",
+      s"select testt.* from app.test")
+    badsqls.foreach(sqlstr =>
+      try {
+        snc.sql(sqlstr)
+        fail(s"expected analysis exception for $sqlstr")
+      } catch {
+        case ae: AnalysisException => // expected ... ignore
+      })
   }
 }
--- a/cluster/src/test/scala/org/apache/spark/sql/MiscTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/MiscTest.scala
@@ -88,4 +88,28 @@ class MiscTest extends SnappyFunSuite with Logging {
       snc.sql(s"set schema app")
     }
   }
+
+  test("SNAP-2440") {
+    snc.sql(s"create table test(col1 int not null, col2 int not null) using column options()")
+    snc.sql(s"insert into test values (1, 2), (4, 5), (6, 7)")
+    val sqlstrs = Seq(s"select app.test.* from app.test",
+      s"select app.test.col1, app.test.col2 from app.test",
+      s"select app.test.col1 from test",
+      s"select col1, col2 from app.test",
+      s"select * from app.test",
+      s"select test.* from test")
+    sqlstrs.foreach(sqlstr => snc.sql(sqlstr).collect().size === 3)
+
+    //    val badsqls = Seq(s"select apppp.test.* from app.test",
+    //      s"select app.test.col99, app.test.col2 from app.test",
+    //      s"select testt.* from app.test",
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -239,6 +242,53 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
     }
   }
 
+  val unresolvedStarRegex = """(cannot resolve ')(\w+).(\w+).*(' give input columns.*)""".r
```
rishitesh: Move all this code to some utility class.

cluster/src/test/scala/org/apache/spark/sql/MiscTest.scala
```diff
@@ -88,4 +88,30 @@ class MiscTest extends SnappyFunSuite with Logging {
       snc.sql(s"set schema app")
     }
   }
+
+  test("SNAP-2440") {
```
rishitesh: Can you add  negative tests for wrong table name, invalid function name etc.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -239,6 +242,53 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
+  val unresolvedColRegex = """(cannot resolve '`)(\w+).(\w+).(\w+)(.*given input columns.*)""".r
+
+  // Hack to fix SNAP-2440 ( TODO: Will return after 102 to see if a better fix is warranted )
+  private def reAnalyzeForUnresolvedAttribute(
+    originalPlan: LogicalPlan, e: AnalysisException,
+    retryCount: Int): Option[LogicalPlan] = {
+    if (!e.getMessage().contains("cannot resolve")) return None
+    val unresolvedNodes = originalPlan.expressions.filter(
+      x => x.isInstanceOf[UnresolvedStar] | x.isInstanceOf[UnresolvedAttribute])
+    if (retryCount > unresolvedNodes.size) return None
```
rishitesh: This can be merged with the first statement of the method to return.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: @rishitesh -- Since this is very specific to this fix and we would certainly want to revisit so I kept it right there. I didn't want to have references to this fix in any other part of the code. In fact I added these regexp in the method itself but then like for every analysis exception I didn't want to create the regexp object.

cluster/src/test/scala/org/apache/spark/sql/MiscTest.scala
kneeraj: It is there. Please check testt.* negative case.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: See, if the exception type is not "cannot resolve" in the message you need not traverse the tree and that is why I kept it like that.

rishitesh: ok.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
rishitesh: Just don't want SnappySession to get riddled with specific handling. But yes if we are revisiting then its fine.

kneeraj: Ok. Let me file a task for this.

sumwale: Can move the regex to SnappySession object so that it is singleton and does not get created for every session.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -239,6 +242,53 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
+          case unresolvedStarRegex(first, schema, table, last) =>
+            if (sessionCatalog.tableExists(new QualifiedTableName(schema, table))) {
+              val qname = s"$schema.$table"
+              if (qname.equalsIgnoreCase(targetString)) {
+                matched = true
+              }
+            }
+          case _ => matched = false
+        }
+        if (matched) UnresolvedStar(None) else us
```
sumwale: This looks incorrect. What if there are multiple qualified names like "app.table1.\*, app.table2.\*"?

kneeraj: @sumwale If there are multiple such comma separated elements then each will have their own UnresolvedAttribute node. If you see the logic the retry is tried for number of such nodes present in the plan. I just tested it and it works fine.
This is what I tried:
select a.b.*, a.c.* from a.b, a.c;
Any other syntax you are worried about?

kneeraj:
>> Not sure why this is being done here and not in parser itself? I think in parser if parts are > 1 then better to just resolve there and substitute required resolution as in the changes below (the UnresolvedStar one needs to be discussed more).

@sumwale - My idea was to just cater to the final error we are getting. Modifying the tree at the time of parsing to me looked a bit more error prone as based on the first two element we can determine whether the table exists or not but may be the overall query correctness is affected. More precisely I am worried about a wrong syntax which will get executed successfully: For example:

There are two schema called: test1 and test2 and there two tables test1.t1 and test2.t1
so a query like this: select test1.t1.* from test2.t1
So if we just check for existence of test1.t1 we will get it exists but if we replace that by only * or t1.* then it will execute successfully.

Having said that, I have filed this ticket  https://jira.snappydata.io/browse/SNAP-2442 to revisit this implementation as I was not comfortable with this implementation. Tomorrow if exception string is modified this will get broken.

So let me know if you think if I should do this in parser itself.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
kneeraj: Ok. Will do. But waiting for the response on the main implementation approach. In the parser if I move the logic then this regexp will anyways not be required.

kneeraj:
Closing this PR as we would like to revisit this and verify how spark handles this. After 102.

PR #1110, opened 2018-07-31
Added Fix to query with optional n argument

PradeepSurale:
In "fetch first n row only" it was expecting to pass 'n' number of first rows.
Now we made n as a optional field, if n didn't pass then it will fetch first 1 row.

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


kneeraj:
Please add a simple test where you test both. Mention the number and another sql without the number mentioned in which case 1 will be default.

PR #1111, opened 2018-08-02
Fix for sqlJoin queries and addition of new queries in NW schema tests for SNAP-2451

sonalsagarwal:
- Fix few queries in sqlJoin test for converted row store tests.
- Add new queries to the NW schema for SNAP-2451

## Patch testing
 - ran sqlJoin.bt from converted tests and northwind.bt

## ReleaseNotes.txt changes
N/A


dtests/src/test/scala/io/snappydata/hydra/SnappyTestUtils.scala
```diff
@@ -85,8 +95,19 @@ object SnappyTestUtils {
...
+  def assertQueryFullResultSet(snc: SnappyContext, sqlString: String, queryNum: String,
+      tableType: String, pw: PrintWriter, sqlContext: SQLContext,
+      usePlanCaching: Boolean): Any = {
+    var snappyDF: DataFrame = null
+    if(!usePlanCaching) {
+      snappyDF = snc.sqlUncached(sqlString)
+    } else {
+      snappyDF = snc.sql(sqlString)
+    }
     var sparkDF = sqlContext.sql(sqlString);
```
swatisawant: Add common code in a separate method and use the same in both the assertQueryFullResultSet methods.

dtests/src/resources/scripts/northWind/nw_queriesWithChangingConstants.sql
```diff
@@ -0,0 +1,321 @@
+elapsedtime on;
```
swatisawant: This should include only queries with changing constants. No need to add all other queries in the bt as it will increase test and analysis time. We are already having coverage for the queries without changing any constants in many tests.

dtests/src/test/java/io/snappydata/hydra/adAnalytics/adAnalytics.conf
```diff
@@ -83,6 +83,12 @@ INITTASK    taskClass   = util.StopStartVMs  taskMethod = StopStart_initTask
     threadGroups = snappyThreads, locatorThreads, snappyStoreThreads, leadThreads;
 
 
+INITTASK   taskClass   = io.snappydata.hydra.installJar.DynamicJarLoadingTest taskMethod  = HydraTask_deployJarUsingJDBC
```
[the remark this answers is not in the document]
sonalsagarwal: The task does not fail as DynamicJarLoadingTest extends SnappyTest. Yes,  will modify the class name to SnappyTest

dtests/src/test/scala/io/snappydata/hydra/northwind/NWTestUtil.scala
```diff
@@ -25,6 +25,8 @@ import org.apache.spark.sql._
 
 object NWTestUtil {
 
+  var executeQueriesByChangingConstants: Boolean = false;
```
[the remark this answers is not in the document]
sonalsagarwal: This, as we decided, will be done later as all northWind jobs and apps will need modification. This variable is used to set the default value which is modified by the method which is called by the new app and job written for query execution.

dtests/src/test/scala/io/snappydata/hydra/SnappyTestUtils.scala
sonalsagarwal: The code is added in the common method, only for signature, a new method is added.

PR #1112, opened 2018-08-02
[SNAP-2351] Fixed result mismatch.

rishitesh:
This issue fixes a problem with row tables and hash joins.
Also, fixes SNAP-2443 due to plan caching of Range based dfs. 

Due to plan caching, we don't re-analyze a SQL plan and re-use the same plan if the queries are differed only by literal constants.
However, row tables need a re-evaluation of filters which they can push down to store, every time they execute. For this filter evaluation is in getPartition() method of RowFormatScanRDD. 
The problem was getPartition is not called again once the partitions are determined. To circumvent this problem we set "numPartition" variable of RDD to null in CachedDataFrame. 

However, this was causing a problem in HashJoin as the build side RDD is not in the dependency graph. 
Moreover, the linkPartitionToBucket flag is an analysis time flag which is used in RDD computation time. This was causing the issue of incorrect partition determination. 

a) Kept reference of build side RDDs in DelegateRDD. This was used to make the numPartition null.
b) Stored the linkPartitionToBucket flag in CachedDataFrame and used it for subsequent execution. 
c) Excluded RangeExec from plan caching. 

## Patch testing

precheckin pending.

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


core/src/test/scala/org/apache/spark/sql/store/SnappyJoinSuite.scala
```diff
@@ -549,6 +549,64 @@ class SnappyJoinSuite extends SnappyFunSuite with BeforeAndAfterAll {
...
+    snc.sql(s"insert into trade.customers values(4,'ccc','2012-02-16','ccc-xyz',1)")
+    snc.sql(s"insert into trade.customers values(5,'ddd','2012-01-16','ddd-xyz',1)")
+    snc.sql(s"insert into trade.customers values(6,'eee','2012-03-17','eee-xyz',1)")
+    snc.sql(s"insert into trade.networth values(1,10.2,11.2,10000,5000,1)")
+    snc.sql(s"insert into trade.networth values(2,10.2,11.2,10000,5000,1)")
+    snc.sql(s"insert into trade.networth values(3,13.2,11.2,15000,8000,1)")
+    snc.sql(s"insert into trade.networth values(4,13.2,11.2,12000,3000,1)")
+    snc.sql(s"insert into trade.networth values(5,13.2,14.2,20000,10000,1)")
+    snc.sql(s"insert into trade.networth values(6,15.2,12.2,25000,15000,1)")
+    snc.sql(s"set spark.sql.autoBroadcastJoinThreshold = -1")
```
kneeraj: I guess you can remove this. The problem is irrespective of this setting. Right?

rishitesh: It will test the DelegateRDD part which is used only in HashJoin.

PR #1119, opened 2018-08-06
Snap 2457

rishitesh:

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
@@ -470,6 +470,7 @@ object SnappySessionPerConnection {
     if (session != null) session
     else {
       val session = SnappyContext().snappySession
+      session.conf.set("snappydata.sql.planCaching", "true")
       val oldSession = connectionIdMap.putIfAbsent(connectionID, session)
       if (oldSession == null) session else oldSession
     }
--- a/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
+++ b/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
@@ -231,7 +231,8 @@ class CachedDataFrame(snappySession: SnappySession, queryExecution: QueryExecuti
     snappySession.sparkContext.setLocalProperty("spark.scheduler.pool", pool)
   }
 
-  private[sql] def refreshDF(): Boolean = {
+
+  private def prepareForCollect(): Boolean = {
     if (prepared) return false
     if (isCached) {
       reset()
@@ -241,11 +242,6 @@ class CachedDataFrame(snappySession: SnappySession, queryExecution: QueryExecuti
     snappySession.linkPartitionsToBuckets(flag = linkPart)
     // Forcibly re-evaluate the partitions.
     reEvaluatePartitions(cachedRDD :: Nil)
-    true
-  }
-
-  private def prepareForCollect(): Boolean = {
-    if (!refreshDF()) return false
     setPoolForExecution()
     // update the strings in query execution and planInfo
     if (currentQueryExecutionString eq null) {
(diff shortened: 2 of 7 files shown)


cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
```diff
@@ -470,6 +470,7 @@ object SnappySessionPerConnection {
     if (session != null) session
     else {
       val session = SnappyContext().snappySession
+      session.conf.set("snappydata.sql.planCaching", "true")
```
sumwale: Better to use Property.PlanCaching: Property.PlanCaching.set(session.sessionState.conf, true)

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
```diff
@@ -59,20 +60,26 @@ trait DynamicReplacableConstant extends Expression {
...
+      val literalValueRef = ctx.addReferenceObj("literal", this,
         classOf[DynamicReplacableConstant].getName)
-      isNull = ctx.freshName("isNull")
-      valueTerm = ctx.freshName("value")
+      val isNull = ctx.freshName("isNull")
+      val valueTerm = ctx.freshName("value")
+      termMap.put(ctx, TermValues(literalValueRef, isNull, valueTerm))
     } else {
-      assert(literalValueRef ne null)
+      assert(termMap.get(ctx).isDefined)
```
sumwale: use termMap.get() only once rather than 4 separate calls here and below.

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
```diff
@@ -59,20 +60,26 @@ trait DynamicReplacableConstant extends Expression {
...
     val addMutableState = !ctx.references.exists(_.asInstanceOf[AnyRef] eq this)
     if (addMutableState) {
-      literalValueRef = ctx.addReferenceObj("literal", this,
+      val literalValueRef = ctx.addReferenceObj("literal", this,
         classOf[DynamicReplacableConstant].getName)
-      isNull = ctx.freshName("isNull")
-      valueTerm = ctx.freshName("value")
+      val isNull = ctx.freshName("isNull")
+      val valueTerm = ctx.freshName("value")
+      termMap.put(ctx, TermValues(literalValueRef, isNull, valueTerm))
```
sumwale: For this case (the common one) don't need a get again below. Store the TermValues as result of the if (addMutableState) block. See the next comment below about the assertion so that result can be stored and the gets below should be removed.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -70,7 +70,13 @@ class CachedDataFrame(snappySession: SnappySession, queryExecution: QueryExecuti
...
 
   private def getExecRDD: RDD[InternalRow] =
-    if (isCached) cachedRDD else queryExecution.executedPlan.execute()
+    if (isCached) {
+      cachedRDD
+    } else {
+      snappySession.sessionState.
+          executePlan(queryExecution.logical).executedPlan.execute()
+    }
+
```
sumwale: This bypasses most of plan caching in any case. We need to discuss and underlying issue must be fixed rather than disabling plan caching altogether. It also needs to be explored why these issues have started cropping up now and before which indicates that there has been some recent change causing these. Has someone tried to check with a git bisect?

rishitesh: It still does not fail with plan caching enabled. One of the plans is the NonRecursive plan which assumes plan caching is enabled, hence it has started failing now. Same with CodeGenFallback.

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
rishitesh: ok

core/src/main/scala/org/apache/spark/sql/catalyst/expressions/ParamLiteral.scala
rishitesh: ok. Will change it.

sumwale:
Lets make sure to document this change.

PR #1128, opened 2018-08-18
[SNAP-2477] Adding an API to get table type for SnappyData tables.

smahajan05:

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
+++ b/core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
@@ -1069,6 +1069,24 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
     }
   }
 
+    def getTableType(table: String): String = {
+        val tableIdent = this.newQualifiedTableName(table)
+        try {
+            val relation: LogicalRelation = getCachedHiveTable(tableIdent)
+            // println(relation)
+            val tableType: ExternalTableType = relation match {
+                case LogicalRelation(mutable: BaseRelation, _, _) =>
+                    snappySession.sessionCatalog.getTableType(mutable)
+                case _ => ExternalTableType.apply("None")
+            }
+            tableType.name
+        } catch {
+            case _: TableNotFoundException | _: NoSuchTableException =>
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -1069,6 +1069,24 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
     }
   }
 
+    def getTableType(table: String): String = {
+        val tableIdent = this.newQualifiedTableName(table)
+        try {
+            val relation: LogicalRelation = getCachedHiveTable(tableIdent)
+            // println(relation)
```
rishitesh: remove println comment.

core/src/test/scala/org/apache/spark/sql/store/ColumnTableTest.scala
```diff
@@ -1430,6 +1430,56 @@ class ColumnTableTest
...
+            "'org.apache.spark.sql.streaming.TweetToRowsConverter')")
+
+        snc.sql("drop table if exists t1")
+        snc.sql(s"create table t1 (c1 integer,c2 string)")
+        snc.sql(s"insert into t1 values(1,'test1')")
+        snc.sql(s"insert into t1 values(2,'test2')")
+        snc.sql(s"insert into t1 values(3,'test3')")
+        val df = snc.sql("select * from t1")
+        df.show
+        val tempPath = "/tmp/" + System.currentTimeMillis()
```
rishitesh: You can use System. getProperty("user.dir") to write the CSV file, rather than a /tmp folder.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -1069,6 +1069,24 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
     }
   }
 
+    def getTableType(table: String): String = {
```
rishitesh: Write proper comments here. Also, write the expected values one would get.

PR #1135, opened 2018-08-25
fixing the policy test failures after the new restrictTableCreation f…

ahshahid:
…lag. The table name SYSIBM.SYSDUMMY1 is to skip permission check. Will be adding new tests with restrictTableCreation flag enabled in due course

skipping the permission check for SYSIBM.SYSDUMMY1 table when doing grant/revoke

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
@@ -189,6 +189,7 @@ object ToolsCallbackImpl extends ToolsCallback with Logging {
             }
           }
           PrivilegeInfo.checkOwnership(currentUser, sd, sd, dd)
+          sd.getAuthorizationId;
         } finally {
           if (contextSet) conn.getTR.restoreContextStack()
         }
--- a/cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
@@ -166,7 +166,7 @@ object ToolsCallbackImpl extends ToolsCallback with Logging {
     ret
   }
 
-  override def checkSchemaPermission(schema: String, currentUser: String): Unit = {
+  override def checkSchemaPermission(schema: String, currentUser: String): String = {
     val ms = Misc.getMemStoreBootingNoThrow
     if (ms != null) {
       var conn: EmbedConnection = null
@@ -181,7 +181,7 @@ object ToolsCallbackImpl extends ToolsCallback with Logging {
             schema, conn.getLanguageConnection.getTransactionExecute, false)
           if (sd == null) {
             if (schema.equals(currentUser)) {
-              if (ms.tableCreationAllowed()) return
+              if (ms.tableCreationAllowed()) return currentUser
               throw StandardException.newException(SQLState.AUTH_NO_ACCESS_NOT_OWNER,
                 schema, schema)
             } else {
@@ -192,7 +192,9 @@ object ToolsCallbackImpl extends ToolsCallback with Logging {
         } finally {
           if (contextSet) conn.getTR.restoreContextStack()
         }
+      } else {
+        currentUser
       }
-    }
+    } else currentUser
   }
 }
--- a/cluster/src/test/scala/org/apache/spark/sql/policy/AlterTableRowLevelSecurityEnableTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/policy/AlterTableRowLevelSecurityEnableTest.scala
@@ -18,6 +18,8 @@ package org.apache.spark.sql.policy
 
 import java.sql.{Connection, DriverManager}
 
+import com.pivotal.gemfirexd.internal.engine.Misc
+import com.pivotal.gemfirexd.internal.engine.store.GemFireStore
 import com.pivotal.gemfirexd.{Attribute, TestUtil}
 import io.snappydata.SnappyFunSuite
 import io.snappydata.core.Data
@@ -48,10 +50,16 @@ class AlterTableRowLevelSecurityEnableTest extends SnappyFunSuite
   val rowTableName: String = s"${tableOwner}.$rowTable"
 
   var ownerContext: SnappyContext = _
-
+  var allowCreateTableFlag: Boolean = _
   override def beforeAll(): Unit = {
-    this.stopAll()
-    System.setProperty("snappydata.RESTRICT_TABLE_CREATION", "false")
+    val ms = Misc.getMemStoreBootingNoThrow
+    if (ms != null) {
+      allowCreateTableFlag = ms.tableCreationAllowed
+    }
+    if (!allowCreateTableFlag) {
+      this.stopAll()
+      System.setProperty("snappydata.RESTRICT_TABLE_CREATION", "false")
+    }
     super.beforeAll()
 
     val seq = for (i <- 0 until numElements) yield {
@@ -78,9 +86,12 @@ class AlterTableRowLevelSecurityEnableTest extends SnappyFunSuite
   override def afterAll(): Unit = {
     ownerContext.dropTable(colTableName, true)
     ownerContext.dropTable(rowTableName, true)
+    val ms = Misc.getMemStoreBootingNoThrow
+    if (allowCreateTableFlag != ms.tableCreationAllowed()) {
+      this.stopAll()
+      System.setProperty("snappydata.RESTRICT_TABLE_CREATION", allowCreateTableFlag.toString)
+    }
     super.afterAll()
-    System.clearProperty("snappydata.RESTRICT_TABLE_CREATION")
-
   }
 
   test("check rls enable/disable for jdbc client") {
--- a/cluster/src/test/scala/org/apache/spark/sql/policy/AlterTableRowLevelSecurityEnableTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/policy/AlterTableRowLevelSecurityEnableTest.scala
@@ -50,16 +50,8 @@ class AlterTableRowLevelSecurityEnableTest extends SnappyFunSuite
   val rowTableName: String = s"${tableOwner}.$rowTable"
 
   var ownerContext: SnappyContext = _
-  var allowCreateTableFlag: Boolean = _
+
   override def beforeAll(): Unit = {
-    val ms = Misc.getMemStoreBootingNoThrow
-    if (ms != null) {
-      allowCreateTableFlag = ms.tableCreationAllowed
-    }
-    if (!allowCreateTableFlag) {
-      this.stopAll()
-      System.setProperty("snappydata.RESTRICT_TABLE_CREATION", "false")
-    }
     super.beforeAll()
 
     val seq = for (i <- 0 until numElements) yield {
@@ -86,11 +78,6 @@ class AlterTableRowLevelSecurityEnableTest extends SnappyFunSuite
   override def afterAll(): Unit = {
     ownerContext.dropTable(colTableName, true)
     ownerContext.dropTable(rowTableName, true)
-    val ms = Misc.getMemStoreBootingNoThrow
-    if (allowCreateTableFlag != ms.tableCreationAllowed()) {
-      this.stopAll()
-      System.setProperty("snappydata.RESTRICT_TABLE_CREATION", allowCreateTableFlag.toString)
-    }
     super.afterAll()
   }
 
--- a/cluster/src/test/scala/org/apache/spark/sql/policy/AlterTableRowLevelSecurityEnableTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/policy/AlterTableRowLevelSecurityEnableTest.scala
@@ -50,6 +50,8 @@ class AlterTableRowLevelSecurityEnableTest extends SnappyFunSuite
   var ownerContext: SnappyContext = _
 
   override def beforeAll(): Unit = {
+    this.stopAll()
+    System.setProperty("snappydata.RESTRICT_TABLE_CREATION", "false")
     super.beforeAll()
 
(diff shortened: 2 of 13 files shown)


core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -781,7 +781,8 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
           // TODO: the authorizationID should be correctly set in SparkSQLExecuteImpl
           // using LCC.getAuthorizationId() itself rather than getUserName()
           val user = snappySession.conf.get(Attribute.USERNAME_ATTR, "")
-          if (user.nonEmpty) {
+          if (user.nonEmpty && !(tableIdent.schemaName == SnappyStoreHiveCatalog.dummyTableSchema
```
ashetkar: .equalsIgnoreCase() instead of == check.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -1457,7 +1458,8 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
 }
 
 object SnappyStoreHiveCatalog {
-
+  val dummyTableSchema = "SYSIBM"
```
ashetkar: SchemaDescriptor.IBM_SYSTEM_SCHEMA_NAME already available.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
ashetkar: Does this not imply that anyone has permissions on SYSIBM.SYSDUMMY1 table?

ahshahid:
I also saw this failure for first time on master. It has to do with some
recent changes. In the grant revoke permission, this dummy schema table is
used and then exception occurs that user does not have permission on that
schema . Hence the check.  If there is another approach pls suggest.
I will implement ur feedback.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1620,7 +1632,7 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
     */
 
     sessionCatalog.registerPolicy(policyName, tableName, policyFor, applyTo, expandedPolicyApplyTo,
-      currentUser, filterStr, filter)
+      owner, filterStr, filter)
```
sumwale: Aren't owner and currentUser exactly same? The callbacks.checkSchemaPermission method returns the same as its argument as per the changes in this PR.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -781,7 +783,9 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
           // TODO: the authorizationID should be correctly set in SparkSQLExecuteImpl
           // using LCC.getAuthorizationId() itself rather than getUserName()
           val user = snappySession.conf.get(Attribute.USERNAME_ATTR, "")
-          if (user.nonEmpty) {
+          if (user.nonEmpty && !(
+              tableIdent.schemaName.equalsIgnoreCase(SchemaDescriptor.IBM_SYSTEM_SCHEMA_NAME )
+              && tableIdent.table.equalsIgnoreCase(SnappyStoreHiveCatalog.dummyTableName))) {
```
sumwale: This will allow all users to write to SYSIBM.DUMMY1 table? I hope Derby layer restricts this somewhere.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1567,9 +1569,19 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
     if (!SecurityUtils.allowPolicyOp(currentUser, tableName, this)) {
       throw new SQLException("Only Table Owner can create the policy", "01548", null)
     }
+    */
+    val callbacks = ToolsCallbackInit.toolsCallback
+    val owner = if (callbacks != null) {
+      // TODO: the authorizationID should be correctly set in SparkSQLExecuteImpl
+      // using LCC.getAuthorizationId() itself rather than getUserName()
+        callbacks.checkSchemaPermission(tableName.schemaName, currentUser)
+    } else {
```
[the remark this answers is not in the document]
ahshahid: The user is already normalized before it reaches this function

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
ahshahid: Not if the owner is a ldap group.

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
ahshahid: I also hope. Bcoz I am also not comfortable with the change. But I have no idea what else can be done at this point to fix the issue & why it started coming now.

cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
```diff
@@ -192,7 +192,9 @@ object ToolsCallbackImpl extends ToolsCallback with Logging {
         } finally {
           if (contextSet) conn.getTR.restoreContextStack()
         }
+      } else {
+        currentUser
       }
-    }
+    } else currentUser
```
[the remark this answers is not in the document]
sumwale: Ok, got it. Missed the PrivilegeInfo.checkOwnership call which now returns the schema owner. I think it is better to explicitly have "sd.getAuthorizationId" call after the PrivilegeInfo.checkOwnership and avoid checkOwnership return that.

cluster/src/test/scala/org/apache/spark/sql/policy/RestrictTableCreationPolicyTest.scala
```diff
@@ -0,0 +1,267 @@
...
+  var ownerContext: SnappyContext = _
+
+  private val sysUser = "gemfire10"
+  var serverHostPort: String = _
+
+
+  override def beforeAll(): Unit = {
+    System.setProperty("snappydata.RESTRICT_TABLE_CREATION", "true")
+    this.stopAll()
+    super.beforeAll()
```
sumwale: I think the property should be set after stopAll() to safe (in case stop depends on presence/absence of property).

cluster/src/test/scala/org/apache/spark/sql/policy/PolicyTest.scala
```diff
@@ -58,6 +55,7 @@ class PolicyTest extends SnappyFunSuite
   }
 
   override def beforeAll(): Unit = {
+    this.stopAll()
```
sumwale: should be reverted

cluster/src/test/scala/org/apache/spark/sql/policy/PolicyTest.scala
```diff
@@ -18,7 +18,7 @@ package org.apache.spark.sql.policy
 
 import com.pivotal.gemfirexd.Attribute
 import com.pivotal.gemfirexd.internal.engine.Misc
-import io.snappydata.SnappyFunSuite
+import io.snappydata.{Constant, SnappyFunSuite}
```
sumwale: can revert

cluster/src/test/scala/org/apache/spark/sql/policy/AlterTableRowLevelSecurityEnableTest.scala
```diff
@@ -77,7 +79,6 @@ class AlterTableRowLevelSecurityEnableTest extends SnappyFunSuite
     ownerContext.dropTable(colTableName, true)
     ownerContext.dropTable(rowTableName, true)
     super.afterAll()
-
   }
```
sumwale: can revert changes to this file

cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
```diff
@@ -181,18 +181,21 @@ object ToolsCallbackImpl extends ToolsCallback with Logging {
...
-              if (ms.tableCreationAllowed()) return
+              if (ms.tableCreationAllowed()) return currentUser
               throw StandardException.newException(SQLState.AUTH_NO_ACCESS_NOT_OWNER,
                 schema, schema)
             } else {
               throw StandardException.newException(SQLState.LANG_SCHEMA_DOES_NOT_EXIST, schema)
             }
           }
           PrivilegeInfo.checkOwnership(currentUser, sd, sd, dd)
+          sd.getAuthorizationId;
```
sumwale: semi-colon can be removed

cluster/src/test/scala/org/apache/spark/sql/policy/SecurityEnabledJdbcClientPolicyTest.scala
```diff
@@ -19,6 +19,7 @@ package org.apache.spark.sql.policy
 import java.sql.{Connection, DriverManager, SQLException}
 import java.util.Properties
 
+import com.pivotal.gemfirexd.internal.engine.Misc
 import com.pivotal.gemfirexd.{Attribute, TestUtil}
```
sumwale: can revert

cluster/src/test/scala/org/apache/spark/sql/policy/SecurityEnabledPolicyTest.scala
```diff
@@ -22,6 +22,8 @@ import com.pivotal.gemfirexd.Attribute
 import java.sql.SQLException
 
 import com.pivotal.gemfirexd.Attribute
+import com.pivotal.gemfirexd.internal.engine.Misc
+import com.pivotal.gemfirexd.internal.iapi.error.StandardException
 import com.pivotal.gemfirexd.security.{LdapTestServer, SecurityTestUtils}
```
sumwale: can revert

cluster/src/test/scala/org/apache/spark/sql/policy/SecurityEnabledPolicyTest.scala
```diff
@@ -53,7 +55,6 @@ class SecurityEnabledPolicyTest extends SnappyFunSuite
   private val sysUser = "gemfire10"
 
   override def beforeAll(): Unit = {
-    this.stopAll()
     super.beforeAll()
```
sumwale: stopAll not required? It will be if test depends on specific SparkConf props.

core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -595,9 +597,11 @@ abstract class SnappyDDLParser(session: SparkSession)
   protected def grantRevoke: Rule1[LogicalPlan] = rule {
     (GRANT | REVOKE | (CREATE | DROP) ~ DISK_STORE | ("{".? ~ CALL)) ~ ANY.* ~>
         /* dummy table because we will pass sql to gemfire layer so we only need to have sql */
-        (() => DMLExternalTable(TableIdentifier("SYSDUMMY1", Some("SYSIBM")),
+        (() => DMLExternalTable(TableIdentifier(SnappyStoreHiveCatalog.dummyTableName,
+          Some(SchemaDescriptor.IBM_SYSTEM_SCHEMA_NAME)),
           LogicalRelation(new execution.row.DefaultSource().createRelation(session.sqlContext,
-            SaveMode.Ignore, Map(JdbcExtendedUtils.DBTABLE_PROPERTY -> "SYSIBM.SYSDUMMY1"),
+            SaveMode.Ignore, Map(JdbcExtendedUtils.DBTABLE_PROPERTY ->
+                s"${SchemaDescriptor.IBM_SYSTEM_SCHEMA_NAME }.${SnappyStoreHiveCatalog.dummyTableName}"),
```
sumwale: this exceeds 100 chars; break the line

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
sumwale: Confirmed that no changes to SYSIBM.DUMMY1 are allowed and is restricted in GemXD layer in any case.

The reason this is coming is because catalog's registerDataSourceTable() now has a checkSchemaPermission call to restrict CREATE EXTERNAL TABLE, VIEW etc as per schema permissions that do not hit storage layers rather are only in hive catalog. Looks like GRANT/REVOKE also are calling registerDataSourceTable() with the dummy table. This should not happen in any case and registerDataSourceTable can return immediately for SYSIBM.DUMMY1 table. Perhaps that is a better fix but can be tracked for later if above changes have already been tested.

cluster/src/test/scala/org/apache/spark/sql/policy/PolicyTest.scala
ahshahid: This is needed because I am setting a new property in conf
spark.sql.crossJoin.enabled", "true"
which requires current cluster to stop.

sumwale: Ok. I have added few cleanups mentioned in my PR which shall be raised shortly.

ahshahid:
I am doing the cleanup

ahshahid:
R u doing the cleanup?

sumwale:
I was doing, but removed after your last note. Please go ahead.

ahshahid:
ok

PR #1136, opened 2018-08-26
Property to disable all hash joins

sumwale:

- added snappydata.sql.disableHashJoin to disable all hash joins including
  for replicated tables; the entire HashJoinStrategies is skipped with this as true
- use a boolean "disableHashJoin" in session to speed up lookup
- clear existing plan cache in case this flag is set
- clean up to remove scalastyle warnings in a bunch of classes

## Patch testing

manual

## ReleaseNotes.txt changes

Document the new property.

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/422

### DIFF
--- a/cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
@@ -19,7 +19,6 @@ package io.snappydata
 import java.io.File
 import java.lang.reflect.InvocationTargetException
 import java.net.URLClassLoader
-import java.util.UUID
 
 import com.pivotal.gemfirexd.internal.engine.Misc
 import com.pivotal.gemfirexd.internal.engine.distributed.utils.GemFireXDUtils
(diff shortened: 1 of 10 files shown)


jramnara:
Why is this property required ? The current hashJoinSize property is not sufficient ?

sumwale:
@jramnara That property is ignored for replicated table joins. This one disables everything (so replicated table joins will usually go the broadcast route from Spark).

jramnara:
The only issue I am aware of has to do with availability of memory for
hashJoins. What problem are we trying to solve?

-----
Jags
SnappyData blog <http://www.snappydata.io/blog>
Download binary, source <https://www.snappydata.io/download>

sumwale:
@jramnara Asif reported a case of incorrect result for a user which was fixed by disabling this internally. We have fixed issues around this in 1.0.2 release and there is no known issue remaining, but since the user test case or data is not available, so this property has been added for the case if it shows up again (while the underlying issue is investigated).

PR #1149, opened 2018-09-10
SNAP-2447: Set seed of random number used in query parameterization 

cruizen:
SNAP-2447: Set seed of random number used in query parameterization for reproducibility

Pass an additional parameter randomSeed in PerfRun.conf for TPC-H . Initialize a random number in TPCH_Queries with this seed for repeatable experiments (with exact same parameter values). 

## Patch testing

Ran TPCHDUnit and ran the query execution job with the same seed locally to verify that query parameters are identical in these two runs. 

## ReleaseNotes.txt changes

## Other PRs 
None

### DIFF
--- a/cluster/src/test/scala/io/snappydata/benchmark/snappy/tpch/QueryExecutor.scala
+++ b/cluster/src/test/scala/io/snappydata/benchmark/snappy/tpch/QueryExecutor.scala
@@ -145,7 +145,7 @@ object QueryExecutor {
         val queryResultsPrintStream: PrintStream = new PrintStream(queryResultsFileStream)
 
         try {
-                var queryToBeExecuted = TPCH_Queries.getQuery(queryNumber, isDynamic, isSnappy = true)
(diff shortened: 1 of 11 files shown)


cluster/src/test/scala/io/snappydata/benchmark/snappy/tpch/QueryExecutor.scala
```diff
@@ -141,7 +145,7 @@ object QueryExecutor {
         val queryResultsPrintStream: PrintStream = new PrintStream(queryResultsFileStream)
 
         try {
-          var queryToBeExecuted = TPCH_Queries.getQuery(queryNumber, isDynamic, isSnappy = true)
+                var queryToBeExecuted = TPCH_Queries.getQuery(queryNumber, isDynamic, isSnappy = true)
```
swatisawant: Seems space got added by mistake.

cruizen: Right. Thanks.

PR #1157, opened 2018-09-20
Structured streaming - default sink call back

vatsalmevada:

- Adding default sink call back implementation 
- Upgrading kafka version to 0.10 from 0.8
- Changed `SnappyStoreHiveCatalog#getKeyColumns` API to return `Seq` of columns instead of `Dataset`

## Patch testing

Manual testing, Unit tests, Dunit tests, precheckin

## Other PRs

`streaming_sink` PR on `snappy-connectors`


cluster/src/test/scala/org/apache/spark/sql/execution/benchmark/StringBenchmark.scala
```diff
@@ -233,12 +233,12 @@ object StringBenchmark {
    */
   def sparkContains(source: UTF8String, target: UTF8String): Boolean = {
     if (target.numBytes == 0) return true
-    val first = target.getByte(0)
-    var i = 0
-    while (i <= source.numBytes - target.numBytes) {
-      if (source.getByte(i) == first && matchAt(source, target, i)) return true
-      i += 1
-    }
+//    val first = target.getBytes(0)
```
kneeraj: @ymahajan Why is this commented out. Not needed?

core/src/test/scala/org/apache/spark/sql/streaming/SnappyStoreSinkProviderSuite.scala
```diff
@@ -0,0 +1,243 @@
...
+
+    kafkaTestUtils.createTopic(topic, partitions = 3)
+
+    val dataBatch1 = Seq(Seq(1, "name1", 30), Seq(2, "name2", 10),
+      Seq(3, "name3", 30))
+
+    kafkaTestUtils.sendMessages(topic, dataBatch1.map(r => r.mkString(",")).toArray)
+
+    val streamingQuery = createAndStartStreamingQuery(topic, testId, withEventTypeColumn = false)
+    Thread.sleep(batchDurationMillis + batchWriteWaitMillis)
```
kneeraj: Instead of sleep poll the sink state table for non emptiness.

core/src/test/scala/org/apache/spark/sql/streaming/SnappyStoreSinkProviderSuite.scala
```diff
@@ -0,0 +1,243 @@
...
+    createTable()
+    val topic = getTopic(testId)
+    kafkaTestUtils.createTopic(topic, partitions = 3)
+
+    val dataBatch1 = Seq(Seq(1, "name1", 20, 0), Seq(2, "name2", 10, 0))
+    kafkaTestUtils.sendMessages(topic, dataBatch1.map(r => r.mkString(",")).toArray)
+
+    val streamingQuery: StreamingQuery = createAndStartStreamingQuery(topic, testId)
+
+    Thread.sleep(batchDurationMillis + batchWriteWaitMillis)
```
kneeraj: Look at the comment above.

cluster/build.gradle
```diff
@@ -41,7 +41,8 @@ dependencies {
     compile project(':snappy-spark:snappy-spark-hive_' + scalaBinaryVersion)
     compile project(':snappy-spark:snappy-spark-repl_' + scalaBinaryVersion)
     compile project(':snappy-spark:snappy-spark-streaming_' + scalaBinaryVersion)
-    compile project(':snappy-spark:snappy-spark-streaming-kafka-0.8_' + scalaBinaryVersion)
+    compile project(':snappy-spark:snappy-spark-streaming-kafka-0.10_' + scalaBinaryVersion)
+    compile project(':snappy-spark:snappy-spark-sql-kafka-0.10_' + scalaBinaryVersion)
```
sumwale: Have we checked if any of the existing customers are still using kafka 0.8 and depend on it? If so we should support both like Spark does. @sudsmenon was going to come back on that for existing customers.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -848,8 +847,7 @@ object SnappyContext extends Logging {
     TOPK_SOURCE -> TOPK_SOURCE_CLASS,
     SOCKET_STREAM_SOURCE -> classOf[SocketStreamSource].getCanonicalName,
     FILE_STREAM_SOURCE -> classOf[FileStreamSource].getCanonicalName,
-    KAFKA_STREAM_SOURCE -> classOf[KafkaStreamSource].getCanonicalName,
-    DIRECT_KAFKA_STREAM_SOURCE -> classOf[DirectKafkaStreamSource].getCanonicalName,
+    KAFKA_STREAM_SOURCE -> classOf[DirectKafkaStreamSource].getCanonicalName,
```
sumwale: We need documentation and release notes that "kafka_stream" now means "directkafka_stream". If there are any existing users of "kafka_stream" then we cannot change this rather will need to support both like before.

core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
```diff
@@ -16,63 +16,135 @@
...
 
-  override def shortName(): String = SnappyContext.DIRECT_KAFKA_STREAM_SOURCE
+  override def shortName(): String = SnappyContext.KAFKA_STREAM_SOURCE
 
   override def createRelation(sqlContext: SQLContext,
-      options: Map[String, String],
-      schema: StructType): BaseRelation = {
+                              options: Map[String, String],
+                              schema: StructType): BaseRelation = {
     new DirectKafkaStreamRelation(sqlContext, options, schema)
```
sumwale: Our default code-style formatting should not lead to this. Please revert these indentation changes here and below and ensure that codeStyleSettings.xml is setup correctly in IDEA as documented.

core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
```diff
@@ -16,63 +16,135 @@
...
 final class DirectKafkaStreamRelation(
-    @transient override val sqlContext: SQLContext,
-    opts: Map[String, String],
-    override val schema: StructType)
-    extends StreamBaseRelation(opts)
+                                       @transient override val sqlContext: SQLContext,
+                                       opts: Map[String, String],
+                                       override val schema: StructType)
+  extends StreamBaseRelation(opts)
     with Logging with Serializable {
```
sumwale: see above about indentation

core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
```diff
@@ -16,63 +16,135 @@
...
 
-  val topicsSet = options("topics").split(",").toSet
+  val topics = options("subscribe").split(",").toList
```
sumwale: Any specific reason to change set to list. Is this dependent on order? Perhaps a LinkedHashSet will be better if it is dependent on order.

core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
```diff
@@ -16,63 +16,135 @@
...
-      // need to call copy() below since there are builders at higher layers
-      // (e.g. normal Seq.map) that store the rows and encoder reuses buffer
-      iter.flatMap(p => rowConverter.toRows(p._2).iterator.map(
-        encoder.toRow(_).copy()))
-    }
+    val consumerStrategies = ConsumerStrategies
+      .Subscribe[String, String](topics, kafkaParams, startingOffsets)
+
+    val stream = KafkaUtils.createDirectStream[String, String](context,
+      preferredHosts, consumerStrategies)
```
sumwale: Always assuming string keys and values in the stream. Why? We should support any key/value types like before. Also the Kafka-0.8 version used key and value decoder classes too. Do you know what Kafka-0.10 version is using for decoding? We don't have a mechanism to register Decoder implementations from SQL so removing the decoders here would mean that custom classes requiring own decoding won't work anymore. How does Kafka 0.10 do key/value decoding in any case?

core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
```diff
@@ -16,63 +16,135 @@
...
+            encoder.toRow(_).copy())
+        })
+      }
+    stream
   }
 }
+
+/**
+  * Copied from org.apache.spark.sql.kafka010.JsonUtils
+  */
```
sumwale: Are there any changes in the copy? If not then I think it might be better to invoke kafka010.JsonUtils by reflection.

core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
```diff
@@ -0,0 +1,184 @@
...
+      if (eventTypeColumnAvailable) {
+        val msg = s"$EVENT_TYPE_COLUMN is present in data but key columns are not defined on table."
+        throw new IllegalStateException(msg)
+      } else {
+        df.write.insertInto(tableName)
+      }
+    }
+    // test hook for validating idempotency
+    if (parameters.contains("internal___failBatch")
+        && parameters("internal___failBatch") == "true") {
```
sumwale: Can move this out as a constant with name indicating test hook like TEST_FAILBATCH_OPTION

core/src/main/scala/org/apache/spark/sql/streaming/WindowPhysicalPlan.scala
```diff
@@ -42,7 +42,7 @@ case class WindowPhysicalPlan(
       override def dependencies = parentStreams.toList
 
       override def slideDuration: Duration = {
-        slide.getOrElse(parentStreams.head.slideDuration)
+        parentStreams.head.slideDuration
```
sumwale: slide is no longer used?

core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
```diff
@@ -0,0 +1,184 @@
...
+    var posDup = false
+
+    if (updated == 0) {
+      try {
+        snappySession.insert(SINK_STATE_TABLE, Row(streamQueryId, batchId))
+        posDup = false
+      }
+      catch {
+        case e: SQLException if e.getSQLState.equals("23505") => posDup = true
+        case exception: Throwable => throw exception
```
sumwale: Don't need to explicitly catch and throw exception. Just remove the line and that will happen either way.

core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
```diff
@@ -0,0 +1,184 @@
...
+
+  override def addBatch(batchId: Long, data: Dataset[Row]): Unit = {
+    val streamQueryId = parameters(STREAM_QUERY_ID).toUpperCase
+
+    val updated = snappySession.sql(s"update $SINK_STATE_TABLE " +
+      s"set batch_id=$batchId where stream_query_id='$streamQueryId' and batch_id != $batchId")
+      .collect()(0).getAs("count").asInstanceOf[Long]
+
+    // TODO: use JDBC connection here
+    var posDup = false
```
sumwale: Come to think of it, using JDBC connection is a required here. We have to use it with transaction isolation level as READ_COMMITTED or better REPEATABLE_READ. There is no way to set those from Spark APIs. Without those there are no atomicity guarantees on any changes on the row table. Use a pooled connection like in JDBCMutableRelation.executeUpdate (but set and reset the transaction isolation level on the connection explicitly).

core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
```diff
@@ -0,0 +1,184 @@
...
+class DefaultSnappySinkCallback extends SnappySinkCallback {
+  def process(snappySession: SnappySession, parameters: Map[String, String],
+              batchId: Long, df: Dataset[Row], posDup: Boolean) {
+
+    df.cache()
+    val snappyTable = parameters(TABLE_NAME).toUpperCase
+    DefaultSnappySinkCallback.log.debug(s"Processing for $snappyTable and batchId $batchId")
+
+    val tableName = parameters(TABLE_NAME).toUpperCase
+    val keyColumnsDefined = !snappySession.sessionCatalog.getKeyColumns(tableName).head(1).isEmpty
```
sumwale: This creates a newDataFrame and will result in yet another job being created which is an unnecessary overhead. Refactor the sessionCatalog.getKeyColumns to return the Seq[Column] (say getKeyColumnsSeq) and use that here.

core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
```diff
@@ -0,0 +1,184 @@
...
+  private val log = Logger.getLogger(classOf[DefaultSnappySinkCallback].getName)
+}
+
+import org.apache.spark.sql.snappy._
+
+class DefaultSnappySinkCallback extends SnappySinkCallback {
+  def process(snappySession: SnappySession, parameters: Map[String, String],
+              batchId: Long, df: Dataset[Row], posDup: Boolean) {
+
+    df.cache()
```
sumwale: Will suggest adding a df.count() immediately after to force cache materialization. Sometimes caching on first query materialization causes unexpected cache due to different kind of plan. Also first execution will have incorrect stats that can be expensive depending on the query, so having a count() is safer. Specifically check if the deleteFrom leads to a SortMergeJoin plan rather than a broadcast under some cases. If so, then count() is definitely required.

core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
suranjan: Is atomicity required here? There is only one thread modifying one row in the table and even the reader thread is same. Unless we want to take care of inconsistency due to all node failures, simple operation should suffice.

Anyways, we have plans to use connection later for performance reasons as suggested earlier. Will track that as JIRA ticket.

sumwale: There are some corner cases where replicated table apply can result in inconsistencies. For instance if lead node fails after sending the update to few nodes but before sending to remaining, then replicas can become inconsistent. Even eventual consistency will not help. Most of these cases are handled for partitioned regions even without transactions, but for replicated tables we do need transactions for atomicity and consistency. READ_COMMITTED is enough for single row changes.

cluster/build.gradle
ymahajan: AFAIK, we don't have customers using 0.8, let @sudsmenon confirm though.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
ymahajan: @sumwale Yes, there will be doc change also there is a syntax change in Kafka param as well. We need to update the docs accordingly.

core/src/main/scala/org/apache/spark/sql/streaming/WindowPhysicalPlan.scala
ymahajan: This was a bug reported on slack, where the parent's slide was ignored with the default one. This change fixes that bug.

core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
vatsalmevada: Updated code to invoke kafka010.JsonUtils using reflection instead of copying the code.

core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
vatsalmevada: Reverted back to using Set again as order doesn't seem to be required here. @ymahajan please let us know in case you faced any issues with the use of Set.

core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
vatsalmevada: Added a JIRA to track this: SNAP-2598

core/src/main/scala/org/apache/spark/sql/streaming/WindowPhysicalPlan.scala
sumwale: Won't this ignore the slide specified for this plan set from parser or otherwise? I think the issue to fix would be why is slide getting set to default one and not remain at None in case no explicit slide has been specified for this plan. Or does parentStreams.head somehow inherit the slide duration of this plan?

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
```diff
@@ -180,6 +180,10 @@ class SnappyStoreHiveCatalog(externalCatalog: SnappyExternalCatalog,
 
   /** API to get primary key or Key Columns of a SnappyData table */
   def getKeyColumns(table: String): Dataset[Column] = {
+    CatalogImpl.makeDataset(getKeyColumnsSeq(table), snappySession)
+  }
```
sumwale: Ok, checked the code and I was wrong about this. It is not used anywhere other than tests. Confused it with the listTables/listColumns functions in CatalogImpl that do return Datasets and perhaps whoever added this used the same template. I think it will be better to remove this and rename getKeyColumnsSeq to getKeyColumns as you suggested (and update the tests).

core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
```diff
@@ -45,34 +43,39 @@ final class DirectKafkaStreamRelation(
...
-  val KD = options.getOrElse("KD", "kafka.serializer.StringDecoder")
-  val VD = options.getOrElse("VD", "kafka.serializer.StringDecoder")
+  private val preferredHosts = LocationStrategies.PreferConsistent
+  private val startingOffsets = getStartingOffsets
+
+  private def getStartingOffsets = {
+    val clazz = Utils.classForName("org.apache.spark.sql.kafka010.JsonUtils")
+    clazz.getMethod("partitionOffsets", classOf[String])
+        .invoke(null, options("startingOffsets"))
+        .asInstanceOf[Map[TopicPartition, Long]]
```
sumwale: You can store the method to avoid looking up in every call. Also setAccessible will be good to be sure:

```scala
private val partitionOffsetMethod = {
  val clazz = Utils.classForName("org.apache.spark.sql.kafka010.JsonUtils")
  val m = clazz.getMethod("partitionOffsets", classOf[String])
  m.setAccessible(true)
  m
}
```

core/src/main/scala/org/apache/spark/sql/hive/SnappyStoreHiveCatalog.scala
suranjan: @sumwale @vatsalmevada This gets used in CDC connector code(Although test but that is an example App too). If we are changing this then we need to change the code there as well and communicate others who may be using this API.
Please have a look at JavaCdcStreamingTestSuite.java

vatsalmevada: Removed the API returning `Dataset` and renamed `getKeyColumnsSeq` to `getKeyColumns`. Also, this change leads to changes in snappy connectors repo. A separate PR is raised for the same.

core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
vatsalmevada: The code is working fine even without explicitly setting accessible to `true` hence not calling `setAccessible`. The other suggestion is incorporated.

cluster/build.gradle
vatsalmevada: I had a word with @sudsmenon and we are not having any customers using Kafka 0.8. So it should be fine to remove support for 0.8.

PR #1159, opened 2018-09-21
[SNAP-2571] add query hint to force join type

sumwale:

- add a wrapper LogicalPlan to carry the query hints for a plan
- analyze above in Snappy join strategies to force hash or broadcast join

## Patch testing

in progress

## ReleaseNotes.txt changes

Note about the query hint

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/io/snappydata/Literals.scala
+++ b/core/src/main/scala/io/snappydata/Literals.scala
@@ -466,6 +466,19 @@ object QueryHint extends Enumeration {
   val JoinOrder = Value("joinOrder")
 
   /**
+   * Query hint to force a join type for the current join. This should appear after
+   * the required table/plan in FROM where the specific join type has to be forced.
+   * Note that this will enable the specific join type only if it is possible
+   * for that table in the join and silently ignore otherwise.
+   *
+   * Possible values are broadcast, hash.
+   *
+   * Example:<br>
+   * SELECT * FROM t1 /`*`+ joinType(broadcast) -- broadcast t1 *`/`, t2 where ...
+   */
+  val JoinType: Value = Value("joinType")
+
+  /**
    * Query hint for SQL queries to serialize STRING type as CLOB rather than
    * as VARCHAR.
    *
--- a/core/src/main/scala/io/snappydata/Literals.scala
+++ b/core/src/main/scala/io/snappydata/Literals.scala
@@ -118,6 +118,13 @@ object Constant {
 
   val MAX_CHAR_SIZE = 254
 
+  // allowed values for QueryHint.JoinType
+  val JOIN_TYPE_BROADCAST = "broadcast"
+  val JOIN_TYPE_HASH = "hash"
+  val JOIN_TYPE_SORT = "sort"
+  val ALLOWED_JOIN_TYPE_HINTS: List[String] =
+    List(JOIN_TYPE_BROADCAST, JOIN_TYPE_HASH, JOIN_TYPE_SORT)
+
   /**
    * Limit the maximum number of rows in a column batch (applied before
    * [[Property.ColumnBatchSize]] property).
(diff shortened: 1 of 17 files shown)


core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
```diff
@@ -262,15 +272,24 @@ private[sql] object JoinStrategy {
    */
   def canBroadcast(plan: LogicalPlan, conf: SQLConf): Boolean = {
     plan.statistics.isBroadcastable ||
-        plan.statistics.sizeInBytes <= conf.autoBroadcastJoinThreshold
+        plan.statistics.sizeInBytes <= conf.autoBroadcastJoinThreshold || (plan match {
+      case LogicalPlanWithHints(_, hints)
+        if hints.get(QueryHint.JoinType.toString).contains("broadcast") => true
```
kneeraj: hint has to be all lowercase or you want to give the flexibility? Also for "broadcast" and other join types you want to keep some constants?

core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
```diff
@@ -262,15 +272,24 @@ private[sql] object JoinStrategy {
...
   /**
    * Matches a plan whose size is small enough to build a hash table.
    */
   def canBuildLocalHashMap(plan: LogicalPlan, conf: SQLConf): Boolean = {
     plan.statistics.sizeInBytes <= ExternalStoreUtils.sizeAsBytes(
-      Property.HashJoinSize.get(conf), Property.HashJoinSize.name, -1, Long.MaxValue)
+      Property.HashJoinSize.get(conf), Property.HashJoinSize.name, -1, Long.MaxValue) ||
+        (plan match {
+          case LogicalPlanWithHints(_, hints)
+            if hints.get(QueryHint.JoinType.toString).contains("hash") => true
```
kneeraj: See the above comment

core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
```diff
@@ -175,6 +176,19 @@ class SnappyParser(session: SnappySession)
...
   }
 
+  private def updatePerPlanQueryHints(plan: LogicalPlan): LogicalPlan = {
+    if (queryHints.isEmpty) return plan
+    // check for join type hint
+    queryHints.remove(QueryHint.JoinType.toString) match {
+      case null => plan
+      case joinTypeHint => plan match {
+        case LogicalPlanWithHints(child, hints) =>
+          LogicalPlanWithHints(child, hints + (QueryHint.JoinType.toString -> joinTypeHint))
```
kneeraj: How are we controlling the join type hint if multiple joins are there in a complex plan? Please explain a bit.

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -465,6 +465,19 @@ object QueryHint extends Enumeration {
...
+  /**
+   * Query hint to force a join type for the current join. This should appear after
+   * the required table/plan in FROM where the specific join type has to be forced.
+   * Note that this will enable the specific join type only if it is possible
+   * for that table in the join and silently ignore otherwise.
+   *
+   * Possible values are broadcast, hash.
+   *
+   * Example:<br>
+   * SELECT * FROM t1 /`*`+ joinType(broadcast) -- broadcast t1 *`/`, t2 where ...
```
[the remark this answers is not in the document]
sumwale: @jramnara The (-- broadcast t1) portion is just a comment of what it means. The syntax is just:

```SELECT * FROM t1 /*+joinType(broadcast) */, t2```

This is similar to Spark's syntax.

> And, the hint can be placed anywhere in the query string ?

It should be placed after the table/sub-query/plan portion which should be broadcast/hash joined. For example in a multi-join query:

```sql
select ...
from table1
inner join table2 on (table1.c1 = table2.c1) /*+joinType(broadcast) */
inner join table3 on (table2.c2 = table3.c2)
```

In this case the result of table1+table2 will be broadcast. It is not clear how it is possible to make hints position independent for such cases. In general the above positional syntax allows the hint to be applied to any portion of the plan. The next LogicalPlan created by parser will pick up the last hints from the stack, so the above hint can be anywhere after "inner join table2" to the end of clause like above.

> Finally, is there an API equivalent like in Spark 2.2 ?

No. We will integrate in Spark's API after 2.3 merge. Currently there is only one hint in Spark 2.1.1 API: Dataset.broadcast().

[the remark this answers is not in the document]
sumwale: @jramnara More about hints in Spark 2.3. The latest 2.3.2 release still has only one hint namely broadcast. The internal mechanism has changed to use "HintInfo" via Dataset.hint but only one possible value exists as of now and in SQL string it looks like: `select ... from t1 /*+broadcastJoin*/ ...`. This is less general than our "joinType" hint so will stick to our format in the SQL string even after 2.3 merge.

PR #1160, opened 2018-09-21
[SNAP-2568] add support for system tables/VTIs

sumwale:
This set of changes is to allow access to store-side system tables/virtual tables from Spark SQL. Also expanded the "SHOW" statements to include SCHEMAS, COLUMN, MEMBERS as well as expanded "DESCRIBE" statements.

- add support for listing and lookup of system tables and VTIs in SYS schema
  in SnappyStoreHiveCatalog using store connection; only the "non-distributed"
  VTIs that do not need to be routed to all nodes i.e. those implementing
  GfxdVTITemplateNoAllNodesRoute; this only leaves out QUERYSTATISTICS, MEMORYANALYTICS,
  STATEMENTPLANS VTIs (SYS.MEMBERS, DISKSTOREIDS, SYS.JARS have been changed on
  store side) none of which are required on Spark side
- add handling of these tables in RowFormatRelation/RDD that may not have an associated region
  (or if it is, then treat LocalRegion as a replicated region since it is expected to be same on all nodes)
- fixed DSID() function to:
  - work in smart connector mode by using "values dsid()" method on a pooled connection
  - mark as NonDeterminstic so that it does not incorrectly push down beyond joins etc
    like a constant expression (since it has not column dependency otherwise)
  - work in non-code generated mode
- added support for CROSS JOIN, DESCRIBE SCHEMA, DESCRIBE FUNCTION, SHOW SCHEMAS,
  SHOW COLUMN IN table, SHOW TBLPROPERTIES, SHOW MEMBERS
- increased heap size for forked java/scala compilers to 2g to fix occasional OOMEs
- exclude org.glassfish.jersey.server.internal.MappableExceptionWrapperInterceptor from hydra
  suspect strings since this may occasionally appear when jersey server initializes on lead
  emitting the dependent classes in normal course

## Patch testing

precheckin -Pstore -Pspark; one failure on Spark side in StreamingContextSuite "SPARK-18560 Receiver data should be deserialized properly" which happens regardless of these changes and will be tracked separately

## ReleaseNotes.txt changes

Document the common virtual tables and their accessibility from Spark. Document the new clauses.

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/426


sumwale:
Merging now. Will take care of any post-merge reviews comments subsequently.

PR #1163, opened 2018-09-25
Snap 2576

ahshahid:
The query plan if it of type Update | Delete | Delete From Table , the policy filter application to plan is to be skipped.
Fixes SNAP-2576

Added bug test

## ReleaseNotes.txt changes

No change in release notes.
## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
+++ b/core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
@@ -423,7 +423,8 @@ class SnappySessionState(snappySession: SnappySession)
 
     def apply(plan: LogicalPlan): LogicalPlan = {
       plan match {
-        case _: BypassRowLevelSecurity => plan
+        case  (_: BypassRowLevelSecurity | _: Update | _: Delete | _: DeleteFromTable) => plan
+
         // TODO: Asif: Bypass row level security filter apply if the command
         // is of type RunnableCommad. Later if it turns out any data operation
         // is happening via this command we need to handle it
--- a/cluster/src/test/scala/org/apache/spark/sql/policy/PolicyJdbcClientTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/policy/PolicyJdbcClientTest.scala
@@ -203,7 +203,7 @@ class PolicyJdbcClientTest extends PolicyTestBase {
     val stmt = conn.createStatement()
     val conn1 = getConnection(Some("UserX"))
     val stmt1 = conn1.createStatement()
-    stmt.execute(s"CREATE TABLE temp (username String, id Int) " +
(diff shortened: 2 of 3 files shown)


dshirish:
@ahshahid 
Can the user update the columns used in policy filter such that those rows become visible to him? If a user can not see rows due to a policy filter, why is he allowed to delete/update those?

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -423,7 +423,8 @@ class SnappySessionState(snappySession: SnappySession)
 
     def apply(plan: LogicalPlan): LogicalPlan = {
       plan match {
-        case _: BypassRowLevelSecurity => plan
+        case  (_: BypassRowLevelSecurity | _: Update | _: Delete | _: DeleteFromTable) => plan
+
```
dshirish: Do we also need to add PutIntoTable?

ahshahid:
I have to give build today to JnJ so merging the PR

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
ahshahid: Thanks Shirish. You are right we would need PutInto to also be bypassed.
I will add it.

sumwale:
@ahshahid Don't understand this change. Like @dshirish mentioned, shouldn't the policy filter be applied for Update and Delete too? Why should users not be able to view the rows as per policy but still update or delete them? This looks like a security hole to me unless I am missing something. Please update the description with a rationale on why it is ok for Update/Delete to bypass row-level security.

ahshahid:
The policy creation ddl spec has following options for policy 
( policy FOR SELECT| UPDATE| DELETE). In the first stage of implementation we agreed to support only "SELECT".  If you try to create a Policy with option any other than SELECT, it will throw exception.
So as per that  , delete/update/putinto is bypassed.
Once we decide on applying policy for other options ( or may be ALL), this code will be changed accordingly.

sumwale:
@ahshahid Ok, but I think its a big security hole that user is not allowed to view some rows but still able to modify them. In SQL normally it is the case that if read is not allowed on an object then write is also disallowed. So even if we do not have explicit grants for UPDATE/DELETE, we should enforce those to be same as SELECT by default. Once we have explicit settings for UPDATE/DELETE then those will override the SELECT policy.

@jramnara What do you say?

jramnara:
Agree.

ahshahid:
If that is so, I will modify the code . Will file a feature ticket for that.
Then I suppose, Select => will apply to Update & Delete too
                            Update => will apply to Delete too.
                            Delete => Only to itself.

sumwale:
You said Update and Delete are not implemented for RLS policy yet, right? Agree that Update should apply to Delete too once it is implemented (and if Delete policy is not explicitly defined).

PR #1166, opened 2018-09-27
Test Cases for the JDBC Pool Driver.

PradeepSurale:

Added a Scala Tests and DUnit Test for the JDBC Client Pool Driver.

 - By using scala test, covered the functional JDBC test cases
 - By using DUnit test, covered the spark-submit scenario

## Patch testing

This only a test patch.

## ReleaseNotes.txt changes

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/428

### DIFF
--- a/cluster/src/dunit/scala/io/snappydata/cluster/SparkJDBCTestJob.scala
+++ b/cluster/src/dunit/scala/io/snappydata/cluster/SparkJDBCTestJob.scala
@@ -1,0 +1,121 @@
+/*
+ *
+ */
+package io.snappydata.cluster
+
+import java.util.Properties
+
+import org.apache.spark.sql.SparkSession
+
+
+object SparkJDBCTestJob {
+
+  val JDBC_URL = "jdbc:snappydata:pool://localhost:" + System.getProperty("locatorPort") + "/"
+  val DRIVER_NAME = "io.snappydata.jdbc.ClientPoolDriver"
+
+  private def getProperties(): Properties = {
+    val connectionProperties = new Properties()
+    connectionProperties.setProperty("pool-maxTotal", "50")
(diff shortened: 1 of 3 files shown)


cluster/src/dunit/scala/io/snappydata/cluster/SparkJDBCTestJob.scala
```diff
@@ -0,0 +1,121 @@
...
+      connectionProperties = getProperties)
+    df.show()
+  }
+
+
+  def jdbcWriteWithPartitionColumn(spark: SparkSession): Unit = {
+
+    println(" =============== jdbcWriteWithPartitionColumn ============== \n")
+
+    val customerDF = spark.read.parquet(s"/home/pradeep/trunk/snappydata/examples/quickstart/data/airlineParquetData")
```
dshirish: Could you change the path here, this will work on your machine only?

sumwale:
@PradeepSurale @dshirish I have not looked at the test code itself but won't it be much better to add the dunit test case to the existing SplitClusterDUnitTest.testSparkShell which creates the split cluster. The split scenario is what we really need to test while scala tests cover the case of single VM.

@dshirish Also I checked the existing SplitClusterDUnitTest.testSparkShell and that is using spark-shell from SnappyData distribution. This should be changed to use the spark-shell from upstream spark (see SplitClusterDUnitTest.productDir for location of upstream Spark that is downloaded by gradle if required).

PradeepSurale:
Hi @sumwale @dshirish Incorporated the suggested changes.

1. Added the test into the existing one which is SplitClusterDUnitTest.testSparkShell
2. Removed the newly added separate test
3. Corrected the existing  SplitClusterDUnitTest.testSparkShell to use the upstream Spark instead of using spark embedded into snappy.

-

PR #1167, opened 2018-10-01
[SNAP-2582] allow for NONE as a valid policy for server-auth-provider

sumwale:
These set of changes (along with those in store-side PR) allow one to specify server-auth-provider=NONE so that gemfire-level p2p authentication is disabled while the auth-provider will set the connection-level authentication (both for client connections and internal connections).

- allow for NONE as a valid security policy in Lead boot
- fixed and cleaned up lead property handling to set the properties correctly for start and stop
- updated calls from snappy layer to use the authentication service from the Database module
  rather than the static version (which has not been removed on the store side)
- wait for sampler to fully initialiaze before proceeding with hive catalog initialization
- changed tests extending ClusterManagerLDAPTestBase to use server-auth-provider=NONE
  and test that authentication has been disabled at jgroups AUTH layer

## Patch testing

precheckin with the new tests; manual testing

## ReleaseNotes.txt changes

Document the possibility of setting "-server-auth-provider=NONE" to disable peer-to-peer authentication

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/429

### DIFF
--- a/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
+++ b/core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
@@ -633,6 +633,8 @@ public class SnappyHiveCatalog implements ExternalCatalog {
           Constant.JDBC_EMBEDDED_DRIVER());
       initCommonHiveMetaStoreProperties(metadataConf);
 
+      // wait for stats sampler initialization
+      Misc.waitForSamplerInitialization();
       final short numRetries = 40;
       short count = 0;
       while (true) {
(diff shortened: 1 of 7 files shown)


core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
```diff
@@ -633,6 +633,8 @@ private void initHMC() {
           Constant.JDBC_EMBEDDED_DRIVER());
       initCommonHiveMetaStoreProperties(metadataConf);
 
+      // wait for stats sampler initialization
+      Misc.waitForSamplerInitialization();
```
kneeraj: Just curious. Why is this required?

sumwale: As mentioned in other PR comments, saw some stats failures in region initialization so this ensures that stats sampler is up before opening up the catalog for users.

PR #1170, opened 2018-10-04
Create a separate module for pooled JDBC driver

sumwale:

- move SnappyDataClientPoolDialect to separate module snappy-jdbc with the dependencies;
  renamed the class to SnappyDataPoolDialect
- avoid double execution by using the current SparkSession and pre-resolve all tables
  possible by looking up the JDBC metadata using current connection
- copied (and slightly modified) the method to convert JDBC metadata to Spark StructType
  from SNAP-2568
- refactored code a bit in core to enable moving JdbcExtendUtils and dependent classes for above
- removed snappydata-client shadowJar creation and instead use snappydata-jdbc which is a superset
- added extension to DataFrameReader, JdbcReader, to allow for registration of pool driver,
  setup the active SparkSession, and adding host:port like properties automatically if possible
  (from snappydata.connection); also added io.snappydata.implicits to create this class
- renamed GemFireXD*Dialect to SnappyData*Dialect

## Patch testing

precheckin and manual testing

## ReleaseNotes.txt changes

Docs and examples of the pool driver with dialect and implicits

## Other PRs 

jdbc-pool-driver PR on store


cluster/src/dunit/scala/io/snappydata/externalstore/CatalogConsistencyDUnitTest.scala
```diff
@@ -150,8 +150,7 @@ class CatalogConsistencyDUnitTest(s: String) extends ClusterManagerTestBase(s) {
     routeQueryDisabledConn.createStatement().execute("drop table column_table1")
 
     // make sure that the table exists in Hive metastore
-    assert(JdbcExtendedUtils.tableExistsInMetaData("APP.COLUMN_TABLE1",
-      routeQueryDisabledConn, GemFireXDClientDialect))
+    assert(JdbcExtendedUtils.tableExistsInMetaData("APP.COLUMN_TABLE1", routeQueryDisabledConn))
```
kneeraj: Oh the third arg was not even being used in the method.

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -18,168 +18,8 @@ package io.snappydata
 
 import scala.reflect.ClassTag
 
-import com.gemstone.gemfire.internal.shared.SystemProperties
-import io.snappydata.collection.ObjectObjectHashMap
-
-import org.apache.spark.sql.collection.Utils
```
kneeraj: Ok. So you just transferred all these Constants from Literals to Constants and the foldable functions to the BaseParser. Just proper places for defining them or some other reason to move?

core/src/main/scala/org/apache/spark/serializer/serializers.scala
```diff
@@ -208,9 +209,9 @@ object ConnectionPropertiesSerializer
     output.writeString(connProps.url)
     output.writeString(connProps.driver)
     connProps.dialect match {
-      case GemFireXDDialect => output.writeByte(0)
-      case GemFireXDClientDialect => output.writeByte(1)
-      case SnappyDataClientPoolDialect => output.writeByte(3)
+      case SnappyDataDialect => output.writeByte(0)
```
kneeraj: @sumwale a general comment just on the naming. The dialect which the Smart Connector should use theoretically should be the derby/gemxd dialect, specially from the executors, rather only from the executors when scanning tables. Do you think keeping the gemfirexd name for such access should be clearly mentioning that dialect as opposed to a client accessing SnappyData store, which will be primarily spark-sql dialect driven.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -377,10 +378,10 @@ abstract class BaseColumnFormatRelation(
         mode match {
           case SaveMode.Ignore =>
 //            dialect match {
-//              case GemFireXDDialect =>
-//                GemFireXDDialect.initializeTable(table,
+//              case SnappyDataDialect =>
```
kneeraj: Why don't we remove it completely?

jdbc/src/main/scala/org/apache/spark/sql/sources/jdbcExtensions.scala
```diff
@@ -167,26 +192,86 @@ object JdbcExtendedUtils extends Logging {
...
+      val cols = new mutable.ArrayBuffer[StructField]()
+      do {
+        // COLUMN_NAME
+        val columnName = rs.getString(4)
+        // DATA_TYPE
+        val jdbcType = rs.getInt(5)
+        // TYPE_NAME
+        val typeName = rs.getString(6)
+        // COLUMN_SIZE
+        val size = rs.getInt(7)
```
kneeraj: For String type what should be expected of size. Can we rely on this size method?

jdbc/src/main/scala/org/apache/spark/sql/sources/jdbcExtensions.scala
```diff
@@ -167,26 +192,86 @@ object JdbcExtendedUtils extends Logging {
...
+  /**
+   * Is the given JDBC type (java.sql.Types) data type signed.
+   */
+  def isSigned(typeId: Int): Boolean = {
+    typeId == Types.INTEGER || typeId == Types.FLOAT || typeId == Types.DECIMAL ||
+        typeId == Types.SMALLINT || typeId == Types.BIGINT || typeId == Types.TINYINT ||
+        typeId == Types.NUMERIC || typeId == Types.REAL || typeId == Types.DOUBLE
+  }
+
+  def getTableSchema(schemaName: String, tableName: String, conn: Connection): Seq[StructField] = {
```
kneeraj: Yeah, this seemed to be the right thing to do to find out metadata from the standard jdbc apis. What was the reason we did not do this to start with.

jdbc/src/main/scala/org/apache/spark/sql/sources/jdbcExtensions.scala
```diff
@@ -167,26 +192,86 @@ object JdbcExtendedUtils extends Logging {
...
+            } else {
+              getCatalystTypeMethod.invoke(JdbcUtils, Int.box(jdbcType),
+                Int.box(size), Int.box(scale),
+                Boolean.box(isSigned(jdbcType))).asInstanceOf[DataType]
+            }
+        }
+        cols += StructField(columnName, columnType, nullable, metadataBuilder.build())
+      } while (rs.next())
+      cols
+    } else Nil
```
kneeraj: Hmm.. I don't think we can fix the problem of schema changing in the embedded data source while the spark side has built the relation using a previous definition. Wish if that was detected too.

core/src/main/scala/org/apache/spark/serializer/serializers.scala
sumwale: Ok, can change this name to SnappyStoreDialect, SnappyStoreClientDialect and SnappyStorePoolDialect. Removing GemFireXD in the name will be good because we have lots of changes now over the erstwhile GemFireXD SQL support.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
sumwale: At first glance I thought this should be uncommented but has been removed for quite long.

jdbc/src/main/scala/org/apache/spark/sql/sources/jdbcExtensions.scala
sumwale: It should be else its a bug.

jdbc/src/main/scala/org/apache/spark/sql/sources/jdbcExtensions.scala
sumwale: This is required only for the case of pre-resolution of table schema to avoid double execution. Otherwise JDBCRDD.resolveTable uses the schema query (which was the problem) and resolves to StructType using the ResultSetMetadata. Perhaps can remove this and instead use "select * from table limit 1" query instead of getColumns.

jdbc/src/main/scala/org/apache/spark/sql/sources/jdbcExtensions.scala
sumwale: Isn't there the way to throw back exception if table has been altered to clear the cached relations? It should be using the local version field on embedded side and compare against what smart connector knows about. Will be a problem in thin connector though.

core/src/main/scala/org/apache/spark/serializer/serializers.scala
kneeraj: ok.

jdbc/src/main/scala/org/apache/spark/sql/sources/jdbcExtensions.scala
kneeraj: Yeah, was talking about the thin connector. Or the jdbc datasource access too. Because first time spark side will fire the  metadata api to find out the schema and then it will cache the relation I guess.

sumwale: I don't think its a problem with jdbc datasource access in most cases at least the way we are proposing to use with query pushdown. Unless user fires the exact same query multiple times, else new query will lead to creation of a new relation.

PR #1173, opened 2018-10-10
SNAP-2508 : Make condition for replicatedTableJoin stricter while creating HashJoinExec

suranjan:
If child plan is already single partitioned then this doesn't cause exchange.
Otherwise, an exchange will ensure that the child is SinglePartition

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


core/src/main/scala/org/apache/spark/sql/execution/joins/HashJoinExec.scala
```diff
@@ -110,7 +110,8 @@ case class HashJoinExec(leftKeys: Seq[Expression],
 
   override def requiredChildDistribution: Seq[Distribution] =
     if (replicatedTableJoin) {
-      UnspecifiedDistribution :: UnspecifiedDistribution :: Nil
+      AllTuples :: AllTuples :: Nil
```
sumwale: Remove space after Nil

sumwale: This change will cause a major performance regression. ReplicatedTableJoin will be true whenever there is even one replicated table involved. Specifically a partitioned table to replicated table join will now cause partitioned table to be exchanged into a single partition which is incorrect. We should check for the case when both sides are replicated tables, and only then fallback to this. Something like:

```scala
  if (replicatedTableJoin) {
    if (allReplicated) AllTuples :: AllTuples :: Nil
    else UnspecifiedDistribution :: UnspecifiedDistribution :: Nil
  }

```
The "allReplicated" flag will be another flag in the case class which will be set to true by caller HashJoinStrategies for the case when "replicatedTableJoin = true" by also checking canLocalJoin for other side. Like:

```scala
          if (canBuildRight(joinType) && canLocalJoin(right)) {
            makeLocalHashJoin(leftKeys, rightKeys, left, right, condition,
              joinType, joins.BuildRight, replicatedTableJoin = true, allReplicated = canLocalJoin(left))

```

Perhaps will be better to rename this to "isReplicated".

Also I think it will be better to remove the JoinStrategy.isLocalJoin condition check in LinkPartitionsToBuckets because now isLocalJoin does not necessarily mean no exchange (and the method itself can also be removed since that is the only place it is used).

PR #1175, opened 2018-10-11
Snap 2491: Column added using alter table add column through snappy shell does not reflects in spark-shell

dshirish:
1. In smart connector SnappySession, calling SnappyStoreHiveCatalog.registerRelationDestroy() so that cached metadata is invalidated when alter table is executed from connector side.
2. To detect alter table on embedded side, in CodegenSparkFallback, check whether error is reported due to catatlog version mismatch between connector side and embedded side. If yes, clear the metadata cache. Also removed redundent code from CachedDataFrame.scala that was added for similar error checking.
   Note that, after metadata cache is cleared any newly created DF will have latest metadata but the existing DF for which error was thrown will continue to throw error for any SQL queries on it. This can be bugnoted.
3. Other option: instead of these changes just allow operations as right now, that is the select in smart connector side will continue to show old schema and we document about it (user will have to explicitly call session.sessionCatalog.invalidateAll()) to refresh metadata.
4. Other changes: In SnappyStoreHiveCatalog, using active session instead of default as default was not set.
 
## Patch testing
Did the tests as reported in bug.
Currently running precheckin.

## Other PRs


core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
```diff
@@ -75,31 +100,40 @@ case class CodegenSparkFallback(var child: SparkPlan) extends UnaryExecNode {
...
-            } finally {
-              session.sessionState.disableStoreOptimizations = false
-            }
-          case None => throw t
+        if (isConnectorCatalogStaleException(t)) {
+          logWarning(s"SmartConnector catalog is not upto date")
+          val session = sqlContext.sparkSession.asInstanceOf[SnappySession]
+          session.sessionCatalog.invalidateAll()
+          SnappySession.clearAllCache()
+          throw t
```
kneeraj: I think we should throw a specialized exception, like CatalogStaleException, and ask the user to reconstruct the dataframe. 
Also we should try to make it transparent by recreating the CachedDataFrame but I understand the challenges. Please file a task and we will see later how to do this.
We may want to mention in docs about how user can handle the exception and instructions for retry.

core/src/main/scala/org/apache/spark/sql/execution/CodegenSparkFallback.scala
```diff
@@ -75,31 +100,40 @@ case class CodegenSparkFallback(var child: SparkPlan) extends UnaryExecNode {
...
-              val result = f(plan)
-              // update child for future executions
-              child = plan
-              result
-            } finally {
-              session.sessionState.disableStoreOptimizations = false
-            }
-          case None => throw t
+        if (isConnectorCatalogStaleException(t)) {
+          logWarning(s"SmartConnector catalog is not upto date")
```
kneeraj: Just add a check to be sure that it is a smart connector mode.

dshirish:
Ran the precheckin. Incorporated the review comments. Will add a task to check  possibility of transparent by recreating the CachedDataFrame transparently.
Merging the changes. Will take care of any review comments from Sumedh after his review.

PR #1176, opened 2018-10-11
Removing duplicate key events while processing streaming batch in default sink callback 

vatsalmevada:

- Removing multiple records with the same key columns from the incoming batch by grouping record by key columns and picking the last record. This approach may work only when data in the in-coming dataframe batch is partitioned by key columns and the record order is maintained on each partition.
 
- An optional parameter with key `conflation` is exposed for `snappysink` default callback implementation. This parameter controls the conflation behavior.  The valid values for this parameter is `true` or `false`. By default conflation is enabled, it can be disabled by explicitly setting this parameter to `false`.

- If `HashAggregateSize` property of `SnappySession` is set to default then it is set to `10m` (10 MB) before processing the batch and resetting it back to default `0` after processing the batch. This is done to fallback to Spark's aggregation when input data size is > 10 MB and hence spark aggregator will use sort aggregation instead of hash aggregation if output data size is actually large at runtime. This is needed as hash aggreation can be expensive in terms of GC hit if the incoming data batch is large. 

## Patch testing

Manual testing, Scalatest, Hydra test with actual CDC like flow is under progress

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
+++ b/core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
@@ -29,7 +29,6 @@ import org.apache.spark.sql.streaming.DefaultSnappySinkCallback.TEST_FAILBATCH_O
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.{DataFrame, Dataset, Row, SnappySession, _}
 import org.apache.spark.util.Utils
-
 /**
  * Should be implemented by clients who wants to override default behavior provided by
  * [[DefaultSnappySinkCallback]].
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
```diff
@@ -107,7 +108,18 @@ case class SnappyStoreSink(snappySession: SnappySession,
       }
     }
 
-    sinkCallback.process(snappySession, parameters, batchId, convert(data), posDup)
+    val hashAggregateSize = Property.HashAggregateSize.get(snappySession.sessionState.conf)
+    val hashAggregateSizeSetToDefault = hashAggregateSize.equals("0")
```
sumwale: Use Property.HashAggregateSize.defaultValue.get instead of "0". Also the name of variable seems a bit other way. Either like "setHashAggregateSizeToDefault" or "hashAggregateSizeChanged"

sumwale:
Haven't looked at test changes. Other reviewers @kneeraj @suranjan please do.

PR #1177, opened 2018-10-11
adding commandline arg to kill on OOM

hemanthmeka:

Related to jira [SNAP-2503](https://jira.snappydata.io/browse/SNAP-2503). adding commandline arguments to register jvmkill agent to kill jvm on OOM.

## Patch testing

Manually tested on local machine by submitting a job which causes OOM

## Other PRs 

Has dependency on [snappy/store](https://github.com/gemxd/gemfirexd-oss/pull/14).

### DIFF
--- a/launcher/src/main/java/io/snappydata/tools/QuickLauncher.java
+++ b/launcher/src/main/java/io/snappydata/tools/QuickLauncher.java
@@ -124,6 +124,10 @@ class QuickLauncher extends LauncherBase {
     // add java path to command-line first
     commandLine.add(System.getProperty("java.home") + "/bin/java");
     commandLine.add("-server");
+    // https://github.com/airlift/jvmkill is added to libgemfirexd.so
+    // adding agentpath helps kill jvm in case of OOM. kill -9 is not
+    // used as it fails in certain cases
+    commandLine.add("-agentpath:" + snappyHome + "/jars/libgemfirexd.so");
     // get the startup options and command-line arguments (JVM arguments etc)
     HashMap<String, Object> options = getStartOptions(args, snappyHome, commandLine, env);
 


launcher/src/main/java/io/snappydata/tools/QuickLauncher.java
```diff
@@ -124,6 +124,10 @@ private int start(final String[] args) throws IOException, InterruptedException
     // add java path to command-line first
     commandLine.add(System.getProperty("java.home") + "/bin/java");
     commandLine.add("-server");
+    // https://github.com/airlift/jvmkill is added to libgemfirexd.so
+    // adding agentpath helps kill jvm in case of OOM. kill -9 is not
+    // used as it fails in certain cases
+    commandLine.add("-agentpath:" + snappyHome + "/jars/libgemfirexd.so");
```
sumwale: Shouldn't this use either libgemfirexd64.so or libgemfirexd.so depending on architecture? You can check for it using SharedLibrary.is64Bit() or Native.is64Bit().

launcher/src/main/java/io/snappydata/tools/QuickLauncher.java
```diff
@@ -124,6 +124,15 @@ private int start(final String[] args) throws IOException, InterruptedException
     // add java path to command-line first
     commandLine.add(System.getProperty("java.home") + "/bin/java");
     commandLine.add("-server");
+    // https://github.com/airlift/jvmkill is added to libgemfirexd.so
+    // adding agentpath helps kill jvm in case of OOM. kill -9 is not
+    // used as it fails in certain cases
+    String arch = System.getProperty("os.arch");
+    if(arch.contains("64") || arch.contains("s390x")){
```
sumwale: @hemanthmeka Ok, since this is launcher which does not have access to SharedLibrary or Native classes, its better to directly use JNA's com.sun.jna.Platform.is64Bit() rather than os.arch check.

PR #1184, opened 2018-10-17
changes for SNAP-2630 and  SNAP-2625

vatsalmevada:

- deleteFrom clause only requires matching key columns in the dataframe and not the complete schema
 - for both row tables and columns tables deleteFrom will fail if dataframe doesn't contain any column(s) matching the key columns

## Patch testing

Enhanced SnappyTableMutableAPISuite with more test scenarios, precheckin


core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -1153,39 +1153,29 @@ private[sql] final class PreprocessTableInsertOrPut(conf: SQLConf)
...
+
+          val childOutput = try {
+            keyColumns.map(_.toUpperCase)
+                .map(childAttributes(_))
+          } catch {
+            case _: NoSuchElementException =>
+              throw new AnalysisException(s"$l requires that the query in the " +
+                  "WHERE clause of the DELETE FROM statement " +
+                  s"must have all the key column(s) ${keyColumns.mkString(",")} but found " +
+                  s"${child.output.mkString(",")} instead.")
```
sumwale: Will be cleaner to match instead of catching exception like:

```scala
val childOutput = keyColumns.map(col => childAttributes.get(col.toUpperCase) match {
  case Some(a) => a
  case None => throw new AnalysisException(...)
})
```

Better way is to use resolveQuoted -- see comment below.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -1153,39 +1153,29 @@ private[sql] final class PreprocessTableInsertOrPut(conf: SQLConf)
...
-          val expectedOutput = l.output
-          if (child.output.length != expectedOutput.length) {
-            throw new AnalysisException(s"$l requires that the query in the " +
-                "WHERE clause of the DELETE FROM statement " +
-                "generates the same number of column(s) as in its schema but found " +
-                s"${child.output.mkString(",")} instead.")
-          }
-          castAndRenameChildOutputForPut(d, expectedOutput, dr, l, child)
+          val relationAttributes = l.output.map(ar => (ar.name.toUpperCase(), ar)).toMap
+          val expectedOutput = keyColumns.map(_.toUpperCase).map(relationAttributes(_))
```
sumwale: The standard way to resolve column names among Attributes both here and above is using LogicalPlan.resolveQuoted like:
```scala
val expectedOutput = keyColumns.map(col => l.resolveQuoted(col, analysis.caseInsensitiveResolution) match {
  case Some(a: Attribute) => a
  case _ => throw ...
})
```

(here analysis is package object in Spark's org.apache.spark.sql.catalyst)
Similarly for child.resolveQuoted() above.

sumwale: To clarify: resolveQuoted matches against the entire tree so returns Option[NamedExpression] which will be an alias in case the matching column is somewhere below or an expression alias itself. However, we need it to be always a direct attribute and hence the match against Some(a: Attribute).

PR #1185, opened 2018-10-17
SNAP-2602 : On snappy UI, add column named "Overflown Size"/ "Disk Size" in Tables.

snappy-sachin:

Feature Request:
 - On snappy UI, add column named "Overflown Size / Disk Size" in Tables, to show how much data is overflown to disk.

Changes:
 - HTML and CSS changes for displaying tables overflown size to disk on UI.
 - TableSummary is updated to hold tables size on disk.
 - SnappyTableStatsProviderDUnitTest is updated accordingly.

(Fill in the changes here)

## Patch testing

 - Tested Manually.

## ReleaseNotes.txt changes

 - YES

## Other PRs 

 - Spark: https://github.com/SnappyDataInc/spark/pull/127
 - Store: https://github.com/SnappyDataInc/snappy-store/pull/440

### DIFF
--- a/cluster/src/dunit/scala/io/snappydata/cluster/SnappyTableStatsProviderDUnitTest.scala
+++ b/cluster/src/dunit/scala/io/snappydata/cluster/SnappyTableStatsProviderDUnitTest.scala
@@ -214,6 +214,7 @@ object SnappyTableStatsProviderDUnitTest {
     entryOverhead *= entryCount
     stats.setSizeInMemory(stats.getSizeInMemory + memSize + entryOverhead)
     stats.setTotalSize(stats.getTotalSize + totalSize + entryOverhead)
+    stats.setSizeOnDisk(stats.getTotalSize - stats.getSizeInMemory)
     stats
   }
 
@@ -249,6 +250,7 @@ object SnappyTableStatsProviderDUnitTest {
     totalSize += overhead * result.getRowCount
     result.setSizeInMemory(totalSize)
     result.setTotalSize(totalSize)
+    result.setSizeOnDisk(0)
     result
   }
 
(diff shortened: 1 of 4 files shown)


snappy-sachin:
> Please change the name to "on Disk size" or "spillover to disk size" ... avoid using 'overflown'.

I have used _Disk Size_ as a column name with tooltip _Tables Size on Disk_. Hope that is fine.
[Click Here](https://jira.snappydata.io/secure/attachment/12401/Table-DiskSize-Column.png) for actual UI component snapshot.

jramnara:
@snappy-sachin here is the thing. The correct value for disk-size is quite different. Given opLogs and lazy compaction. From what I call, you are simply subtracting the memory size from the total size, right? Also, irrespective of overflow, the entire data is always on disk. 
So, maybe, this should be 'Spill to disk size'.

snappy-sachin:
> @snappy-sachin here is the thing. The correct value for disk-size is quite different. Given opLogs and lazy compaction. From what I call, you are simply subtracting the memory size from the total size, right? Also, irrespective of overflow, the entire data is always on disk.
> So, maybe, this should be 'Spill to disk size'.

OK Jags, I will rename the column name to "Spill-To-Disk Size". Is that fine?

snappy-sachin:
> > @snappy-sachin here is the thing. The correct value for disk-size is quite different. Given opLogs and lazy compaction. From what I call, you are simply subtracting the memory size from the total size, right? Also, irrespective of overflow, the entire data is always on disk.
> > So, maybe, this should be 'Spill to disk size'.
> 
> OK Jags, I will rename the column name to "Spill-To-Disk Size". Is that fine?

Done, 
_**Disk Size**_ is renamed to _**Spill-To-Disk Size**_ and _**Memory Size**_ is renamed to _**In-Memory Size**_.

PR #1186, opened 2018-10-23
Fix for SNAP-2645:

vatsalmevada:

Making startingOffsets property optional as spark's streaming API takes care of that by using committed offset or kafka param "auto.offset.reset" if no explicit value is set for startingOffsets

## Patch testing

Updated SnappyStreamingKafkaSuite test scenario to test streaming without passing "startingOffsets" param

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
+++ b/core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
@@ -23,6 +23,7 @@ import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.encoders.RowEncoder
 import org.apache.spark.sql.sources.{BaseRelation, DataSourceRegister}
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/streaming/DirectKafkaStreamSource.scala
```diff
@@ -65,8 +68,13 @@ final class DirectKafkaStreamRelation(
   private val startingOffsets = getStartingOffsets
 
   private def getStartingOffsets = {
-    DirectKafkaStreamRelation.partitionOffsetMethod.invoke(null, options("startingOffsets"))
-        .asInstanceOf[Map[TopicPartition, Long]]
+
+    if (options.contains(STARTING_OFFSETS_PROP)) {
+      partitionOffsetMethod.invoke(null, options(STARTING_OFFSETS_PROP))
+          .asInstanceOf[Map[TopicPartition, Long]]
+    } else {
```
sumwale: I prefer using a single lookup in maps instead of contains+apply (esp since this is case-insensitive map so has even more operations). Though perf does not matter here, but a good practise nevertheless leading to cleaner pattern matching and to avoid cases where one may miss that it can have some perf impact:

```scala
options.get(STARTING_OFFSETS_PROP) match {
  case Some(offsets) => ...
  case None => ...
}
```

PR #1187, opened 2018-10-24
Snap 2503 kill on OOM 

hemanthmeka:

changes related to [SNAP-2503](https://jira.snappydata.io/browse/SNAP-2503).
1: Enabling gemfirexd lib copy to product jars folder in case of mac/linux
2: adding agent registration for all services to kill on OOM

## Patch testing

tested locally on ubuntu14.04, centos6 and mac

## ReleaseNotes.txt changes

No

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/442
https://github.com/SnappyDataInc/snappy-aqp/pull/173


launcher/src/main/java/io/snappydata/tools/QuickLauncher.java
```diff
@@ -134,6 +134,12 @@ private int start(final String[] args) throws IOException, InterruptedException
       } else {
         commandLine.add("-agentpath:" + snappyHome + "/jars/libgemfirexd.so");
       }
+    }else if(Platform.isMac()){
```
sumwale: spaces after/before braces and after if

launcher/src/main/java/io/snappydata/tools/QuickLauncher.java
```diff
@@ -125,15 +125,36 @@ private int start(final String[] args) throws IOException, InterruptedException
...
+        } else {
+          String agentPath = snappyHome + "/jars/libgemfirexd.dylib";
+          System.load(agentPath);
+          commandLine.add("-agentpath:" + agentPath);
+        }
       }
+    } catch (UnsatisfiedLinkError e) {
+      System.out.println("WARNING: agent not loaded. service might not be killed on OutOfMemory");
+    }catch (SecurityException e){
+      System.out.println("WARNING: agent not loaded. service might not be killed on OutOfMemory");
```
sumwale: Java8 allows combining common code for exceptions. Also you could print out the exception for information:

```java
} catch (UnsatisfiedLinkError | SecurityException e) {
  System.out.println("WARNING: agent not loaded due to " + e +
     ". Service might not be killed on OutOfMemory. Build jvmkill.c on your platform using build.sh script" +
     " from source on your platform and replacing the library in product jars directory to enable the agent.");
}
```

gradle.properties
```diff
@@ -10,3 +10,4 @@ ossrhUsername=
 ossrhPassword=
 archivaUsername=
 archivaPassword=
+snappydata.enterprise=true
```
sumwale: @hemanthmeka this is not correct. We should always keep the project gradle.properties as OSS. Add in your ~/.gradle/gradle.properties if required.

hemanthmeka: didn't notice this being pushed while merging. removed this line in new commit. Thanks

PR #1189, opened 2018-10-27
[SNAP-2656] check for underlying Attribute with joins on aggregate columns

sumwale:
For queries like:

```sql
  select * from t1 join t2 on (t1.c1 = t2.c1)
  where t1.c2 = (select max(c2) from t2 where t1.c1 = t2.c1)
```

when t1, t2 are partitioned on c1, should not result in a shuffle. This PR adds a change to
determine underlying attributes across aggregate projection alias columns.

- drill down into the underlying attribute beyond aggregate plans when checking for
  partitioning columns matching with join keys
- skip join hint for hash/broadcast in case estimated size is too large because
  it can result in OOME with warning that session properties need to be set explicitly

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


sumwale:

PR #1190, opened 2018-10-28
[SNAP-2657] added caching for hive catalog lookups

sumwale:
Meta-data queries with large number of tables take quite long because of nested loop joins between
SYSTABLES and HIVETABLES for most meta-data queries. Additionally each row in SYSTABLES
looks up hive-metastore separately to determine if the table is a column table or not. Overall this
results in close to a million hive-metastore lookups for each meta-data query when there are tables
in the range of hundreds.

The changes below add caching for hive meta-store lookups (with invalidation) which improves the
performance by nearly two orders of magnitude (though the design itself can use some improvement).

- added caching in SnappyHiveCatalog lookups to improve meta-data lookup performance
  for large number of tables
- separate bounded caches for successful and unsucessful lookups; latter are kept separate
  to not push out former if large (unsuccessful lookups are common for internal column store regions)
- added a new RefreshMetadata function for meta-data changes on all executors; this replaces
  most uses of Utils.mapExecutors in the product to avoid dependency on Spark scheduling and also
  because mapExecutors can fail with timeout in case scheduling never gets a core whereas
  GemFire Function execution will keep trying until it goes through if a node is part of the cluster
- added invalidation of the new cache in SnappyHiveCatalog as part of sendRegisterRelationDestroy
  and added appropriate calls to new table/policy creation to invoke it
- also fixed cases where PartitionedRegion.needsBatching may have been cached incorrectly before
  the column store for a region was created
- removed isRowTable/isColumnTable hive operations from SnappyHiveCatalog and instead
  determine directly from HiveTable already looked up (or use GET_TABLE if not)
- cleaned up SNAPPY_URL_PREFIX and JDBC_URL_PREFIX handling removing latter which was
  incorrectly named
- fixed some scalastyle warnings in LeadImpl and few others

## Patch testing

precheckin; manual with large number of tables (700)

## ReleaseNotes.txt changes

NA

## Other PRs 

related store PR https://github.com/SnappyDataInc/snappy-store/pull/444


core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
```diff
@@ -67,13 +71,30 @@
...
   private final ExecutorService hmsQueriesExecutorService;
 
   public SnappyHiveCatalog() {
+    this.nonExistentTables = CacheBuilder.newBuilder().maximumSize(2000).build();
+    this.tableCache = CacheBuilder.newBuilder().maximumSize(2000).build(
+        new CacheLoader<Map.Entry<String, String>, Table>() {
+          @Override
+          public Table load(@Nonnull Map.Entry<String, String> table) throws Exception {
+            Hive client = Hive.get();
+            return client.getTable(table.getKey(), table.getValue());
```
kneeraj: The first lookup will bear the cost and subsequent is fast. Good. But for demo purpose etc we should have run few queries beforehand otherwise the same latency will show up if used first time. Just a note.

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
```diff
@@ -180,18 +201,19 @@ public Table getTable(String schema, String tableName, boolean skipLocks) {
...
   }
 
   @Override
-  public boolean isRowTable(String schema, String tableName, boolean skipLocks) {
-    HMSQuery q = getHMSQuery();
-    q.resetValues(HMSQuery.ISROWTABLE_QUERY, tableName, schema, skipLocks);
-    Future<Object> f = this.hmsQueriesExecutorService.submit(q);
-    return (Boolean)handleFutureResult(f);
+  public boolean isColumnTable(Object hiveTable) {
+    return !ExternalTableType.getTableType((Table)hiveTable).equalsIgnoreCase(
```
kneeraj: What happens if isColumnTable is called for an external table. !Row will make it column?

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
```diff
@@ -263,8 +285,34 @@ public String catalogSchemaName() {
...
+      AbstractMap.SimpleEntry<String, String> key = new AbstractMap.SimpleEntry<>(
+          Utils.toUpperCase(schemaName), Utils.toUpperCase(tableName));
+      this.nonExistentTables.invalidate(key);
+      this.tableCache.invalidate(key);
+      // also clear "isRowBuffer" since it may have been cached incorrectly
+      // when column store was still being created
+      Region<?, ?> region = Misc.getRegion(Misc.getRegionPath(key.getKey(),
+          key.getValue(), null), false, true);
+      if (region instanceof PartitionedRegion) {
+        ((PartitionedRegion)region).clearIsRowBuffer();
```
kneeraj: When is the isRowBuffer set BTW?

cluster/src/main/scala/org/apache/spark/scheduler/cluster/SnappyCoarseGrainedSchedulerBackend.scala
```diff
@@ -113,11 +113,8 @@ class BlockManagerIdListener(sc: SparkContext)
...
     SnappyContext.clearBlockIds()
 
-  private def handleNewExecutorJoin(bid: BlockManagerId) = {
+  private def handleNewExecutorJoin(bid: BlockManagerId): Unit = {
     val uris = SnappySession.getJarURIs
-    Utils.mapExecutors[Unit](sc, () => {
-      ToolsCallbackInit.toolsCallback.addURIsToExecutorClassLoader(uris)
-      Iterator.empty
-    }, 30, Seq(bid))
+    RefreshMetadata.executeOnAllEmbedded(RefreshMetadata.ADD_URIS_TO_CLASSLOADER, uris)
```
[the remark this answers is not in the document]
sumwale: This is invoked in onExecutorAdded callback.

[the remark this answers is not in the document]
sumwale: Ok, you mean the previous Seq(bid) to send to only bid. Will revert this one.

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
sumwale: I didn't see any significant slowdown even in first explicit call. Likely due to internal calls by stats provider etc.

cluster/src/main/scala/org/apache/spark/scheduler/cluster/SnappyCoarseGrainedSchedulerBackend.scala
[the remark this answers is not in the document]
kneeraj: Yes.

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
sumwale: Yeah, not sure why this was done in previous code. Tracking it down, this should be true for column and sample, but since earlier we had only row, column and sample so perhaps this made sense. Will change to be column+sample.

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
sumwale: Its initialised lazily on first access. See store PR: https://github.com/SnappyDataInc/snappy-store/pull/444/files#diff-0bd50a396f29b0cd9abd2865f498a5b2L2526

That was a bug because it could access before child column store regions are created then it will be permanently wrong.

sumwale: I saw the problem in manual testing while manually querying systables while tables were being created. Then the tableType would go permanently wrong for few tables, but the same could happen due to internal calls too and user will probably not notice (except that rollover, snapshot would no longer happen).

core/src/main/java/io/snappydata/impl/SnappyHiveCatalog.java
sumwale: Column+sample+index

PR #1196, opened 2018-11-01
temporary fix for ensuring aqp bootstrap code does not copy the buffe…

ahshahid:
…r variable in snappyhashaggregate exec

Temporarily disabling the copy of the buffer variables ( for closed form as well as bootstrap)  if the aqp aggregate functions are present ( & bootstrap or closed form analysis is needed).

This change  will be rectified properly after understanding the reason for copying buffer variable. The bootstrap code, especially relies on the holder object ( the buffer variable) reference check to decide whether the group by key has changed or not.

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 


sumwale:
Closing this as AQP has a better fix now that works fine.

PR #1198, opened 2018-11-06
[SNAP-2462] enable common-subexpression elimination for ParamLiterals

sumwale:
ParamLiterals are not constants, so common-subexpression elimination phases of optimizer
may fail when values inside them are actually equal.

This adds a new RefParamLiteral that points to a ParamLiteral which is used when a constant value is
found that is same as one found before.
The hashCode mirrors that of original ParamLiteral while equals is reference equality with original.
This enables a ParamLiteral to be equal to a RefParamLiteral in a different position.

When considering two different plans for equality, the RefParamLiterals have to be in exactly same
positions to be evaluated as equal since RefParamLiterals are never equal to other ParamLiterals.
For two RefParamLiterals, they are equal iff their original ParamLiterals are.

This allows expressions like "a = 4 and b = 4" and "a = 6 and b = 6" to be equal after tokenization
in plan caching, but not "a = 4 and b = 4" and "a = 6 and b = 5" which will have different plans.

Also added a test for above including check for filter push down.

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


sumwale:
Merging. Will take care of any review comments subsequently.

PR #1201, opened 2018-11-15
[SNAP 2634] - providing snappy data specific strategies in IncrementalExecution

vatsalmevada:

Providing snappy data specific strategies in IncrementalExecution (which is used for streaming queries). As part of these changes, we have removed Snappy's DefaultPlanner and instead of passing Snappydata specific strategies as part of `experimentalMethods.extraStrategies` of `SnappySessionState`.

Also `SnappyAggregationStrategy` is disabled for stream queries as it is clashing with `StatefulAggregationStrategy` which is added by Spark as part of `IncrementalExecution`. This is achieved by setting `snappydata.sql.hashAggregateSize` to `-1` in `SnappySession`'s configuration.
This implies that Snappydata aggregation optimization will be turned off for any usage of that particular `SnappySession` even for non-streaming queries.

## Patch testing

- Manual testing
- scalatest
- precheckin

## Other PRs

https://github.com/SnappyDataInc/snappy-aqp/pull/176


core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
```diff
@@ -124,7 +122,8 @@ private[sql] trait SnappyStrategies {
           s"Expected one of ${Constant.ALLOWED_JOIN_TYPE_HINTS}")
     }
 
-    def apply(plan: LogicalPlan): Seq[SparkPlan] = if (isDisabled || session.disableHashJoin) {
+    def apply(plan: LogicalPlan): Seq[SparkPlan] = if (isDisabled ||
+        snappySession.disableHashJoin) {
```
sumwale: unnecessary line break or was it exceeding 100 chars?

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -52,25 +53,40 @@ import org.apache.spark.sql.internal.SQLConf.SQLConfigBuilder
...
+    case BypassRowLevelSecurity(child) => PlanLater(child) :: Nil
+    case _ => Nil
+  }
+
   protected lazy val snappySharedState: SnappySharedState = snappySession.sharedState
 
   private[internal] lazy val metadataHive = snappySharedState.metadataHive().newSession()
 
+  override lazy val streamingQueryManager: StreamingQueryManager = {
+    // Disabling `SnappyAggrerateStrategy` for streaming queries as it clashes with
```
sumwale: typo in SnappyAggrerateStrategy

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -52,25 +53,40 @@ import org.apache.spark.sql.internal.SQLConf.SQLConfigBuilder
...
+    // `StatefulAggregationStrategy` which is applied by spark for streaming queries. This
+    // implies that Snappydata aggregation optimisation will be turned off for any usage of
+    // this session including non-streaming queries.
+
+    HashAggregateSize.set(snappySession.sessionState.conf, "-1")
+    new StreamingQueryManager(snappySession)
```
sumwale: Will be better to call super.streamingQueryManager

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -701,13 +714,22 @@ class SnappySessionState(snappySession: SnappySession)
...
   override final def executePlan(plan: LogicalPlan): QueryExecution = {
+    ingestSnappyStrategies
     clearExecutionData()
     beforeExecutePlan(plan)
     val qe = newQueryExecution(plan)
     if (enableExecutionCache) executionCache.put(plan, qe)
     qe
   }
 
+  private lazy val ingestSnappyStrategies = {
```
sumwale: better to just name "snappyStrategies"

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
vatsalmevada: I tried "super.streamingQueryManager" earlier. However, it fails at compile time stating "super may not be used on lazy value streamingQueryManager". Based on my brief analysis it is possibly due to https://issues.scala-lang.org/browse/SI-1938.

core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
vatsalmevada: it was exceeding 100 chars.

PR #1205, opened 2018-11-26
SNAP-2719 - falling back to uncached SQL flow for streaming queries

vatsalmevada:

- Falling back to uncached SQL flow for streaming queries.
- Disabling `planCaching` by default except for JDBC flow
- Removing `planCachingAll` property as this propery was merely cleaning plan cache when set to `false`. Setting this propery to `true` again had no effect at all. This hehavor was a bit confising to the name of the property and after setting default `planCaching` to `false` this property wasn't adding much value.
- Changing unit test frameworks to randomly change `planCaching` value for different tests so that queries with or without plan caching will be tested eventually without adding extra test execution time

Changes from https://github.com/SnappyDataInc/snappydata/pull/1119 are also merged with this PR as they are related.

## Patch testing

- `SnappyStreamSuite#"sql queries"` test is passing which was failing earlier (and was ignored)
- precheckin


sumwale:
I think we had discussed that it will be better to change the Property.PlanCaching to false by default on all sessions other than those coming from JDBC. For now it can be done explicitly on the streaming session because doing on all will likely require many test changes that assume otherwise. The "sqlUncached" API can remain as internal, undocumented one.

cluster/src/main/scala/io/snappydata/gemxd/SparkSQLExecuteImpl.scala
```diff
@@ -482,6 +482,7 @@ object SnappySessionPerConnection {
     if (session != null) session
     else {
       val session = SnappyContext().snappySession
```
sumwale: For corner cases, this should better check that SparkContext is not set instead of throwing NPE like:
```scala
val session = SparkContext.globalSparkContext match {
   // use a CancelException to force failover by client to another lead if available
  case null => throw new CacheClosedException("No SparkContext ...")
  case sc => new SnappySession(sc)
}
```

cluster/src/test/scala/org/apache/spark/sql/NorthWindTest.scala
```diff
@@ -57,6 +57,30 @@ class NorthWindTest
     validatePartitionedColumnTableQueries(snc)
   }
 
+  // enable if transformations are supported in plan-caching.
+  ignore("SNAP-2451"){
+    createAndLoadColumnTables(snc)
```
sumwale: I think this test should be enabled. Do you see failure in this?

core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -190,7 +190,7 @@ object Property extends Enumeration {
 
   val PlanCaching: SQLValue[Boolean] = SQLVal[Boolean](
     s"${Constant.PROPERTY_PREFIX}sql.planCaching",
-    "Property to set/unset plan caching", Some(true))
+    "Property to set/unset plan caching", Some(false))
```
sumwale: This change needs to be documented. And recommendations for users under which conditions should planCaching be used, where it is useful etc.

core/src/test/scala/io/snappydata/core/LocalTestData.scala
```diff
@@ -81,6 +81,7 @@ object LocalSparkConf {
     val conf = new SparkConf()
         .setIfMissing("spark.master", "local[4]")
         .setIfMissing("spark.memory.debugFill", "true")
+        .setIfMissing("snappydata.sql.planCaching", "true")
```
sumwale: I think you will also need to set it for dunit tests -- see bootProps/sysProps in ClusterManagerTestBase

cluster/src/test/scala/org/apache/spark/sql/NorthWindTest.scala
vatsalmevada: This test is failing when planCaching is enabled.

core/src/main/scala/io/snappydata/Literals.scala
vatsalmevada: A documentation task is added for this https://jira.snappydata.io/browse/SNAP-2855

cluster/src/test/scala/org/apache/spark/sql/SnappySQLQuerySuite.scala
```diff
@@ -433,6 +434,9 @@ class SnappySQLQuerySuite extends SnappyFunSuite {
 
   test("Push down TPCH Q19") {
     session.sql("set spark.sql.autoBroadcastJoinThreshold=-1")
+    session.sql("set snappydata.sql.planCaching=true").collect()
```
suranjan: Always setting true?

core/src/test/scala/org/apache/spark/sql/store/DisableTokenizationTest.scala
```diff
@@ -33,14 +34,18 @@ class DisableTokenizationTest
         with BeforeAndAfterAll {
 
   val table = "my_table"
+  var planCaching : Boolean = false
 
   override def beforeAll(): Unit = {
     snc.sql(s"set snappydata.sql.tokenize = false")
+    planCaching = Property.PlanCaching.get(snc.sessionState.conf)
+    Property.PlanCaching.set(snc.sessionState.conf, true)
```
suranjan: Setting true for some reason?

core/src/test/scala/org/apache/spark/sql/store/TokenizationTest.scala
```diff
@@ -39,18 +39,22 @@ class TokenizationTest
...
   val table2 = "my_table2"
   val all_typetable = "my_table3"
+  var planCaching : Boolean = false
 
   override def beforeAll(): Unit = {
     // System.setProperty("org.codehaus.janino.source_debugging.enable", "true")
     System.setProperty("spark.sql.codegen.comments", "true")
     System.setProperty("spark.testing", "true")
+    planCaching = Property.PlanCaching.get(snc.sessionState.conf)
+    Property.PlanCaching.set(snc.sessionState.conf, true)
```
suranjan: Same, changing to true for all tests?

cluster/src/test/scala/org/apache/spark/sql/SnappySQLQuerySuite.scala
vatsalmevada: yes, that is done because as part of `testTPCHQ19` method we are asserting analyzed plan which works only with `planCaching` enabled.

core/src/test/scala/org/apache/spark/sql/store/TokenizationTest.scala
vatsalmevada: Several tests from this suite is having assertions similar to this which starts failing when planCaching disabled:
```
val cacheMap = SnappySession.getPlanCache.asMap()
assert(cacheMap.size() == 1
```

PR #1209, opened 2018-11-29
SNAP-2745 - processing all delete events prior to conflating the dataframe

vatsalmevada:

processing all delete events prior to conflating the dataframe as skipping processing delete events might lead to extraneous records in the column table and will cause application failure in case of row tables due to the primary key constraint violation.

For example, if the target table contains some record with key k1 and the next incoming batch contains following a delete event followed by an insert event then due to conflation the delete event will be skipped and only insert will be processed. Since the existing record didn't get deleted, this will result into an extra record in case of column table or will cause failure due to unique constraint violation for row table.

## Patch testing

- existing tests in SnappyStoreSinkProviderSuite are passing and also added new tests in the same suite

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
+++ b/core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
@@ -163,11 +163,12 @@ class DefaultSnappySinkCallback extends SnappySinkCallback {
         s", eventTypeColumnAvailable:$eventTypeColumnAvailable,possible duplicate: $posDup")
 
     if (keyColumns.nonEmpty) {
-      val dataFrame: DataFrame = if (conflationEnabled) getConflatedDf else df
+      val dataFrame: DataFrame = if (conflationEnabled) {
+        groupByKeyColumnsAndGetLast(df).cache()
+      } else df
       if (eventTypeColumnAvailable) {
-        val deleteDf = dataFrame.filter(dataFrame(EVENT_TYPE_COLUMN) === EventType.DELETE)
-            .drop(EVENT_TYPE_COLUMN)
-        deleteDf.write.deleteFrom(tableName)
+        // processing all deletes prior to conflating to cater SNAP-2745
+        processDeletes()
         if (posDup) {
           val upsertEventTypes = List(EventType.INSERT, EventType.UPDATE)
           val upsertDf = dataFrame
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
```diff
@@ -148,7 +148,6 @@ import org.apache.spark.sql.snappy._
 class DefaultSnappySinkCallback extends SnappySinkCallback {
```
sumwale: hashAggregateSizeChanged name at line 111 is confusing. Should be "hashAggregateSizeIsDefault"

core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
```diff
@@ -218,11 +217,20 @@ class DefaultSnappySinkCallback extends SnappySinkCallback {
...
+        // if event type of the last event for a key is insert and there are more than one
+        // events for the same key, then convert inserts to put into
+        val columns = df.columns.filter(_ != EVENT_TYPE_COLUMN).map(col) ++
+            Seq(when(col(EVENT_TYPE_COLUMN) === EventType.INSERT && col(EVENT_COUNT_COLUMN) > 1,
+              EventType.UPDATE).otherwise(col(EVENT_TYPE_COLUMN)).alias(EVENT_TYPE_COLUMN))
+
         df.groupBy(keyCols.head, keyCols.tail: _*)
             .agg(exprs.head, exprs.tail: _*)
+            .select(columns: _*)
             .select(df.columns.head, df.columns.tail: _*)
```
sumwale: Is the second select() required given that all required columns have already been projected in first?

core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
```diff
@@ -163,6 +162,7 @@ class DefaultSnappySinkCallback extends SnappySinkCallback {
         s", eventTypeColumnAvailable:$eventTypeColumnAvailable,possible duplicate: $posDup")
 
     if (keyColumns.nonEmpty) {
+      df.cache().count()
```
sumwale: There is a way to avoid creating unnecessary tasks in the SQL tab. See SnappyCacheTableCommand:

```scala
        val memoryPlan = df.queryExecution.executedPlan.collectFirst {
          case plan: InMemoryTableScanExec => plan.relation
        }.get
        memoryPlan.cachedColumnBuffers.count()
```

On the other hand if we do want to show the incoming DF plan that got cached, then the df.cache().count() won't show it rather shows only the count aggregate plan on InMemoryRelation after it has been created. So I think it will be better to refactor SnappyCacheTableCommand from the portion after DF is created in first few lines, and use here which will show the incoming DF plan that matters and also take care of using off-heap for the cache if configured.

vatsalmevada: Tracked by https://jira.snappydata.io/browse/SNAP-2761.

PR #1212, opened 2018-12-05
RefParamLiteral being checked for equality against another RefParamLiteral

ahshahid:
RefParamLiteral of RefParamLiteral was being created so referenceEquals was returning false.
Correct the search to ensure that RefParamLiterals are skipped and also added an assertion that
RefParamLiteral should never wrap another RefParamLiteral.

Added a test with views to check above.

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
@@ -462,6 +462,7 @@ class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
         " b.longtext "
     snappy.sql(q)
     snappy.sql(s" CREATE OR REPLACE VIEW v2 as $q")
+    snappy.sql("select count(*) from v2").collect()
 
     stmt.execute("drop view v3")
     stmt.execute("drop view v2")
--- a/cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
@@ -422,4 +422,57 @@ class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
     conn.close()
     TestUtil.stopNetServer()
   }
+
+  test("Bug SNAP-2758 . view containing aggregate function & join throws error") {
+    snc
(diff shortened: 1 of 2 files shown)


sumwale:
I think the semanticEquals addition is fine, but the other one that adds equality based on ParamLiterals is not correct. The issue is the following:

ParamLiterals are equal based on their position. So two plans having same type of ParamLiterals in respective positions will match (assuming rest of LogicalPlan matches). This allows the second plan to use the same physical plan as first. Since the actual value contained inside them can be different, so they do not match by value (at least once they are marked as "tokenized") and are not foldable like Literals. Only with these conditions will the physical plans for two tokenized plans match iff the parsed plans match.

RefParamLiterals are to exploit common sub-expression optimizations. If two ParamLiterals have same value in parsed plan, then second one is turned into a RefParamLiteral pointing to first. It depends on the fact that values during parsing were the same.

For example an optimization rule might change (hypothetical example because this actually will get constant folded):
select cast(4 as decimal), c * cast(4 as decimal)
to
select cast(4 as decimal) e1, c * e1. 

This depends on the fact that the constant "4" matched in both projections. Thus the plan for:
"select cast(4 as decimal), c * cast(5 as decimal)" should be different and not match above because it will not reduce to the second form. If ParamLiterals contained inside were to be matched then the tokenized version of the two plans would be the same: select cast(? as decimal), c * cast(? as decimal) and would be equal. Perhaps a unit test to check this should be added (maybe already there in TokenizationTest).

Will look at the view problem later today.

sumwale:
@ahshahid The issue was that RefParamLiteral of RefParamLiteral was being created so referenceEquals was returning false. Corrected the search to ensure that RefParamLiterals are skipped and also added an assertion that RefParamLiteral should never wrap another RefParamLiteral. Have reverted the previous code changes. The new test now passes as well as existing TokenizationTest (that failed with previous changes which were not correct -- please see the previous comment on the rationale and semantics of RefParamLiteral).

Please verify precheckin and merge.

sumwale:
@ahshahid I think you meant to add some JDBC queries in the new test? Will be good to enhance the test with JDBC queries and also some data with result verification.

sumwale:
@ahshahid updated the title and description of this PR accordingly

PR #1213, opened 2018-12-05
Fixes for issues seen in different putInto tests

sumwale:

- switched all intermediate arrays in decoders/encoders to use BufferAllocator so that it can
  be accounted propertly (only in off-heap config for now)
- added proper close methods everywhere for both encoders and decoders for above and invoke
  the close in ColumnTableScan generated code
- force a repartition on the implicit BATCHID column in column update/delete when there is
  no explicit partitioning in the table so that all changes for a batch are together and get
  applied as a single delta rather than spread across all partitions
- use KEY_COLUMNS as partitioning columns if none have been specified expliciitly
- use SnappyCacheTableCommand for putInto intermediate cache to show the plan of cache properly
- moved task interrupt checks to both column and row iterators so that task execution can
  stop quickly; removed the check from ColumnTableScan generated code

## Patch testing

precheckin in progress

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/447

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnDelta.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnDelta.scala
@@ -98,12 +98,16 @@ final class ColumnDelta extends ColumnFormatValue with Delta {
           // ignore if either of the buffers is empty (old placeholder of 4 bytes
           // while UnsafeRow based data can never be less than 8 bytes)
           if (existingBuffer.limit() <= 4 || newBuffer.limit() <= 4) {
-            oldColValue
+            // returned value should be the original and not oldColValue which may be a temp copy
+            oldValue
           } else {
             val merged = mergeStats(existingBuffer, newBuffer, schema)
             if (merged ne null) {
               new ColumnFormatValue(merged, oldColValue.compressionCodecId, isCompressed = false)
-            } else oldColValue
+            } else {
+              // returned value should be the original and not oldColValue which may be a temp copy
+              oldValue
+            }
           }
         } else {
           val tableColumnIndex = ColumnDelta.tableColumnIndex(columnIndex) - 1
--- a/cluster/build.gradle
+++ b/cluster/build.gradle
@@ -34,6 +34,31 @@ dependencies {
   compile 'org.slf4j:jcl-over-slf4j:' + slf4jVersion
   compile 'org.slf4j:jul-to-slf4j:' + slf4jVersion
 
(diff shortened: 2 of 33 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnDelta.scala
```diff
@@ -98,12 +98,16 @@ final class ColumnDelta extends ColumnFormatValue with Delta {
...
+            // returned value should be the original and not oldColValue which may be a temp copy
+            oldValue
           } else {
             val merged = mergeStats(existingBuffer, newBuffer, schema)
             if (merged ne null) {
               new ColumnFormatValue(merged, oldColValue.compressionCodecId, isCompressed = false)
-            } else oldColValue
+            } else {
+              // returned value should be the original and not oldColValue which may be a temp copy
+              oldValue
```
kneeraj: This was a permanent leak. Right? as the old value is replaced from the map.

sumwale: This new value will be put into the region so its not a problem otherwise. Released buffer being put into the region has other major problems though. Its reference count will always remain zero, so all reads will internally read from disk. Plus the accounting has gone wrong because releaseBuffer does no accounting assuming value has already been removed from region.

Either way off-heap has a PhantomReference way of getting released so its never a permanent leak unless someone is holding on to the ByteBuffer itself.

kneeraj: >> Released buffer being put into the region has other major problems though. Its reference count will always remain zero, so all reads will internally read from disk.
Isn't OVERFLOW_TO_DISK or similar Token is expected to be in the region so that is read from the disk? Are we indicating OVERFLOW in some other way now for off-heap.

ok Phantomref is good to know.

sumwale: If the value overflows to disk and is faulted back in, then the reference count will get corrected. Didn't get the question.

kneeraj: Trying to understand this. If some value is evicted then what does the value field in the vm is for those entries. It used to be some Token like REMOVED_PHASE1 or something similar. Is that the case for this as well. Or we have a ColumnFormatValue object with a released bytebuffer.
Pardon my little understanding here.

PR #1214, opened 2018-12-05
Store the temporary join result in offheap

suranjan:

Check if there are servers with offheap setting and if so store them in offheap.
## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/ddl.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/ddl.scala
@@ -24,8 +24,10 @@ import com.pivotal.gemfirexd.internal.engine.store.GemFireStore
 import com.pivotal.gemfirexd.internal.iapi.reference.{Property => GemXDProperty}
 import com.pivotal.gemfirexd.internal.impl.jdbc.Util
 import com.pivotal.gemfirexd.internal.shared.common.reference.SQLState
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/execution/ddl.scala
```diff
@@ -367,16 +370,31 @@ case class SnappyCacheTableCommand(tableIdent: TableIdentifier,
...
         val df = Dataset.ofRows(session, lp)
+        val isOffHeap = SnappyContext.getClusterMode(sparkSession.sparkContext)
+        match {
+          case _: ThinClientConnectorMode =>
+            SparkEnv.get.memoryManager.tungstenMemoryMode == MemoryMode.OFF_HEAP
+          case _ =>
+            try {
+              SnappyTableStatsProviderService.getService.getMembersStatsOnDemand.
+                  values.exists(member => (member.isDataServer
+                  && (member.getOffHeapMemorySize > 0)))
```
sumwale: I think we should check that all data servers should have off-heap, else persist with off-heap will fail if even one does not: values.forall(member => !member.isDataServer || member.getOffHeapMemorySize > 0)

suranjan: Changed..

PR #1216, opened 2018-12-11
Fixes for SNAP-2512

smahajan05:

- Added code to check if table exists or not in catalog.

## Patch testing

Manually tested on snappy shell

## ReleaseNotes.txt changes

NA

## Other PRs 

None


kneeraj:
Please add a test for the scenario which was failing, which now passes with your change. Rejecting this PR as even the branch name is SNAP-2707 which is a different issue.

kneeraj:
Ok the branch is correct but the comment is wrong, please rectify that. And as said earlier add a test.

PR #1217, opened 2018-12-12
[SNAP-2773] improve GUI display of put into

sumwale:

- pass the query string along to SnappyCacheTableCommand to display in GUI
- removed the extra DS.count() and instead pass back the count of cached batches from cache query
- put into will pass "CACHE FOR (sql)" string if SQL is available else use "PUT INTO <plan>"
  when using putInto API
- also changed CACHE TABLE <table> without query to use off-heap route if enabled,
  and display the full plan in GUI as done for the query case
- set SQL text as Spark job description to show in the jobs/stages tabs in the UI

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


sumwale:
Attaching screenshots of how put into from SQL looks like after these changes including the job description change.

screenshot1
screenshot2
screenshot3

suranjan:
Looks good to me.

PR #1219, opened 2018-12-16
Catalog cleanup and integration with Spark's HiveExternalCatalog

sumwale:
This replaces the current session catalog implementations (SnappyStoreHiveCatalog,
SnappyConnectorExternalCatalog) with a single new implementation SnappySessionCatalog that
is common for both embedded and connector modes. All real work has now been shifted to
ExternalCatalog implementations SnappyHiveExternalCatalog and ConnectorExternalCatalog
respectively. Former now extends HiveExternalCatalog and has minimal overrides to behave as close to
Spark as possible.

Smart connector catalog has been re-written to use two common procedures to get and update
catalog meta-data passing generic arguments encapsulated in thrift classes. The older separate
procedures for DDL executions are now all removed. It no longer uses a hive client for queries either.
These changes allow the smart connector to execute all kinds of catalog updates available in
the Spark interface including thing like altering buckets/partitions for Spark parquet/ORC tables.

## Other changes proposed in this pull request

- enabled a bunch of compatibility tests; this will closed multiple open issues that will be done
  separately when this branch merges
- added implicit retry for catalog stale exception in queries
- invalidate entire cache of connector for a create/drop/alter since the version
  stored for other relations in RelationInfo will also certainly be stale

Replace koloboke maps with eclipse collections:
- The koloboke project has been dead and unmaintained for a couple of years now
  so replaced with eclipse collections though latter are a bit slower for some operations
  and also add significant bulk (~10M).

- change tokenize to be session property rather than global

## Patch testing

"precheckin -Pstore" with many compatibility tests now enabled

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/452
https://github.com/SnappyDataInc/spark/pull/132


sumwale:
Merging. Will take care of any future review comments subsequently.

PR #1224, opened 2018-12-22
initial changes to support Null Data Type for column tables . Not sur…

ahshahid:
The view sql which previously contained null virtual columns ( mapping to spark's NullType, I suppose ) are no longer working in current master.
The change I have done today is a very quick fix type of thing . It may or may not be correct. I will spend some more time on it on monday or as soon as I get bandwidth. Pls modify or fix the code as per best approach. 
The BugTest is checked in.
(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 
[spark pr](https://github.com/SnappyDataInc/spark/pull/133)


sumwale:
Closing this as behaviour is identical to Spark which is what is required in product.

PR #1232, opened 2019-01-08
Fixing issues seen in hydra test after gradle upgrade 5.0

sonalsagarwal:
- Fixing classpath for store jars.
- Code to ignore the 0.01 decimal mismatch in spark and snappy results.

## Patch testing
Full snappy regression.

## ReleaseNotes.txt changes
N/A

## Other PRs 
N/A


dtests/src/test/scala/io/snappydata/hydra/SnappyTestUtils.scala
```diff
@@ -94,6 +94,38 @@ object SnappyTestUtils {
...
+          val value1: String = actualArray(i)
+          val value2: String = expectedArray(i)
+          if (!value1.equals(value2)) {
+            try {
+              val val1: Double = value1.toDouble
+              val val2: Double = value2.toDouble
+              if (val1 > val2) diff = val1.-(val2).doubleValue
+              else diff = val2.-(val1).doubleValue
+              println("diff is " + diff)
+              if (diff <= 0.01) isIgnorable = true
```
swatisawant: Add the diff <=0.1 for ignore condition instead of 0.01. The round function used in snappy tests results in the difference of 0.1 for most of the queries with the double data type.

PR #1234, opened 2019-01-08
Snap 2760

ahshahid:
Providing a new sql function in spark which returns an array of  ldap groups of which the current user is member of.
As the return type of this function is array, it can be used with array_contains function for condition check .
This functionality will be used by customer for implementing the RowLevelSecurity where a user's ad group will be present.

## Patch testing

tested using a basic functionality test, more tests will be added.

## ReleaseNotes.txt changes

it needs to be added in release notes & documentation
## Other PRs 
[Store change](https://github.com/SnappyDataInc/snappy-store/pull/455)


ahshahid:
Pls review this CR, as currently I have given this patch to JnJ for RLS as they were waiting on it..

core/src/main/scala/org/apache/spark/sql/policy/policyFunctions.scala
```diff
@@ -49,3 +52,32 @@ case class CurrentUser() extends LeafExpression with CodegenFallback {
...
+       map(UTF8String.fromString(_))
+    ArrayData.toArrayData(array)
+  }
+
+  override def prettyName: String = "current_user_ldap_groups"
+}
+
+object LdapGroupsOfCurrentUser {
+  val dataType: DataType = ArrayType(StringType)
+}
```
sumwale: Where is this used?

ahshahid: You mean to ask where this function is being used?

ahshahid: If so, this is being used by customer in implementing their Row Level security logic.
Where they want to filter out rows based on condition if the user is part of particular ldap group .

PR #1248, opened 2019-01-23
SNAP-2886 - executing streaming queries in separate scheduler pool when configured

vatsalmevada:

Executing streaming queries as part of custom scheduler pool if passed by  `snappydata.scheduler.pool` property.

## Patch testing

manual, scala tests

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/136


core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -216,11 +216,13 @@ class CachedDataFrame(snappySession: SnappySession, queryExecution: QueryExecuti
     var pool = snappySession.sessionState.conf.activeSchedulerPool
     // Check if it is pruned query, execute it automatically on the low latency pool
     if (isLowLatencyQuery && pool == "default") {
-      if (snappySession.sparkContext.getPoolForName(Constant.LOW_LATENCY_POOL).isDefined) {
-        pool = Constant.LOW_LATENCY_POOL
+        if (snappySession.sparkContext.getPoolForName(Constant.LOW_LATENCY_POOL).isDefined) {
```
suranjan: Does this cause all the query in SnappySink also to go in streaming pool?

cluster/conf/fairscheduler.xml
```diff
@@ -27,4 +27,8 @@
     <weight>2</weight>
     <minShare>2</minShare>
   </pool>
+  <pool name="streaming">
```
suranjan: By default we will comment this. User will enable it if required.

core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
vatsalmevada: This is done to avoid overwriting `streaming` pool from this code. We are setting streaming pool when the streaming thread started. If any of the queries inside the sink will execute this code, then `streaming` pool was getting overwritten.

vatsalmevada: this particular change is not required and hence reverted.

PradeepSurale:

PR #1250, opened 2019-01-25
Fixing the index of the byte used for setting & retrieveing the proje…

ahshahid:
…ction bit set. Fixes SNAP-2890

Correcting the position of the byte used to set the projection bit in ColumnFormatIterator
As the index is 0 based, using column number -1, to identify the position of the byte &  the bit for masking.
The issue is that say if the max column in projection is 128, the number of bytes needed to store bits is 16 ( 0 to 15). But for 128th column, the current code gets index as 16 ( instead of 15)
It is possible that the fix may be needed at other places too?

Added bug test


ahshahid:
@sumwale @kneeraj @suranjan @ashetkar , can you pls review this asap, as the customer project is stuck on this issue.

ahshahid:
@sumwale @kneeraj @suranjan @ashetkar , pls review...

cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
```diff
@@ -19,10 +19,15 @@ package org.apache.spark.sql.store
 import java.io.{BufferedReader, FileReader}
 import java.sql.{DriverManager, SQLException}
 
+import scala.tools.nsc.interpreter
+import scala.tools.nsc.interpreter.session
+
```
sumwale: these imports are required?

cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
```diff
@@ -475,5 +480,76 @@ class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
...
+      "using column ")
+
+    val params = Array.fill(numCols)('?').mkString(",")
+    val insertStr = s"insert into test1 values (${params})"
+    val ps = conn.prepareStatement(insertStr)
+    for (i <- 0 until 1000) {
+      for (j <- 1 until numCols + 1) {
+        ps.setString(j, j.toString)
+      }
+      ps.executeUpdate()
```
sumwale: better to do batch insert which will be much more efficient and reduce test running time: ps.addBatch(), and ps.executeBatch() outside the loop

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatIterator.scala
```diff
@@ -75,7 +75,7 @@ final class ColumnFormatIterator(baseRegion: LocalRegion, projection: Array[Int]
       val maxProjection = Ints.max(projection: _*)
       val bitset = new Array[Long](UnsafeRow.calculateBitSetWidthInBytes(maxProjection) >>> 3)
       for (p <- projection) {
-        BitSet.set(bitset, Platform.LONG_ARRAY_OFFSET, p)
+        BitSet.set(bitset, Platform.LONG_ARRAY_OFFSET, p-1)
```
sumwale: missing spaces: p - 1

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatIterator.scala
```diff
@@ -217,7 +217,7 @@ final class ColumnFormatIterator(baseRegion: LocalRegion, projection: Array[Int]
           val tableColumn = ColumnDelta.tableColumnIndex(columnIndex)
           tableColumn > 0 &&
               BitSet.isSet(projectionBitSet, Platform.LONG_ARRAY_OFFSET,
-                tableColumn, projectionBitSet.length)
+                tableColumn -1 , projectionBitSet.length)
```
sumwale: missing space after -

cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
ahshahid: will remove it..

PR #1252, opened 2019-02-04
Fix for SNAP-2887. Checking for primary key columns also to identify …

ahshahid:
…handled filters

At the time of building the scan, use the PK columns also along with index columns to identify the handled filters.

## Patch testing

Added bug test & validated the fix through the hydra test


cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
```diff
@@ -475,5 +475,68 @@ class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
 
   }
 
+  test("Bug SNAP-2887") {
+    snc
+    var serverHostPort2 = TestUtil.startNetServer()
+    var conn = DriverManager.getConnection(s"jdbc:snappydata://$serverHostPort2")
+    var stmt = conn.createStatement()
+    val snappy = snc.snappySession
+
```
kneeraj: Why so many spaces throughout the test file.

cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
```diff
@@ -475,5 +475,68 @@ class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
 
   }
 
+  test("Bug SNAP-2887") {
```
kneeraj: The intention of the test just looks sanity on inserts. Do you wish to fire queries which uses push down when conditions are on indexed or pk columns?

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatRelation.scala
```diff
@@ -113,7 +113,8 @@ class RowFormatRelation(
 
   override def buildUnsafeScan(requiredColumns: Array[String],
       filters: Array[Expression]): (RDD[Any], Seq[RDD[InternalRow]]) = {
-    val handledFilters = filters.flatMap(ExternalStoreUtils.handledFilter(_, indexedColumns))
+    val handledFilters = filters.flatMap(ExternalStoreUtils.handledFilter(_, indexedColumns
+      ++ pushdownPKColumns(filters)) )
```
kneeraj: I am not sure that this is a right change. This is because for pk columns we have hash indexes i.e. the region itself. Other than equality how is it suppossed to work for greater than, less than etc. And equality is already handled.
Isn't it?

cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
ahshahid: removed those

cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
ahshahid: removed those.

core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatRelation.scala
ahshahid: Yes. and so if the filter condition has equality with correct columns ( so that they form primary key) then they should be part of handled filters. Pls take a look at bug , where I have described the exact issue. The check of handled/unhandled filters is done at two places.
1) To identify the unhandled filters
2) To identify the handled filters.
Both should work exactly same way. which was not the case. Unhandled filters wa staking into account primary key cols, while handled filter was not.

PR #1253, opened 2019-02-05
Snap 2237

ahshahid:
Added a new Resolution Rule to handle the case where the group by column refers to the projection using the alias. The code replaces the group by alias name with the actual expression in the projection.

## Patch testing
Added bug test

[AQP-PR](https://github.com/SnappyDataInc/snappy-aqp/pull/180)


ahshahid:
I will check this in & close the bug. When the merge happens, we will accordingly take care of it...

PR #1255, opened 2019-02-05
Fix for SNAP-2368, handling the case when SnappyDataBaseDialect is us…

ahshahid:
…ed to determine table schema with the table name not containing schema name

Handling the case when Dialect is used to obtain table's schema name, but the connection passed is null & session is also null. In such case, the returned name is not fully qualified table name.
This should be fine, as snappydata on server will use the default schema name.
## Patch testing
Added bug test


dshirish:
Hi @ahshahid 
The change to avoid NPE when the table name does not contain schema looks good.

The error - "TYPE 'TEXT' does not exist", is reproducible as SnappyStoreClientDialect is not getting registered with Spark because of which Spark picks wrong dialect.   If we run the test in spark-shell the issue reproduces (probably is not seen in junit/scalatest as dialect is registered in  static block of SnappySession via call to SnappyStoreDialect.init())

If we launch spark-shell using ```bin/spark-shell --master local[4]``` and run the test, it will error out
```
import org.apache.spark.sql.types.{StringType, StructField, StructType}
import org.apache.spark.sql.{Row, SaveMode, SparkSession}

val schema = StructType(List(StructField("name", StringType, nullable = true)))
val data = Seq(Row("abc"), Row("def"))
val namesDF = spark.createDataFrame(spark.sparkContext.parallelize(data), schema)
namesDF.createOrReplaceTempView("names")
Class.forName("io.snappydata.jdbc.ClientDriver")
spark.table("names").write.mode(SaveMode.Overwrite).jdbc("jdbc:snappydata://localhost:1527/", "names", new java.util.Properties())
```
If we add the statement ```jdbcDialects.registerDialect(org.apache.spark.sql.SnappyStoreClientDialect$.MODULE$)``` in the above test it runs fine.
Do we have to document this?

ahshahid:
Thanks Shirish.
Cant' we move this Dialect file in shared folder for both core & client jars  & register it in the driver instance creation?

ahshahid:
I agree. We can document it...
Or should we keep the Dialect related files in the driver jar & try to register it with spark in static block?.If spark packages are not available we eat the exception...?

PR #1270, opened 2019-03-13
[SNAP-2052] Skip implicit cast for update

vatsalmevada:
### throwing analysis exception in the following cases
- binary arithmetic operation is performed before the assignment and one of the operands is a string type
- trying to assign a value of some different type which doesn't match the column's data type. Although the assignment is allowed in the following cases even if the data type doesn't match: 
    - assigning null value
    - assigning narrower decimal to a wider decimal
    - assigning narrower numeric type to wider numeric type as far as precision is not compromised
    - assigning expression of any datatype to a string type column

This change was required as spark performs fail-safe casting which can lead to target column(s) getting updated with `NULL` if the cast fails (e.g. when the string value is not a number). 

User can explicitly cast the value to match the types appropriately. However, it is important to make sure that the values being cast is of an appropriate type otherwise the cast will again result into `NULL`.

## Patch testing

manual, scalatests, precheckin -Pspark

## Other PRs 

https://github.com/SnappyDataInc/snappy-aqp/pull/181


core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -102,14 +102,18 @@ class SnappySessionState(val snappySession: SnappySession)
...
             false
           } else {
             true
           }
           case _ => true
-        }): _*)
+        }).flatMap(_ match {
+          case PromoteStrings => Seq(StringPromotionCheckForUpdate, PromoteStrings)
+          case a => Seq(a)
+        })
```
sumwale: Can't the above filter rule be combined in flatMap instead of traversing twice?

```scala
        }).flatMap(_ match {
          case PromoteStrings => if (sqlParser.sqlParser.questionMarkCounter > 0) Nil
                                 else StringPromotionCheckForUpdate :: PromoteStrings :: Nil
          ...
```
Its slightly more efficient to use "::" for small sequences (a::Nil) and Nil avoids object creation and always better than Seq.empty

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -138,12 +142,55 @@ class SnappySessionState(val snappySession: SnappySession)
...
+    override lazy val batches: Seq[Batch] = snappyAnalyzer.batches.map {
+      case batch  if batch.name.equalsIgnoreCase("Resolution") =>
+        val rules = batch.rules.flatMap {
+          // Prepending PreUpdateTypeConversionCheck rule before PromoteString rule
+          case PromoteStrings =>
+            Seq(StringPromotionCheckForUpdate.asInstanceOf[Rule[LogicalPlan]],
+              PromoteStrings)
+          case r => Seq(r)
+        }
+
```
sumwale: a bit better to use "::"

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -138,12 +142,55 @@ class SnappySessionState(val snappySession: SnappySession)
...
+        case Update(_, _, _, _, _) => plan resolveExpressions {
+          case e if !e.childrenResolved => e
+          case BinaryArithmetic(_@StringType(), _) | BinaryArithmetic(_, _@StringType()) =>
+            throw new AnalysisException("Implicit type casting is not performed for update" +
+                " statements")
+          case e => e
+        }
+        case _ => plan
+      }
+    }
```
sumwale: I think we discussed that this should be applied only for updateExpressions, not in other places for consistency right?

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -628,19 +675,11 @@ class SnappySessionState(val snappySession: SnappySession)
...
-              // avoid unnecessary copy+cast when inserting DECIMAL types
-              // into column table
-              expr.dataType match {
-                case _: DecimalType
-                  if attr.dataType.isInstanceOf[DecimalType] => expr
-                case _ => Alias(Cast(expr, attr.dataType), attr.name)()
-              }
+            if (!attr.dataType.sameType(expr.dataType)) {
+              throw new AnalysisException("Implicit type casting is not performed for update" +
+                  " statements")
```
sumwale: As we discussed, this will lead to frequent DECIMAL type mismatches throwing exceptions so user will have to add precision/scala casts unnecessarily. If you add some tests with decimal types then it should immediately show the problem (in set clause and where clause). Implicit cast for DECIMAL types should be retained IMO.

vatsalmevada: Done changes in this logic as follows: 

trying to assign value of some different type which doesn't match the column's data type will fail except for following cases:  

- assigning null value
- assigning narrower decimal to a wider decimal
- assigning narrower numeric type to wider numeric type as far as precision is not compromised

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -138,12 +136,56 @@ class SnappySessionState(val snappySession: SnappySession)
...
+  // conversion when type of two operands does't match and this can lead to null values getting
+  // populated in the table.
+  object StringPromotionCheckForUpdate extends Rule[LogicalPlan] {
+
+    override def apply(plan: LogicalPlan): LogicalPlan = {
+      plan match {
+        case Update(table, child, keyColumns, updateColumns, updateExpressions) =>
+          updateExpressions.foreach {
+            case BinaryArithmetic(_@StringType(), _) | BinaryArithmetic(_, _@StringType()) =>
+              throw new AnalysisException("Implicit type casting is not performed for update" +
```
vatsalmevada: Should we have some more specific exception message here? Something like "Implicit casting of string type to numeric type is not performed for update expressions."

PradeepSurale:

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
vatsalmevada: Updated the message.

dshirish:
The change looks good to me. As discussed with @kneeraj, we can merge this PR.  We can check with @sumwale,  if there is a better approach instead of creating a new SnappyAnalyzer class and incorporate those changes later if required.

This change is useful for prepared statements as the PromoteStrings rule causes issues in it by replacing ParamLiteral with NULL (in case of of BinaryComparison node where left side is String and right side is ParamLiteral which is NullType). There also we need to add a new rule just before the PromoteStrings rule to avoid ParamLiteral getting converted to NULL.

PR #1271, opened 2019-03-14
Added ShowSchemaCommand class

bisenpiyush:
During the process of optimizing snappydata for Spotfire, found that after selecting schema from auth window, table list of that schema is not getting fetched. After an investigation found that, result of command Show schemas, and show tables, which is default in Uppercase in Snappydata causing the issue, so to resolve this issue converting the result into the lowercase

(Fill in the changes here)

## Patch testing
Tested with Spotfire as well as Tableau, to make sure, this patch is not affecting snappydata while using with other applications. 

## ReleaseNotes.txt changes

## Other PRs 


core/src/main/scala/org/apache/spark/sql/execution/ddl.scala
```diff
@@ -407,14 +407,14 @@ class ShowSnappyTablesCommand(session: SnappySession, schemaOpt: Option[String],
   override val output: Seq[Attribute] = {
     if (hiveCompatible) AttributeReference("name", StringType, nullable = false)() :: Nil
     else {
-      AttributeReference("schemaName", StringType, nullable = false)() ::
+      AttributeReference("database", StringType, nullable = false)() ::
```
dshirish: Could you update the class comment to reflect this change?

core/src/main/scala/org/apache/spark/sql/execution/ddl.scala
```diff
@@ -407,14 +407,14 @@ class ShowSnappyTablesCommand(session: SnappySession, schemaOpt: Option[String],
...
           AttributeReference("tableName", StringType, nullable = false)() ::
           AttributeReference("isTemporary", BooleanType, nullable = false)() :: Nil
     }
   }
 
   override def run(sparkSession: SparkSession): Seq[Row] = {
-    if (!hiveCompatible) return super.run(sparkSession)
+    if (!hiveCompatible) return super.run(sparkSession).map(r => Row(r.getString(0).toLowerCase, r.getString(1).toLowerCase, r.getBoolean(2)))
```
dshirish: The change is specific to make Spotfire work. 
Do we know whether we have to convert to lower case even when hive compatibility is turned on?

sumwale:
> Do we know this is not a Spotfire issue?

One can say so since it breaks with upper-case names. However, since Spark always returns lower-case so its not necessarily a Spotfire issue since this seems to be behaviour with only its Apache Spark connector.

core/src/main/scala/org/apache/spark/sql/execution/ddl.scala
sumwale: Has been reverted

core/src/main/scala/org/apache/spark/sql/execution/ddl.scala
sumwale: Now changed Catalog listDatabases/listTables itself to return the result as received from hive client (which is lower-case) so that it is Spark compatible. Other code changes to deal with this otherwise have been added.

PR #1279, opened 2019-04-03
Cluster stability MALLOC settings and SNAP-2959

suranjan:

[SNAP-2959]
Changed the weight of lean node to 17. Effectively it makes the weight to 27.

Set the following malloc settings by default after tests demonstrated the expected behavior
MALLOC_ARENA_MAX = 4
MALLOC_MMAP_THRESHOLD_=131072 
MALLOC_MMAP_MAX_= 2147483647

## Patch testing

Prechecking and manual testing.

## ReleaseNotes.txt changes

## Other PRs 


launcher/src/main/java/io/snappydata/tools/QuickLauncher.java
```diff
@@ -387,6 +387,10 @@ private String getHostNameAndDir() {
     commandLine.add("-Dorg.apache.commons.logging.Log=" +
         "org.apache.commons.logging.impl.Log4JLogger");
     setDefaultVMArgs(options, hostData, commandLine);
+    // For lead node set the member-weight to be 17
+    if (launcherClass.equals("io.snappydata.tools.LeaderLauncher")) {
+      vmArgs.add("-Dgemfire.member-weight=17");
+    }
```
sumwale: Since you have added to LeaderLauncher, why is this required?

Also if you keep this too for some reason then can you add a short unit test somewhere to check this weight (maybe just a check in an existing dunit test)? In future this namespace can change from io.snappydata and its likely that changing this may be missed.

cluster/bin/load-snappy-env.sh
```diff
@@ -32,6 +32,18 @@ if [ -z "$SNAPPY_ENV_LOADED" ]; then
...
+  fi
+
+  if [ -z "$MALLOC_MMAP_THRESHOLD_" ]; then
+    export MALLOC_MMAP_THRESHOLD_=131072
+  fi
+
+  if [ -z "$MALLOC_MMAP_MAX_" ]; then
+    export MALLOC_MMAP_MAX_=2147483647
+  fi
+
```
sumwale: Better to move this to snappy-env.sh rather than load-snappy-env.sh. Also it looks like the bin/snappy script loads only spark-env.sh and not snappy-env.sh. Can you update the snappy script to also load snappy-env.sh? Also moving to spark-env.sh will be better so that the same settings apply to Spark clusters too.

suranjan: From yesterday's observation, it should work as it is. I have added loading of snappy env in snappy script too.

launcher/src/main/java/io/snappydata/tools/QuickLauncher.java
suranjan: Tested it manually was working. All the scripts use QuickLauncher which extends LauncherBase calling setDefaultVMArgs.

Will add a test.

sumwale: Ok but you have already this in the override of LeaderLauncher.setDefaultVMArgs, so why is this also required here?

suranjan: LeaderLauncher extends LauncherBase and QuickLauncher extends LauncherBase too.
So any one starting with LeaderLauncher without QuickLauncher would need this like SnappyUtilLauncher.

suranjan: Tested it in a dunit test, the property is not reflected? Checking why.

sumwale: Ultimately QuickLauncher invokes the "launcherClass" i.e. LeaderLauncher so setting as vmArgs inside it will be too late but can be set as system property. I think the cleaner way is to set as system property in LeadImpl.start

suranjan: Quicklauncher invokes ""launcherClass" but uses the vmargs of its own. 
For dunit, will have to make change in LeaderImpl.

sumwale: Why can't you set as a system property in LeadImpl?

suranjan: Yes doing that. It will not require any other changes.

cluster/src/main/scala/io/snappydata/tools/LeaderLauncher.scala
```diff
@@ -16,6 +16,7 @@
  */
 package io.snappydata.tools
 
+import java.util
```
sumwale: not required

PR #1285, opened 2019-04-10
Set default value for spark.sql.files.maxPartitionBytes to 32mb

dshirish:
Set default value for spark.sql.files.maxPartitionBytes in snappy to 32mb. User can change the value if required. 

## Patch testing
Verified the patch manually
The setting of 32mb has been verfied by Swati/Sonal using parquet, csv and ORC formats

### DIFF
--- a/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
@@ -91,6 +91,8 @@ class LeadImpl extends ServerImpl with Lead
     bootProperties.remove("isTest")
     val authSpecified = Misc.checkLDAPAuthProvider(bootProperties)
 
+    ServiceUtils.setCommonBootDefaults(bootProperties, false)
+
     // prefix all store properties with "snappydata.store" for SparkConf
 
     // first the passed in bootProperties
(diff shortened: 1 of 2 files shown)


cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -142,6 +142,10 @@ class LeadImpl extends ServerImpl with Lead
     bootProperties.setProperty(serverGroupsProp, groups)
     bootProperties.setProperty(STORE_PREFIX + Attribute.GFXD_HOST_DATA, "false")
     bootProperties.setProperty(STORE_PREFIX + Attribute.GFXD_PERSIST_DD, "false")
+    // default value for spark.sql.files.maxPartitionBytes in snappy is 32mb
+    if (bootProperties.getProperty("spark.sql.files.maxPartitionBytes") == null) {
+      bootProperties.setProperty("spark.sql.files.maxPartitionBytes", "33554432")
+    }
```
sumwale: It might be better to keep this in ServiceUtils.setCommonBootDefaults so that all such SD defaults are in one place.

dshirish: Actually had earlier added this in ServiceUtils.setCommonBootDefaults, but realized that the property wasn't taking effect on lead as only `bootProperties `are being set in Spark conf by a call to `conf.setAll(bootProperties.asScala)` in `LeadImpl#start` Let me know if I am missing something here.

sumwale: Oh ok. Then perhaps LeadImpl.start should have an explicit call to ServiceUtils.setCommonBootDefaults at the start like in ServerImpl.start. Otherwise the other Spark properties being set may not have been applying correctly either.

dshirish: Changed the code as per the review comment. Please have a look.

PR #1288, opened 2019-04-12
[SNAP-2975] fix SEGV in putInto

sumwale:
Issue was incorrect number of columns being determined in ColumnTableScan for the base table
due to output projection not having the extra virtual columns projected.

- use relation schema to determine put/update case rather than output projection in ColumnTableScan
- removed duplicate methods across SharedUtils and Utils

## Patch testing

precheckin

## ReleaseNotes.txt changes

NA

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
@@ -46,7 +46,7 @@ import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.dsl.expressions._
 import org.apache.spark.sql.catalyst.expressions._
 import org.apache.spark.sql.catalyst.expressions.codegen.{CodegenContext, ExprCode, ExpressionCanonicalizer}
-import org.apache.spark.sql.collection.Utils
+import org.apache.spark.sql.collection.{SharedUtils, Utils}
 import org.apache.spark.sql.execution._
 import org.apache.spark.sql.execution.columnar.encoding._
 import org.apache.spark.sql.execution.columnar.impl.{BaseColumnFormatRelation, ColumnDelta}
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/ColumnTableScan.scala
```diff
@@ -496,8 +496,10 @@ private[sql] final case class ColumnTableScan(
     val statsRow = ctx.freshName("statsRow")
     val deltaStatsRow = ctx.freshName("deltaStatsRow")
     val colNextBytes = ctx.freshName("colNextBytes")
-    val numTableColumns = if (ordinalIdTerm eq null) relationSchema.size
-    else relationSchema.size - ColumnDelta.mutableKeyNames.length // for update/delete
+    val numTableColumns = if (relationSchema.exists(
```
suranjan: How is it different?

dshirish: Why was the previous one wrong. I mean the ordinalId term check?
What we understood is that perhaps ordinalIDTerm == null meant no upate/delete has happened on it so just take the relationSchema.size else subtract the virtual columns.

Also why did you change size to length? Any issue there?

Please explain how was it wrong earlier?

sumwale: ordinalIDTerm will be set based on output, so if the output projection does not contain the extra columns then this check will go wrong. Specifically since putInto uses the same plan object for join/anti-join, there is a scan there which goes into the joins and is not fed to ColumnUpdateExec directly but still has those additional columns.

sumwale:
@dshirish @suranjan As explained in the comments above, output projection may not have the extra columns used for update/delete with putInto. Specifically since putInto plan uses/combines same plan objects for join/anti-join one of the scans has the additional columns but does not get fed directly to ColumnUpdateExec rather is used only in the join, so output projection does not have those columns but the relation's schema (created from ColumnFormatRelation) has them.

PR #1289, opened 2019-04-13
Quote table name in some commands before sending to GemXD

sumwale:
allows support for "default" as schema name without explicit quotes
which is reserved keyword in GemXD parser

- quote table name in INSERT/PUT and ALTER TABLE before executing on GemXD connection

## Patch testing

manual

## ReleaseNotes.txt changes

NA

## Other PRs 

NA


cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
```diff
@@ -770,4 +772,50 @@ class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
...
+
+    session.sql("insert into default.t1 values (1, 'name1'), (2, 'name2')")
+    var res = session.sql("select * from default.t1 order by id").collect()
+    assert(res === Array(Row(1L, "name1"), Row(2L, "name2")))
+    res = session.sql("select * from default.t1 where id = 1").collect()
+    assert(res === Array(Row(1L, "name1")))
+    res = session.sql("select * from `DEFAULT`.t1 where id = 2").collect()
+    assert(res === Array(Row(2L, "name2")))
+
+    stmt.execute("insert into default.t2 values (1, 'name1'), (2, 'name2')")
```
kneeraj: Can you please add one insert with the table name itself in the backticks and see if that works too.
Like even this should work.

insert into "backtick"default"backtick"."backtick"t1"backtick" values (1, 'name1'), (2, 'name2')

Not able to format with backtick so wrote backtick with quotes.

sumwale: Ok. Btw, you can add code in triple backticks and also language i.e. ```sql and then end with triple backticks like below ("Edit" from options to see the formatting).

```sql
insert into `default`.`t1` values (1, 'name1'), (2, 'name2')
```

kneeraj: ok. Thanks. I did try the triple backtick but may be I goofed up something.

PR #1291, opened 2019-04-18
* Versioning related changes for upcoming 1.1.0 release.

ashetkar:
- Bumped up versions as below:
snappydata 1.0.2.1   ->   1.1.0
snappy-store 1.6.2.1   ->   1.6.3
spark 2.1.1.4   ->   2.1.1.5
spark-jobserver 0.6.2.8   ->   0.6.2.9
zeppelin interpreter 0.7.3.4   ->   0.7.3.5

- Updated archive names and org names.
- Changes in submodules (store, spark, aqp and snappy-connectors) are similar but minor ones.

## Patch testing
Basic testing with cluster launch and CREATE DDL

## ReleaseNotes.txt changes
NA

## Other PRs 
Changes in build.gradle of other repos. Will be added soon.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,7 @@ allprojects {
   apply plugin: "build-time-tracker"
 
   group = 'io.snappydata'
-  version = '1.0.2.2'
+  version = '1.1.0'
 
   // apply compiler options
   tasks.withType(JavaCompile) {
@@ -97,12 +97,12 @@ allprojects {
   javadoc.options.charSet = 'UTF-8'
 
   ext {
-    productName = 'SnappyData'
-    vendorName = 'SnappyData, Inc.'
+    productName = 'TIBCO ComputeDB'
+    vendorName = 'TIBCO, Inc.'
     scalaBinaryVersion = '2.11'
     scalaVersion = scalaBinaryVersion + '.8'
     sparkVersion = '2.1.1'
-    snappySparkVersion = '2.1.1.4'
+    snappySparkVersion = '2.1.1.5'
     sparkDistName = "spark-${sparkVersion}-bin-hadoop2.7"
     sparkCurrentVersion = '2.3.2'
     sparkCurrentDistName = "spark-${sparkCurrentVersion}-bin-hadoop2.7"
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -1129,11 +1133,17 @@ buildDeb {
 }
 
 distTar {
+  archiveName = 'TIB_compute-ce_' + version + '_linux.tar.gz'
+  if (isEnterpriseProduct) {
+    archiveName = 'TIB_compute_' + version + '_linux.tar.gz'
+  }
```
sumwale: Instead of hardcoding linux, use osName to derive like in the ext block after osName declaration: osName.getFamilyName().replace(' ', '').toLowerCase()

build.gradle
```diff
@@ -1142,11 +1152,17 @@ distTar {
 }
 
 distZip {
+  archiveName = 'TIB_compute-ce' + '_' + version + '_linux' + '.zip'
+  if (isEnterpriseProduct) {
+    archiveName = 'TIB_compute' + '_' + version + '_linux' + '.zip'
+  }
```
sumwale: see the comment above about using osName

build.gradle
```diff
@@ -1101,7 +1101,9 @@ ospackage {
 
 buildRpm {
   dependsOn ':packageVSD'
-  dependsOn ':packageZeppelinInterpreter'
+  if (rootProject.hasProperty('enablePublish')) {
+    dependsOn ':packageZeppelinInterpreter'
+  }
   requires('glibc')
```
sumwale: Shouldn't the name be changed for rpm and deb too? I think its "packageName". Since those are only for Linux, so there is no need to add "_linux" to the name.

ashetkar: Yes, names are changed for them too. No '_linux' affix for them. Debian package had some restrictions on naming. So the naming is like "tib-compute_1.1.0-1_all.deb" and "tib-compute-1.1.0-1.noarch.rpm"

build.gradle
ashetkar: Done for this and distZip.

PR #1297, opened 2019-04-23
Start hive-thriftserver by default in background

sumwale:

- changed default of hive-thriftserver to true
- to reduce startup time, the default behaviour is to launch the thrift server
  in background with startup message stating: Starting ... (without host/port information)
- if snappydata.hiveServer.enabled=true is explicitly provided then the hive thriftserver
  is launched in foreground with proper host/port information like before

## Patch testing

precheckin

## ReleaseNotes.txt changes

docs update to mention that hive-thriftserver is started by default in background

## Other PRs 

NA

### DIFF
--- a/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
+++ b/cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
@@ -252,8 +252,8 @@ class LeadImpl extends ServerImpl with Lead
       var jobServerConfig: Config = null
       var startupString: String = null
       if (Property.JobServerEnabled.get(conf)) {
-        jobServerWait = Property.JobServerWaitForInit.get(conf) ||
-            (!startHiveServerDefault && startHiveServer)
+        jobServerWait = (!startHiveServerDefault && startHiveServer) ||
+            Property.JobServerWaitForInit.get(conf)
         confFile = conf.getOption("jobserver.configFile") match {
           case None => Array[String]()
           case Some(c) => Array(c)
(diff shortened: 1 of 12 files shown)


cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -284,13 +294,11 @@ class LeadImpl extends ServerImpl with Lead
       SnappyTableStatsProviderService.start(sc, url = null)
 
       if (startHiveServer) {
-        val useHiveSession = Property.HiveServerUseHiveSession.get(conf)
         val hiveService = SnappyHiveThriftServer2.start(useHiveSession)
-        val sessionKind = if (useHiveSession) "session=hive" else "session=snappy"
-        SnappyHiveThriftServer2.getHostPort(hiveService) match {
-          case None => addStartupMessage(s"Started hive thrift server ($sessionKind)")
+        if (jobServerWait) SnappyHiveThriftServer2.getHostPort(hiveService) match {
```
suranjan: How does it start in background?

sumwale: Earlier code had: jobserverWait = startHiveServer || Property.JobServerWaitForInit.get(conf). So when hive server was configured to start, then it used to wait for entire initialization to finish. Now the new variable "startHiveServerDefault" will be true if nothing has been configured and in that case jobServerWait will be false.

So the code added before that adds startup message when startHiveServerDefault is true will display the message but the host/port won't be shown in that case because it is obtained only after startup of hive server (which can read from SparkConf or HadoopConf).

suranjan: Looked like start() call above is blocking. Just that some service may have been invoked in the background so no information of host:port immediately.

PR #1298, opened 2019-04-24
Worakaround fix for SNAP-2440.

kneeraj:
For single table this workaround will work. But for tables with same name
but in different schema it will have issues. Will be revisited as part of
SNAP-2442

Catching Analysis exception and checking if it is a resolution issue and if UnresolvedStar or UnresolvedAttribute kind of node is present in the execution plan. If yes then replace them by just star or minimally qualified column name.

## Patch testing

Unit test added. Precheckin underway.

## ReleaseNotes.txt changes

## Other PRs 

None.

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/MiscTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/MiscTest.scala
@@ -88,4 +88,31 @@ class MiscTest extends SnappyFunSuite with Logging {
       snc.sql(s"set schema app")
     }
   }
+
+  test("SNAP-2440") {
+    snc.sql(s"create table test(col1 int not null, col2 int not null) using column options()")
+    snc.sql(s"insert into test values (1, 2), (4, 5), (6, 7)")
+    val sz = snc.sql(s"select * from app.test").collect().size
+    val sqlstrs = Seq(s"select app.test.* from app.test",
+      s"select app.test.col1, app.test.col2 from app.test",
+      s"select col1, col2 from app.test",
+      s"select * from app.test",
+      s"select test.* from test")
+    sqlstrs.foreach(sqlstr => {
+      val res = snc.sql(sqlstr).collect()
+      assert(res.size === 3)
+      assert(res(0).get(0) != res(0).get(1))
+    })
+
+    val badsqls = Seq(s"select apppp.test.* from app.test",
+      s"select app.test.col99, app.test.col2 from app.test",
+      s"select testt.* from app.test")
+    badsqls.foreach(sqlstr =>
+      try {
+        snc.sql(sqlstr)
+        fail(s"expected analysis exception for $sqlstr")
+      } catch {
+        case ae: AnalysisException => // expected ... ignore
+      })
+  }
 }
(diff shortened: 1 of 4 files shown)


sumwale:
Since this is a temporary solution, why not simply trim the name off at Parser instead of this exception catching etc. Tried this change and your test passes fine. It also checks for existence of schema.

```
diff --git a/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala b/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
index c2b130057..3361f0cbc 100644
--- a/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyParser.scala
@@ -904,9 +904,21 @@ class SnappyParser(session: SnappySession)
         '.' ~ ws ~ (identifier. +('.' ~ ws) ~ ('.' ~ ws ~ '*' ~ push(true) ~ ws).? ~> {
           (i1: String, rest: Any, s: Any) =>
             if (s.asInstanceOf[Option[Boolean]].isDefined) {
-              UnresolvedStar(Option(i1 +: rest.asInstanceOf[Seq[String]]))
+              val remaining = rest.asInstanceOf[Seq[String]]
+              val qualifiedStar = if (remaining.isEmpty) i1 :: Nil else {
+                // schema should exist
+                session.sessionCatalog.requireSchemaExists(i1)
+                remaining
+              }
+              UnresolvedStar(Option(qualifiedStar))
             } else {
-              UnresolvedAttribute(i1 +: rest.asInstanceOf[Seq[String]])
+              val remaining = rest.asInstanceOf[Seq[String]]
+              val qualifiedName = if (remaining.length > 1) {
+                // schema should exist
+                session.sessionCatalog.requireSchemaExists(i1)
+                remaining
+              } else i1 +: remaining
+              UnresolvedAttribute(qualifiedName)
             }
         } | '*' ~ ws ~> { (i1: String) => UnresolvedStar(Some(Seq(i1)))
         }) |
diff --git a/core/src/main/scala/org/apache/spark/sql/internal/SnappySessionCatalog.scala b/core/src/main/scala/org/apache/spark/sql/internal/SnappySessionCatalog.scala
index 087ccc06c..cf38808a8 100644
--- a/core/src/main/scala/org/apache/spark/sql/internal/SnappySessionCatalog.scala
+++ b/core/src/main/scala/org/apache/spark/sql/internal/SnappySessionCatalog.scala
@@ -339,6 +339,14 @@ class SnappySessionCatalog(val externalCatalog: SnappyExternalCatalog,
     super.createDatabase(schemaDefinition, ignoreIfExists)
   }

+  private[sql] def requireSchemaExists(schemaName: String,
+      ignoreIfNotExists: Boolean = false): Unit = {
+    if (!databaseExists(schemaName)) {
+      if (ignoreIfNotExists) return
+      else throw new AnalysisException(s"Schema $schemaName not found")
+    }
+  }
+
   /**
    * Drop all the objects in a schema. The provided schema must already be formatted
    * with a call to [[formatDatabaseName]].
@@ -349,10 +357,7 @@ class SnappySessionCatalog(val externalCatalog: SnappyExternalCatalog,
       throw new AnalysisException(s"$schemaName is a system reserved schema")
     }

-    if (!externalCatalog.databaseExists(schemaName)) {
-      if (ignoreIfNotExists) return
-      else throw new AnalysisException(s"Schema $schemaName not found")
-    }
+    requireSchemaExists(schemaName, ignoreIfNotExists)
     checkSchemaPermission(schemaName, table = "", defaultUser = null, ignoreIfNotExists)

     if (cascade) {
```

PR #1300, opened 2019-04-25
Fix buffer reference count for couple of cases

sumwale:

- release compression buffer if compression has been determined to be ineffective
- release positions array buffer created by ColumnDeltaEncoder

## Patch testing

precheckin -Pstore

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/474


sumwale:
> Looks good. So if unreleased, all this gets gced when full gc runs? It seems sparks keep running System.gc() periodically.

Yes it will be released on full GC. We also invoke explicit GC before throwing LME if heap-size is <= 16GB and no eviction was possible (see SnappyUnifiedManager.tryExplicitGC) but System.gc() is unreliable and may not cause a full GC.

PR #1306, opened 2019-05-07
Make SnappyConf key name search case-insensitive

sumwale:

- search through all pre-defined key names case-insensitively for set/unset of SQLConf properties
- added check for planCaching property for the same

## Patch testing

precheckin

## ReleaseNotes.txt changes

mention that SQLConf property names are case-insensitive

## Other PRs 

NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
+++ b/core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
@@ -806,7 +806,9 @@ class SnappyConf(@transient val session: SnappySession)
     case SQLConf.AUTO_BROADCASTJOIN_THRESHOLD.key |
          Property.HashJoinSize.name |
          Property.HashAggregateSize.name |
-         Property.ForceLinkPartitionsToBuckets.name => session.clearPlanCache(); key
+         Property.ForceLinkPartitionsToBuckets.name =>
+      session.clearPlanCache()
+      key
     case SQLConf.SHUFFLE_PARTITIONS.key =>
       // stop dynamic determination of shuffle partitions
       if (doSet) {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -800,12 +800,13 @@ class SnappyConf(@transient val session: SnappySession)
...
 
-  private def keyUpdateActions(key: String, value: Option[Any], doSet: Boolean): Unit = key match {
+  private def keyUpdateActions(key: String, value: Option[Any],
+      doSet: Boolean, search: Boolean = true): String = key match {
     // clear plan cache when some size related key that effects plans changes
     case SQLConf.AUTO_BROADCASTJOIN_THRESHOLD.key |
          Property.HashJoinSize.name |
          Property.HashAggregateSize.name |
-         Property.ForceLinkPartitionsToBuckets.name => session.clearPlanCache()
+         Property.ForceLinkPartitionsToBuckets.name => session.clearPlanCache(); key
```
kneeraj: Putting the key on the next line will be better I think for better readability.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionState.scala
```diff
@@ -837,22 +842,25 @@ class SnappyConf(@transient val session: SnappySession)
...
     case Property.DisableHashJoin.name =>
       value match {
         case Some(boolVal) => session.disableHashJoin = boolVal.toString.toBoolean
         case None => session.disableHashJoin = Property.DisableHashJoin.defaultValue.get
       }
       session.clearPlanCache()
+      key
 
-    case SQLConf.WHOLESTAGE_CODEGEN_ENABLED.key => session.clearPlanCache()
+    case SQLConf.WHOLESTAGE_CODEGEN_ENABLED.key => session.clearPlanCache(); key
```
kneeraj: Putting the key on the next line will be better I think for better readability.

PR #1307, opened 2019-05-08
Invalidating the cached query plans if set command contains aqp prope…

ahshahid:
…rties

Invalidating the cache plans for the session if the property being set has key starting with spark.sql.aqp.
This fixes the bug AQP-285. 
The bug test is written in the aqp master.
## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
@@ -46,6 +46,7 @@ import org.apache.spark.sql.streaming.StreamPlanProvider
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.{SnappyParserConsts => Consts}
 import org.apache.spark.streaming._
+import org.apache.spark.sql.execution.InvalidateCachedPlans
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/SnappyDDLParser.scala
```diff
@@ -702,7 +703,11 @@ abstract class SnappyDDLParser(session: SparkSession)
           if (separatorIndex >= 0) {
             val key = rest.substring(0, separatorIndex).trim
             val value = rest.substring(separatorIndex + 1).trim
-            SetCommand(Some(key -> Option(value)))
+            if (key.startsWith("spark.sql.aqp.")) {
+              new SetCommand(Some(key -> Option(value))) with InvalidateCachedPlans
+            } else {
+              SetCommand(Some(key -> Option(value)))
+            }
```
sumwale: @ahshahid This is not the way we deal with actions on key set/unset. Code currently handles this in SnappyConf.keyUpdateActions so this should be moved there and the new InvalidateCachedPlans should be removed.

PR #1320, opened 2019-05-24
Fix of bug SNAP-2644

kneeraj:

The PromoteString rule is removed from the analyzer phase. It was intended to be removed
but the questionMarkCounter additional filter was sometimes causing the
rule to not get excluded. Since this is specific to prepare phase so removing it completely is not an issue.

## Patch testing

Added a multi thread dunit test.

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -42,7 +42,7 @@ import org.apache.spark.annotation.{DeveloperApi, Experimental}
 import org.apache.spark.jdbc.{ConnectionConf, ConnectionUtil}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.SnappySession.CACHED_PUTINTO_UPDATE_PLAN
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -201,10 +201,18 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
-    sessionState.analyzerPrepare.execute(logical)
+    var ap: Analyzer = null
+    if (!skipPromote) {
+      ap = sessionState.analyzerPrepare
+    } else {
+      ap = sessionState.analyzerWithoutPromote
+    }
+    // logInfo(s"KN: Batches ${ap.batches.filter(
+    //  _.name.equalsIgnoreCase("Resolution")).mkString("_")}")
+    ap.execute(logical)
```
dshirish: Do we need analyzerPrepare now or can we just use org.apache.spark.sql.internal.SnappySessionState#analyzer?

PR #1325, opened 2019-06-10
Spark multiversion support

sumwale:

- support for multiple Spark versions from same code base
- SparkInternals interface to abstract out the internal APIs used by SnappyData/AQP layers that have
  changed between 2.1 to 2.4 with implementations for 2.1.0/2.1.1/2.3.2
- updated build to allow for "spark.connector.version" property to build smart connector for non-default
  Spark version

## Patch testing

precheckin

## ReleaseNotes.txt changes

Documentation for multiple spark version support in smart connector mode

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/478
https://github.com/SnappyDataInc/snappy-aqp/pull/187


ashishkshukla:
@sumwale - I was  looking into this code base and trying to make a build for spark 2.3.2.
I noticed that we  have defined a def newSnappySessionState(snappySession: SnappySession): SnappySessionState which creates an instance of SnappySessionState for the given spark version.
It seems this needs to implemented for every spark version we support, as it has direct call from SnappySession sessionState .
We have its implementation in Spark210Internals.scala  for spark 2.1.1 and spark 2.1.0 but its implementation is  missing for spark 2.3.2. Do we need to work on the implementation and we are handing in different way as I can not see its implementation in Spark232Internals.scala .

PR #1329, opened 2019-06-18
Added code to copy the configuration files into other member of cluster

bisenpiyush:
As per requirement, added code in snappy-nodes.sh to copy the configuration files present in "conf" directory into other members of the cluster. Files present in "conf" are log4j.properties, spark-env.sh, spark-defaults.conf,snappy-env.sh, slaves, servers, leads, locators, metrics.properties,fairscheduler.xml, debug.conf.template,docker.properties. Along with these files if the user has its own configuration files, copy those files as well.
Following use case has been taken care:
1- If cluster running in localhost then do not copy, even though the configuration file configured with IP address and host-name.
2- Do not copy files with ".template" extension
3- Replace the leads, locators, servers files, with existing one other member of the cluster but inform the user about it and keep the backup of it.
4- Replace the other configuration files if it exists in other node/member of the cluster but inform the user about it and keep backup of it.

## Patch testing
To test this patch created the Virtual Machine with Ubuntu 16.04. Ran the multinode cluster to test and copy the files from the main machine to VM. Please let me know if there any test suite available for testing scripts.

## ReleaseNotes.txt changes
NA

### DIFF
--- a/cluster/sbin/snappy-nodes.sh
+++ b/cluster/sbin/snappy-nodes.sh
@@ -244,7 +244,7 @@ function execute() {
     esac
   done
   #copy the conf files into other node before starting the launch processs
-  if [[ -n "$isStart" && $NO_COPYCONF -eq 0 ]]; then
+  if [[ -n "$isStart" && $SKIP_CONF_COPY -eq 0 ]]; then
     copyConf "$@"    
   fi
   if [ "$host" != "localhost" ]; then
@@ -360,7 +360,7 @@ function copyConf() {
 	  else
 	      backupDir="backup"
 	      if [[ ! -z $(ssh $host "cat $entry" | diff - "$entry") ]] ; then
-		backupFileName=${fileName}_${TIME_STAMP_OF_START}
+		backupFileName=${fileName}_${START_ALL_TIMESTAMP}
 		echo "INFO: Copied $filename from this host to $host. Moved the original $filename on $host to $backupFileName."
                 (ssh "$host" "{ if [ ! -d \"${SPARK_CONF_DIR}/$backupDir\" ]; then  mkdir \"${SPARK_CONF_DIR}/$backupDir\"; fi; } ")
 		ssh $host "mv ${SPARK_CONF_DIR}/$fileName ${SPARK_CONF_DIR}/$backupDir/$backupFileName"		    
--- a/cluster/sbin/snappy-nodes.sh
+++ b/cluster/sbin/snappy-nodes.sh
@@ -339,8 +339,11 @@ function getNumLeadsOnHost() {
 }
 
 function scpConf() {
- var=$(ip addr | grep 'state UP' -A2 | tail -n1 | awk '{print $2}' | cut -f1  -d'/')
- if [[ "$host" != "$var" && "$host" != "localhost" ]] ; then
+ 
+ currentNodeIpAddr=$(ip addr | grep 'state UP' -A2 | head -n3 | tail -n1 | awk '{print $2}' | cut -f1  -d'/')
+ currentNodeHostName=$(uname -n)
+
+ if [[ "$host" != "$currentNodeIpAddr" && "$host" != "localhost" && "$host" != $currentNodeHostName ]] ; then
   	#loop to get the all the files avaliable in Conf directory
 	for entry in "${SPARK_CONF_DIR}"/*
 	do
@@ -351,14 +354,21 @@ function scpConf() {
    	    template=".template"
 	    #check the extension, interested in files those doesn't have template extension
 	    if [[ ! "$fileName" = @(*.template) ]]; then	       	
-		if ! ssh $host "test -e $entry"; then #"File does not exist." 		  
-		  scp -i ~/.ssh/id_rsa.pub ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR}
+		if ! ssh $host "test -e $entry"; then #"File does not exist."	  
+		  scp ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR}
 		else
-		  #file exits but do not override/copy..since user might have different configuration setting for different node.
-      		  # except			 
-		  if [[ "$fileName" == "locators" || "$fileName" == "servers" || "$fileName" == "leads"  ]]; then		 
-		  scp -i ~/.ssh/id_rsa.pub ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR}   
-	          fi								
+		  #file exits but before replacing with new one, create a backup file and warn user about it then replace
+      		  # except locators,leads and servers file has to replace all the time			 
+		  if [[ "$fileName" == "locators" || "$fileName" == "servers" || "$fileName" == "leads"  ]]; then
+		    scp ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR} 
+		  else
+	          	if [[ ! -z $(ssh $host "cat $entry" | diff - "$entry") ]] ; then
+				backupFileName=${fileName%.*}_`date +%Y_%m_%d_%H_%M_%S`.${fileName#*.}
+				echo "###### As $fileName differ from the $fileName present on $host, creating a backup file with the name $backupFileName before replacing #######"
+				ssh $host "mv ${SPARK_CONF_DIR}/$fileName ${SPARK_CONF_DIR}/$backupFileName"		    
+		    		scp ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR} 
+		    	fi  
+          fi												
 		fi        				
 	    fi # end of if, check extension
 	  fi # end of if to get each file
@@ -419,7 +429,11 @@ if [ -n "${HOSTLIST}" ]; then
       execute "$@"
       if [ -n "$isStart" ]; then
     	count=$(echo $MEMBERS_FILE | grep -o -i "$host" "$MEMBERS_FILE" | wc -l)
-	if [ $count -eq 2 ]; then  # 2 ,since in a single line there wiil be two match of ip address
+	# checking count is 2, since in a single line of member.txt, there will be two matches of ip address/hostname of a component.
+	# entry happpens in member file sequentially. So for example, if on some node 1, both server and lead is running, then by that time lead will start,
+	# member file will have already an entry for ipaddress/hostname of that node 1 after starting the server  
+	# it kind of cross checking if component with same ipaddress already present in member file or not.
+	if [ $count -eq 2 ]; then
     	scpConf "$@"
   	fi
       fi
--- a/cluster/sbin/snappy-nodes.sh
+++ b/cluster/sbin/snappy-nodes.sh
@@ -338,12 +338,41 @@ function getNumLeadsOnHost() {
   echo $numLeadsOnHost
 }
 
+function scpConf() {
+ var=$(ip addr | grep 'state UP' -A2 | tail -n1 | awk '{print $2}' | cut -f1  -d'/')
+ if [[ "$host" != "$var" && "$host" != "localhost" ]] ; then
+  	#loop to get the all the files avaliable in Conf directory
+	for entry in "${SPARK_CONF_DIR}"/*
+	do
+	  if [ -f "$entry" ];then
+	    #${file%.*} to get the filename without the extension and ${file##*.} to get the extension alone
+	    fileName=$(basename $entry)
+	    #echo "fileName:  $fileName"
+   	    template=".template"
+	    #check the extension, interested in files those doesn't have template extension
+	    if [[ ! "$fileName" = @(*.template) ]]; then	       	
+		if ! ssh $host "test -e $entry"; then #"File does not exist." 		  
+		  scp -i ~/.ssh/id_rsa.pub ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR}
+		else
+		  #file exits but do not override/copy..since user might have different configuration setting for different node.
+      		  # except			 
+		  if [[ "$fileName" == "locators" || "$fileName" == "servers" || "$fileName" == "leads"  ]]; then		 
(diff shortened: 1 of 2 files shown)


kneeraj:
Please mention specific files which needs to be copied at this point of time.
For example you have mentioned log4j.properties.
Other could be 'snappy-env.sh', 'spark-env.sh'?

Please check all the required files and mention explicitly in the PR.

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +338,41 @@ function getNumLeadsOnHost() {
   echo $numLeadsOnHost
 }
 
+function scpConf() {
+ var=$(ip addr | grep 'state UP' -A2 | tail -n1 | awk '{print $2}' | cut -f1  -d'/')
```
kneeraj: Please use a meaningful variable name so that we can get the idea of what is being tried to be taken out from the conf.

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +338,41 @@ function getNumLeadsOnHost() {
...
+ if [[ "$host" != "$var" && "$host" != "localhost" ]] ; then
```
kneeraj: The user can give the actual hostname instead of 'localhost' or 'ip' then in that case this condition will be incorrect.

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +338,41 @@ function getNumLeadsOnHost() {
...
+	for entry in "${SPARK_CONF_DIR}"/*
+	do
+	  if [ -f "$entry" ];then
+	    #${file%.*} to get the filename without the extension and ${file##*.} to get the extension alone
+	    fileName=$(basename $entry)
+	    #echo "fileName:  $fileName"
+   	    template=".template"
+	    #check the extension, interested in files those doesn't have template extension
+	    if [[ ! "$fileName" = @(*.template) ]]; then	       	
+		if ! ssh $host "test -e $entry"; then #"File does not exist."
```
kneeraj: This assumes that the installation directory is the same on all the hosts. Probably thats correct but please check if there could be a deviation here.

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +338,41 @@ function getNumLeadsOnHost() {
...
+	do
+	  if [ -f "$entry" ];then
+	    #${file%.*} to get the filename without the extension and ${file##*.} to get the extension alone
+	    fileName=$(basename $entry)
+	    #echo "fileName:  $fileName"
+   	    template=".template"
+	    #check the extension, interested in files those doesn't have template extension
+	    if [[ ! "$fileName" = @(*.template) ]]; then	       	
+		if ! ssh $host "test -e $entry"; then #"File does not exist." 		  
+		  scp -i ~/.ssh/id_rsa.pub ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR}
```
kneeraj: passwordless ssh is assumed to exists for the user. Direct scp should work without the '-i' option.

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +338,41 @@ function getNumLeadsOnHost() {
...
+		else
+		  #file exits but do not override/copy..since user might have different configuration setting for different node.
```
kneeraj: exists. spell mistake.

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +338,41 @@ function getNumLeadsOnHost() {
...
+      		  # except			 
+		  if [[ "$fileName" == "locators" || "$fileName" == "servers" || "$fileName" == "leads"  ]]; then
```
kneeraj: In some sense we are giving preference to the node on which we are invoking the start script because for these cases we are going ahead and overwriting. If that is the idea then the documentation should have strict recommendation as to which host should be used for invoking the script.
Also if you prefer lead machines for example or locators then at least we should check that the conf mentioned for one particular types should all match.

cluster/sbin/snappy-nodes.sh
```diff
@@ -385,6 +417,12 @@ if [ -n "${HOSTLIST}" ]; then
         args="$args -J-Dsnappydata.numLeadsOnHost=$(getNumLeadsOnHost "$host")"
       fi
       execute "$@"
+      if [ -n "$isStart" ]; then
+    	count=$(echo $MEMBERS_FILE | grep -o -i "$host" "$MEMBERS_FILE" | wc -l)
```
kneeraj: I think number of occurrence of host can be more than 2 also because he may use the host in the path somewhere. If you want to just take the host count then take the first word from each line.
Also it is better to take care of empty lines or comments line explicitly, though I guess for this file comments are not there since it is generated.

kneeraj:
Another comment would be that if files are existing and they have different contents then we should warn and let the user know so that inadvertently unwanted settings are not applied.

cluster/sbin/snappy-nodes.sh
bisenpiyush: Modified as per review

cluster/sbin/snappy-nodes.sh
bisenpiyush: checking count is 2, since, in a single line of member.txt, there will be two matches of IP address/hostname of a component. Entry happens in member file sequentially. So for example, if on some node 1, both server and lead are running, then by that time lead will start, member file will have already an entry for IP address/hostname of that node 1 after starting the server. It kind of cross-checking if a component with the same IP address already present in member file or not.

cluster/sbin/snappy-nodes.sh
bisenpiyush: resolved

cluster/sbin/snappy-nodes.sh
bisenpiyush: Resolved

cluster/sbin/snappy-nodes.sh
bisenpiyush: Removed -i option

bisenpiyush:
> Another comment would be that if files are existing and they have different contents then we should warn and let the user know so that inadvertently unwanted settings are not applied.

Added code to take care of this scenario.

cluster/sbin/snappy-nodes.sh
ashetkar: Yes, checking ofr count 2 may not always work as mentioned by Neeraj.

Also, we need to copy the conf files **before** the process starts, otherwise, those will not be processed.

We should do the copying just before the process is launched on respective nodes (e.g. in execute() function, just before `if [ "$host" != "localhost" ]; then`  block. This could be done by maintaining the hostnames in a array and updating the array with every iteration and avoid copying the files if the hostname is already present in it.

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +338,51 @@ function getNumLeadsOnHost() {
...
+		else
+		  #file exits but before replacing with new one, create a backup file and warn user about it then replace
+      		  # except locators,leads and servers file has to replace all the time			 
+		  if [[ "$fileName" == "locators" || "$fileName" == "servers" || "$fileName" == "leads"  ]]; then
+		    scp ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR} 
+		  else
+	          	if [[ ! -z $(ssh $host "cat $entry" | diff - "$entry") ]] ; then
+				backupFileName=${fileName%.*}_`date +%Y_%m_%d_%H_%M_%S`.${fileName#*.}
+				echo "###### As $fileName differ from the $fileName present on $host, creating a backup file with the name $backupFileName before replacing #######"
+				ssh $host "mv ${SPARK_CONF_DIR}/$fileName ${SPARK_CONF_DIR}/$backupFileName"
```
ashetkar: Also, as discussed, please move the backed up files inside a directory (say, "backup") in the remote conf directory.

bisenpiyush: Modified as per review comment

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +338,51 @@ function getNumLeadsOnHost() {
...
+		  scp ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR}
+		else
+		  #file exits but before replacing with new one, create a backup file and warn user about it then replace
+      		  # except locators,leads and servers file has to replace all the time			 
+		  if [[ "$fileName" == "locators" || "$fileName" == "servers" || "$fileName" == "leads"  ]]; then
+		    scp ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR} 
+		  else
+	          	if [[ ! -z $(ssh $host "cat $entry" | diff - "$entry") ]] ; then
+				backupFileName=${fileName%.*}_`date +%Y_%m_%d_%H_%M_%S`.${fileName#*.}
+				echo "###### As $fileName differ from the $fileName present on $host, creating a backup file with the name $backupFileName before replacing #######"
```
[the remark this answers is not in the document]
bisenpiyush: Modified as per review comment

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +338,51 @@ function getNumLeadsOnHost() {
...
+		if ! ssh $host "test -e $entry"; then #"File does not exist."	  
+		  scp ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR}
+		else
+		  #file exits but before replacing with new one, create a backup file and warn user about it then replace
+      		  # except locators,leads and servers file has to replace all the time			 
+		  if [[ "$fileName" == "locators" || "$fileName" == "servers" || "$fileName" == "leads"  ]]; then
+		    scp ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR} 
+		  else
+	          	if [[ ! -z $(ssh $host "cat $entry" | diff - "$entry") ]] ; then
+				backupFileName=${fileName%.*}_`date +%Y_%m_%d_%H_%M_%S`.${fileName#*.}
```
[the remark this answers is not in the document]
bisenpiyush: Modified as per review comment

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +338,51 @@ function getNumLeadsOnHost() {
...
+	    #echo "fileName:  $fileName"
+   	    template=".template"
+	    #check the extension, interested in files those doesn't have template extension
+	    if [[ ! "$fileName" = @(*.template) ]]; then	       	
+		if ! ssh $host "test -e $entry"; then #"File does not exist."	  
+		  scp ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR}
+		else
+		  #file exits but before replacing with new one, create a backup file and warn user about it then replace
+      		  # except locators,leads and servers file has to replace all the time			 
+		  if [[ "$fileName" == "locators" || "$fileName" == "servers" || "$fileName" == "leads"  ]]; then
```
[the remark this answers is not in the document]
bisenpiyush: Modified as per review comment

cluster/sbin/snappy-nodes.sh
bisenpiyush: Modified as per review comment

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +346,44 @@ function getNumLeadsOnHost() {
...
+	fileName=$(basename $entry)
+ 	template=".template"
+	#check the extension, interested in files those doesn't have template extension
+	if [[ ! "$fileName" = @(*.template) ]]; then	       	
+	  if ! ssh $host "test -e $entry"; then #"File does not exist."	  
+            scp ${SPARK_CONF_DIR}/$fileName  $host:${SPARK_CONF_DIR}
+	  else
+	      backupDir="backup"
+	      if [[ ! -z $(ssh $host "cat $entry" | diff - "$entry") ]] ; then
+		backupFileName=${fileName}_`date +%Y_%m_%d_%H_%M_%S`
```
ashetkar: Can initialize it once and use it for each execute() call.

cluster/sbin/snappy-nodes.sh
```diff
@@ -243,6 +243,14 @@ function execute() {
       -*) postArgs="$postArgs $arg"
     esac
   done
+  #copy the conf files into other node before starting the launch processs
+  if [ -n "$isStart" ]; then
+   #check $host is present in memberArr or not. If not then copy else do not copy
+    if [[ ! " ${memberArray[@]} " =~ " ${host} " ]]; then
```
ashetkar: I think we can remove memberArray and instead rely on the file diff you do before copy, to avoid copying the file multiple times.

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +346,44 @@ function getNumLeadsOnHost() {
...
+  fi # end of if 	
+}
+
 if [ -n "${HOSTLIST}" ]; then
   declare -a arr
   declare -a hosts
   declare -a counts
   isStartOrStatus=
-
+  count=0
```
ashetkar: Unused.

cluster/sbin/snappy-nodes.sh
```diff
@@ -338,12 +346,44 @@ function getNumLeadsOnHost() {
   echo $numLeadsOnHost
 }
 
+function scpConf() {
```
ashetkar: You can add a one line comment describing what the function does. Also, name can be more intuitive.

cluster/sbin/snappy-nodes.sh
bisenpiyush: Removed

cluster/sbin/snappy-nodes.sh
bisenpiyush: Added comment

cluster/sbin/snappy-nodes.sh
bisenpiyush: Removed memberArray

cluster/sbin/snappy-nodes.sh
bisenpiyush: Added timestamp in snappy-start-all.sh

PR #1332, opened 2019-06-20
examples module can be used as independent gradle project and extende…

paresh-p11:
…d further as is.

* Moved SnappyTestRunner from snappy-core to snappy-examples
* Removed snappy-core dependency from examples/build.gradle file.
* Added snappy-examples dependency to core/build.gradle and dtests/build.gradle

## Patch testing

Did manual testing. Precheckin yet to be done

## ReleaseNotes.txt changes

## Other PRs 


core/src/test/scala/io/snappydata/CommandLineToolsSuite.scala
```diff
@@ -24,6 +24,7 @@ import org.apache.commons.io.output.TeeOutputStream
 import org.apache.spark.sql.collection.Utils
 
 import scala.sys.process._
+import io.snappydata.SnappyTestRunner
```
ashetkar: Is this required? It's the same package.

examples/build.gradle
```diff
@@ -24,7 +24,53 @@ sourceSets.main.java.srcDirs = []
...
+    sparkJobServerVersion = '0.6.2.9'
+  }
+
+  repositories {
+    mavenCentral()
+    maven { url 'https://dl.bintray.com/big-data/maven' }
+    maven { url "https://repo.spring.io/libs-release" }
+    maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
+    maven { url 'http://repository.snappydata.io/repository/internal' }
+    maven { url 'http://repository.snappydata.io/repository/snapshots' }
```
ashetkar: Needs to be cleaned up. e.g. repository.snappydata.io does not exist any more.

examples/build.gradle
```diff
@@ -24,7 +24,53 @@ sourceSets.main.java.srcDirs = []
...
+    maven { url 'https://app.camunda.com/nexus/content/repositories/public' }
+  }
+
+  dependencies {
+
+    compile 'org.scala-lang:scala-library:' + '2.11.8'
```
ashetkar: Don't see these many dependencies in else block below. (log4j, slf4j, scala-library, scala-reflect, junit, snappy-core, spark-*). Doesn't snappy-cluster pull in those?

core/src/test/scala/io/snappydata/CommandLineToolsSuite.scala
paresh-p11: yes, that's redundant. Thanks

examples/build.gradle
paresh-p11: ok

kneeraj:
The idea is that the user should be able to setup the gradle project in ide, idea and also eclipse and he should be able to run the examples, make changes and play with that.
Please complete a document task as to how the user can go about those in simple steps. Please explain with one of the bundled example.

PR #1334, opened 2019-06-26
Prepared stmt on views gives error [SNAP-2254]

dshirish:

In case of views, to return LogicalPlan, SnappySessionCatalog#lookupRelation() parses the SQL text for view using sessionState.sqlParser.  For PreparedStatements this resets SnappyParser#_questionMarkCounter thereby causing the inconsistency. Now using a new instance SnappySqlParser to parse the view text.

## Patch testing
Added unit tests for prepared statements on views
Also added a test for SNAP-3007
precheckin

## ReleaseNotes.txt changes
NA


dshirish:
> We should test some complex views and preparedstatement combination. Please take care of this.
Yes, basic unit tests added in this PR. More tests to be also covered in SNAP-2515

PR #1337, opened 2019-07-01
[Snap-3065] Addressing issue in prepared statement with subquery

dshirish:

- Spark PromoteStrings rule causes issues in prepared statements by replacing ParamLiteral
with NULL in case of BinaryComparison with left node being StringType and right being
ParamLiteral (or vice-versa) as by default ParamLiteral datatype is NullType. Added a rule called SnappyPromoteStrings that converts ParmaLiteral type to StringType to prevent it being replaced by NULL
- Added a dunit test that runs all TPCH queries as prepared statement
- Removed redundent analyzerWithoutPromote object

## Patch testing
New unit tests
Precheckin

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


cluster/src/dunit/scala/org/apache/spark/sql/TPCHDUnitTest.scala
```diff
@@ -77,6 +78,259 @@ class TPCHDUnitTest(s: String) extends ClusterManagerTestBase(s)
...
+          prepStatement.setString(5, parameters(4))
+          prepStatement.setString(6, parameters(5))
+          prepStatement.setString(7, parameters(6))
+          prepStatement.setString(8, parameters(7))
+          prepStatement.setString(9, parameters(8))
+          prepStatement.setString(10, parameters(9))
+          prepStatement.setString(11, parameters(10))
+          prepStatement.setString(12, parameters(11))
+          prepStatement.setString(13, parameters(12))
+          prepStatement.setString(14, parameters(13))
```
kneeraj: Generic comment - May be for few queries instead of using setString, we can use setObject or other setter methods as per the data types and then execute.
Please make sure, as discussed, that other kinds of setter combinations are covered elsewhere in the tests. Please update this PR with some information on those.

cluster/src/dunit/scala/org/apache/spark/sql/TPCHDUnitTest.scala
```diff
@@ -77,6 +78,259 @@ class TPCHDUnitTest(s: String) extends ClusterManagerTestBase(s)
...
+      result += row
+    }
+    // scalastyle:off println
+    println(s"Number of rows : $count")
+    // scalastyle:on println
+
+    val expectedFile = sc.textFile(getClass.getResource(
+      s"/TPCH/RESULT/$expectedResultFile").getPath)
+    val expectedNoOfLines = expectedFile.collect().size
+    assert(count == expectedNoOfLines)
```
kneeraj: Just dump the actual results in a csv and may be you can compare the actual results as well.

dshirish: Done

cluster/src/dunit/scala/org/apache/spark/sql/TPCHDUnitTest.scala
dshirish: Done, now using other setter methods in some of the queries. Other tests in PreparedQueryRoutingSingleNodeSuite and PreparedQueryRoutingDUnitTest cover  setter methods such as setInt, setDouble, setString, setTimestamp, setDate, setBigDecimal, setLong.

PR #1340, opened 2019-07-03
Added a fix and test for SNAP-3024

dshirish:
- In CachedDataFrame#collectWithHandler, for one case, query/job is run using sc.runJob(). Here if sc.runJob fails due to stale catalog, there was no code invalidate the cached metadata and retry the query. Added this code.
- Added a dunit test

## Patch testing
New dunit test
precheckin

## ReleaseNotes.txt changes
NA

## Other PRs 
NA

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
+++ b/core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
@@ -17,6 +17,7 @@
 package org.apache.spark.sql
 
 import java.nio.ByteBuffer
+import java.sql.SQLException
 
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/sql/CachedDataFrame.scala
```diff
@@ -401,13 +404,32 @@ class CachedDataFrame(snappySession: SnappySession, queryExecution: QueryExecuti
...
+                  val execution =
+                    snappySession.getContextObject[() => QueryExecution](SnappySession.ExecutionKey)
+                  execution match {
+                    case Some(exec) =>
+                      val execRDD = exec().executedPlan.execute()
+                      runAsJob(execRDD, processPartition, resultHandler, sc)
+                    case _ => throw t
+                  }
+                } catch {
+                  case t: Throwable
```
suranjan: Should we retry for a few times or indefinitely before giving up? The same strategy in other code paths as well. Otherwise, the user has to catch this exception always.

dshirish: Yes, changed code to retry multiple times in case of failure as discussed.

PR #1344, opened 2019-07-05
SNAP-2636 Updating the UDF jar causes cluster to be unusable on restart

ashetkar:
Changes:
- Users can recreate a function with the same jar without restarting the cluster.
- DROP FUNCTION now removes jar from server work directories also.
- LIST JARS lists the UDFs also.

## Patch testing
Manually tested and added a new testcase.
precheckin is running.

## ReleaseNotes.txt changes
TODO

## Other PRs 
NA


core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitSecurityTest.scala
```diff
@@ -874,6 +875,226 @@ class SplitClusterDUnitSecurityTest(s: String)
...
+    }).foreach(x => println(s"BEFORE DROP  [snappy-jars]: ${x.getAbsolutePath}"))
+
+
+    // Drop a function of jdbcUser2
+    executeSQL(stmt2, s"drop function myUDF")
+
+    // Verify function does not execute
+    stmt2 = user2Conn.createStatement()
+    try {
+      stmt2.execute(s"select myUDF('abcd')")
```
kneeraj: Check if user1.myUdf works.

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitSecurityTest.scala
```diff
@@ -874,6 +875,226 @@ class SplitClusterDUnitSecurityTest(s: String)
...
+    rows = 0
+    if (rs ne null) {
+      while (rs.next()) {
+        rows += 1
+        val udfName = rs.getString(1)
+        assert(udfName.equalsIgnoreCase(s"[UDF]$jdbcUser1.myUDF"),
+          s"Unexpected UDF name $udfName")
+      }
+    }
+
```
kneeraj: just assert count == 1 too for sanity.

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitSecurityTest.scala
```diff
@@ -874,6 +875,226 @@ class SplitClusterDUnitSecurityTest(s: String)
...
```
kneeraj: Same comment as above. assert cnt == 1

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitSecurityTest.scala
```diff
@@ -874,6 +875,226 @@ class SplitClusterDUnitSecurityTest(s: String)
...
+    stmt1 = user1Conn.createStatement()
+    executeSQL(stmt1, s"DROP FUNCTION myUDF")
+
+    stmt2 = user1Conn.createStatement()
+    try {
+      stmt2.execute(s"select myUDF('abcd')")
+      assert(false, s"Expected an exception!")
+    } catch {
+      case _: SQLException => // expected
```
kneeraj: Please check for actual error expected. May be later but please note.

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitSecurityTest.scala
```diff
@@ -874,6 +875,226 @@ class SplitClusterDUnitSecurityTest(s: String)
...
+      case t: Throwable => assert(false, s"Unexpected exception $t")
+    }
+
+    // Verify list jars
+    stmt2.execute(s"list jars")
+    rs = stmt2.getResultSet
+    if (rs ne null) {
+      while (rs.next()) {
```
kneeraj: you can assert just rs.next() == false.

rs.next() == true is an illegal state.

core/src/main/scala/org/apache/spark/sql/internal/JarUtils.scala
```diff
@@ -98,6 +116,35 @@ object ContextJarUtils extends Logging {
     if (!jarDirectory.exists()) jarDirectory.mkdir()
     jarDirectory
   }
+
+  def addToTheListInCmdRegion(k: String, item: String, head: String): Unit = {
```
kneeraj: I won't suggest to take care of making it transactional but at least put in the docs that parallel 'Create/Drop Function' from different threads should not be done.

core/src/main/scala/org/apache/spark/sql/internal/SnappySessionCatalog.scala
```diff
@@ -752,7 +753,7 @@ class SnappySessionCatalog(val externalCatalog: SnappyExternalCatalog,
   private def removeFromFuncJars(funcDefinition: CatalogFunction,
       qualifiedName: FunctionIdentifier): Unit = {
     funcDefinition.resources.foreach { r =>
-      ContextJarUtils.deleteFile(funcDefinition.identifier.toString(), r.uri)
+      ContextJarUtils.deleteFile(funcDefinition.identifier.toString(), r.uri, isEmbeddedMode())
```
kneeraj: Dropping from smart connector will not clean the lead file neither the embedded server's local file. We need to clean both smart connector cluster as well as embedded cluster. You can call a stored procedure or something for the smart connector. Please check GET_DEPLOYED_JARS procedure. We can have a similar procedure for this cleanup.

ashetkar:
Will take care of review comments in the next PR targeted for SNAP 3069 or SNAP 3062.

cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -83,16 +84,21 @@ class SnappyExecutor(
...
+        if (!ContextJarUtils.checkItemExists(ContextJarUtils.droppedFunctionsKey, appName)) {
+          logInfo(s"Creating ClassLoader for $appName" +
+              s" with dependencies $appDependencies")
+          urls = appDependencies.map(name => {
+            val localName = name.split("/").last
+            logInfo(s"Fetching file $name for App[$appName]")
+            Utils.fetchFile(name, new File(SparkFiles.getRootDirectory()), conf,
+              env.securityManager, hadoopConf, -1L, useCache = !isLocal)
+            val url = new File(SparkFiles.getRootDirectory(), localName).toURI.toURL
+            Misc.getMemStore.getGlobalCmdRgn.put(ContextJarUtils.functionKeyPrefix + appName, name)
```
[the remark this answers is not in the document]
ashetkar: These entries need to be retained, simply to be in-sync with function definitions which are persistent.

core/src/dunit/scala/io/snappydata/cluster/SplitClusterDUnitSecurityTest.scala
ashetkar: Yes, that would be cleaner, will do. BTW, I'm asserting FALSE right after `rs.next()` which ensures that `rs.next() == true` is invalid.

PR #1347, opened 2019-07-10
Deploy command allows names with hyphens and dots [SNAP-2355]

paresh-p11:

* A new rule that allows '-' and '.' in package names

## Patch testing

Ran precheckin.

## ReleaseNotes.txt changes

## Other PRs 


core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
```diff
@@ -311,12 +311,26 @@ abstract class SnappyBaseParser(session: SparkSession) extends Parser {
     quotedIdentifier
   }
 
+  protected final def packageIdentifierPart: Rule1[String] = rule {
+    atomic(capture((Consts.identifier | Consts.hyphen | Consts.dot).+)) ~ delimiter ~> { (s: String) =>
```
kneeraj: delimiter rule is not required I guess. Just ws( whitespaces ) should be fine, otherwise wrong sytaxes will also start getting parsed. Please check.

cluster/src/dunit/scala/io/snappydata/cluster/SplitSnappyClusterDUnitTest.scala
```diff
@@ -257,6 +257,53 @@ class SplitSnappyClusterDUnitTest(s: String)
...
+
+    val sns = new SnappySession(sc)
+    sns.sql("deploy package  mongo_spark 'org.mongodb.spark:mongo-spark-connector_2.11:2.2.2'")
+    sns.sql("deploy package mongo-spark_v1.0  'org.mongodb.spark:mongo-spark-" +
+        "connector_2.11:2.2.2'")
+    sns.sql("deploy package app.mongo-spark_v1.1  'org.mongodb.spark:mongo-spark-" +
+        "connector_2.11:2.2.2'")
+    sns.sql("deploy package testsch.mongo-spark_v1.2  'org.mongodb.spark:mongo-spark" +
+        "-connector_2.11:2.2.2'")
+    sns.sql("deploy package \"testsch\".\"mongo-spark_v1.3\"  'org.mongodb.spark:mongo" +
```
kneeraj: Use s instead of backslash.

PR #1351, opened 2019-07-11
Snap 2228 fix

dshirish:

Added check for schema version when insert/update/delete operation are done from smart connector mode. On schema version change, now CatalogStaleException will be thrown and cached catalog tables will be cleared so that schema is fetched from embedded side.

Added new unit tests

## Patch testing
New unit tests
precheckin


dshirish:
> Looks good.
> One test for rebalancing could be added as that also triggers version change.

Thanks!  Yes will test for rebalancing scenario as well (tracked thru SNAP-3025)

PR #1352, opened 2019-07-11
Snap 3061

ahshahid:
The column table insert code takes a diskstore level read lock in the begin transaction & releases it in the commit  or rollback.
The idea is that if the tx thread finds, that the standby oplog is in use, an InsufficientDiskSpaceException is thrown at oplog level to indicate the transaction to be rolledback. Also at oplog level when the oplog is switched to standby oplog, an asynchronous thread is started to shut the down the diskstore. This asynch thread seeks to take a write lock on the diskstore. The idea is that the diskstore should not be shut so long as any active transactions are not rolled back. Once the transaction is ended, the diskstore read lock is released, allowing the asynch diskstore shutdown thread to close the diskstore & regions using that diskstore.

## Patch testing

Ran the store:check target.

## Other PRs 
[store-change](https://github.com/SnappyDataInc/snappy-store/pull/485)


ahshahid:
For some reason, I am getting Exception with this change. The batch update fails. Not sure what wrong I am doing , as I am taking the lock in the begin tx part & releasing on commit/rollback

ahshahid:
Looks like for some partition task, the table name is coming in lower case, causing the problem.

PR #1354, opened 2019-07-15
[SNAP-3023] Handle CatalogStaleException with retries in snappy sink

vatsalmevada:

Retrying streaming batch processing for the snappy sink when `CatalogStaleException` is encountered.  
Currently, number of attempts are set to 10 wtih backoff interval of `100 * attempt_number` millis.   
Also exposed an internal property (`internal___attempts`) to overwrite this for tests.

## Patch testing

Added a Dunit test, added scala tests

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
+++ b/core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
@@ -23,14 +23,12 @@ import java.util.NoSuchElementException
 import com.pivotal.gemfirexd.internal.shared.common.reference.SQLState.{LOGIN_FAILED, SNAPPY_CATALOG_SCHEMA_VERSION_MISMATCH}
 import io.snappydata.Property._
 import io.snappydata.util.ServiceUtils
-import org.apache.log4j.Logger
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/sql/streaming/SnappySinkCallback.scala
```diff
@@ -149,15 +151,40 @@ case class SnappyStoreSink(snappySession: SnappySession,
...
+  private def processBatchWithRetries(batchId: Long, data: Dataset[Row], possibleDuplicate: Boolean,
+      retries: Int = 5) : Unit = {
+    try {
+      sinkCallback.process(snappySession, parameters, batchId, convert(data), possibleDuplicate)
+    } catch {
+      case ex: Exception if retries == 0 || !isRetriableException(ex) => throw ex
+      case ex: Exception =>
+        logWarning(s"Encountered a retriable exception. Will retry processing batch." +
+            s" Retries left: $retries" , ex)
+        processBatchWithRetries(batchId, data, possibleDuplicate = true, retries - 1)
```
suranjan: Please use backoff for retry.

PR #1360, opened 2019-07-19
Support interval expressions

sumwale:
Add expressions in addition to interval literals. Currently product has no way to construct a
CalendarInterval value given an expression (except for string concatenation and cast from string to
interval). With this change expressions like "timestamp + interval hour(time) hours" can be used.

- added new IntervalExpression implementation that generates the code required for a set of expressions
  each representing a long value with units specified separately
- parser support for the new interval expressions as well as specifying multiple intervals in a single
  clause e.g. interval 4 hours 3 minutes
- unit test for interval expressions and literals
- split out some independent classes into separate files from ParamLiteral.scala
- add "hadoop." and "javax.jdo." prefixes to be passed through as system properties

## Patch testing

precheckin

## ReleaseNotes.txt changes

document support for interval literals and expressions

## Other PRs 

NA


sumwale:
Merging now to ease subsequent PRs. Will take care of any review comments subsequently.

PR #1362, opened 2019-07-19
[Snap 2828] Serialize the write ops on Column Table

suranjan:

Take Write lock on table in case of Insert, PutInto, Update and Delete.
Release them on completion of operation. This is to serialize the write operations to avoid inconsistency.
Handles the smart connector case where we use procedure to take and release lock for a write operation on the table.

## Patch testing

precheckin and hydra tests, manual tests.
## ReleaseNotes.txt changes

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/488


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -472,12 +471,11 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
-      case _: ThinClientConnectorMode => null
-      case _ =>
-        PartitionedRegion.getRegionLock("PUTINTO_" + table, GemFireCacheImpl.getExisting)
-    }
-    if (lock ne null) lock.lock()
+    val (schemaName: String, _) =
+      JdbcExtendedUtils.getTableWithSchema(table, conn = null, Some(sqlContext.sparkSession))
+    val lock = grabLock(table, schemaName, defaultConnectionProps)
+
+    if (lock.isInstanceOf[RegionLock]) lock.asInstanceOf[RegionLock].lock()
```
kneeraj: For uniformity I think you can move the lock acquisition inside grabLock method itself even for embedded mode. Cosmetic change, not insisting.

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -265,22 +271,59 @@ abstract class BaseColumnFormatRelation(
...
+          stmt.close()
+          result
+        } finally {
+          connection.commit()
+          connection.close()
+        }
+      }
+    }
+    finally {
+      logDebug(s"Added the ${lock} object to the context. in InsertRows")
```
kneeraj: This log might be a little misleading as even on failure this might get logged. Right? May be move it to actually where the lock is being taken.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -518,18 +511,100 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
+              if (retrycount == 2) {
+                throw e
+              }
+            }
+          }
+          finally {
+            retrycount = retrycount + 1
+            // conn.close()
+          }
+        } while (!locked)
```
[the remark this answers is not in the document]
suranjan: retrycount is in exception handling.
Without exception, the locked value has to be true.

PR #1365, opened 2019-07-22
Snap 3045

ahshahid:
The bug in the case of outer join is due to the middle while(true) loop in HashJoinExec  being broken as soon as the extra filter condition evaluates to false, with the marker entryVar set to null.
The expected behaviour is that if say the stream side row has join condition fullfilled with say 5 build side rows, then the middle while(true) loop should check if the build side row has filter condition fullfilled by any or all of those 5 rows ( & each inclusion becoming part of results).
Iff the stream side row does not get included with any of the 5 rows, then only it should be included in the resultset as outer join row , with build side tuple as null.
What was happening was that if  for the first build side row , the  filter condition was evaluated as false, then the other 4 rows were not being considered at all, & the stream side row was getting added to the result with build side row as null.

Though the change in this PR fixes the issue seen, I am uncertain about the following behaviour 
Lets say 
T1 outer join T2 on T1.x = T2.x and T2. a is not null and T1.b = 'abd'

type of query.
In the above case lets say a particular T1 row is not getting included because of absence of join or due to filter condition.
In such case, if that T1 row is inlcuded, then all fields of T2 will be null, which will violate the condition of the query.
I am not sure if the right approach is to re-evaluate the filter conditions with all build side variable as null & then includ/exclude the row.
Similar case I suppose can happen with T1.b condition.

In fact, my thinking of outer join working was that if an outer Table T1 's row has an equi join match with any available build side row, but is getting eliminated due to the filter condition of any sort, then that outer row should not be inlcuded at all. But with that behaviour the query has some missing results as compared to the sort based join ...

So right now the fix includes the outer row if it is not at all included due to filter or join condition not matching.

## Patch testing

will be running precheckin

In aqp test one query inovolving outer join & hash aggregate has failed due to NPE. The cause is most likely due to 3045 change, but I think the issue lies in the code of hash aggregate excec ( this is seen in both current master as well as new code) so I suspect something like, the attribute says not null, while the outer join causes a null to be returned.. I will  be looking into it tomorrow. 

## ReleaseNotes.txt changes

## Other PRs 


ahshahid:
The failure in AQP is due to some bug in the hamdling of null, in Bootstrap Sum code which I am working on fixing.

ahshahid:
The AQP query failure has been fixed with checkin done on master ( ticket ENT-61)

ahshahid:
However, this PR would require some new tests . Because Jags verified that in PostGres extra conditions if present are used to see if the outer join row with accompanying null counter part  passes through the filter  condition or not. In our case looks like we are not reapplying the filter. It is possible that there are issues in other forms of Outer Joins too. Will be working on it tomorrow.

ahshahid:
Ok. I have verified there are no issues either in this PR & the other form of outer joins.
I was not aware of the difference in the behaviour of conditions specified in the "ON" clause vis a vis "where" clause. 
Any condition present in where clause is applied on the out put of the outer join results  conceptually.
While if the outer table row does not get selected due to any condition failing in "ON" clause, then that outer row is selected with inner row as null

PR #1368, opened 2019-07-23
Fixing [SNAP-2653]

paresh-p11:
* Mask credentials (in case of s3 URI) in Describe extended/formatted output.
* Mask credentials in case of s3 on UI for external tables.
* Disallow access non-admin user to the tables in SNAPPY_HIVE_METASTORE.
* Added tests for the fixes.

(Fill in the changes here)

## Patch testing

Precheckin is in progress

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
+++ b/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
@@ -209,15 +209,8 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
         // External Tables
         hiveTables.collect {
           case table if table.tableType == CatalogObjectType.External.toString =>
-            val tableSrcPath = table.dataSourcePath
-            val maskedSrcPath = if (tableSrcPath.toLowerCase().startsWith("s3a://") ||
-                tableSrcPath.toLowerCase().startsWith("s3://") ||
-                tableSrcPath.toLowerCase().startsWith("s3n://") ) {
-              tableSrcPath.replace(tableSrcPath.slice(tableSrcPath.indexOf("//") + 2,
-                tableSrcPath.indexOf("@")), "****:****")
-          } else tableSrcPath
             new SnappyExternalTableStats(table.entityName, table.tableType, table.schema,
-              table.shortProvider, table.externalStore, maskedSrcPath, table.driverClass)
+              table.shortProvider, table.externalStore, table.dataSourcePath, table.driverClass)
         }
       }
       catch {
--- a/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
+++ b/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
@@ -209,8 +209,15 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
         // External Tables
         hiveTables.collect {
           case table if table.tableType == CatalogObjectType.External.toString =>
+            val tableSrcPath = table.dataSourcePath
+            val maskedSrcPath = if (tableSrcPath.toLowerCase().startsWith("s3a://") ||
+                tableSrcPath.toLowerCase().startsWith("s3://") ||
+                tableSrcPath.toLowerCase().startsWith("s3n://") ) {
+              tableSrcPath.replace(tableSrcPath.slice(tableSrcPath.indexOf("//") + 2,
+                tableSrcPath.indexOf("@")), "****:****")
+          } else tableSrcPath
             new SnappyExternalTableStats(table.entityName, table.tableType, table.schema,
-              table.shortProvider, table.externalStore, table.dataSourcePath, table.driverClass)
+              table.shortProvider, table.externalStore, maskedSrcPath, table.driverClass)
         }
       }
       catch {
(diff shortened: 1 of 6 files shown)


core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -209,8 +209,15 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
...
+            val tableSrcPath = table.dataSourcePath
+            val maskedSrcPath = if (tableSrcPath.toLowerCase().startsWith("s3a://") ||
+                tableSrcPath.toLowerCase().startsWith("s3://") ||
+                tableSrcPath.toLowerCase().startsWith("s3n://") ) {
+              tableSrcPath.replace(tableSrcPath.slice(tableSrcPath.indexOf("//") + 2,
+                tableSrcPath.indexOf("@")), "****:****")
+          } else tableSrcPath
             new SnappyExternalTableStats(table.entityName, table.tableType, table.schema,
-              table.shortProvider, table.externalStore, table.dataSourcePath, table.driverClass)
+              table.shortProvider, table.externalStore, maskedSrcPath, table.driverClass)
```
sumwale: Will be better to change in StoreHiveCatalog.getDataSourcePath so that both here and sys.hivetables will mask the path. Also will handle masking passwords in sourcePath, if any. Patch to mask for all cases:

```
diff --git a/core/src/main/scala/io/snappydata/sql/catalog/impl/StoreHiveCatalog.scala b/core/src/main/scala/io/snappydata/sql/catalog/impl/StoreHiveCatalog.scala
index 4921e6df9..00b8214ee 100644
--- a/core/src/main/scala/io/snappydata/sql/catalog/impl/StoreHiveCatalog.scala
+++ b/core/src/main/scala/io/snappydata/sql/catalog/impl/StoreHiveCatalog.scala
@@ -444,16 +447,19 @@ class StoreHiveCatalog extends ExternalCatalog with Logging {
       }
     }

+    private def maskPassword(s: String): String =
+      SnappyExternalCatalog.PASSWORD_MATCH.replaceAllIn(s, "xxx")
+
     private def getDataSourcePath(properties: scala.collection.Map[String, String],
         storage: CatalogStorageFormat): String = {
       properties.get("path") match {
-        case Some(p) if !p.isEmpty => p
+        case Some(p) if !p.isEmpty => maskPassword(p)
         case _ => properties.get("region.path") match { // for external GemFire connector
-          case Some(p) if !p.isEmpty => p
+          case Some(p) if !p.isEmpty => maskPassword(p)
           case _ => properties.get("url") match { // jdbc
             case Some(p) if !p.isEmpty =>
               // mask the password if present
-              val url = SnappyExternalCatalog.PASSWORD_MATCH.replaceAllIn(p, "xxx")
+              val url = maskPassword(p)
               // add dbtable if present
               properties.get(SnappyExternalCatalog.DBTABLE_PROPERTY) match {
                 case Some(d) if !d.isEmpty => s"$url; ${SnappyExternalCatalog.DBTABLE_PROPERTY}=$d"

```

core/src/main/scala/io/snappydata/sql/catalog/impl/StoreHiveCatalog.scala
```diff
@@ -434,24 +434,40 @@ class StoreHiveCatalog extends ExternalCatalog with Logging {
...
+    }
+
+    // Mask access key and secret access key in case of S3 URI
+    private def maskLocationURI(locURI: String): String = {
+      val maskedSrcPath = if (locURI.toLowerCase().startsWith("s3a://") ||
+          locURI.toLowerCase().startsWith("s3://") ||
+          locURI.toLowerCase().startsWith("s3n://") ) {
+        locURI.replace(locURI.slice(locURI.indexOf("//") + 2,
+          locURI.indexOf("@")), "****:****")
+      } else locURI
```
sumwale: change to "else maskPassword(locURI)" since location URI can also contain passwords

core/src/main/scala/io/snappydata/sql/catalog/impl/StoreHiveCatalog.scala
```diff
@@ -434,24 +434,40 @@ class StoreHiveCatalog extends ExternalCatalog with Logging {
       }
     }
 
+    private def maskPassword(s: String): String = {
+      SnappyExternalCatalog.PASSWORD_MATCH.replaceAllIn(s, "xxx")
```
sumwale: add "|secret" to SnappyExternalCatalog.PASSWORD_MATCH regex

core/src/main/scala/io/snappydata/sql/catalog/impl/StoreHiveCatalog.scala
```diff
@@ -434,24 +434,40 @@ class StoreHiveCatalog extends ExternalCatalog with Logging {
...
-        case Some(p) if !p.isEmpty => p
+        case Some(p) if !p.isEmpty => maskPassword(p)
         case _ => properties.get("region.path") match { // for external GemFire connector
-          case Some(p) if !p.isEmpty => p
+          case Some(p) if !p.isEmpty => maskPassword(p)
           case _ => properties.get("url") match { // jdbc
             case Some(p) if !p.isEmpty =>
               // mask the password if present
-              val url = SnappyExternalCatalog.PASSWORD_MATCH.replaceAllIn(p, "xxx")
+              val url = maskPassword(p)
```
sumwale: After changing maskLocationURI() to invoke maskPassword() as suggested, change all maskPassword calls to maskLocationURI. Specifically the "path" property can be present with "s3" URL in addition to storage.locationUri

PR #1373, opened 2019-07-23
Code changes for SNAP-2779 and SNAP-1338:

snappy-sachin:

 - Adding Redundancy column in Tables List to view count of redundant copies.
 - Adding Redundancy Status column in Tables List to monitor redundancy has been satisfied or broken.
 - Changes for maintaining Redundancy and isRedundancyImpaired details for count of redundant
   copies and redundancy satisfaction/broken status.

## Patch testing

 - Tested manually

## ReleaseNotes.txt changes

 - Yes, to be updated.

## Other PRs 

 **Spark:** https://github.com/SnappyDataInc/spark/pull/160
 **Store:** https://github.com/SnappyDataInc/snappy-store/pull/495

### DIFF
--- a/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
+++ b/core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
@@ -26,6 +26,7 @@ import scala.collection.JavaConverters._
 import scala.collection.mutable
 import scala.language.implicitConversions
 import scala.util.control.NonFatal
+import scala.util.control.Breaks._
 
 import com.gemstone.gemfire.CancelException
 import com.gemstone.gemfire.cache.execute.FunctionService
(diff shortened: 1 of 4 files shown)


core/src/main/scala/io/snappydata/SnappyTableStatsProviderService.scala
```diff
@@ -227,6 +228,27 @@ object SnappyEmbeddedTableStatsProviderService extends TableStatsProviderService
...
+          val PrRegRedProvider = tablePrRegion.getRedundancyProvider
+          if (PrRegRedProvider != null) {
+            rs.setRedundancyImpaired(PrRegRedProvider.isRedundancyImpaired)
+            rs.setRedundancy(tablePrRegion.getRedundantCopies)
+          }
+
+          val numBuckets = tablePrRegion.getPartitionAttributes.getTotalNumBuckets
+          breakable {
+            for (i <- 0 until numBuckets) {
+              val idm = tablePrRegion.getNodeForBucketRead(i)
```
kneeraj: Test this for time. Use a schema which has multiple tables with a lot of buckets and please check the time it takes to complete all calls. Since this is going to be called for every run so we should check the cost of this.
For example take at least 100 tables with default number of bucket which means 12800 calls will be done.

snappy-sachin: Acknowledged, will check it and will enhance the code in next release.

PR #1380, opened 2019-07-26
Snap 3055

hemanthmeka:

review changes for previous changes

## Patch testing

manually tested suite and failure scenarios

## ReleaseNotes.txt changes

no

## Other PRs 

None


core/src/main/scala/org/apache/spark/sql/hive/SnappyHiveExternalCatalog.scala
```diff
@@ -388,17 +384,13 @@ class SnappyHiveExternalCatalog private[hive](val conf: SparkConf,
...
         // AnalysisException not thrown while getting table. suspecting that wrong table
         // name is passed. throwing exception as a precaution.
-        throw StandardException.newException(
-          SQLState.LANG_UNEXPECTED_USER_EXCEPTION, null, "Table retrieved successfully. To " +
-              "continue to drop this table change FORCE_DROP argument in procedure to true");
+        throw new AnalysisException("Table retrieved successfully. To " +
+            "continue to drop this table change FORCE_DROP argument in procedure to true");
       }
     } catch {
       case a: AnalysisException if (a.message.contains("might be inconsistent in hive catalog")) =>
```
dshirish: It might be better to have the message stored in one constant and then use it at places where required. Or create a new exception of and catch it here.

PR #1381, opened 2019-07-26
Use synchronizedMap for DynamicReplacableConstant#termMap

dshirish:

Use synchronizedMap for DynamicReplacableConstant#termMap to avoid it getting corrupted when same ParamLiteral is used concurrently in multiple threads. 
This fixes SNAP-2764 (seen in case of subqueries). 

## Patch testing
precheckin

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


dshirish:
Merging   the change and it has been already discussed with @sumwale

PR #1392, opened 2019-07-29
SNAP-3069 Fixing a test failure seen in precheckin.

ashetkar:
* Do not fetch jar file of a dropped function, even if the jar url is part of existing function's jar urls.
 This is because we set all jar urls as session dependency and in some cases those may end up being fetched by the executor, triggering the failure.

## Patch testing
precheckin

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


cluster/src/main/scala/org/apache/spark/executor/SnappyExecutor.scala
```diff
@@ -89,15 +89,28 @@ class SnappyExecutor(
...
             val localName = name.split("/").last
-            logInfo(s"Fetching file $name for App[$appName]")
-            Utils.fetchFile(name, new File(SparkFiles.getRootDirectory()), conf,
-              env.securityManager, hadoopConf, -1L, useCache = !isLocal)
-            val url = new File(SparkFiles.getRootDirectory(), localName).toURI.toURL
-            Misc.getMemStore.getGlobalCmdRgn.put(ContextJarUtils.functionKeyPrefix + appName, name)
-            url // points to the jar in executor's work directory
+            var fetch = true
+            val firstHyphen = localName.indexOf("-")
+            if (firstHyphen > -1) {
```
kneeraj: Just write the logic in comment so that anyone doing a maintenance here can easily understand.

PR #1394, opened 2019-07-29
Generic support for INSERT/PUT ... VALUES with partial column spec

sumwale:
Changes allow the following to be supported in addition to currently supported ones on all tables:

```sql
// simple constructs
INSERT (INTO|OVERWRITE) <table> VALUES ...
PUT INTO <table> VALUES ...
```

```sql
// with partial columns that can be out of order with table definition
INSERT (INTO|OVERWRITE) <table>(<col1>...) VALUES ...
INSERT (INTO|OVERWRITE) <table>(<col1>...) SELECT ...
PUT INTO <table>(<col1>...) VALUES ...
PUT INTO <table>(<col1>...) SELECT ...
```

- add SnappyParser "inlineTable" rule to insert/put and removed "subSelectQuery" rule
  that was specifically added to avoid the "inlineTable" rule
- enhanced the AnalyzeMutableOperations rule to deal with the case of "partial columns"
  which is parsed as a TableValueFunction; locate the columns in target table and
  add appropriate Project on child
- capture the query string for possible later use in "partial column" construct for
  row tables where explicit non-null DEFAULT value for a column may have been specified;
  for this case switch to DMLExternalTable which may fail if query contains Spark-only functions
- removed custom PutIntoValuesColumnTable and use consistent way for both INSERT and PUT
- handle the QuestionMark case in VALUES() in a generic way as LocalRelation resolved by Spark
- removed ResolveRelationsExtended and instead made table as proper child of DMLExternalTable
  so that it is resolved in the normal way by ResolveRelations

New unit tests to be added shortly.

## Patch testing

precheckin

## ReleaseNotes.txt changes

Document support for VALUES and partial column specification in INSERT/PUT. Limitation of
not being able to use Spark-only functions with partial column specification for ROW tables.

## Other PRs 

NA


dshirish (commented):
For complex datatypes,  will we still continue using "insert into<> select <>" syntax?

dshirish (approved):
Looks like unit tests mentioned are to be pushed.
However the code change itself looks good to me.

sumwale:
> For complex datatypes, will we still continue using "insert into<> select <>" syntax?

No, ARRAY/STRUCT/MAP will work now in VALUES like they do in SELECT.

PR #1400, opened 2019-07-31
Snap 3048 2

suranjan:

Added a wrapper catalog for PreWriteCheck to esnsure it gets r esolved LogicalRelation instead of SimpleCatalogRelation.
Changed the equals and hashcode for JDBCMutableRelation and JDBCAppendableRelation for correct equality checks.

## Patch testing

Added test and precheckin.

## ReleaseNotes.txt changes

## Other PRs 

https://github.com/SnappyDataInc/snappy-aqp/pull/190


sumwale:
@suranjan Can you explain why Spark's PreWriteCheck does not work? In fact the code you added seems almost identical to that in PreWriteCheck.

suranjan:
> @suranjan Can you explain why Spark's PreWriteCheck does not work? In fact the code you added seems almost identical to that in PreWriteCheck.

Not able to exactly debug due to IntelliJ issue but, I could see that the plan returned by catalog.lookupRelation is not matching the rules there.

Plan returned by catalog.lookupRelation:

'SimpleCatalogRelation app, CatalogTable(
        Table: `app`.`columntable`
        Owner: skumar
        Created: Wed Jul 31 12:38:07 IST 2019
        Last Access: Thu Jan 01 05:30:00 IST 1970
        Type: EXTERNAL
        Schema: [StructField(col1,IntegerType,true), StructField(col2,IntegerType,true), StructField(col3,IntegerType,true)]
        Provider: column
        Properties: [transient_lastDdlTime=1564556887]
        Storage(Location: , InputFormat: org.apache.hadoop.mapred.SequenceFileInputFormat, OutputFormat: org.apache.hadoop.hive.ql.io.HiveSequenceFileOutputFormat, Serde: org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe, Properties: [BUCKETS=1, serialization.format=1, column_max_delta_rows=10000, PARTITION_BY=Col1, dbtable=app.columntable, schemaddl.numParts=1, schemaddl.part.0=(Col1 INT ,Col2 INT, Col3 INT), column_batch_size=25165824]))

We may be missing more rules due to this, so still investigating while this can be additional check.

sumwale:
@suranjan Do you mean the comparison fails due to upper-case table name? If so, then isn't a better fix to override toString in JDBCAppendableRelation and JDBCMutableRelation (and use equalsIgnoreCase on resolveName and also compare the mode).

Btw, about debugging in IDEA do you use "-DideaBuild" option to Gradle VM options under "Build, Execution, Deployment -> Build Tools -> Gradle"? This changes the import to use snappy-spark for core module and dependents instead of upstream Apache Spark. Just set it in gradle and refresh which should be enough and confirm that dependencies of snappy-core module show snappy-spark-core etc instead of spark-core and all.

suranjan:
No, the match is failing because we return SimpleCatalogRelation in case of the snappy table too, whereas spark returns the plan. In case of spark, SimpleCatalogRelation is returned in case of view only.

suranjan:
Thanks for the IntelliJ recommendation it worked.

sumwale:
> No, the match is failing because we return SimpleCatalogRelation in case of the snappy table too, whereas spark returns the plan. In case of spark, SimpleCatalogRelation is returned in case of view only.

Do you mean the result of catalog.lookupRelation? But SessionCatalog.lookupRelation also returns SimpleCatalogRelation which is what we followed. So what has changed between Spark and Snappy? This is assuming that if you create a table in spark-shell using spark, for example, then insert into the same table reading from itself fails. If it does, then need to see which rule applies there and why it doesn't in our case.

suranjan:
Spark returns the table relation itself it is a datasource table and returns SimpleCatalogRelation in case of views.
Lokked at other rules where it handles SimpleCatalogRelation but not in PreWriteCheck, because we can't write to views.

sumwale:
> Spark returns the table relation itself it is a datasource table and returns SimpleCatalogRelation in case of views.
> Lokked at other rules where it handles SimpleCatalogRelation but not in PreWriteCheck, because we can't write to views.

Here is the code in Spark's SessionCatalog: https://github.com/SnappyDataInc/spark/blob/snappy/branch-2.1/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/catalog/SessionCatalog.scala#L575

It clearly returns a SimpleCatalogRelation in all cases. VIEW only adds the additional argument to it. What am I missing?

That said I understand why the rule in CreateTable won't apply. It will not apply to Spark tables either. In any case it isn't useful for tables to begin with because you cannot create a table selecting from a table that does not exist. My question is why the PreWriteCheck rule for InsertIntoTable does not apply. It is doing the simple thing of collecting all BaseRelations from query and searching for the table. The only difference I see in your code is table name upper-case comparison which is easily fixable by overriding the equals in JDBCMutableRelation and JDBCAppendableRelation.

sumwale:
> That said I understand why the rule for CreateTable won't apply. It will not apply to Spark tables either. In any case it isn't useful for tables to begin with because you cannot create a table selecting from a table that does not exist.

Take that back. CreateTable with overwrite plan will be created for the saveAsTable() API though cannot happen using SQL. So question is what is different between Spark's SessionCatalog.lookupRelation and SnappySessionCatalog.lookupRelation.

sumwale:
Ok I understand now. SparkSession when using the default in-memory catalog will have the same problem of truncating the target table with overwrite -- tested with JDBC source. But with hive catalog, it does return the LogicalRelation in HiveMetastoreCatalog.cachedDataSourceTables.load.

I think we are better off mimicking HiveSessionCatalog behaviour rather than changing analysis rules because there may be more. The code in SnappySessionCatalog's resolveRelationWithAlias can be part of lookupRelation() last else block (line 844 in current code) and former method can be removed.

suranjan:
Yes that's better, Will do that. I was looking for ways to return the LogicalRelation to match the spark behavior.

I tried locally on spark, although, a different version and it failed for saveAsTable and insertInto (for insert into it failed from a new code path, not in 2.1.1).

sumwale:
> Yes that's better, Will do that. I was looking for ways to return the LogicalRelation to match the spark behavior.

Sorry, changed my mind after looking at Spark 2.3/2.4 code. Both of those don't have any special handling for HiveSessionCatalog vs SessionCatalog and return same UnresolvedCatalogRelation, so the code on the spark-multiversion-support branch assumes the same. Searching through code, only PreWriteCheck seems to lack handling for CatalogRelation and assumes resolved one. So the CreateTable check should be fine but insert one should not be required. Will review the PR accordingly.

> I tried locally on spark, although, a different version and it failed for saveAsTable and insertInto (for insert into it failed from a new code path, not in 2.1.1).

Perhaps you tried from spark-shell that enables hive by default.

core/src/main/scala/org/apache/spark/sql/internal/session.scala
```diff
@@ -515,6 +515,21 @@ private[sql] final class PreprocessTable(state: SnappySessionState) extends Rule
...
+        EliminateSubqueryAliases(queryOpt.get) match {
+          case plan: LogicalPlan =>
+            val srcRelationNames = plan.collect {
+              case LogicalRelation(src: MutableRelation, _, _) => src.table
+            }
+            // we store table as Uppercase in relation
+            if (srcRelationNames.contains(tableIdent.unquotedString.toUpperCase)) {
+              throw new AnalysisException(
+                s"Cannot overwrite table ${tableDesc.identifier} that is also being read from")
+            }
```
sumwale: Why only check for MutableRelation because the behaviour seems to be a problem for other data sources too. Since PreWriteCheck is already applied that does the catalog lookup, I think the safest option is to add a simple SessionCatalogWrapper around SnappySessionCatalog that overrides lookupRelation to instead invoke resolveRelationWithAlias. This wrapper catalog can be the one passed to PreWriteCheck constructor in SnappySessionState/SnappyAQPSessionState.getExtendedCheckRules.

core/src/main/scala/org/apache/spark/sql/internal/session.scala
```diff
@@ -565,6 +580,23 @@ private[sql] final class PreprocessTable(state: SnappySessionState) extends Rule
...
+            // we store table as Uppercase in relation
+            if (srcRelationNames.contains(r.table.toUpperCase)) {
+              throw new AnalysisException(
+                s"Cannot overwrite table ${r.table} that is also being read from")
+            }
+          case _ =>
+        }
+      i
+    }
```
sumwale: Why would the rule in PreWriteCheck for InsertIntoTable not apply? If due to table upper case issue then the equals of JDBCMutableRelation and JDBCAppendableRelation can be overridden.

core/src/main/scala/org/apache/spark/sql/hive/SnappySessionState.scala
```diff
@@ -681,6 +681,18 @@ class SnappySessionState(val snappySession: SnappySession)
...
 
+  lazy val wrapperCatalog: SessionCatalogWrapper = {
+    new SessionCatalogWrapper(
+      catalog.externalCatalog,
+      snappySession,
+      snappySession.sharedState.globalTempViewManager,
+      functionResourceLoader,
+      functionRegistry,
+      conf,
+      newHadoopConf(),
```
sumwale: Use catalog.hadoopConf instead of new

core/src/test/scala/org/apache/spark/sql/store/ColumnTableBatchInsertTest.scala
```diff
@@ -69,6 +67,42 @@ class ColumnTableBatchInsertTest extends SnappyFunSuite
...
+    }
+    try {
+      result.write.format("column").mode(SaveMode.Overwrite).saveAsTable(tableName)
+      fail("Expected AnalysisException while overwriting table which is also being read from")
+    }
+    catch {
+      case ae: AnalysisException => assert(ae.getMessage().contains("Cannot overwrite table"))
+      case t: Throwable => fail("Unexpected Exception ", t)
+    }
+
```
sumwale: Also add check for SQL: INSERT OVERWRITE and even INSERT INTO should fail when done on the same table.

PR #1407, opened 2019-08-01
Fixed an issue with lock release in case of deleteFrom 

suranjan:
In the case of deleteFrom, lock was not getting released.
Fixed the tests that were using the same session for all threads.
Added more tests and refactored code/tests

## Patch testing

precheckin
## ReleaseNotes.txt changes

## Other PRs 


core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
```diff
@@ -262,41 +262,47 @@ case class ExecutePlan(child: SparkPlan, preAction: () => Unit = () => ())
...
+      if (shuffleIds.nonEmpty) {
+        sc.cleaner match {
+          case Some(c) => shuffleIds.foreach(c.doCleanupShuffle(_, blocking = false))
+          case None =>
+        }
+      }
+      result
+    }
+    finally {
+      logDebug(" Unlocking the table in execute of ExecutePlan ")
```
sumwale: Could add the table name to log here. Perhaps available in session else maybe child.treeString(verbose = false)

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -278,7 +278,6 @@ abstract class BaseColumnFormatRelation(
       case None => snc.grabLock(table, schemaName, connProperties)
       case Some(a) => null // Do nothing as putInto will release lock
     }
-    if ((lock != null) && lock.isInstanceOf[RegionLock]) lock.asInstanceOf[RegionLock].lock()
```
sumwale: These explicit lock acquire were redundant and grabLock itself takes the lock?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
```diff
@@ -315,13 +314,12 @@ abstract class BaseColumnFormatRelation(
...
       case Some(_) => null // Do nothing as putInto will release lock
     }
-    if ((lock != null) && lock.isInstanceOf[RegionLock]) lock.asInstanceOf[RegionLock].lock()
     try {
       f()
     }
     finally {
-      logDebug(s"Added the ${lock} object to the context.")
       if (lock != null) {
+        logDebug(s"Added the ${lock} object to the context.")
```
sumwale: Will be good to add the table being locked in these log lines

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/ColumnFormatRelation.scala
suranjan: Just moved the code inside the grablock.

PR #1413, opened 2019-08-02
Change in the exit status if already running is in "waiting"

kneeraj:
and if already running is in "running" state. This differentiation is required in start all script for starting lead members.
The start up script carefully decides based on the exit status of the servers
where at least one server is in "running" state and then only starts lead members.

(Fill in the changes here)

## Patch testing

Manuall testing
## ReleaseNotes.txt changes

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/511

### DIFF
--- a/cluster/sbin/snappy-nodes.sh
+++ b/cluster/sbin/snappy-nodes.sh
@@ -280,6 +280,7 @@ function execute() {
     fi
   fi
 
+  childPids+=($LAST_PID)
   df=${dirfolder}
   if [ -z "${df}" ]; then
     df=$(echo ${dirparam} | cut -d'=' -f2)
@@ -297,6 +298,8 @@ index=1
 isServerStart=
 declare -a leadHosts
 declare -a leadCounts
+declare -a childPids
+
 if [ "$componentType" = "server" -a -n "$(echo $"${@// /\\ }" | grep -w start)" ]; then
   isServerStart=1
 fi
@@ -448,4 +451,18 @@ else
   fi
   execute "$@"
 fi
-wait
+
+if [ "$isServerStart" ]; then
+  # server status file
+  SERVERS_STATUS_FILE="$SNAPPY_HOME/work/serversstatus"
+  if [ -f $SERVERS_STATUS_FILE ]; then
+    rm $SERVERS_STATUS_FILE
+  fi
+  touch $SERVERS_STATUS_FILE
+  for pid in "${childPids[@]}"; do
+    wait $pid
+    echo $? >> $SERVERS_STATUS_FILE
+  done
+else
+  wait
+fi
--- a/cluster/sbin/snappy-nodes.sh
+++ b/cluster/sbin/snappy-nodes.sh
@@ -280,7 +280,7 @@ function execute() {
     fi
   fi
 
-  childPids+=($LAST_PID)
+  childPids[$LAST_PID]="$host"
   df=${dirfolder}
   if [ -z "${df}" ]; then
     df=$(echo ${dirparam} | cut -d'=' -f2)
@@ -454,14 +454,14 @@ fi
 
 if [ "$isServerStart" ]; then
   # server status file
-  SERVERS_STATUS_FILE="$SNAPPY_HOME/work/serversstatus"
+  SERVERS_STATUS_FILE="$SNAPPY_HOME/work/members-status.txt"
   if [ -f $SERVERS_STATUS_FILE ]; then
     rm $SERVERS_STATUS_FILE
   fi
   touch $SERVERS_STATUS_FILE
-  for pid in "${childPids[@]}"; do
+  for pid in "${!childPids[@]}"; do
     wait $pid
-    echo $? >> $SERVERS_STATUS_FILE
+    echo "$? ${childPids[${pid}]}" >> $SERVERS_STATUS_FILE
   done
 else
   wait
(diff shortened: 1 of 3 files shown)


cluster/sbin/snappy-nodes.sh
```diff
@@ -448,4 +451,18 @@ else
   fi
   execute "$@"
 fi
-wait
+
+if [ "$isServerStart" ]; then
+  # server status file
+  SERVERS_STATUS_FILE="$SNAPPY_HOME/work/serversstatus"
```
sumwale: There is already a members.txt, so better name would be members-status.txt

cluster/sbin/snappy-nodes.sh
```diff
@@ -448,4 +451,18 @@ else
...
+  if [ -f $SERVERS_STATUS_FILE ]; then
+    rm $SERVERS_STATUS_FILE
+  fi
+  touch $SERVERS_STATUS_FILE
+  for pid in "${childPids[@]}"; do
+    wait $pid
+    echo $? >> $SERVERS_STATUS_FILE
```
sumwale: For user readability will be good to also add the host in this file

cluster/sbin/snappy-start-all.sh
```diff
@@ -80,5 +95,13 @@ export SKIP_CONF_COPY
 
 # Start Leads
 if [ "$clustermode" != "rowstore" ]; then
-  "$sbin"/snappy-leads.sh $CONF_DIR_ARG start
+  checkIfOkayToStartLeads
```
sumwale: empty parens will be good to add to indicate function call

cluster/sbin/snappy-start-all.sh
```diff
@@ -80,5 +95,13 @@ export SKIP_CONF_COPY
...
+  if [ "$CAN_START_LEAD" = "1" ]; then
+    "$sbin"/snappy-leads.sh $CONF_DIR_ARG start
+  else
+    echo "Cannot start lead components. At least one server should be"
+    echo "in running state. Please check server logs."
+    echo "Try running snappy-status-all script to check servers status"
+    echo "Try starting lead once at least one server comes in running state."
```
sumwale: "Try running lead with snappy-start-all again once ...

launcher/src/main/java/io/snappydata/tools/QuickLauncher.java
```diff
@@ -164,7 +165,16 @@ private int start(final String[] args) throws IOException, InterruptedException
     // See bug 32574.
     String msg = verifyAndClearStatus();
     if (msg != null) {
+      final Status status = getStatus();
```
sumwale: Better to avoid getting the status again here by having verifyAndClearStatus return the exit code and do the System.err.println() itself. Here a non-zero return code will be checked and returned. The same change of using the proper exit code for System.exit() call can be also done in CacheServerLauncher for cases where QuickLauncher is not being used (though will only happen with a special flag but handling in both places will be good for consistency).

cluster/sbin/snappy-nodes.sh
kneeraj: Ok.

cluster/sbin/snappy-nodes.sh
kneeraj: ok.

cluster/sbin/snappy-start-all.sh
kneeraj: ok.

cluster/sbin/snappy-start-all.sh
kneeraj: Ok.

launcher/src/main/java/io/snappydata/tools/QuickLauncher.java
kneeraj: ok.

cluster/sbin/snappy-start-all.sh
kneeraj: Well this does not seem to work:
checkIfOkayToStartLeads()

If I put parenthesis then the finction is not called. What is wrong here. At other places also I checked for zero arg function just the name is mentioned.

launcher/src/main/java/io/snappydata/tools/QuickLauncher.java
```diff
@@ -163,13 +163,11 @@ private int start(final String[] args) throws IOException, InterruptedException
 
     // Complain if a cache server is already running in the specified working directory.
     // See bug 32574.
-    String msg = verifyAndClearStatus();
-    if (msg != null) {
-      final Status status = getStatus();
-      System.err.println(msg);
+    int state = verifyAndClearStatus();
+    if (state != 0) {
```
sumwale: Better to check against Status.SHUTDOWN rather than hard-coded 0.

cluster/sbin/snappy-start-all.sh
```diff
@@ -34,6 +34,21 @@ if [ -f "${MEMBERS_FILE}" ]; then
...
+CAN_START_LEAD=0
+function checkIfOkayToStartLeads() {
+  while  read -r line || [[ -n "$line" ]]; do
+    read exitstatus member <<< $line
+    if [ "x$exitstatus" = "x0" -o "x$exitstatus" = "x10" ]; then
+      CAN_START_LEAD=1
+    fi
+  done < $SERVERS_STATUS_FILE
+}
+
```
sumwale: @kneeraj will be better to move this to snappy-lead.sh or snappy-leads.sh because k8s setup does not use start-all. Check with @dshirish as to which one it actually uses so will be better to move this code block and its caller to that script.

PR #1422, opened 2019-08-08
SNAP-3134: Ensure package is deployed as a package and not as a jar, during cluster startup.

ashetkar:
* SNAP-3134: Ensure package is deployed as a package and not as a jar, during cluster startup.
* System property FAIL_ON_JAR_UNAVAILABILITY now defaults to false, allowing cluster to come up even if a package or jar could not be deployed. Also, the jar files are removed from executors work directory.
* Added a testcase.

## Patch testing
precheckin

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


sumwale:
@ashetkar Not sure what this change is trying to do, but isn't the correct way to use the cached jars for deploy in restart? Sometimes maven packages may no longer be present (esp when using custom maven repos), or network download temporarily out. Besides we don't want startup to wait for downloads all over again, so the proper way is to use the already locally copied jars.

ashetkar:
@sumwale DEPLOY PACKAGE resolves the coordinates and downloads all the jars into local jar cache path (default ~/.ivy2, if nothing specified). And the lead makes these available to servers on `RpcEnvFileServer`. Upon startup, AFAIK, the jars are not downloaded again if those are already available in the jar cache path. This is something taken care of in Spark itself (`SparkSubmitUtils.resolveMavenCoordinates()`).

Here, I'm simply making sure that the we do not treat a maven coordinate as an absolute jar path.

sumwale:
@ashetkar I understand that but afaik Spark will try remote resolution first, compare timestamps or similar with local cache and only use the local cache if it is current. My concern is that in restart of a long running cluster, if the remote repo no longer has the artifacts or repo itself no longer valid, then won't the cluster startup fail and no way to fix it unless one re-creates remote repo again the way it was? This will be moot if Spark resolution uses local cache first and treats it as a master copy without checking remote repo copy in all cases.

kneeraj:
@sumwale there is a system flag added which can be made as FALSE which will take care of restarting by ignoring these kind of errors.
FAIL_ON_JAR_UNAVAILABILITY

sumwale:
@kneeraj That should not be required and is a major usability problem. From what I understand the jars are already copied to lead/server working directory. To be cleaner, they can be inside a sub-directory "deployed-jars" and on restart the lead/servers should simply add all those jars to their classpath without rerunning deploy. There were other issues discussed with Amogh & Swati which need cleanups/fixes that we should track as JIRA tasks (e.g. add name to jar name to make it unique, Spark's existing ADD jar etc).

sumwale:
@kneeraj DEPLOY is needed only to the extent of registering in map without actual jar shipping so that undeploy works

kneeraj:
@sumwale Currently the 'deploy package' is a re-run at restart. The 'path' which is mentioned is a cache path. So even if it runs again it fetches locally in lead itself. And then distributes to other servers. IMO this is fine. Why is this a usability issue?

For JARS, they are supposed to be present at the same place where you initially mentioned during deploying. On restart it just distributes to the servers. Looks ok to me.
We can discuss in the huddle tomorrow.

jramnara:
@kneeraj @sumwale @ashetkar  Not sure if I shared this earlier, the lack of finding artifacts should NEVER stop the cluster from coming up. It is very painful as it is very difficult for the user to track down where the jars are located. There should be no system property that the user needs to know. I think, neeraj, we have discussed in the past that when a user deploy a Jar/Package the system needs to be fully "own" this functionality. If I move my DB to a different set of nodes or take a backup, etc, all this should always work. In fact, in my test, I deliberately avoided specifying a path in the hope that it will target a well established folder on the lead to manage the artifacts. To me, this should be no different than when a user deploys a Job - the jar is only dispatched to the lead and automatically made available to the servers, right ? Like suggested, maybe, all we do is dump into a DB sub-folder (by DB I mean 'work' or 'disk store folder') and this gets added to the classpath , always?

kneeraj:
As per the discussion in the huddle and above comments we think we need to think a bit more on the managing of the deployed jars and packages. The other part which was anyways to be handled later which was unifying UDFs with deployed jars is there. So we would need to revisit with an integrated design. For now we will make the restart not fail by default but log warnings and do the necessary cleanups in the globalcommand entries if artifacts related to them cannot be loaded for some bad configuration or absence of files.

ashetkar:
@kneeraj Please review the latest commit which sets the system property to false and also deletes the failed jars from server's work directory.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1124,6 +1144,21 @@ object SnappyContext extends Logging {
...
+  private def failOnJarUnavailability(k: String, jars: Array[String], e: Throwable): Unit = {
+    GemFireXDUtils.waitForNodeInitialization()
+    Misc.getMemStore.getGlobalCmdRgn.remove(k)
+    if (!jars.isEmpty) {
+      val mutable = new ArrayBuffer[String]()
+      mutable += "__REMOVE_FILES_ONLY__"
+      jars.foreach(e => mutable += e)
+      RefreshMetadata.executeOnAll(globalSparkContext, RefreshMetadata.REMOVE_URIS_FROM_CLASSLOADER,
+        mutable.toArray)
+    }
```
dshirish: Question: Should we also catch and log exception here so that restart does not fail due to any exception here?

core/src/dunit/scala/org/apache/spark/sql/streaming/SnappySinkProviderDUnitTest.scala
```diff
@@ -149,6 +149,7 @@ class SnappySinkProviderDUnitTest(s: String)
     Files.deleteIfExists(Paths.get(snappyProductDir, "conf", "locators"))
     Files.deleteIfExists(Paths.get(snappyProductDir, "conf", "leads"))
     Files.deleteIfExists(Paths.get(snappyProductDir, "conf", "servers"))
+    Files.move(Paths.get(snappyProductDir, "work"), Paths.get(snappyProductDir, "work-SnappySinkProviderDUnitTest"))
```
dshirish: Line length limit exceeds.

ashetkar: Done

PR #1423, opened 2019-08-09
[SNAP-3110] Changes for SNAP-3110 with fix for backward compatibility 

vatsalmevada:
Re-merging changes for SNAP-3110 with the fix for backward compatibility.

Earlier we introduced a new parameter for the procedure
`CREATE_OR_DROP_RESERVOIR_REGION`. This can break the upgrade from older
versions while DDL replay. Instead of exposing extra parameter for removal of sampler,
we are not removing it always along with the reservoir region.

Fixing failing test `AQPCatalogConsistencyDUnitTest`

## Patch testing
- `AQPCatalogConsistencyDUnitTest` passing
- will trigger `precheckin -Pstore`

## Other PRs 

https://github.com/SnappyDataInc/snappy-aqp/pull/194
https://github.com/SnappyDataInc/snappy-store/pull/514

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
@@ -16,12 +16,14 @@
  */
 package org.apache.spark.sql.execution.columnar.impl
 
+import java.lang.reflect.Method
 import java.net.URLClassLoader
 import java.sql.SQLException
 import java.util.Collections
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
+import scala.util.control.NonFatal
 
 import com.gemstone.gemfire.cache.{EntryDestroyedException, RegionDestroyedException}
 import com.gemstone.gemfire.internal.cache.lru.LRUEntry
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/StoreCallbacksImpl.scala
```diff
@@ -581,6 +584,36 @@ object StoreCallbacksImpl extends StoreCallbacks with Logging with Serializable
...
+    // `java.lang.Class.getMethod` method.
+    samplerClass.getMethods.filter(m => {
+      val paramTypes = m.getParameterTypes
+      if (m.getName.equals("removeSampler") && m.getParameterCount == 2 &&
+          paramTypes(0).equals(classOf[String]) && paramTypes(1).getName.equals("boolean")) {
+        true
+      } else false
+    }).head
+  }
+
```
sumwale: samplerClass.getMethod("removeSampler", classOf[String], classOf[java.lang.Boolean]) works fine

vatsalmevada: Somehow it's not working on my machine. I already tried earlier and tried again. Failing with: ```java.lang.NoSuchMethodException: org.apache.spark.sql.execution.StratifiedSampler$.removeSampler(java.lang.String, java.lang.Boolean)```

Could this be a JVM version/vendor specific issue?  Java vesion on my machine: `openjdk version "1.8.0_222"`

sumwale: Just tried and the following worked: `getMethod("removeSampler", classOf[String], classOf[Boolean])`

Also don't really need to go to the singleton class since bytecode of the main class also has static method, so:

```scala
    val samplerClassName = "org.apache.spark.sql.execution.StratifiedSampler"

    ...
    samplerClass.getMethod("removeSampler", classOf[String], classOf[Boolean])
    ...
    removeSamplerMethod.invoke(null, sampleTableName, Boolean.box(false))
```

PR #1428, opened 2019-08-14
Snap 3126 [Take Write lock only after executors are assigned]

suranjan:
Proceed to take write lock only when executors are assigned.
Provided conf property for disabling the write lock
Provided the conf property for providing timeout for lock

The deadlock was following scenario in smart connector mode:

- The test submits multiple spark jobs. The default mode of submission is client mode.
- In client mode the driver process runs outside the cluster. From the code it is seen ( we were not aware of this ) that the resource allocation(executor creation) to each submitted job happens in a different thread asynchronously.
- The main thread then proceeds ahead and grabs the dlock.
- Now this asynchronous behavior becomes reason for a deadlock because of contention of two resources, 1) Executors being required to be spawned for submitted spark app and 2) Dlock on the region on which write needs to happen.
- There can be a scenario that for a job which has been allocated executors asynchronously, the main thread might not get the dlock on the region, and therefore the allocated executors will never become free.
- Another job's asynchronous thread will wait for the executors to become free but its main thread can go ahead and grab the region dlock.
- The first job is waiting for the lock whereas the second job is waiting for the executors.

## Patch testing

precheckin running.

## ReleaseNotes.txt changes

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/517

### DIFF
--- a/core/src/main/scala/io/snappydata/Literals.scala
+++ b/core/src/main/scala/io/snappydata/Literals.scala
@@ -188,6 +188,17 @@ object Property extends Enumeration {
     s"${Constant.PROPERTY_PREFIX}sql.planCaching",
     "Property to set/unset plan caching", Some(false))
 
+  val SerializedWrites: SQLValue[Boolean] = SQLVal[Boolean](
+    s"${Constant.PROPERTY_PREFIX}sql.serializedWrites",
+    "Property to set/unset serialized writes", Some(true))
+
+  val SerializedWriteLockTimeOut: SQLValue[Int] = SQLVal[Int](
+    s"${Constant.PROPERTY_PREFIX}sql.serializedWriteLockTimeOut",
+    "Property to specify the lock timeout for write ops. If the write operation" +
+      "doesn't get lock for write within this time period then operation will fail." +
+      "Default value is 300 sec", Some(300))
+
+
   val Tokenize: SQLValue[Boolean] = SQLVal[Boolean](
     s"${Constant.PROPERTY_PREFIX}sql.tokenize",
     "Property to enable/disable tokenization", Some(true))
--- a/core/src/main/scala/io/snappydata/Literals.scala
+++ b/core/src/main/scala/io/snappydata/Literals.scala
@@ -17,8 +17,10 @@
 package io.snappydata
 
 import scala.reflect.ClassTag
+
 import com.gemstone.gemfire.internal.shared.unsafe.DirectBufferAllocator
 import com.pivotal.gemfirexd.internal.engine.GfxdConstants
+
 import org.apache.spark.sql.execution.columnar.ExternalStoreUtils
 import org.apache.spark.sql.internal.{AltName, SQLAltName, SQLConfigEntry}
 
@@ -189,7 +191,9 @@ object Property extends Enumeration {
 
   val SerializeWrites: SQLValue[Boolean] = SQLVal[Boolean](
     s"${Constant.PROPERTY_PREFIX}sql.serializeWrites",
-    "Property to set/unset serialized writes on column table", Some(true))
+    "Property to set/unset serialized writes on column table." +
+      "There will be a global lock which will ensure that at a time only" +
+      "one write operation is active on the column table.", Some(true))
 
   val SerializedWriteLockTimeOut: SQLValue[Int] = SQLVal[Int](
     s"${Constant.PROPERTY_PREFIX}sql.serializedWriteLockTimeOut",
--- a/core/src/main/scala/io/snappydata/Literals.scala
+++ b/core/src/main/scala/io/snappydata/Literals.scala
@@ -17,9 +17,8 @@
 package io.snappydata
 
 import scala.reflect.ClassTag
-
 import com.gemstone.gemfire.internal.shared.unsafe.DirectBufferAllocator
-
+import com.pivotal.gemfirexd.internal.engine.GfxdConstants
 import org.apache.spark.sql.execution.columnar.ExternalStoreUtils
 import org.apache.spark.sql.internal.{AltName, SQLAltName, SQLConfigEntry}
 
@@ -188,16 +187,16 @@ object Property extends Enumeration {
     s"${Constant.PROPERTY_PREFIX}sql.planCaching",
     "Property to set/unset plan caching", Some(false))
 
-  val SerializedWrites: SQLValue[Boolean] = SQLVal[Boolean](
-    s"${Constant.PROPERTY_PREFIX}sql.serializedWrites",
-    "Property to set/unset serialized writes", Some(true))
+  val SerializeWrites: SQLValue[Boolean] = SQLVal[Boolean](
+    s"${Constant.PROPERTY_PREFIX}sql.serializeWrites",
+    "Property to set/unset serialized writes on column table", Some(true))
 
   val SerializedWriteLockTimeOut: SQLValue[Int] = SQLVal[Int](
     s"${Constant.PROPERTY_PREFIX}sql.serializedWriteLockTimeOut",
-    "Property to specify the lock timeout for write ops. If the write operation" +
-      "doesn't get lock for write within this time period then operation will fail." +
-      "Default value is 300 sec", Some(300))
-
+    "Property to specify the lock timeout for write ops in seconds. If the" +
+      " write operation doesn't get lock for write within this time period" +
+      s" then operation will fail. Default value is ${GfxdConstants.MAX_LOCKWAIT_DEFAULT/1000} sec",
+    Some(GfxdConstants.MAX_LOCKWAIT_DEFAULT/1000))
 
   val Tokenize: SQLValue[Boolean] = SQLVal[Boolean](
     s"${Constant.PROPERTY_PREFIX}sql.tokenize",
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -26,7 +26,6 @@ import scala.concurrent.Future
 import scala.language.implicitConversions
 import scala.reflect.runtime.universe.{TypeTag, typeOf}
 import scala.util.control.NonFatal
-
 import com.gemstone.gemfire.internal.GemFireVersion
 import com.gemstone.gemfire.internal.cache.PartitionedRegion.RegionLock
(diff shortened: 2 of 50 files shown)


core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -188,6 +188,17 @@ object Property extends Enumeration {
     s"${Constant.PROPERTY_PREFIX}sql.planCaching",
     "Property to set/unset plan caching", Some(false))
 
+  val SerializedWrites: SQLValue[Boolean] = SQLVal[Boolean](
+    s"${Constant.PROPERTY_PREFIX}sql.serializedWrites",
```
kneeraj: I think more appropritae will be "serializeWrites" ( omitted d )

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -538,32 +551,52 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
     }
   }
 
+
+
   private[sql] def clearWriteLockOnTable(): Unit = {
+    logInfo(s"cleanWriteLockOnTable for app ${sqlContext.sparkContext.appName}",
+      new Throwable("ClearWrite"))
```
kneeraj: You will remove this Throwable. Right?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -538,32 +551,52 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
 
   private[sql] def grabLock(table: String, schemaName: String,
-      connProperties: ConnectionProperties): Any = {
+      connProperties: ConnectionProperties): Option[Any] = {
     SnappyContext.getClusterMode(sparkContext) match {
       case _: ThinClientConnectorMode =>
+        while (!SnappySession.executorAssigned) {
+          // Don't expect this case usually unless lots of applications are submitted in parallel
+          try {
+            Thread.sleep(100)
```
kneeraj: Should this also timeout? Otherwise while loop can be indefinite sometime?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -538,32 +551,52 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
             rs.next()
             locked = rs.getBoolean(1)
             ps.close()
-            logDebug("Took lock on server. ")
+            logDebug(s"Took lock on server. for string " +
+              s"${SnappySession.WRITE_LOCK_PREFIX + table} and " +
+              s"app ${sqlContext.sparkContext.appName}")
           }
           catch {
             case e: Throwable =>
```
kneeraj: How are you handling timeout? I mean if the timeout happens then some exception will come? Is it ok to eat the full Throwable or just handle just the timeout? Moreover on timeout are we retrying or should it fail?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -577,28 +610,32 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
       case _ =>
         logDebug(s"Taking lock in " +
-            s" ${Thread.currentThread().getId} ")
+            s" ${Thread.currentThread().getId} and " +
+          s"app ${sqlContext.sparkContext.appName}")
         val regionLock = PartitionedRegion.getRegionLock(SnappySession.WRITE_LOCK_PREFIX + table,
           GemFireCacheImpl.getExisting)
-        regionLock.lock()
-        regionLock
+        regionLock.lock(serializedWriteLockTimeOut * 1000)
```
kneeraj: Timeout is in seconds? I think in the property description more clarity can be given in the description.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -577,28 +610,32 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
+          logInfo(s"Going to unlock the lock object bulkOp $lock and " +
+            s"app ${sqlContext.sparkContext.appName}")
           lock.asInstanceOf[PartitionedRegion.RegionLock].unlock()
         }
       case (conn: Connection, id: TableIdentifier) =>
         var unlocked = false
         try {
-          logDebug(s"releasing lock on the server. ${id.table}")
+          logDebug(s"Going to unlock the lock on the server. ${id.table} for " +
+            s"app ${sqlContext.sparkContext.appName}")
```
kneeraj: Can unlock fail? If yes then we should perhaps catch the exception and at least log a WARN. Right?

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1987,6 +2041,7 @@ private class FinalizeSession(session: SnappySession)
   override protected def clearThis(): Unit = {
     sessionId = SnappySession.INVALID_ID
   }
+
```
kneeraj: Extra line can be removed.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
suranjan: At spark level too we see that run job is blocked waiting. So, I thought indefinite wait is fine.
We can use wait/notify too.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1967,6 +2004,23 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
+}
+
+// A listener to let the session know if at least one executor has been
+// added to this application.
+private class AppSparkListener extends SparkListener with Logging {
+  override def onExecutorAdded(execList: SparkListenerExecutorAdded): Unit = {
+    logInfo(s"AppSparkListener: onExecutorAdded: added $execList")
+    SnappySession.executorAssigned = true
+  }
 }
```
sumwale: There is already a SparkContextListener in object SnappyContext. This method can be moved there and there is no need to add another listener in every session.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -1967,6 +2004,23 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
   }
+
+
+  SnappyContext.getClusterMode(_sc) match {
+    case ThinClientConnectorMode(_, _) => {
+      _sc.listenerBus.addListener(new AppSparkListener)
+    }
+    case _ => // Nothing to do
+  }
+}
```
sumwale: This will end up adding the same listener for every session and will end up creating too many listeners. See comment below.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -2011,6 +2065,8 @@ object SnappySession extends Logging {
     GemFireVersion.isEnterpriseEdition
   }
 
+  var executorAssigned = false
```
sumwale: make private[sql]

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -306,6 +305,12 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
   @transient
   private[sql] var planCaching: Boolean = Property.PlanCaching.get(sessionState.conf)
 
+  @transient
+  private[sql] var serializedWrites: Boolean = Property.SerializedWrites.get(sessionState.conf)
+
+  @transient
+  private[sql] var serializedWriteLockTimeOut: Int = Property.SerializedWriteLockTimeOut.get(sessionState.conf)
+
```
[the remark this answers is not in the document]
sumwale: Such variables also need to be updated in SnappyConf.keyUpdateActions else they won't be set/unset with "SET" command. This PR is not doing that so this is incorrect. Please change to using property directly as noted above, or you can change the "var" declaration to "def".

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -16,11 +16,14 @@
  */
 package org.apache.spark.sql
 
+import java.rmi.ServerException
```
sumwale: unnecessary import

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -16,11 +16,14 @@
...
 import java.sql.{Connection, SQLException, SQLWarning}
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.{Calendar, Properties}
 
+import com.gemstone.gemfire.cache.LockTimeoutException
```
sumwale: unnecessary import

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -566,16 +559,24 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
+            while (!SnappyContext.executorAssigned) {
+              if (!SnappyContext.executorAssigned) {
+                try {
+                  SnappyContext.resourceLock.wait(100)
+                }
+                catch {
+                  case _: InterruptedException =>
+                    logWarning("Interrupted while waiting for executor.")
+                }
+              }
```
sumwale: Should the code also break out of this loop and fail after Property.SerializedWriteLockTimeOut?

sumwale: I guess this is similar to Neeraj's question before. Perhaps you are expecting that lock will timeout and then this can proceed, right?

suranjan: This is a case where even spark just submits the job and is seen stuck on runJob. When an executor gets added to the app, it proceeds. So thinking of keeping similar behavior as spark will keep accepting Application without failure. This will automatically proceed once other App either times out or finishes.

suranjan:
> Minor comments and a question. LGTM otherwise.
> 
> Spark-side change linked to this PR is still a concern.

Deleted the branch. That was not needed after listener approach.

PR #1432, opened 2019-08-14
Fix for SNAP-3141. This was a case where the input variables were eva…

ahshahid:
…luated inside a block to generate the group by key. However same base input variables were also needed by aggregate function. The fix is to first evaluate the input var code & then generate Key Expression code on the input var.

Evaluating the input variables code , before generating the KeyExpressions code of group by Key.
This ensures that if aggregate functions refer to variables of input attribute , they are available.
Previously the input variables were getting evaluated inside the code block of key expression generation, making the input variables unavailable inside aggregate function block.

## Patch testing

precheckin being run

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
+++ b/cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
@@ -1202,4 +1202,20 @@ class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
     assert(metrics(numRowsMetric.accumulatorId) ===
         SQLMetrics.stringValue(numRowsMetric.metricType, numRows :: Nil))
   }
+
+  test("SNAP-3141: code gen failure") {
+    // TODO: new SHA code generation fails for query below
+    val session = snc.snappySession.newSession()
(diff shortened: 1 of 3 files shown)


cluster/src/test/scala/org/apache/spark/sql/store/BugTest.scala
```diff
@@ -1202,4 +1202,20 @@ class BugTest extends SnappyFunSuite with BeforeAndAfterAll {
...
+
+    val numRows = 1000000
+
+    session.sql("create table test1 (id long, data string) using column " +
+      s"options (buckets '8') as select id, 'data_' || id from range($numRows)")
+    val ds = session.sql(
+      "select avg(id) average, id % 10 from test1 group by id % 10 order by average")
+
+    ds.collect()
+  }
```
sumwale: Instead of adding new test you can just remove/change the set calls in the previous test for SNAP-3123 and add a comment there that it also tests for SNAP-3141 (or name of test can be like "SNAP-3123_3141: check GUI plans and codegen failure")

PR #1437, opened 2019-08-23
temporary fix for HashJoin on sample table with non sample table not …

ahshahid:
…working.  There is  still a bug where by if the sample table is broadcasted for Hash Join , in my test I see 1 row missing. I suspect the reservoir region is not being broadcasted or something . so if the relation involves sample table in join it wont be broadcasted but will continue to be used in hashjoin.

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
@@ -396,7 +396,10 @@ private[sql] object JoinStrategy {
   def allowsReplicatedJoin(plan: LogicalPlan): Boolean = {
     plan match {
       case PhysicalScan(_, _, child) => child match {
-        case LogicalRelation(t: PartitionedDataSourceScan, _, _) => !t.isPartitioned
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/SnappyStrategies.scala
```diff
@@ -392,7 +396,10 @@ private[sql] object JoinStrategy {
   def allowsReplicatedJoin(plan: LogicalPlan): Boolean = {
     plan match {
       case PhysicalScan(_, _, child) => child match {
-        case LogicalRelation(t: PartitionedDataSourceScan, _, _) => !t.isPartitioned
+        case LogicalRelation(t: PartitionedDataSourceScan, _, _) => !t.isPartitioned && (t match {
+          case _: SamplingRelation => false
+          case _ => true
+        })
```
sumwale: Simpler: !t.isPartitioned && !t.isInstanceOf[SamplingRelation]

PR #1443, opened 2019-09-24
Snap 3152

smahajan05:

Added code changes for the package deployed in snappy job should be accessible within the job

## Patch testing

Added dunit test.

## ReleaseNotes.txt changes

NA

## Other PRs 

None

### DIFF
--- a/cluster/src/dunit/scala/io/snappydata/cluster/jobs/SNAP3152TestJob.scala
+++ b/cluster/src/dunit/scala/io/snappydata/cluster/jobs/SNAP3152TestJob.scala
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2018 SnappyData, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you
+ * may not use this file except in compliance with the License. You
+ * may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License. See accompanying
+ * LICENSE file.
+ */
+package io.snappydata.cluster.jobs
+
+import com.typesafe.config.Config
+import org.apache.spark.sql._
+
+object SNAP3152TestJob extends SnappySQLJob {
+
+  override def isValidJob(sc: SnappySession, config: Config): SnappyJobValidation = SnappyJobValid()
+
+  override def runSnappyJob(sc: SnappySession, jobConfig: Config): Any = {
+    // scalastyle:off println
+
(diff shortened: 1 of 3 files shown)


cluster/src/dunit/scala/io/snappydata/cluster/jobs/SNAP3152TestJob.scala
```diff
@@ -0,0 +1,65 @@
...
+    sc.sql("create external table customer12 " +
+        "using org.apache.spark.sql.cassandra options" +
+        " (table 'customer', keyspace 'test', spark.cassandra.input.fetch.size_in_rows '200000'," +
+        " spark.cassandra.read.timeout_ms '10000')")
+    assert(sc.sql("select * from customer12").count() == 3)
+    sc.sql("undeploy cassJar")
+    assert(sc.sql("list packages").count() == 0)
+    println("Undeployed successfully")
+    sc.sql("drop table customer12")
+
```
ashetkar: Create the external table again and assert it fails here.

core/src/dunit/scala/io/snappydata/cluster/CassandraSnappyDUnitTest.scala
```diff
@@ -453,4 +454,13 @@ class CassandraSnappyDUnitTest(val s: String)
...
   }
+
+  def doTestSNAP3152(): Unit = {
+    logInfo("Running testSNAP3152")
+    submitAndWaitForCompletion("io.snappydata.cluster.jobs." +
+        "SNAP3152TestJob",
+      s"--conf spark.cassandra.connection.host=localhost" +
+          s" --conf connectorJar=$cassandraConnectorJarLoc")
+    logInfo("Job Completed Successfully")
+  }
```
ashetkar: Could also check that the deployed jar is accessible from outside the job as well.

core/src/dunit/scala/io/snappydata/cluster/CassandraSnappyDUnitTest.scala
```diff
@@ -462,5 +462,11 @@ class CassandraSnappyDUnitTest(val s: String)
       s"--conf spark.cassandra.connection.host=localhost" +
           s" --conf connectorJar=$cassandraConnectorJarLoc")
     logInfo("Job Completed Successfully")
+    stmt1.execute("list packages")
+    assert(getCount(stmt1.getResultSet) == 1)
+    stmt1.execute("undeploy cassJar")
```
ashetkar: I mean you could also verify that external table creation works here too.

cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
```diff
@@ -117,6 +117,7 @@ object ToolsCallbackImpl extends ToolsCallback with Logging {
         case ite: InvocationTargetException => assert(ite.getCause.isInstanceOf[SecurityException])
       }
     }
+    Thread.currentThread().setContextClassLoader(lead.urlclassloader)
```
kneeraj: I don't think this is the right fix. Please test with other jars and packages which is passed through the job. With those a classloader is created which is specific to those coordinates and jars passed. They will be ignored.

PR #1444, opened 2019-09-24
Snap 3120

smahajan05:

Undeploy package or jar command does not removes uris from 'files' variable of type ConcurrentHashMap because of which deploying package with different name and path but same repos throws exception. So added code changes to remove uri from ConcurrentHashMap variable 'files' while undeploying package.

## Patch testing

Added dunit tests

## ReleaseNotes.txt changes

NA

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/176


smahajan05:
> Looks good.
> I'd suggest running this test in iterations and check logs for any stacktraces. Just to ensure removing file uri from the fileserver does not affect anything else.

Ran test in iterations. Not found any stacktraces in logs because of this change.

PR #1463, opened 2019-11-07
Monitoring SnappyData with metrics feature code changes

smahajan05:

- Added code changes to publish snappydata custom metrics of cluster stats, member stats, table stats, external table stats to the sinks that are supported by spark.
- Generated unique and stable cluster uuid to identify the cluster
- Added code to store the cluster uuid and all members diskstore uuid into region.
- Modified the ListPackageJarsCommand class to not show the cluster uuid and members diskstore uuid stored into region while executing "list packages" query.
- Changed getGlobalCmdRgn method name to getMetadataCmdRgn.

## Patch testing

Manually testes. Writing dunit test and also testing with large data.

## Other PRs 

https://github.com/SnappyDataInc/snappy-store/pull/524

https://github.com/SnappyDataInc/spark/pull/181


cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -182,6 +183,7 @@ class LeadImpl extends ServerImpl with Lead
           set("spark.scheduler.mode", "FAIR").
           setIfMissing("spark.memory.manager",
             ExecutorInitiator.SNAPPY_MEMORY_MANAGER)
+          .set("spark.metrics.namespace", "")
```
snappy-sachin: User can provide his own namespace using property "spark.metrics.namespace" in conf. Making it explicitly "" may hamper spark's own metric system behaviour for sparks nodes, right ?

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -316,6 +318,12 @@ class LeadImpl extends ServerImpl with Lead
       // update the Spark UI to add the dashboard and other SnappyData pages
       ToolsCallbackInit.toolsCallback.updateUI(sc)
 
+      // store every member diskStore ID to metadataCmdRgn
+      SnappyMetricsClass.putMembersDiskStoreIdInRegion()
```
snappy-sachin: can SnappyMetricsClass.putMembersDiskStoreIdInRegion() be shifted inside SnappyMetricsClass.init(sc) method ?

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
smahajan05: No need of setting this parameter. Will remove it.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
smahajan05: Yes, it can be shift to SnappyMetricsClass.init(sc) method.

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
```diff
@@ -377,13 +381,19 @@ class LeadImpl extends ServerImpl with Lead
...
         // noinspection SimplifyBooleanMatch
         startStatus match {
           case true =>
             logInfo("Primary lead lock acquired.")
 
+            // store unique cluster id to metadataCmdRgn
+            var clusterUuid = Misc.getMemStore.getMetadataCmdRgn.get("__ClusterID__")
+            if (clusterUuid == null) {
+              val region = Misc.getMemStore.getMetadataCmdRgn
+              region.put("__ClusterID__", UUID.randomUUID().toString)
```
kneeraj: As I said earlier, I feel user defined cluster name is much more useful than random uuid. If there are multiple systems then UUID will give no help to the user to identify a particular cluster. It should not be difficult to add.

@jramnara, @sudsmenon  please comment.

cluster/src/main/scala/io/snappydata/metrics/SnappyMetricsClass.scala
```diff
@@ -0,0 +1,207 @@
...
+  def init(sc: SparkContext): Unit = {
+
+    // initialize metric system with cluster id as metrics namespace
+    val clusterUuid = region.get("__ClusterID__")
+    UserMetricsSystem.initialize(sc, clusterUuid)
+
+    // store every member diskStore ID to metadataCmdRgn
+    putMembersDiskStoreIdInRegion()
+
+    val memberEntries = region.getAll(region.keySet())
```
kneeraj: Its name should be all meta entries or something as at his point you are collecting everything.

cluster/src/main/scala/io/snappydata/metrics/SnappyMetricsClass.scala
```diff
@@ -0,0 +1,207 @@
...
+
+    val timeInterval = 1000
```
kneeraj: Please remove the hard coding. What is the logic of 1 second?
I would suggest to at least put it in a constant which are defined at one place. Better would be to use a property to make it configurable from outside. You can choose to do it later. But this is 1 sec and stats provider service is 5 sec. Is there a correlation. Or rather should there be a correlation?

cluster/src/main/scala/io/snappydata/metrics/SnappyMetricsClass.scala
```diff
@@ -0,0 +1,207 @@
...
+
+  def createHistogram(metricName: String, metricValue: Long): Unit = {
+    lazy val tempHistogram: SparkHistogram = UserMetricsSystem.histogram(metricName)
+    tempHistogram.update(metricValue)
+  }
+
+  def setMetricsForClusterStatDetails(): Unit = {
+    val csInstance = ClusterStatistics.getInstance()
+    createGauge("ClusterMetrics.totalCores", csInstance.getTotalCPUCores)
+    updateHistogram("ClusterMetrics.timeLineTrend",
```
kneeraj: What is a timeline trend?

cluster/src/main/scala/io/snappydata/metrics/SnappyMetricsClass.scala
```diff
@@ -0,0 +1,207 @@
...
+    createGauge(s"TableMetrics.columnTablesCount", columnTablesCount)
+    createGauge(s"TableMetrics.rowTablesCount", rowTablesCount)
+
+    for ((k, v) <- tableBuff) {
+      SnappyTableMetrics.convertStatsToMetrics(k, v)
+    }
+  }
+
+  def setMetricsForExternalTableStatDetails(externalTableBuff:
+                                            Map[String, SnappyExternalTableStats]): Unit = {
```
kneeraj: Please correct indentation.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -1112,7 +1112,8 @@ object SnappyContext extends Logging {
                   val commandSet = ToolsCallbackInit.toolsCallback.getGlobalCmndsSet
                   commandSet.forEach(new Consumer[Entry[String, String]] {
                     override def accept(t: Entry[String, String]): Unit = {
-                      if (!t.getKey.startsWith(ContextJarUtils.functionKeyPrefix)) {
+                      if (!t.getKey.startsWith(ContextJarUtils.functionKeyPrefix)
+                        && !t.getKey.startsWith("__")) {
```
kneeraj: Please put the pattern used as a single place so that anyone going to use the meta region for new purpose is aware of what patterns to use. Please define a set of prefix Constants which one can use for their purpose and use that instead of herd coding. Like It is being used for jars and packages now, So something like jar_name or package_name can be the key. You can use it with prefix clusterid__key .. Going forward others can use it too provided they have a guideline to do it.

core/src/main/scala/org/apache/spark/sql/execution/ddl.scala
```diff
@@ -561,30 +561,32 @@ case class ListPackageJarsCommand(isJar: Boolean) extends RunnableCommand {
...
-          val jarfiles = jars.map(f => {
-            val lastIndexOf = f.lastIndexOf('/')
-            val length = f.length
-            if (lastIndexOf > 0) f.substring(lastIndexOf + 1, length)
-            else {
-              f
-            }
-          })
-          rows += Row(alias, jarfiles.mkString(","), false)
+        if (!t.getKey.startsWith("__")) {
```
kneeraj: Use constants defined at one place.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
ashetkar: Also, add this pattern in the `if` case at the end of `RecoveryService.getAllDDLs()`. Check all such references of `ContextJarUtils.functionKeyPrefix` and make sure you include your new pattern also, wherever required.

cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
```diff
@@ -185,19 +185,19 @@ object ToolsCallbackImpl extends ToolsCallback with Logging {
 
   override def getAllGlobalCmnds: Array[String] = {
     GemFireXDUtils.waitForNodeInitialization()
-    val r = Misc.getMemStore.getGlobalCmdRgn
+    val r = Misc.getMemStore.getMetadataCmdRgn
     val keys = r.keySet().asScala.filter(p => !p.startsWith(ContextJarUtils.functionKeyPrefix))
```
ashetkar: Add `&& !p.startsWith("__")`

cluster/src/main/scala/io/snappydata/impl/LeadImpl.scala
smahajan05: It would be surely taken up for next release . Jira ticket for this task is https://jira.snappydata.io/browse/SNAP-3230

cluster/src/main/scala/io/snappydata/metrics/SnappyMetricsClass.scala
smahajan05: Added suggested change

core/src/main/scala/org/apache/spark/sql/execution/ddl.scala
smahajan05: Added suggested change

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
smahajan05: Added suggested change

cluster/src/main/scala/io/snappydata/ToolsCallbackImpl.scala
smahajan05: Added suggested change

cluster/src/main/scala/io/snappydata/metrics/SnappyMetricsClass.scala
smahajan05: Changed timeInterval to 5 sec instead of 1

cluster/src/main/scala/io/snappydata/metrics/SnappyMetricsClass.scala
smahajan05: Added suggested change

PR #1465, opened 2019-11-08
New interpreter support through new script snappy-shell

kneeraj:

Now scala code can be run from snappy-shell similar to spark-shell.
As soon as you connect you get a snappysession object and a sparkcontext handle similar to spark.

For example:

<shell>./bin/snappy-shell
snappy-shell> connect client 'localhost:1527'
snappy-shell>snappy.sql("create table tt(col1 int not null)");
snappy-shell>snappy.sql("insert into tt values(1)");
snappy-shell>val df1 = snappy.sql("select * from tt");
snappy-shell>val df2 = snappy.sql("select * from tt");

snappy-shell>val df = snappy.sql("select * from tt union
  select * from tt");
snappy-shell>println(df.count);

real output after running the above:
<nekumar@nekumar-ThinkPad-P52:~/snappy/snappydata/build-artifacts/scala-2.11/snappy$ ./bin/snappy-shell 
Welcome to
    ______                            __       ____  ____
   / ____/___  ____ ___  ____  __  __/ /____  / __ \/ __ )
  / /   / __ \/ __ `__ \/ __ \/ / / / __/ _ \/ / / / __  |
 / /___/ /_/ / / / / / / /_/ / /_/ / /_/  __/ /_/ / /_/ /  version 1.2 on Spark version 2.1.1.7
 \____/\____/_/ /_/ /_/ .___/\__,_/\__/\___/_____/_____/
                     /_/

Using Scala version 2.11.8 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_221)

Connect to an existing cluster using the connect command
Example: connect client 'localhost:1527';

Once connected, type in expressions to have them evaluated.

Spark context Web UI available at localhost:5050
Spark context available as 'sc'
Snappy session available as 'snappy'.

snappy-shell> connect client 'localhost:1527';
19/11/08 15:52:32.009 IST main<tid=0x1> INFO ClientService: Starting client on '172.17.0.1' with ID='25662|2019/11/08 15:52:31.986 IST' Source-Revision=745fdcf0d5d1e85997e59aace8d0cd44a28336c7
Using CONNECTION0
snappy-shell> snappy.sql("create table tt(col1 int not null)");
res1: org.apache.spark.sql.DataFrame = []                                                                                       

snappy-shell> snappy.sql("insert into tt values(1)");
res2: org.apache.spark.sql.DataFrame = [count: int]                                                                             

snappy-shell> val df1 = snappy.sql("select * from tt");
df1: org.apache.spark.sql.DataFrame = [col1: int]                                                                               

snappy-shell> val df2 = snappy.sql("select * from tt");
df2: org.apache.spark.sql.DataFrame = [col1: int]                                                                               

snappy-shell> 
> val df = snappy.sql("select * from tt union
>   select * from tt");
df: org.apache.spark.sql.DataFrame = [col1: int]                                                                                

snappy-shell> println(df.count);
1                                                                                                                               

snappy-shell> 
>
## Patch testing

Manual testing.

## ReleaseNotes.txt changes

Yes.

## Other PRs 
https://github.com/SnappyDataInc/snappy-store/pull/526


cluster/src/main/scala/io/snappydata/tools/SnappyUtilLauncher.scala
```diff
@@ -171,9 +178,82 @@ object SnappyUtilLauncher extends StoreCallback {
...
+    val arglist = args.toList
+    type ArgsMap = Map[Symbol, Any]
+
+    def isSwitch(s : String) = (s(0) == '-')
+
+    def nextOption(map : ArgsMap, list: List[String]) : ArgsMap = {
+
+      list match {
+        case Nil => map
+        case ("--snappydata.connection" | "-conn" | "--connection" ) :: value :: tail =>
```
jramnara: - Make sure this is included in the Spec... and, hence in the docs. 
- why are there so many ways to connect ? shouldn't we stick to the convention of --conf spark.snappydata.connection or --conf snappydata.connection  only ?

cluster/src/main/scala/io/snappydata/remote/interpreter/RemoteInterpreterStateHolder.scala
```diff
@@ -221,7 +229,8 @@ class RemoteILoop(spw: StringPrintWriter, intpHelper: RemoteInterpreterStateHold
 
 object RemoteILoop {
   private val notTBeInheritedCommandNames = Set(
-    "h?", "edit", "line", "load", "paste", "power", "quit", "replay", "reset", "settings")
+    "h?", "edit", "line", "load", "paste", "power",
```
jramnara: Providing an examples for loading external scala file would be nice? 
Also, are all these commands documented in Spark guides ?

cluster/src/main/scala/io/snappydata/tools/SnappyUtilLauncher.scala
```diff
@@ -171,9 +178,82 @@ object SnappyUtilLauncher extends StoreCallback {
     }
   }
 
+  val intpUsage = s"\nUsage:\n" +
+    s"\n${SCRIPT_NAME} [--snappydata.connection | -conn | --connection locatorhost:port]" +
```
jramnara: this usage model should be in the specs/docs.

cluster/src/main/scala/io/snappydata/tools/SnappyUtilLauncher.scala
kneeraj: cleaned up the help. Avoiding --conf thingy as users might get confused that they can pass in spark properties as well, which is not the case.

PR #1468, opened 2019-11-15
SNAP-3260: Add support for external data stores like Azure blob storage (WASB, not ADLS) and GCS 

cruizen:
Support for Azure blob storage

Compile hadoop-azure package (version same as the hadoop version used in the product) as part of product build (similar to hadoop-aws SDK) to enable Azure blob storage access 

## Patch testing

Manually. From Snappy shell and spark shell (smart connector mode)
 
## ReleaseNotes.txt changes

Does this change require an entry in ReleaseNotes.txt? Haven't added a note. 

## Other PRs 


nikhilbandi:
Looks good. Need to run pre-checkin and few regression tests to make sure anything is broken due to this change.

cruizen:
Will remove the hadoop 3.2.1 comments. That was a trial for ADLS gen2 support

cluster/build.gradle
```diff
@@ -133,6 +133,30 @@ dependencies {
     exclude(group: 'org.apache.directory.server', module: 'apacheds-kerberos-codec')
   }
 
+  compile(group: 'org.apache.hadoop', name: 'hadoop-azure', version: hadoopVersion) {
+    // exclude some dependencies like in hadoop-aws to avoid conflicts with other deps
+    exclude(group: 'asm', module: 'asm')
```
ashetkar: Curious to know how did you arrive at these exclusions? Is this the complete list of exclusions or is it something we'll have to update continuously as and when we see any conflicts during test runs, etc.

cruizen: Initially, I just added Guava since we explicitly fetch it elsewhere. However, I saw a problem in statistics collection in the web app, since a library used by Jetty had a different version (indicating that we need to exclude additional dependencies). I later just excluded everything that hadoop-aws excludes since it would be safe to do so.

The list all transitive dependencies of the hadoop-azure library can be obtained following this
https://solidsoft.wordpress.com/2014/11/19/gradle-tricks-tracking-down-not-expected-transitive-dependencies/  

This task prints the dependency tree for snappy-cluster 
`./gradlew :snappy-cluster_2.11:dependencies
`
Snippet from the output for hadoop-aws and hadoop-azure
....
+--- org.apache.hadoop:hadoop-aws:2.7.7
|    +--- org.apache.hadoop:hadoop-common:2.7.7 (*)
|    +--- com.fasterxml.jackson.core:jackson-databind:2.2.3 -> 2.6.7.2 (*)
|    +--- com.fasterxml.jackson.core:jackson-annotations:2.2.3 -> 2.6.7
|    \--- com.amazonaws:aws-java-sdk:1.7.4
|         +--- org.apache.httpcomponents:httpclient:4.2 -> 4.5.6 (*)
|         +--- commons-codec:commons-codec:1.3 -> 1.11
|         +--- com.fasterxml.jackson.core:jackson-core:2.1.1 -> 2.6.7
|         +--- com.fasterxml.jackson.core:jackson-databind:2.1.1 -> 2.6.7.2 (*)
|         +--- com.fasterxml.jackson.core:jackson-annotations:2.1.1 -> 2.6.7
|         \--- joda-time:joda-time:[2.2,) -> 2.10.1
\--- org.apache.hadoop:hadoop-azure:2.7.7
     +--- org.apache.hadoop:hadoop-common:2.7.7 (*)
     +--- com.fasterxml.jackson.core:jackson-core:2.2.3 -> 2.6.7
     +--- org.apache.httpcomponents:httpclient:4.2.5 -> 4.5.6 (*)
     +--- com.microsoft.azure:azure-storage:2.0.0
     |    +--- com.fasterxml.jackson.core:jackson-core:2.2.3 -> 2.6.7
     |    +--- org.slf4j:slf4j-api:1.7.5 -> 1.7.25
     |    \--- org.apache.commons:commons-lang3:3.3.2 -> 3.8.1
     \--- com.google.guava:guava:11.0.2 -> 14.0.1
....

build.gradle
```diff
@@ -117,6 +117,8 @@ allprojects {
     junitVersion = '4.12'
     mockitoVersion = '1.10.19'
     hadoopVersion = '2.7.7'
+    gcsHadoop2ConnectorVersion='hadoop2-2.0.0'
+    gcsHadoop3ConnectorVersion='hadoop3-2.0.0'
```
ashetkar: Why do we need `gcsHadoop3ConnectorVersion` ?

cruizen: Since there are separate versions of the connector from Google for Hadoop2 and Hadoop3, I have defined 2 different identifiers. We don't need gcsHadoop**3**ConnectorVersion now, but we will need it when we start supporting Hadoop 3.x

nikhilbandi: Can't we do away with the Hadoop3ConnectorVersion for now and add it when we have the related support in place? We might have to remove the Hadoop2ConnectorVersion at that time.

PR #1474, opened 2019-11-21
Added fix for SNAP-3240

smahajan05:

Earlier putMembersDiskStoreIdInRegion() method was putting members diskStore ids in region only once. Because of that if new member gets added the disk store id does not get added in metadataCmdRgn and it was throwing NPE while retrieving new member id from region. So changed method call into thread so that if new member gets added metadataCmdRgn also gets updated.

Removed allMetaEntries variable which is not required.

Added fix for deleting metrics if table gets deleted (SNAP-3248)

## Patch testing

Manually tested. Writing dunit test

## ReleaseNotes.txt changes

## Other PRs 


cluster/src/main/scala/io/snappydata/metrics/SnappyMemberMetrics.scala
```diff
@@ -21,19 +21,19 @@ package io.snappydata.metrics
...
-  def convertStatsToMetrics(member: String, memberDetails: MemberStatistics,
-       allMetaEntries: util.Map[String, String]) {
+  def convertStatsToMetrics(member: String, memberDetails: MemberStatistics) {
+
     val shortDirName = memberDetails.getUserDir.substring(
       memberDetails.getUserDir.lastIndexOf(System.getProperty("file.separator")) + 1)
 
-    val memberUuid = allMetaEntries.
-        get(Constant.MEMBER_ID_PREFIX + shortDirName + member + "__").toString
+    val memberUuid = Misc.getMemStore.getMetadataCmdRgn.
```
ashetkar: Please add a note here saying in future, `memberUuid` may get replaced by user provided name.
Otherwise, this statement doesn't make sense as `memberUuid` and `member` are one and the same.

cluster/src/main/scala/io/snappydata/metrics/SnappyMetricsSystem.scala
```diff
@@ -85,7 +81,6 @@ object SnappyMetricsSystem {
   }
 
   def putMembersDiskStoreIdInRegion(): Unit = {
-
     val membersBuff = SnappyTableStatsProviderService.getService.getMembersStatsFromService
     for ((k, v) <- membersBuff) {
       val shortDirName = v.getUserDir.substring(
```
ashetkar: Better to move `System.getProperty("file.separator")` outside the for loop. Also, to take a reference of `Misc.getMemStore.getMetadataCmdRgn` outside the for loop.

cluster/src/main/scala/io/snappydata/metrics/SnappyMemberMetrics.scala
smahajan05: Currently I am putting memberUuid in both key and value. It will be change into a future for storing user provided member unique id. So in that case the value for member will also gets changed.

cluster/src/main/scala/io/snappydata/metrics/SnappyMetricsSystem.scala
smahajan05: Added suggested code changes

cluster/src/main/scala/io/snappydata/metrics/SnappyMemberMetrics.scala
ashetkar: Please put it as a comment in the code.

PR #1484, opened 2019-11-26
Snap 3223 Test case for com.snappydata.jdbc.ClientDriver class

schiwate:

Adding test case for changes in store. Chose a test suite that had a cluster running as pre-condition.

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 
store: https://github.com/SnappyDataInc/snappy-store/pull/534


core/src/test/scala/io/snappydata/CommandLineToolsSuite.scala
```diff
@@ -309,4 +312,25 @@ class CommandLineToolsSuite extends SnappyTestRunner {
 
     }
   }
+
+  test ("SNAP-3223 Routing query using driver class com.snappydata.jdbc.ClientDriver"){
```
dshirish: Instead of "Routing query" say "Executing query"

PR #1490, opened 2019-11-29
Provide option to not retry Insert operation

suranjan:
Default retry of spark tasks on failure can cause duplicate in case of insert operations.
  Provided a user property snappydata..maxRetryAttemptsForWrite, which user can set to 0 to avoid this scenario.

  Other, operations as usual will retry without causing any consistency issue.

(Fill in the changes here)

## Patch testing

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/io/snappydata/Literals.scala
+++ b/core/src/main/scala/io/snappydata/Literals.scala
@@ -322,6 +322,12 @@ object Property extends Enumeration {
     s"${Constant.PROPERTY_PREFIX}sql.codeSplitFunctionParamsSizeInSHA",
     s"The number of group by keys or aggregates which should be used as parameters at a time" +
       s" for code splitting function", Some(5))
+
+  // By default it will be same as MAX_TASK_FAILURES
+  // User should set it to 0, if they want insert to fail on any failure.
+  val MaxRetryAttemptsForWrite: SQLValue[Int] = SQLVal[Int](
+    s"${Constant.PROPERTY_PREFIX}maxRetryAttemptsForWrite",
+    s"The number of times a write task should be retried before all tasks failing." , Some(4))
 }
 
 // extractors for properties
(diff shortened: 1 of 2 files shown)


core/src/main/scala/io/snappydata/Literals.scala
```diff
@@ -322,6 +322,14 @@ object Property extends Enumeration {
     s"${Constant.PROPERTY_PREFIX}sql.codeSplitFunctionParamsSizeInSHA",
     s"The number of group by keys or aggregates which should be used as parameters at a time" +
       s" for code splitting function", Some(5))
+
+  // By default it will be same as MAX_TASK_FAILURES
+  // User should set it to 0, if they want insert to fail on any failure.
+  // this is used in TaskSchedulerImpl.SNAPPY_WRITE_RETRY_PROP Any change here
+  // should be done reflected there too.
+  val MaxRetryAttemptsForWrite: SQLValue[Int] = SQLVal[Int](
```
kneeraj: Why don't we set it to '0' only?

core/src/main/scala/org/apache/spark/sql/execution/ExistingPlans.scala
```diff
@@ -262,8 +261,16 @@ case class ExecutePlan(child: SparkPlan, preAction: () => Unit = () => ())
 
   protected[sql] lazy val sideEffectResult: Array[InternalRow] = {
     val session = sqlContext.sparkSession.asInstanceOf[SnappySession]
+    val sc = session.sparkContext
+    // Insert operations shouldn't retry.
+    child.collectFirst {
+      case _: ColumnInsertExec | _: RowInsertExec => {
```
kneeraj: PutInto will also have an Insert node. So it will cover that as well. Right?

core/src/main/scala/io/snappydata/Literals.scala
suranjan: Putting '0'  by default behaviour may cause some existing to fail so for compatibility setting it to earlier values.

PR #1491, opened 2019-11-29
Fixes for SNAP-3253:

snappy-sachin:

**Problem:** [SNAP-3253] Few Cluster and AQP tests are failing with InvocationTargetException after introducing structured streaming ui feature

**Cause:** 
  Structured streaming UI tab is created when SparkSession instance is created. But while creating SnappySession (which extends SparkSession class) instance, it requires gemfire cache be created before its instance is created. But SparkSession makes calls to add structured streaming tab before instance is initialised and which becomes bottle neck here.  

**Changes:**
  - Adding method updateStructuredStreamingUITab which calls parent classes method
    updateUIWithStructuredStreamingTab using reflection.
    This method creates or updates Structured Streaming UI Tab.
  - Adding call to updateStructuredStreamingUITab for each SnappySession object creation.

## Patch testing

 - Tested manually. Ran precheckin, no relevant failures.

## ReleaseNotes.txt changes

## Other PRs 

**Spark:** https://github.com/SnappyDataInc/spark/pull/187

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappySession.scala
@@ -16,8 +16,8 @@
  */
 package org.apache.spark.sql
 
+import java.lang.reflect.Method
 import java.sql.{Connection, SQLException, SQLWarning}
-import java.util.Calendar
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.{Calendar, Properties}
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -2154,6 +2154,24 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
+      val updateUIMethod: Method = super.getClass.getMethod("updateUIWithStructuredStreamingTab")
+      updateUIMethod.invoke(this)
+    } catch {
+      case e: NoSuchMethodException =>
+        logWarning("Unable to add Structured Streaming UI Tab because" +
+            " updateUIWithStructuredStreamingTab method is not found in SparkSession class.")
+    }
+  }
+  // Call to update Structured Streaming UI Tab
+  updateStructuredStreamingUITab;
```
dshirish: Remove the semicolon. Also add empty paranthesis in the function call and function definition

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
```diff
@@ -2154,6 +2154,24 @@ class SnappySession(_sc: SparkContext) extends SparkSession(_sc) {
...
+     smart connector mode using SnappyData's Spark distribution
+  */
+  def updateStructuredStreamingUITab: Unit = {
+    try {
+      val updateUIMethod: Method = super.getClass.getMethod("updateUIWithStructuredStreamingTab")
+      updateUIMethod.invoke(this)
+    } catch {
+      case e: NoSuchMethodException =>
+        logWarning("Unable to add Structured Streaming UI Tab because" +
+            " updateUIWithStructuredStreamingTab method is not found in SparkSession class.")
```
dshirish: It will be good to indicate that in the message that this means snappy-spark distribution is not used.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
snappy-sachin: Done.

core/src/main/scala/org/apache/spark/sql/SnappySession.scala
snappy-sachin: Yes, done. 
As discussed updated the log message to indicate it is not snappy-spark distribution if UI is unable to add structured streaming tab.

snappy-sachin:
All feedback has been taken care. Hence merging.

PR #1492, opened 2019-11-30
SNAP-3162

paresh-p11:

*  Setting the property "spark.locality.wait.process" to a higher value by default in recovery mode
    to enforce process locality.
*  Throw exception in case compute is launched on non-preferred host in recovery mode.
    This way user can know the default locality wait time value is not sufficient and has to be increased.

## Patch testing
Manual testing, precheckin.

## ReleaseNotes.txt changes

## Other PRs 


core/src/main/scala/org/apache/spark/sql/execution/oplog/impl/OpLogRdd.scala
```diff
@@ -528,6 +529,16 @@ class OpLogRdd(
   override def compute(split: Partition, context: TaskContext): Iterator[Row] = {
     logDebug(s"starting compute for partition ${split.index} of table $fqtnUpper")
     try {
+      val currentHost = SparkEnv.get.executorId
+      val removePattern = "(executor_).*(_)".r
+      var expectedHost  = bucketHostMap.getOrElse(split.index,"")
+      require(expectedHost.nonEmpty, s"No preferred host found.")
```
kneeraj: Give more informative reasons for requirement. Like a hint to user to what went wrong and point him to corrective action.

core/src/main/scala/org/apache/spark/sql/execution/oplog/impl/OpLogRdd.scala
```diff
@@ -528,6 +529,16 @@ class OpLogRdd(
...
+
+      expectedHost = removePattern.replaceAllIn(expectedHost, "")
```
kneeraj: Shouldn't we make the expectedHost patten removal while creating the relation itself so that in each rdd compute is not required and code clooks cleaner as well.
Also give corrective action to the user as well in the exception message.

core/src/main/scala/org/apache/spark/sql/execution/oplog/impl/OpLogRdd.scala
paresh-p11: yes, will modify the statement.

core/src/main/scala/org/apache/spark/sql/execution/oplog/impl/OpLogRdd.scala
paresh-p11: Right, makes sense. will move pattern removal in OpLogFormatRelation. And qualify exception message.

core/src/main/scala/org/apache/spark/sql/execution/oplog/impl/OpLogRdd.scala
```diff
@@ -528,6 +529,19 @@ class OpLogRdd(
   override def compute(split: Partition, context: TaskContext): Iterator[Row] = {
     logDebug(s"starting compute for partition ${split.index} of table $fqtnUpper")
     try {
+      val currentHost = SparkEnv.get.executorId
+      val expectedHost  = bucketHostMap.getOrElse(split.index,"")
+      require(expectedHost.nonEmpty, s"No preferred host found." +
+          s" Error while getting executor host," +
+          s" please check RecoveryService logs for more information.")
```
kneeraj: What is a RecoveryService logs. Are we creating a separate log for this mode?

paresh-p11: We have TraceRecoveryMode flag for debugging. I meant to ask the user to check logs from RecoveryService class, which is visible in the logs. Will improve the statement to be clearer.

PR #1495, opened 2019-12-04
[Snap 3251]

paresh-p11:

* Fixing logic to find index of complex datatype columns in the byte array, read from the disk.
* Refactoring

## Patch testing

Manually. Running relevant unit test.
Running Precheckin( some unrelated failures, analysis is in progress)

## ReleaseNotes.txt changes

## Other PRs 


cluster/src/main/scala/io/snappydata/gemxd/ClusterCallbacksImpl.scala
```diff
@@ -183,7 +183,7 @@ object ClusterCallbacksImpl extends ClusterCallbacks with Logging {
     tablesArr.foreach(f = table => {
       Try {
         val tableData = session.sql(s"select * from $table;")
-        logDebug(s"EXPORT_DATA procedure is writing table: $table.")
+        logInfo(s"EXPORT_DATA procedure is exporting table: $table.")
```
kneeraj: put more info if you want to log this. Like destination and format etc. Otherwise I don't see much value.

paresh-p11: The thought was to let user know which tables have been picked up for exporting, when ALL option is given instead of list of table names in the procedure args. Adding destination and format makes it better. Will make the changes. 
Thanks..

PR #1496, opened 2019-12-04
[SNAP-3262] Fixing createTable API for python

vatsalmevada:

`createTable` API was failing when used without passing explicit values for
 provider and schema.
By default API defaults provider parameter to `parquet` and the schema can
also be retrieved from parquet file.

Usage example: snappy.createTable("table1",path ="/path/to/file.parquet")

## Patch testing

Manually tested using `pyspark` shell. Have written unit test but somehow 
`checkPython` target is not picking up the test. Will analyze further and add it
later.

### DIFF
--- a/python/pyspark/sql/snappy/snappysession.py
+++ b/python/pyspark/sql/snappy/snappysession.py
@@ -61,9 +61,9 @@ class SnappySession(SparkSession):
         :return: :class:`DataFrame`
         """
         if provider is None:
-            provider = self.getConf("spark.sql.sources.default", "org.apache.spark.sql.parquet")
+            provider = self.conf.get("spark.sql.sources.default", "org.apache.spark.sql.parquet")
         if schema is None:
-            df = self._jsparkSession.createTable(tableName, provider, allowExisting, options)
+            df = self._jsparkSession.createTable(tableName, provider, options, allowExisting)
         else:
             if isinstance(schema, str):
                 df = self._jsparkSession.createTable(tableName, provider, schema, options, allowExisting)


python/pyspark/sql/snappy/snappysession.py
```diff
@@ -61,9 +61,9 @@ def createTable(self, tableName, provider=None, schema=None, allowExisting=True,
         :return: :class:`DataFrame`
         """
         if provider is None:
-            provider = self.getConf("spark.sql.sources.default", "org.apache.spark.sql.parquet")
+            provider = self.conf.get("spark.sql.sources.default", "org.apache.spark.sql.parquet")
         if schema is None:
-            df = self._jsparkSession.createTable(tableName, provider, allowExisting, options)
+            df = self._jsparkSession.createTable(tableName, provider, options, allowExisting)
         else:
```
cruizen: Good catch. :-) The createTable method should preferably match the signature of the Scala equivalent class https://github.com/SnappyDataInc/snappydata/blob/master/core/src/main/scala/org/apache/spark/sql/SnappySession.scala

cruizen:
LGTM. We should run the checkPython target to find out if any Python tests are running at all, not just this test. Also, Python2 will reach end of life by 2020/01/01 and Spark might stop supporting Python 2. We should make sure that Python 3 is also supported and the tests, examples use Python 3 runtime.

python/pyspark/sql/snappy/snappysession.py
vatsalmevada: May be options are kept as the last argument as options are accepted as varargs? Unlike Scala counterpart where options are accepted as map. Not sure whether Scala has syntactic sugar to accept varargs as key-value pairs as the argument list.

PR #1497, opened 2019-12-05
Snap 3255

paresh-p11:

* Decompressing delta region entry values

## Patch testing

Manual testing, Running relevant unit tests.

## ReleaseNotes.txt changes

## Other PRs 

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/oplog/impl/OpLogRdd.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/oplog/impl/OpLogRdd.scala
@@ -402,16 +402,16 @@ class OpLogRdd(
    * @param phdrCol PlaceHolderDiskRegion of column batch
    */
   def iterateColData(phdrCol: PlaceHolderDiskRegion): Iterator[Row] = {
-    val directBuffers = mutable.ListBuffer.empty[ByteBuffer]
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/sql/execution/oplog/impl/OpLogRdd.scala
```diff
@@ -468,14 +467,20 @@ class OpLogRdd(
...
             } else null
 
             val deltaEntry2 = regMap.getEntry(k.withColumnIndex(deltaColIndex - 1))
             val delta2 = if (deltaEntry2 ne null) {
-              DiskEntry.Helper.readValueFromDisk(deltaEntry2.asInstanceOf[DiskEntry],
-                phdrCol).asInstanceOf[ColumnFormatValue].getBuffer
+              val buffer = DiskEntry.Helper.readValueFromDisk(deltaEntry2.asInstanceOf[DiskEntry],
+                phdrCol).asInstanceOf[ColumnFormatValue]
+                  .getValueRetain(FetchRequest.DECOMPRESS).getBuffer
+              if(buffer.isDirect) directBuffers += buffer
```
suranjan: Just check if directBuffers size is not too large. Otherwise, we can just return the buffer and keep it alive till the row is decoded.

paresh-p11: Thanks. Will check.

PR #1498, opened 2019-12-05
Fix for SNAP-3261.

smahajan05:

Earlier code was wrong as it was removing the metrics for table name and was taking null value after the table gets deleted. So I modified code to remove table metrics only once and not generate the null value.
 
Added some test changes in dunit test
- Added code to match the table metrics values with the expected values
- Added code to match member metrics values with the expected values

## Patch testing
Manuall tested. Ran hydra test

## ReleaseNotes.txt changes

## Other PRs 


cluster/src/main/scala/io/snappydata/metrics/SnappyMetricsSystem.scala
```diff
@@ -234,8 +234,10 @@ object SnappyMetricsSystem {
   def removeTableMetricsIfTableDeleted(stringToStats: Map[String, SnappyRegionStats]): Unit = {
     for (t <- tables) {
       if (!stringToStats.exists(x => x._1 == t)) {
-        SnappyTableMetrics.removeTableMetrics(t)
-        tables -= t
+        if (t != null) {
```
ashetkar: In which case, can the table-name be null? Is that a valid case or a bug?
If a valid case, it's better to check for null at the time of adding the entry itself in `tables`, in `setMetricsForTableStatDetails`?

cluster/src/dunit/scala/io/snappydata/metrics/SnappyMetricsSystemDUnitTest.scala
```diff
@@ -79,10 +71,10 @@ class SnappyMetricsSystemDUnitTest(s: String)
 
   override def afterClass(): Unit = {
     logInfo((snappyProductDir + "/sbin/snappy-stop-all.sh").!!)
-    s"rm -rf $snappyProductDir/work".!!
-    Files.deleteIfExists(Paths.get(snappyProductDir, "conf", "locators"))
-    Files.deleteIfExists(Paths.get(snappyProductDir, "conf", "leads"))
-    Files.deleteIfExists(Paths.get(snappyProductDir, "conf", "servers"))
+    // s"rm -rf $snappyProductDir/work".!!
```
ashetkar: What's the reason to retain the work directory?

cluster/src/dunit/scala/io/snappydata/metrics/SnappyMetricsSystemDUnitTest.scala
```diff
@@ -60,16 +53,15 @@ class SnappyMetricsSystemDUnitTest(s: String)
...
+        s"-dir=$snappyProductDir/work/lead2 -spark.ui.port=3334")
     pw1.close()
     val pw2 = new PrintWriter(new File(s"$confDir/servers"))
-    pw2.write(s"localhost -locators=localhost[$port] -client-port=$netPort2" +
-        s" -dir=$snappyProductDir/work/server1\n")
-    pw2.write(s"localhost -locators=localhost[$port] -client-port=$netPort3" +
-        s" -dir=$snappyProductDir/work/server2\n")
-    pw2.write(s"localhost -locators=localhost[$port] -client-port=$netPort4" +
-        s" -dir=$snappyProductDir/work/server3")
+    pw2.write(s"localhost -locators=localhost[$port] -dir=$snappyProductDir/work/server1\n")
```
ashetkar: Don't need to specify `-dir` property. Scripts automatically create these directories.

cluster/src/dunit/scala/io/snappydata/metrics/SnappyMetricsSystemDUnitTest.scala
```diff
@@ -60,16 +53,15 @@ class SnappyMetricsSystemDUnitTest(s: String)
         s" -peer-discovery-port=$port -client-port=$netPort")
     pw.close()
     val pw1 = new PrintWriter(new File(s"$confDir/leads"))
-    pw1.write(s"localhost -locators=localhost[$port] -dir=$snappyProductDir/work/lead1\n")
-    pw1.write(s"localhost -locators=localhost[$port] -dir=$snappyProductDir/work/lead2")
+    pw1.write(s"localhost -locators=localhost[$port] " +
```
ashetkar: Why specify `spark.ui.port` for each lead? Does it cause any conflict?

cluster/src/dunit/scala/io/snappydata/metrics/SnappyMetricsSystemDUnitTest.scala
```diff
@@ -119,13 +111,11 @@ class SnappyMetricsSystemDUnitTest(s: String)
...
-  def getConnection(): Connection = {
-    val driver = "io.snappydata.jdbc.ClientDriver"
-    Utils.classForName(driver).newInstance
-    DriverManager.getConnection(s"jdbc:snappydata:thrift://localhost[$netPort]")
-  }
+  def getConnection(netPort: Int): Connection =
+    DriverManager.getConnection(s"${Constant.DEFAULT_THIN_CLIENT_URL}localhost:$netPort")
 
   def testMetricsMonitoring(): Unit = {
+    Thread.sleep(10000)
```
ashetkar: Why is this sleep for and can we instead use the `WaitCriterion`?

cluster/src/dunit/scala/io/snappydata/metrics/SnappyMetricsSystemDUnitTest.scala
```diff
@@ -319,51 +309,35 @@ class SnappyMetricsSystemDUnitTest(s: String)
         assertEquals("APP.TEST2", v)}
       if (containsWords(k, Array("TableMetrics.APP.TEST2", "isReplicatedTable"))) {
         assertEquals(false, v)}
-      if (containsWords(k, Array("TableMetrics.APP.TEST2", "sizeSpillToDisk"))) {
```
ashetkar: Please explain these changes in the test as a PR description.

cluster/src/main/scala/io/snappydata/metrics/SnappyMetricsSystem.scala
smahajan05: Added the changes you suggested

cluster/src/dunit/scala/io/snappydata/metrics/SnappyMetricsSystemDUnitTest.scala
smahajan05: Mistakenly added comment for these lines. Removed it

cluster/src/dunit/scala/io/snappydata/metrics/SnappyMetricsSystemDUnitTest.scala
smahajan05: yes its causing conflict and i am getting bind exception So i added this spark.ui.port for getting metrics on that port

cluster/src/dunit/scala/io/snappydata/metrics/SnappyMetricsSystemDUnitTest.scala
smahajan05: Sleep is required to take effects in metrics values as the metrics values gets updated by 5 sec

cluster/src/dunit/scala/io/snappydata/metrics/SnappyMetricsSystemDUnitTest.scala
smahajan05: Explained the changes in pr description

cluster/src/dunit/scala/io/snappydata/metrics/SnappyMetricsSystemDUnitTest.scala
smahajan05: I need that option for getting full and short directory names for members work directory.

PR #1500, opened 2019-12-06
Fixes for SNAP-3264:

snappy-sachin:

 - call HashAggregateSize.set(conf, "-1") has been shifted from
   SnappySessionState.streamingQueryManager to SnappyStreamingQueryManager.

## Patch testing

 - Ran failing tests again and verified that those are not failing again.

## ReleaseNotes.txt changes

 - None

## Other PRs 

 - None


snappy-sachin:
Merging.

ahshahid:
I suppose this means that if the streaming query is using aqp , most likely
it will fail.
@Supriya Pillai <suppilla@tibco.com>  I think we will have to test it . I
will file a bug for the same....

PR #1502, opened 2019-12-07
masking creds before capturing into catalog

hemanthmeka:

we are capturing exact ddl string provided and adding it to corresponding catalog table. This is needed so that EXPORT_DDLS procedure can return original ddls.
This PR masks passwords if any provided in ddls

## Patch testing

Precheckin looks fine

## ReleaseNotes.txt changes

no

## Other PRs 

no

### DIFF
--- a/core/src/main/scala/io/snappydata/util/ServiceUtils.scala
+++ b/core/src/main/scala/io/snappydata/util/ServiceUtils.scala
@@ -177,4 +177,34 @@ object ServiceUtils {
         }
     }
   }
+
+  /**
+   * We capture table ddl string and add it to table properties before adding to catalog.
+   * This will also contain passwords in the string such as jdbc connection string or
+   * s3 location etc. This method masks passwords
+   * @param str DDL string that might contain jdbc/s3 passwords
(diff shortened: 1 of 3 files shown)


core/src/main/scala/io/snappydata/util/ServiceUtils.scala
```diff
@@ -177,4 +177,34 @@ object ServiceUtils {
...
+
+  /**
+   * We capture table ddl string and add it to table properties before adding to catalog.
+   * This will also contain passwords in the string such as jdbc connection string or
+   * s3 location etc. This method masks passwords
+   * @param str DDL string that might contain jdbc/s3 passwords
+   * @return similar string with masked passwords
+   */
+  def maskPasswordsInString(str: String): String = {
+    val jdbcPattern1 = ".*jdbc.*(?i)\\bpassword\\b=([^);&]*).*".r
```
suranjan: Factor out list of patterns at common place as constant.
See if we can add more or these patterns are sufficient. e.g other blob storage.

PR #1506, opened 2019-12-12
[SNAP-1849] Partial fix for embedded case.

ashetkar:
Check authorization for `SELECT COUNT(*) ...`  on row table for embedded mode.

## Patch testing
- New testcase.
- precheckin - Few test failures seen. Some of them fail on master and some others passed when run individually.

## ReleaseNotes.txt changes
NA

## Other PRs 
NA


core/src/main/scala/org/apache/spark/sql/execution/row/RowFormatScanRDD.scala
```diff
@@ -267,6 +268,10 @@ class RowFormatScanRDD(@transient val session: SnappySession,
    */
   override def compute(thePart: Partition, context: TaskContext): Iterator[Any] = {
 
+    if (Misc.getMemStoreBootingNoThrow != null && useResultSet) {
+      SecurityUtils.authorizeTableOperation(tableName, projection,
+        Authorizer.SELECT_PRIV, Authorizer.SQL_SELECT_OP, connProperties)
+    }
```
sumwale: What is the difference between this check and the previous one in !pushProjections block? As far as I remember pushProjections is true only for smart connector which is the same check as above.

sumwale: It seems that pushProjections is also true for local system table scans (see RowFormatRelation:127). Still doesn't seem to effect the count case for which the bug was filed.

ashetkar: The issue was originally reported for row tables in smart connector mode two years ago. But since start of this year, it started failing in embedded mode as well.

The PR is only for embedded case. Observed that `SELECT COUNT(*) ...` on row table always enters the `if (pushProjections)` block which does not have authz check.
Hence moved the authz check at the top. Added null check on memstore to avoid NPE in smart connector mode.

PR #1515, opened 2019-12-24
Snap 3275 - incrementing the index variable for null values for reference type

vatsalmevada:

Index variable tracking reference type fields was not getting incremented
when a reference type column is set to null. Fixing this behavior by
incrementing the index variable at the correct place.

Note that this issue was affecting JDBC connection mode only. Queries run
using `SnappySession#sql` were not affected.

## Patch testing

- added dunit test


vatsalmevada:
Merging the branch. Will take up changes to have the column name as key for complex type JSON results separately. https://jira.snappydata.io/browse/SNAP-3283 is raised to track the same.

PR #1516, opened 2020-01-01
SNAP-3279

paresh-p11:

* Date and timestamp are stored as integer and long respectively,
  hence it requires special handling.
* Added unit tests for the same

## Patch testing

Manual testing, running relevant test, precheckin in progress

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/execution/oplog/impl/OpLogRdd.scala
+++ b/core/src/main/scala/org/apache/spark/sql/execution/oplog/impl/OpLogRdd.scala
@@ -236,8 +236,9 @@ class OpLogRdd(
               if (daysSinceEpoch == null) null else DateTimeUtils.toJavaDate(daysSinceEpoch)
             })
           } else if (arrayType.elementType == TimestampType) {
-            data.toArray[Long](LongType).map(nsSinceEpoch => {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/sql/execution/oplog/impl/OpLogRdd.scala
```diff
@@ -231,16 +231,88 @@ class OpLogRdd(
...
       case mapType: MapType =>
         val map = valueArr(complexFieldIndex)
         val data = new SerializedMap()
+        // todo: refactor following code.
         if (map != null) {
           data.pointTo(map, Platform.BYTE_ARRAY_OFFSET)
-          val jmap = new java.util.HashMap[Any, Any](data.numElements())
-          data.foreach(mapType.keyType, mapType.valueType, (k, v) => jmap.put(k, v))
-          jmap
+          val jMap = new java.util.HashMap[Any, Any](data.numElements())
```
suranjan: Can refactor this part of code.

paresh-p11: Will take up in the next iteration. Will keep this PR open until then.
Thanks.

PR #1517, opened 2020-01-01
Fixes for SNAP-3280 and SNAP-3284

ashetkar:
- Used `catalogTable.database` instead of `catalogTable.owner` to get the table owner (schema).
- Fixed the issue in extracting the schema name from identifier.
- Used capitalized form of LDAP group name in `PermissionChecker.addRemoveUserForKey()` . This fixed an issue where if a user specified non-capitalized LDAP group name in GRANT/REVOKE DDL for external tables, it failed.
- If a username is found to be null during checking authz for external tables, set it to empty string and proceed to authz instead of returning without authz.
- Added several new testcases in the dunit test.

## Patch testing
Added new testcases.
Precheckin is clear.

## ReleaseNotes.txt changes
NA

## Other PRs 
NA

### DIFF
--- a/cluster/src/main/scala/io/snappydata/remote/interpreter/SnappyInterpreterExecute.scala
+++ b/cluster/src/main/scala/io/snappydata/remote/interpreter/SnappyInterpreterExecute.scala
@@ -251,13 +251,13 @@ object SnappyInterpreterExecute {
         val commaSepVals = users.split(",")
         commaSepVals.foreach(u => {
           if (isGrant) {
-            if (u.startsWith(Constants.LDAP_GROUP_PREFIX))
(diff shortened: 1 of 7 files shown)


cluster/src/main/scala/io/snappydata/remote/interpreter/SnappyInterpreterExecute.scala
```diff
@@ -250,13 +250,12 @@ object SnappyInterpreterExecute {
         // expand the users list. Can be a mix of normal user and ldap group
         val commaSepVals = users.split(",")
         commaSepVals.foreach(u => {
+          val uUC = u.toUpperCase
```
dshirish: StringUtil.SQLToUpperCase(u) can be used similar to REFRESH_LDAP_GROUP() procedure.

PR #1519, opened 2020-01-03
* [SNAP-3282] SYS.REFRESH_LDAP_GROUP now updates metadata region also.

ashetkar:
* [SNAP-3282] Procedure SYS.REFRESH_LDAP_GROUP now updates the entries in metadata region which are used in authorization of external tables.
* Also fixed an issue where system user could not grant permission on external table of another user.

## Patch testing
- Tested manually.
- Added a testcase for second issue.
- precheckin is fine.

## ReleaseNotes.txt changes
NA

## Other PRs 
NA

### DIFF
--- a/cluster/src/main/scala/io/snappydata/remote/interpreter/SnappyInterpreterExecute.scala
+++ b/cluster/src/main/scala/io/snappydata/remote/interpreter/SnappyInterpreterExecute.scala
@@ -31,8 +31,9 @@ import io.snappydata.gemxd.SnappySessionPerConnection
 import org.apache.log4j.Logger
 
 import org.apache.spark.Logging
-import org.apache.spark.sql.execution.InterpretCodeCommand
+import org.apache.spark.sql.execution.{GrantRevokeOnExternalTable, InterpretCodeCommand}
 import org.apache.spark.sql.execution.columnar.ExternalStoreUtils
+import scala.collection.JavaConverters._
 import scala.collection.mutable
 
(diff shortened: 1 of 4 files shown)


cluster/src/main/scala/io/snappydata/remote/interpreter/SnappyInterpreterExecute.scala
```diff
@@ -141,6 +142,25 @@ object SnappyInterpreterExecute {
...
+  }
+
+  private def updateMetaRegion(group: String): Unit = {
+    val r = Misc.getMemStore.getMetadataCmdRgn
+    val allAuthKeys = r.keySet().asScala.filter(s =>
+      s.startsWith(GrantRevokeOnExternalTable.META_REGION_KEY_PREFIX))
+    allAuthKeys.foreach(k => {
+      val p = r.get(k)
+      if (p != null) {
+        p.asInstanceOf[PermissionChecker].addLdapGroup(getNameWithLDAPPrefix(group), true)
```
dshirish: As an optimization, LDAP groups members iterator can be obtained only once and used everywhere while updating the list (instaed of getting it once in "permissions.refreshOnLdapGroupRefresh(group)" and again in "SnappyInterpreterExecute.updateMetaRegion" via call to "addLdapGroup". 
Can be done later.

ashetkar: Yes, will take care of it post 1.2.0.

PR #1524, opened 2020-01-15
Throw exception on interpreter getting offending line.

kneeraj:

Changed the name of the object 'snappy' having implicits to 'snappydata' to avoid clash with 'snappy'variable defined in snappy repl.

Throw run time error if interpretation of a line cannot be done.

## Patch testing

Manual

## ReleaseNotes.txt changes

No

## Other PRs 

No.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/SnappyImplicits.scala
+++ b/core/src/main/scala/org/apache/spark/sql/SnappyImplicits.scala
@@ -29,7 +29,7 @@ import org.apache.spark.{Partition, TaskContext}
  * Implicit conversions used by Snappy.
  */
 // scalastyle:off
-object snappy extends Serializable {
+object snappydata extends Serializable {
 // scalastyle:on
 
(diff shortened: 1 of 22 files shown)


core/src/main/scala/org/apache/spark/sql/SnappyImplicits.scala
```diff
@@ -29,7 +29,7 @@ import org.apache.spark.{Partition, TaskContext}
  * Implicit conversions used by Snappy.
  */
 // scalastyle:off
-object snappy extends Serializable {
+object snappydata extends Serializable {
```
dshirish: Does this change require customer/user code base to be changed/recompiled in order to upgrade to 1.2 if they are using implicits like putInto?

kneeraj: That's correct.
After discussion with Jags, Shirish and Nikhil we decided that let's change the Symbol name only for now instead of changing the name of the object which contains the implicits. Will close the PR.

PR #1536, opened 2020-02-24
Snap 3326

ahshahid:
If a table ( row or column) is being created using CTAS syntax, with option  stringtypeas 'varchar', then if the source table has schema with any structfield which is of type String Type and if ( the metadata associated with the struct field is null OR metadata DOES NOT contain key "base" OR metadata contains  "base" as key with  value = "STRING") ,  then that StructField is modified by adding a metadata key "base" = "varchar" and size = "32672", so that the struct field is mapped to SQL Type Varchar by the snappy driver.  

## Patch testing

added test

## ReleaseNotes.txt changes

yes . this will require addition to release notes.

## Other PRs 
[store-pr](https://github.com/SnappyDataInc/snappy-store/pull/548)
PR of the other subprojects that are related to this change)


jramnara:
To be consistent with the how options are specified for column tables, rename the option to string_type_as.

core/src/main/scala/org/apache/spark/sql/SnappyBaseParser.scala
```diff
@@ -298,7 +300,8 @@ abstract class SnappyBaseParser(session: SparkSession) extends Parser {
   }
 
   protected final def columnCharType: Rule1[DataType] = rule {
-    VARCHAR ~ '(' ~ ws ~ digits ~ ')' ~ ws ~> ((d: String) => VarcharType(d.toInt)) |
+    VARCHAR ~ ('(' ~ ws ~ digits ~ ')').? ~ ws ~> ((d: Any) => VarcharType(
```
jramnara: @ahshahid there is another requirement to support just Varchar as the column specification. i.e. allow `create table (a int, b varchar, c varchar(10)) using column`. Seems like we could modify this rule to support this

core/src/main/scala/org/apache/spark/sql/execution/CommonUtils.scala
```diff
@@ -0,0 +1,68 @@
...
+import io.snappydata.Constant
+
+import org.apache.spark.sql.{SnappyParser, SnappySession}
+import org.apache.spark.sql.types.{CharType, DataType, MetadataBuilder, StringType, StructType, VarcharType}
+
+
+object CommonUtils {
+  def modifySchemaIfNeeded(options: Map[String, String], session: SnappySession,
+    schema: StructType): StructType = {
+    options.get("stringtypeas").map(stringTypeAs => {
```
jramnara: use 'string_type_as' instead of 'stringtypeas'. Also, a nit - this constant should be defined as a token somewhere else?

core/src/main/scala/org/apache/spark/sql/execution/columnar/impl/DefaultSource.scala
```diff
@@ -102,8 +110,9 @@ final class DefaultSource extends ExternalSchemaRelationProvider with SchemaRela
   }
 
   private[sql] def createRelation(session: SnappySession, mode: SaveMode,
-      options: Map[String, String], specifiedSchema: StructType): BaseColumnFormatRelation = {
-
+      optionS: Map[String, String], specifiedSchema: StructType): BaseColumnFormatRelation = {
+   // remove the property stringtypeas as it has been utilized
+    val options = optionS.filterNot(tup => tup._1.equalsIgnoreCase("stringtypeas"))
```
jramnara: same nit - define the constant in one place :-)

ahshahid:
Right. I have to define "stringtypeas"  as constant.
The requirement of supporting just "varchar" is implemented in the change.

PR #1539, opened 2020-03-19
[SDENT-175] Support for cancelling query via JDBC

vatsalmevada:

Changes to support query cancellation using JDBC.

## Patch testing

- added new test
- precheckin clean

## Other PRs

https://github.com/SnappyDataInc/snappy-store/pull/551

### DIFF
--- a/cluster/src/dunit/scala/io/snappydata/externalstore/JDBCPreparedStatementDUnitTest.scala
+++ b/cluster/src/dunit/scala/io/snappydata/externalstore/JDBCPreparedStatementDUnitTest.scala
@@ -18,7 +18,7 @@ package io.snappydata.externalstore
 
 import java.sql.{PreparedStatement, SQLException}
 import java.util.concurrent.atomic.AtomicInteger
-import java.util.concurrent.{CountDownLatch, Executors}
+import java.util.concurrent.{CountDownLatch, Executors, TimeoutException}
 
 import scala.concurrent.{Await, ExecutionContext, ExecutionContextExecutor, Future}
 import scala.util.Try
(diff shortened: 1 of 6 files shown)


cluster/src/dunit/scala/io/snappydata/externalstore/JDBCPreparedStatementDUnitTest.scala
```diff
@@ -445,4 +451,47 @@ class JDBCPreparedStatementDUnitTest(s: String) extends ClusterManagerTestBase(s
     assertEquals(0, rscnt.getInt(1))
     assertEquals(100, deletedRecords.get)
   }
+
+  def testQueryCancellation(): Unit = {
```
kneeraj: I don't understand the test.
What is the use of the latch? I don't see it getting used.
Sleep should be avoided. I think perhaps latch is what you wanted for the purpose for what you introduced sleep?
I don't think that the test is properly checking whether the query was cancelled or not. It will pass even if it was not. Isn't it?

vatsalmevada: @kneeraj yes I was trying the latch approach to avoid sleep but later I realized that it won't work because we need to cancel the query before the execution of select is complete. Hence I reverted the latch approach but the declaration of latch left by mistake. The test should not pass if the cancel didn't work as we have `Assert.fail` statement after the execution of the select query.

P.S. Still looking for some approach to avoid sleep.

kneeraj: assert outside rs.next loop will be preferable.

PR #1542, opened 2020-03-27
[SDENT 180] support for fail fast type casting

vatsalmevada:

Apache Spark handles type casting failures in a fail safe manner. Instead of
failing the entire query, Spark populates `null` values whenever type casting
failure occurs. Snappydata inherits the same behavior from Spark.

With this changes we are exposing a session level configuration to enable fail
fast behaviour for casting errors. i.e. instead of returning `null` values for
casting failures, the query will be failed altogether whenever first type cast
error occurs. 

The configuration property is `snappydata.failOnCastError` which defaults
to `false`. Setting this property true will turn on fail fast casting behaviour.

Note that these changes won't be available on smart connector mode as the
changes are done at the Spark layer.

## Patch testing

- added new tests

## Other PRs 

https://github.com/SnappyDataInc/spark/pull/201

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -72,7 +72,7 @@ allprojects {
   apply plugin: "build-time-tracker"
 
   group = 'io.snappydata'
-  version = '1.2.0'
+  version = '1.3.0-SNAPSHOT'
 
   // apply compiler options
   tasks.withType(JavaCompile) {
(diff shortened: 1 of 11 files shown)


kneeraj (approved):
Can you please add just one test for cluster and set the property on jdbc connection and check if all looks good.
Also why have you named in 'fail fast'? Just 'fail on cast error'. Curious?

vatsalmevada:
> Also why have you named in 'fail fast'? Just 'fail on cast error'. Curious?
@kneeraj To keep it shorter. Also it depicted the overall behaviour which is opposite of the existing "fail safe" behaviour. I am fine with "fail on cast error" too if "fail fast" sounds confusing.

PR #1556, opened 2021-04-30
Fix "hostname -I" to use ifconfig/ip in snappy-nodes.sh

sumwale:
- "hostname -I" to get all IP addresses is non-standard and fails on many Linux distros
- instead use ifconfig or "ip addr" whichever is available

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ When speed is essential, applications can selectively copy the external data int
 
 
 ## Downloading and Installing SnappyData
-You can download and install the latest version of SnappyData from [github](https://github.com/SnappyDataInc/snappydata/releases) or you can download the enterprise version that is TIBCO ComputeDB from [here](https://edelivery.tibco.com/storefront/index.ep).
+You can download and install the latest version of SnappyData from [github](https://github.com/TIBCOSoftware/snappydata/releases) or you can download the enterprise version that is TIBCO ComputeDB from [here](https://edelivery.tibco.com/storefront/index.ep).
 Refer to the [documentation](docs//install.md) for installation steps.
 
 ## Getting Started
@@ -65,7 +65,7 @@ You can find more information on options for running SnappyData [here](docs/quic
 
 ## Quick Test to Measure Performance of SnappyData vs Apache Spark
 
-If you are already using Apache Spark, you can experience upto 20x speedup for your query performance with SnappyData. Try this [test](https://github.com/SnappyDataInc/snappydata/blob/master/examples/quickstart/scripts/Quickstart.scala) using the Spark Shell.
+If you are already using Apache Spark, you can experience upto 20x speedup for your query performance with SnappyData. Try this [test](https://github.com/TIBCOSoftware/snappydata/blob/master/examples/quickstart/scripts/Quickstart.scala) using the Spark Shell.
 
 ## Documentation
 To understand SnappyData and its features refer to the [documentation](http://snappydatainc.github.io/snappydata/).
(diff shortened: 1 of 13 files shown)


sumwale:
@ashetkar

Have also updated the spark package reference though I have no idea if the package has been re-published with TIBCOSoftware (for ex see the changes to examples/README.md). I guess upcoming release might be, but it would also be good to re-publish under TIBCOSoftware since SnappyDataInc is no longer around. Let me know if that change should be kept or reverted.

sumwale:
Also updated hive version (to a more recent one than the other PR).

README.md
```diff
@@ -48,7 +48,7 @@ When speed is essential, applications can selectively copy the external data int
 
 
 ## Downloading and Installing SnappyData
-You can download and install the latest version of SnappyData from [github](https://github.com/SnappyDataInc/snappydata/releases) or you can download the enterprise version that is TIBCO ComputeDB from [here](https://edelivery.tibco.com/storefront/index.ep).
+You can download and install the latest version of SnappyData from [github](https://github.com/TIBCOSoftware/snappydata/releases) or you can download the enterprise version that is TIBCO ComputeDB from [here](https://edelivery.tibco.com/storefront/index.ep).
```
ashetkar, line 51: Have done similar changes in another branch SDENT-196 which would be merged into master once all changes are completed and the process is approved by Legal.

build.gradle
```diff
@@ -127,7 +128,7 @@ allprojects {
     kryoVersion = '4.0.1'
     thriftVersion = '0.9.3'
     jacksonVersion = '2.9.5'
-    hiveVersion = '1.21.2.3.1.2.0-4'
+    hiveVersion = '1.21.2.3.1.2.2-1'
```
ashetkar, line 131: Me and Swati had earlier identified 1.21.2.3.1.2.1-1 as the version which is closest to the one being replaced. Have also ran precheckins with these two versions to understand if we have serious issues with change. Please see this PR https://github.com/TIBCOSoftware/snappydata/pull/1553

Anything particular about 1.21.2.3.1.2.2-1 or should we continue with 1.21.2.3.1.2.1-1 ?

core/build.gradle
```diff
@@ -291,7 +291,7 @@ task sparkPackagePom(dependsOn: shadowJar) { doLast {
   file("${rootProject.buildDir}/distributions").mkdirs()
   pom {
     project {
-      groupId 'SnappyDataInc'
+      groupId 'TIBCOSoftware'
```
ashetkar, line 294: Spark packages weren't published for 1.2.0-HF-1 and not sure if we can do that for upcoming 1.3.0.

core/src/main/scala/org/apache/spark/sql/SnappyContext.scala
```diff
@@ -413,7 +413,7 @@ class SnappyContext protected[spark](val snappySession: SnappySession)
    * @param provider Provider name such as 'COLUMN', 'ROW', 'JDBC' etc.
    * @param schema   Table schema
    * @param options  Properties for table creation. See options list for different tables.
-   *              https://github.com/SnappyDataInc/snappydata/blob/master/docs/rowAndColumnTables.md
+   *              https://github.com/TIBCOSoftware/snappydata/blob/master/docs/rowAndColumnTables.md
```
ashetkar, line 416: The filename does not exit anymore. Will need to find the alternative file and update here.
I think we have identified that in SDENT-196_new branch.

examples/README.md
```diff
@@ -69,8 +69,8 @@ You can run the example programs as jobs in the SnappyData cluster using either
   ```
   $SNAPPY_HOME/bin/spark-submit --class io.snappydata.examples.AirlineDataSparkApp \
   --master spark://<hostname>:7077 --conf snappydata.connection=localhost:1527 \
-  --packages SnappyDataInc:snappydata:1.1.0-HF-1-s_2.11 \
-  build/libs/snappydata-examples_2.11-1.1.0-HF-1.jar`
+  --packages TIBCOSoftware:snappydata:1.2.0-HF-1-s_2.11 \
```
ashetkar, line 72: Isn't available for 1.2.0-HF-1 but the latest is 1.2.0

examples/quickstart/scripts/Quickstart.scala
```diff
@@ -17,9 +17,7 @@
 /**
  * This script demonstrate the performance difference between Spark and Snappydata.
  * To execute this script on spark you can use below command:
- * ./spark-shell --driver-memory 4g --master local[*] --packages "SnappyDataInc:snappydata:0.7
- * .2-s_2.11" -i Quickstart
- * .scala
+ * ./spark-shell --driver-memory 4g --master local[*] --packages "TIBCOSoftware:snappydata:1.2.0-HF-1-s_2.11" -i Quickstart.scala
```
ashetkar, line 20: Same as above, the latest available is 1.2.0

README.md
sumwale: Ok will revert

build.gradle
sumwale: Nothing particular about this version apart as far as I recall apart from being a bit newer. The other PR has not been merged so if you plan to merge that, then I will revert this. Else will change to that version if that is better tested. Either of these two changes need to go in otherwise the product does not build as of now. Let me know.

examples/README.md
sumwale: Ok will revert assuming these will be taken care of in SDENT-196

build.gradle
ashetkar: Yes, will merge the above PR soon.

ashetkar (approved):

ashetkar (approved):

